PR #8, opened 2013-04-29
Update Workque to 0.9.64 and add wmagent spec files

ericvaandering:
Please take 0.9.64 for WMAgent, Workqueue


geneguvo:
Have you tested this? It is missing, for instance, py2-zmq.spec and zeromq.spec.

From the https://twiki.cern.ch/twiki/bin/viewauth/CMS/DMWMBuildsStatusReleases page, your last wmagent tag seems to be:
bld_1366294257_wmagent-dev_slc5_amd64_gcc461

After getting py2-zmq and zeromq from there, I could build wmagent-dev successfully (on SLC5) and I'll therefore commit the changes. However, please note that the bld_\* tag above had some outdated dependencies. By synching with comp/HEAD, you will get the updates, which is probably good, but keep in mind that there are changes there. You probably want to revalidate the wmagent RPMs you get by building from the comp branch.

geneguvo:
Done.

ericvaandering:
Thanks Diego.

I got a 30 minute crash course from Steve yesterday on building releases since he's leaving. This is going to be a rough process for the moment. How do I "test" the build process since, as far as I know, it doesn't build anything locally and only instructs the infrastructure to make a build.

Thanks,

Eric

PR #66, opened 2013-05-30
add stopm dependency for ASO

spigad:


geneguvo:
Ciao,
Was the inclusion of this new dependency discussed somewhere? Am I missing something?
Cheers,
Diego

spigad:
No never discussed. It's needed for building the ASO in crab3-panda mode.  Is it a problem to include it?

geneguvo:
If this is needed to talk to Panda, then ok. I suppose we don't have any other option...

The problem with new dependencies is always the same:
- it is one more external for us to manage: to make sure is update, to make sure it build/validate for SLC5/SLC6/OSX and newer python versions;
- it is one more thing to port in case of infrastructure changes: i.e. SSL/grid security support
- it is one more different stuff we use in our projects, therefore one more thing to audit wrt security.
- it is one more thing we don't yet know how it would behave in production: bugs that defeat us, performance issues, etc..

So before embarking with any project design that needs new externals, I'd like people to discuss first. We'll then analyse the use case and see if there aren't other ways to achieve the same thing with the tools we already have and reasonable effort. Of course if the new thing gives us clear benefits over what we already have we would consider. There is always a tradeoff to play.

 I'll include it, but it will be renamed to py2-stomp to follow the naming convention of other python packages.
Cheers,
Diego.

spigad:
Hi Diego, 
  I see all your points and I fully agree. Few things: 
- I presented the architecture many many times in more than one forum. Probably I missed the right one (BTW which would be the right one ?). 
- There is a constraint (PanDA) which enforce us to use the ActiveMQ message bus for PanDA communication (ASO callback panda).
- I'm not aware of other smart solution instead of stomp so I suggested Hassen to use it for this purpose. Obviously I'm opened to anything better so if there is a suggestion from you / anybody I'm happy to consider it. 
  Cheers and thanks 
       Daniele

geneguvo:
The way to proceed when you planning to use a new dependency is to write a message to webInterfaces with that explicit topic "inclusion of library XYZ on June release for use in crab", then justify your use case. Then people can comment if we already have any solutions for the same problem, or if they know about better libs, or if they see any problems with it, etc.

It is very hard for us to catch these things from project architecture presentations, or even when people talk about those things on meetings. Even talking to me directly is not enough. It is possible we have some other project responsible planning to use ActiveMQ and have already evaluated some other tools and concluded it would be better to use something other than stomppy. Who knows...

spigad:
All right. Nothing prevent to start this process now.. I'm fine with it. Please let me know if you are ok, and I'll send the mail or even better I'll ask Hassen to do it :)

cheers Daniele

geneguvo:
https://github.com/cms-sw/cmsdist/commit/4032fc9b4af63746bb1cfe49e75b7d3e56014b7e

PR #82, opened 2013-06-11
0.9.69a

dballesteros7:
This is the June production stable WMAgent release. Please build an RPM for this.


geneguvo:
Eric requested for wmagent 0.9.76 on #80. So, which one should I pick up for wmagent, 0.9.76 or 0.9.69a? I'm going with 0.9.69a, then if needed I can build 0.9.76 later.

dballesteros7:
Thanks! For WMAgent 0.9.69a was the one we needed the RPM built.

PR #83, opened 2013-06-13
WMCore 0.9.77

ericvaandering:
Please pull to build WMCore/WMAgent 0.9.77. Needed for CRAB3, nothing else.


geneguvo:
Hi Eric,

Your pull requests have been very confusing. What do you mean by "Needed for CRAB3" here? if it is needed by crab3, you need to change that on the crab*.spec files, not on workqueue or wmagent. The fact you tag WMCore with something has nothing to do with requesting service-specific RPMs like for Crab.

Also, I've noticed you always bump a new wmagent version together with workqueue. They are independent project RPMs deployed on different places. The workqueue one gets only deployed centrally and is included in the cmsweb releases, so you only need to care about it when targeting the monthly cmsweb deployments. I know wmagent has a workqueue "component", but that's already shipped with the wmagent RPM AFAICT. 

The only thing it makes sense for you to request RPMs for is wmagent{,-dev}. For workqueue, reqmgr, reqmon, crab, etc, their own responsibles should speak for. If you stomp other project requests, I have no way to tell which one is the right one to be picked up.

On a separate notice, I see you mistakenly requested to merge to "master", not to "comp", and therefore there is a monster amount of changes...

Cheers,
Diego.

ericvaandering:
Hi Diego. Lots of issues here:

1) Sorry about the request to master. Once I know the correct thing to do (rest of message) let's close this and I'll do it right.

2) We need to get a new version of the WMCore RPM built for CRAB3 to include in its dependencies. This is how I know how to do it. If there is a better way, let me know and I'll try to do that.

3) Similarly, building new WMCore/WMAgent/WorkQueue RPMs at the same time is basically what I know how to do. I don't see that it harms anything since I explicitly pointed out it should not be deployed. (And of course I did not open TRAC ticket.)

4) AFAIK, I am the only person that is going to be requesting WorkQueue RPMs and deployment of them. I took that over from Steve F when he left.

Thanks,

Eric

geneguvo:
Hi Eric,

Your work flow is wrong. We don't build a RPM for WMCore anymore since a while ago, therefore no projects have a dependency on it. Each project itself (crab, wmagent, workqueue, etc) builds its own independent RPM and includes only the WMCore pieces it needs. If you need a new Crab RPM, you need to change the crabserver.spec file.

However, if a crab RPM is needed, it must come from their responsibles, Mattia or Daniele in this case.

And yes, creating intermediate (junk) RPMs pisses me off. It blows up the RPM repository with plenty of unneeded RPMs that only make it slower and the repository operations difficult. It also generates confusion and chances the HTTP group picks the wrong version are higher. If that happens and go to production, we can have other serious problems. So please only request a new RPM when it is really needed. If you need an RPM for development/testing, you can build it yourself and upload to your private repository as the tutorial tells. You don't need to waste time requesting this and we don't get interrupted all the time.

Thanks,
Diego.

geneguvo:
Hi Eric,

I apologize for being so rude/sharp here. I'm just stressed again on a friday afternoon with various other things bugging on my head... 

Really sorry!

PR #91, opened 2013-06-24
[IB/CMSSW_6_2_X/stable] Enable GIT by default; Move to G4 9.6.p02

davidlt:


ktf:
+1. Go ahead and merge it everywhere. Notice we still need to update cms-git-tools and add the to cmssw-tool-conf before we proceed with 6_2_0_pre8.

PR #120, opened 2013-07-13
Tauolapp113patch

inugent:
Adding spec and toolfile for Tauola 1.1.3


davidlt:
Good to be approved, but needs to go together with cms-sw/cmssw#94. Tested on `CMSSW_7_0_X_2013-07-12-1400`.

PR #143, opened 2013-07-25
add spec for taskworker and adjust crabserver and client

spigad:


geneguvo:
I've picked this but with the following small changes:
- keep the crabclient rpm package name instead of crab-client3. Using the "3" on the package name is redundant since it is already in the package version. For DBS we allowed  that because we need to have both the dbs2 and 3 rpms at the same time, but there is no such crab2 rpm. I don't object if you guys prefer "crab-client3" anyway, but if that's the case we need to rename the spec file accordingly too;
- do always keep the %n on the output part of the source url for sources shared with other packages like wmcore. Otherwise there is a race condition when building those packages in parallel (both processes will try to write the same file on disk but one will fail). This is a known issue with pkgtools.

PR #153, opened 2013-08-01
[CMSSW_7_0_X][48] Update for GCC, fixes for Darwin platform, dead sources clean up, data packages update

davidlt:
**ChangeLog**
- Final changes for GCC 4.8.
- Fixes for Darwin platform (tested on _macms09_).
- Clean up dead sources.
- Build data packages by download files directly instead of looking into _download.url_ files in CMSSW packages.


ktf:
Not only related to this one. Can we try to minimize the number of branches we have? We are diverging again in N different directions for no particular reason IMHO.

davidlt:
Exactly. This is a merge between two branches tested to some extend on many architectures. Single branch for GCC 4.8, which becomes the main branch for CMSSW_7_0_X once we say yes to 4.8.1.

PR #162, opened 2013-08-20
comp.pre slc5_amd64_gcc461 dbs3 dbs3-client dbs3-migration dbs3-pycurl-client

giffels:
Hi,

could you build dbs3 rpms (comp.pre, slc5_amd64_gcc461) for the new DBS release 3.1.6, please?

Thanks,
Manuel


geneguvo:
Heya,

Just a small comment. Was the inclusion of py2-docutils in the dependencies intentional? This is usually used only during build time to build the documentation, so there is no need to include it as a dependency (there is already a BuildRequires for py2-sphinx, which includes py2-docutils).

If this was added just to conveniently have it on the dev environment, then instead you could deploy "admin/devtools" together with dbs3 in you devvm. This would bring up sphinx and docutils once you source the admin/init.sh script (see https://cms-http-group.web.cern.ch/cms-http-group/tutorials/environ/rpm-dev.html).

Let me know your decision... don't need to make a new request, I can change that.

giffels:
Hi Diego,

sorry for the late answer, I just returned from vacation. The inclusion of py2-docutils is intentional. We are still using the old WMCore REST API. This API generating a documentation page from the docstrings of the methods. Since we are using Sphinx syntax in docstrings, it looks quite ugly. See https://cmsweb.cern.ch/dbs/dev/global/DBSReader/ for example. 
DBS 3 is now using a different cheetah template, which is parsing the sphinx syntax in the docstring using docutils dynamically. The result is a much nicer documentation page. See https://dbs3-dev01.cern.ch/dbs/dev/global/DBSReader/ for example.

I hope that is compliant with the cmsweb regulations. 

Cheers,
Manuel

giffels:
Hi Diego (@geneguvo),

Valentin has discovered a bug in the listDataset API (https://svnweb.cern.ch/trac/CMSDMWM/ticket/4260), which has been fixed in 3.1.6a. Could you re-pick the specs from the updated branch above, please?

Thanks,
Manuel

geneguvo:
Yes. No problem. I was just about to close the release, so still in time for this change.

PR #169, opened 2013-08-23
reqmgr2 update for september (based on reqmgr2_0.9.79 WMCore version)

ticoann:


geneguvo:
This can't possibly work. You are pointing to a branch that doesn't exist on my private (geneguvo) clone of WMCore. 

I suppose the idea was to point to the reqmgr2 branch of WMCore under the dmwm organization, right? however, that can't work too since nobody included the Dashboard fix we did for the 0.8.78a release in the previous cycle. Could you please include https://github.com/geneguvo/WMCore/commit/b80b7cc1248fafc5fc0606cb33b2ef2dddc3403d to main master/WMCore and also reqmgr2/WMCore branches?

On a separate topic, why reqmgr2 branched? Is it doing any sort of incompatible change with reqmgr1 ? note that reqmgr2 depends on reqmgr1 and therefore shares its couch databases. For this reason they must always be in sync to whatever data formats and api changes. I.e. if you change a request parameter you must change on both at the same time. Failing to do that is a recipe for problems in the near future.

geneguvo:
Also, there is no point for this reqmgr2_0.9.79 tag as it points exatcly to the same 0.9.79 (the last commit prior to the branching). I suppose you whish to make catch the changes you added afterwards to the reqmgr2 branch, otherwise it wouldn't make any sense to have this branch and this tag.

ticoann:
Sorry Diego, I meant to use dmwm/WMCore reqmgr2 branch. I changed the spec and added your patch on Dashboard dependencies. (I didn't notice there were dashboard dependencies in current reqmgr2. 
I couldn't find the code which calls dashboard directly from reqmgr2. It might be somewhere in spec api) Thanks for pointer.

Yes you are right. The reason uses reqmgr2 branch is there is due to the  incompatibility to reqmgr1.
Although it shares some db (couchdb), I realized that we can't run at both at the same time. 
(Both can be run but one has to be read only, also data needs to be manually merged from oracle to couchdb, and couchdb to oracle (if reqmgr2 is allow to write and  if both needs to sync exactly). Anyway, I thought it might be better to create the branch instead of try to maintain both versions of reqmgr which requires some hacks probably needs to be removed later when reqmgr2 switch happens. 

Actually, it is possible to deploy reqmgr2 just on testbed without updating production. Current reqmgr2 includes which writes on db. I am sure no one would use that intentionally in production but it doesn't need to be in due to the reason I mentioned above. (One has to be read only)

ticoann:
Forgot to mention that I changed the spec to point dmwm/WMCore reqmgr2 branch now.

geneguvo:
If they share the same couch db, you can't have reqmgr1 pushing the couchapp and then reqmgr2 comes in and pushes something else. For this reason we made sure in the deploy scripts that reqmgr2 don't push any couchapp that reqmgr1 already did.

If you want to share the couchapps and  the data on the couch database, you must make them compatible. Of course you still don't want reqmgr2 writing anything there while it has not been validated. Otherwise chances you'll disrupt reqmgr1 are high.

ticoann:
I see. Then please ignore this request. I will create the new one which is compatible. It doesn't have to be in the production this time. Thanks,

PR #174, opened 2013-08-30
Build PhEDEx on SLC6 with https git-access. Also updated lifecycle-dataprovider and some p5-* modules

TonyWildish:
This includes a few minor bugfixes for PhEDEx, but the main feature is the changes for building on SLC6. 

This means getting the code from git via HTTPS, because of the zlib problem with native git commands. This applies to the PhEDEx spec files but also to the lifecycle-dataprovider.

To build on SLC6, a number of perl modules needed to be recompiled too. These have had a dummy line added to the spec file to trigger a rebuild.

So, could this request be merged please, and a build made in the comp repository?

### DIFF
--- a/PHEDEX-admin.spec
+++ b/PHEDEX-admin.spec
@@ -27,6 +27,8 @@ Provides: perl(Net::Twitter::Lite)
 
 %prep
 %setup -n %{downloadn}-admin
+rm -rf Build
+rm -rf Contrib
 rm -rf Documentation/ACAT2008
 rm -rf Documentation/DC04PostMortem
 rm -rf Documentation/DC04Stats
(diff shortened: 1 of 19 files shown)


davidlt:
SLC6 boostrap contains _zlib 1.2.3_, which does not compile in version symbols required by GIT. _cmsBuild_ uses GIT from system and not from CMSDIST. The problem is that _cmsBuild_ initializes bootstrap environment and _zlib_ from CMSDIST ends up as a first library for dynamic linker. Thus GIT fails with _zlib_ related errors. For CMSSW it's fixed in `slc6_amd64_gcc481` arch.

TonyWildish:
just checking: is this waiting for me to do something more, or is it in someone's queue?

davidlt:
It's for computing, thus it's now up to Diego to merge or not to merge it.

PR #185, opened 2013-09-10
Do not use $OSG_APP if the installation is not actually there.

ktf:


ktf:
As requested by @bbockelm  in https://hypernews.cern.ch/HyperNews/CMS/get/sw-develtools/1916/1/1/1/1.html.

ktf:
@davidlt can you build and upload it, please?

PR #192, opened 2013-09-17
llvm,llvm-gcc-toolfile: update to work with GCC 4.8.1

davidlt:
Remove `-Wno-unused-local-typedefs' GCC option on LLVM/Clang, it's
not supported. Our (SLC{5,6}) triplet is x86_64-redhat-linux-gnu,
which is not available in Clang. Add it.


ktf:
+1

PR #194, opened 2013-09-18
Second pythia8

mkirsano:


vciulli:
@Mikhail in cmssw-tool-conf.spec please add 
Requires: pythia8-175-toolfile
otherwise pythia8-175 is not built.

I also see in tauolapp.spec 
Requires: pythia8
./configure --prefix=%{i} --with-hepmc=$HEPMC_ROOT --with-pythia8libs=$PYTHIA_ROOT --with-lhapdf=$LHAPDF_ROOT CXX="%cms_cxx" CXXFLAGS="%cms_cxxflags"

I think it would be better to compile it with pythia8-175
But it is not mandatory now and can be changed later, as tauolapp it is only linked to have tauspinner and can not be used in 53X for the moment.

mkirsano:
Done (the first one)
                        Mikhail

PR #199, opened 2013-09-27
rpm,apt: use new bootstrap-bundle and make gcc a build require

smuzaffar:
This allows us to bundle all the needed externals of rpm/apt into one rpm (bootstrap-bundle) and make gcc a build time dependency. This reduces the bootstrap size to 35MB from 560MB.

Note, this just touches apt.spec and rpm.spec (plus new *-bootstraps.spec) only. So it is safe to merge it in to stable branch of 70X. This is not going to trigger a re-built or any thing (unless one explicitly build apt/rpm or bootstrap-driver).

I have tested it for slc6_amd64_gcc481, slc5_amd64_gcc481, osx108_amd64_gcc481, fc18_armv7hl_gcc481 and slc6_mic_gcc481.

NOTE: I am having problems forking the cmsdist repo. So I have created a branch in cms-sw/cmsdist. Feel free to delete this branch once merge is done.


ktf:
This is great. Merging this: we can debug it if we have issues but this is clearly the way to go.

PR #212, opened 2013-10-08
Use system python on MIC.

ktf:
Our python only works on MIC itself... We should probably use perl here...


smuzaffar:
I think this should be other way round. We should use system python for mic build and our python for rest

PR #214, opened 2013-10-08
Typo in ordering...

ktf:


ktf:
As suggested by @smuzaffar in #212 , we need to use the system python on MIC.

PR #252, opened 2013-10-29
Bugfix/add nss deps

PerilousApricot:
Nss was sucking in zlib and sqlite dependencies from the system instead of from ourself

### DIFF
--- a/nss-3.12.9-add-SQLITE-LIBS-DIR.patch
+++ b/nss-3.12.9-add-SQLITE-LIBS-DIR.patch
@@ -1,0 +1,22 @@
+diff -rupN nss-3.12.9.orig/mozilla/security/nss/cmd/platlibs.mk nss-3.12.9/mozilla/security/nss/cmd/platlibs.mk
+--- nss-3.12.9.orig/mozilla/security/nss/cmd/platlibs.mk	2012-01-28 21:16:37.000000000 +0100
++++ nss-3.12.9/mozilla/security/nss/cmd/platlibs.mk	2012-01-28 21:18:45.000000000 +0100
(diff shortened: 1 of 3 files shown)


davidlt:
You can always drop `NSS_USE_SYSTEM_SQLITE` and `USE_SYSTEM_ZLIB`. In that case embedded zlib and sqlite3 is used.

PerilousApricot:
Ah, I didn't realize that. Let me finish pushing all my local changes
upstream then try building the other way to make sure it clicks.

-Melo

davidlt:
It will click correctly, but the question would be: is using embedded zlib and sqlite3 for nss trigger any compatibility issues, or not? I am moving towards directly that it doesn't. You might also want to enable

```
export BUILD_OPT=1
export NSS_NO_PKCS11_BYPASS=1
```

And we should try to build with `NSS_BUILD_WITHOUT_SOFTOKEN`. We shouldn't ship _softoken_ or _freebl_ libraries. _glibc_ depends on _freebl_.

I did some experimentation and went back into thinking mode: https://github.com/davidlt/cmsdist/commit/2ec18b73dc3bb195f4ff2fc6cb763a0e87f62406

PerilousApricot:
Interesting. Let me try and build with all those options in a bit and deploy a server with it. Can you think of a good test that would hit a potential compatibility snag?

davidlt:
Well:

libnss-sqlite is a Name Service Switch (NSS) library which allows you to use a SQLite database as backend for UNIX authentication. This way, you can store users and groups in a database rather than in files.

This sounds that might not be needed, and if I recall correctly it's only used by _softoken_.

PR #262, opened 2013-11-06
Package HTCondor along with perl build-time dependencies.

bbockelm:
This adds HTCondor to comp -- builds cleanly (and results in a functioning set of HTCondor clients) for me on SL5.  I think more work is needed for SL6.

### DIFF
--- a/condor.spec
+++ b/condor.spec
@@ -1,0 +1,92 @@
+### RPM external condor 8.0.4
+## INITENV +PATH LD_LIBRARY_PATH %i/lib
+## INITENV +PATH LD_LIBRARY_PATH %i/lib/condor
+## INITENV +PATH PYTHONPATH %i/lib/python
+
+%define _default_patch_fuzz 2
+%define condortag V8_0_3
+
+%global __requires_exclude libkeyutils\\.so.*|libsasl2\\.so.*|libselinux\\.so.*
+
+Source: git://github.com/htcondor/htcondor.git?obj=master/%{condortag}&export=condor-%{realversion}&output=/condor-%{realversion}.tar.gz
+Patch0: cms-htcondor-build
+
+Requires: openssl zlib
+Requires: cmake gcc
+Requires: python
+Requires: libtool
+Requires: p5-archive-tar
+Requires: expat
+Requires: pcre
+
+%prep 
+%setup -n %n-%{realversion}
+%patch0 -p1
+sed -i "s,P5_ARCHIVE_TAR_ROOT,$P5_ARCHIVE_TAR_ROOT," externals/bundles/globus/5.2.1/CMakeLists.txt
+sed -i "s,P5_IO_ZLIB_ROOT,$P5_IO_ZLIB_ROOT," externals/bundles/globus/5.2.1/CMakeLists.txt
+sed -i "s,P5_PACKAGE_CONSTANTS_ROOT,$P5_PACKAGE_CONSTANTS_ROOT," externals/bundles/globus/5.2.1/CMakeLists.txt
+sed -i "s,ENVLD_LIBRARY_PATH,$LD_LIBRARY_PATH," externals/bundles/globus/5.2.1/CMakeLists.txt
+sed -i "s,LIBTOOL_ROOT,$LIBTOOL_ROOT,g" externals/bundles/globus/5.2.1/CMakeLists.txt
+sed -i "s,OPENSSL_ROOT,$OPENSSL_ROOT,g" externals/bundles/globus/5.2.1/CMakeLists.txt
+#sed -i "s,EXPAT_ROOT,$EXPAT_ROOT,g" externals/bundles/voms/2.0.6/CMakeLists.txt
+sed -i "s,EXPAT_ROOT,$PWD/cmake_build/voms_expat,g" externals/bundles/voms/2.0.6/CMakeLists.txt
+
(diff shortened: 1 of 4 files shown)


geneguvo:
Hi Brian,

It looks like you forgot to commit the cms-htcondor-build.patch libtool.spec and possibly other files. Also, you did the pull request against cmsdist/master, not cmsdist/comp.

I see there is a branch called "comp_htcondor" in your private github copy of cmsdist, and it looks all files are there. Should I use it, instead?

Cheers,
Diego.

bbockelm:
Yes - go ahead and use the comp_htcondor version.  I was just trying to clean things up for you.

Didn't realize that we wanted to merge against comp, not master.

geneguvo:
Ok. I've reviewed it and did various changes:
- condortag was V8_0_3, while version on the header was 8.0.4. Changed to grab the version tag automatically to avoid inconsistencies;
- LD_LIBRARY_PATH for %i/lib is already added by default. Removed it;
- set the PYTHONPATH appropriately;
- some difficulty to understand what is build dependency and what was runtime dependency. Put cmake, gcc build dependency, but looks like more cleaning could be done;
- pass other dependencies paths to CMAKE_LIBRARY_PATH so that cmake can detect them correctly;
- do not build boost, but use the one we already have;
- removed hacks that were not needed (expat, the fuzz option to patch, etc);
- got rid of hardcoded references to /usr/bin/python on the produced condor package;
- generate the dependencies shell script.

I'm about to commit it and release the official rpms to comp and comp.pre (I'm just waiting for the build to finish).

On a separate issue, we should try to build voms and globus as a separate package and just make condor depend on them (do not do  the "make externals"). There are many reasons to do  that:
- they are not being configuration managed when getting built as part of the condor package, which means there is no way from someone to find they already exist, nor to know their versions, nor to know how they were built (i.e. which enabled options); 
- we are not keeping track of their sources, so they don't get cached on the RPM repository;
- even if we only need to rebuild condor itself for some adjustment, we always have to wait it to rebuild globus and voms, which is annoying;
- we foresee using voms standalone, regardless of condor needing it or not.

But in general this is good enough to start with. Hopefully I have not cleaned too much stuff.

bbockelm:
Hi Diego,

Thanks for the help!

Everything looks good.  Two side notes:
- If we wish, we can include the Globus tarballs in the source RPM.  HTCondor can take a base "source URL" for externals, which includes "file:///"
- I started this package 2 days before 8.0.4 was tagged, but knew I wanted 8.0.4; this is the explanation behind calling the version 8.0.4 and using 8.0.3 sources.  I swear I'm not crazy.

Brian

geneguvo:
Yeah, that's fine.

The RPMs are now available in the "comp" and the "comp.pre" RPM repositories.

```
# Bootstrap the software distribution area for the first time
$ (ROOT=/data/compsw; mkdir -p $ROOT; cd $ROOT; wget http://cmsrep.cern.ch/cmssw/comp/bootstrap.sh &>/dev/null; sh -x ./bootstrap.sh -architecture slc5_amd64_gcc461 -path $ROOT -repository comp setup &> bootstrap.log)

# Install the version 8.0.4 on existing area
$ (ROOT=/data/compsw; source $ROOT/slc5_amd64_gcc461/external/apt/*/etc/profile.d/init.sh; apt-get update; apt-get install -y external+condor+8.0.4)
...
Executing RPM (/data/compsw/slc5_amd64_gcc461/external/apt/429-comp/bin/rpm-wrapper -Uvh -r /data/compsw --force --prefix /data/compsw --ignoreos --ignorearch --force --prefix /data/compsw --ignoreos --ignorearch --oldpackage)...
Preparing...                ########################################### [100%]
   1:external+expat+2.0.1-co########################################### [  6%]
   2:external+zlib+1.2.8    ########################################### [ 13%]
   3:external+openssl+0.9.8e########################################### [ 19%]
   4:external+p5-extutils-ma########################################### [ 25%]
   5:external+gdbm+1.8.3-com########################################### [ 31%]
   6:external+sqlite+3.7.10 ########################################### [ 38%]
   7:external+python+2.6.8-c########################################### [ 44%]
   8:external+boost+1.51.0-c########################################### [ 50%]
   9:external+curl+7.28.0-co########################################### [ 56%]
  10:external+libxml2+2.7.7-########################################### [ 63%]
  11:external+pcre+7.9-comp ########################################### [ 69%]
  12:external+libtool+2.4.2 ########################################### [ 75%]
  13:external+p5-io-zlib+1.1########################################### [ 81%]
  14:external+p5-package-con########################################### [ 88%]
  15:external+p5-archive-tar########################################### [ 94%]
  16:external+condor+8.0.4  ########################################### [100%]
Done.

# Run it standalone
$ (ROOT=/data/compsw; source $ROOT/slc5_amd64_gcc461/external/condor/8.0.4/etc/profile.d/init.sh; condor_q)

Neither the environment variable CONDOR_CONFIG,
/etc/condor/, nor ~condor/ contain a condor_config source.
Either set CONDOR_CONFIG to point to a valid config source,
or put a "condor_config" file in /etc/condor or ~condor/
Exiting.
```

For crab3, however, putting "Requires: condor" should be enough. So when crab3 is deployed it installs condor as dependency, and when the crab3 environment is sourced, it has condor tools/libraries available too.

I'm closing this request.
Cheers,
Diego.

PR #275, opened 2013-11-20
Fix crabserver and taskworker spec after repository migration. Push crabserver 3.3.0.rc1 version

mmascher:


crabserver.spec
```diff
@@ -1,36 +1,30 @@
...
 %build
-
 cd ../WMCore-%{wmcver}
 python setup.py build_system -s crabserver
 PYTHONPATH=$PWD/build/lib:$PYTHONPATH
-
-cd ../CAFUtilities-%{crabutils}
-perl -p -i -e "s{<VERSION>}{%{realversion}}g" doc/crabutilities/conf.py
-python setup.py build_system -s crabserver
+touch $PWD/build/condor_config
```
PerilousApricot: These condor_configs need to be put above line 24 (the WMCore setup.py) and it also needs to be in the %install section. also, $PWD/build doesn't necessarily always exist.

mmascher: I applied your fixes, it should be alright now

PR #280, opened 2013-11-26
HG1312 DAS update changes

vkuznet:
Please take new DAS version along with changes required to properly generate DAS maps for cmsweb-testbed.


amaltaro:
Thanks Valentin, we are handling it right now.
I have a suggestion for this code, but then it's better to properly test and come with it for the next cycle (2014). You could run sed directly on the file you want to get changed, so we could get rid off the tmp files and maybe the tmp dir. It would be something like:
sed -i "s+cmsweb.cern.ch+cmsweb-testbed.cern.ch+" yourFile

PR #288, opened 2013-12-02
Generate runtime tarball for TaskWorker.

bbockelm:
Generate runtime tarball during the TaskWorker RPM generation.  This allows the TaskWorker to be self-contained and not rely on an external server.

### DIFF
--- a/crabclient.spec
+++ b/crabclient.spec
@@ -1,14 +1,14 @@
-### RPM cms crabclient 3.3.0.pre3
+### RPM cms crabclient 3.3.0.rc1
 ## INITENV +PATH PATH %i/xbin
 ## INITENV +PATH PYTHONPATH %i/$PYTHON_LIB_SITE_PACKAGES
 ## INITENV +PATH PYTHONPATH %i/x$PYTHON_LIB_SITE_PACKAGES
 
-%define wmcver 0.9.83
+%define wmcver 0.9.84.htcondor2
 %define webdoc_files %{installroot}/%{pkgrel}/doc/
-%define crabserver 3.3.0.pre30
+%define crabserver 3.3.0.rc6
 
 
-Source0: git://github.com/dmwm/WMCore.git?obj=master/%{wmcver}&export=WMCore-%{wmcver}&output=/WMCore-%{n}-%{wmcver}.tar.gz
+Source0: git://github.com/bbockelm/WMCore.git?obj=master/%{wmcver}&export=WMCore-%{wmcver}&output=/WMCore-%{n}-%{wmcver}.tar.gz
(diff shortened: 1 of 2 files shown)


geneguvo:
Hi,

I'd not agree with doing things like this. It goes against the idea of using RPMs for the software packaging and distribution. You basically have your own "configuration management" inside this htcondor_make_runtime.sh script, which makes things hard to maintain and to version control. Each package shall have only its own files, and use RPM dependencies for what it needs.

Would you mind to clarify what's the use case ? I don't understand what you mean with "self-contained and not rely on external server".

Cheers,
Diego.

PerilousApricot:
Hi Diego,

These tarballs are downloaded/unpacked on the worker nodes at runtime and contain the needed dependencies. Currently, we run the make_runtime script and drop the tarball onto an HTTP server somewhere and have the jobs curl the tarball on startup. By packaging that tarball with the server, we can simply transfer it within condor from the task worker to the worker node, dropping the need to have HTTP servers around to simply serve up a tarball. This is the same behavior as WMAgent's sandbox creation facilities.

Cheers,
Andrew

bbockelm:
Hi @geneguvo -

Any updates on this?

Brian

geneguvo:
Hi guys,
Sorry, I couldn't reply before as I'm busy.

I see the use case now, but even though it is not good to make these things. There are a bunch of issues that can arise from mixing package versions from different places. This really needs to be configuration managed somehow.

I'd say this is the kind of things to be distributed through CVMFS to the worker nodes. You deploy a RPM just like people do for CMSSW releases, and then the job sources the init.sh to have all the environment ready for use.

Alternatively, if you need/prefer to have it distributed through condor, then I'd at least expect it to be on a spec file using the RPMS/sources defined by cmsdist, not hardcoded on an external script executed during build time. I.e. create a taskworker-sandbox.spec, put "BuildRequires" for every dependency, then cp only the needed files and/or create the tarball there. Once the sandbox RPM is installed on the taskworker you have the files and/or the tarball available to ship to the worker nodes.

Cheers,
Diego.

bbockelm:
Hi Diego,

I agree that CVMFS is one eventual path - but CMS is not making that a hard requirement for sites until April 2014.  So, we cannot proceed to CVMFS immediately.

I think, short-term, migration of the functionality out of the script into the RPM is more than doable (not only maintaining the exact dependency list in the RPM, but also using RPM to stage the input files).  This was already our intent.  It's not completely straightforward, as we also want developers to be able to generate this outside of RPM for their own development instances.

However, we need a starting point to get everyone on the same page before we rewrite the script.  So, I ask you to take the pull request for now with the understanding that further cleanup will happen throughout the month which will get things into a better state.

Thanks,

Brian

geneguvo:
Yes. Agreed.

PR #289, opened 2013-12-03
add parsing of the custom CMS c++11 attributes cms::ThreadSafe and cms::...

gartung:
...ThreadGuard("guard var name")


gartung:
@Dr15Jones @ktf this is the patch to allow clang to parse our custom attributes and add them to the clang AST.

PR #294, opened 2013-12-12
rivet 19 and yamlcpp

mkirsano:


ktf:
No, it was done in #343 which adapts to the yaml-cpp tool file we already had in 70X.

PR #296, opened 2013-12-12
new scram version, config tag, html log zip instead of tar gz

smuzaffar:
see PR #295

### DIFF
--- a/SCRAMV1.spec
+++ b/SCRAMV1.spec
@@ -1,4 +1,4 @@
-### RPM lcg SCRAMV1 V2_2_5_pre4
+### RPM lcg SCRAMV1 V2_2_5_pre5
 ## NOCOMPILER
 %define GitHubVersion %(echo SCRAM-%realversion | sed 's|-V|-|')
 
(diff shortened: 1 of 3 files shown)


davidlt:
@smuzaffar have you tested it locally?

smuzaffar:
I build cmssw.spec and every thing worked fine. I must admit I did not 
build cmssw-ib.spec (where I have changes for uzing zip instead of tar).

davidlt:
Any particular reason for uzing zip instead of *nix friendly tarballs?

smuzaffar:
Giulio had noticed that for few days IB web pages are very slow 
(specailly when are go to runTheMatrix logs). I noticed that extrating 
the files list from the tar file was very slow as compare to zip and 
also extraing single file out of tgz was more than 1000 time slower, so 
we diced to use zip.

smuzaffar:
you can see the difference by going to

logs with zip:
http://cmssdt.cern.ch/SDT/cgi-bin//showMatrixTestLogs.py/slc5_amd64_gcc481/www/thu/7.0-thu-02/CMSSW_7_0_X_2013-12-12-0200/pyRelValMatrixLogs/run/

vs logs in tgz
https://cmssdt.cern.ch/SDT/cgi-bin//showMatrixTestLogs.py/slc6_amd64_gcc481/www/thu/7.0-thu-02/CMSSW_7_0_X_2013-12-12-0200/pyRelValMatrixLogs/run/

ktf:
tarballs do not have an index, zip does.

davidlt:
I did some digging. You are correct that standard tarball does not contain index, while zip contains index and compress each file individually.

ktf:
I know I'm correct, as always...;)

PR #304, opened 2013-12-19
Cleanup CRAB3 runtime generation

bbockelm:
This is the first set of improvements of the CRAB3 runtime generation.  With this patch, all dependencies except one are now taken from the build RPM environment instead of external servers.


bbockelm:
@geneguvo @amaltaro - can I get this merged?

amaltaro:
We're working on it. RPMs should be available by tomorrow morning.

amaltaro:
RPMs available in comp.pre. It'll make it to comp rep when a new tag to testbed is deployed. Thanks

PR #313, opened 2014-01-08
Add yaml-cpp

ghost:


smuzaffar:
Please change YAMLCPP_BASE to YAML_CPP_BASE

smuzaffar:
Also please drop 
%define online %(case %{cmsplatf} in (_onl____) echo true;; (_) echo false;; esac)

ghost:
Updated as requested.

PR #320, opened 2014-01-11
Update aso spec

HassenRiahi:
As per subject.


amaltaro:
Hassen, note that we have to apply a few changes on top of your request, otherwise we would not be able to build the RPMs.
One patch against WMCore was created to solve a dependency issue, could you please make the same changes, test and make a pull request against WMCore? When that's done, then we can remove this patch from cmsdist and update the asyncstageout.spec file.
Last but not least, now a RPM package is being created only for the documentation. Thanks

Issue #310, opened 2014-01-07
New patch for Geant4 9.6p02


civanch:
During winter break number of SIM production jobs were lost due to the new Geant4 bug. The fix should be applied for CMSSW_7_0_X, CMSSW_6_2_X, and CMSSW_6_2_X_SLHC.

The patch is very simple and should not affect any CMS object but only fix crashes in production. It can be accessed via url: https://vnivanch.web.cern.ch/vnivanch/geant4-9.6p02-cms01.patch

ktf:
@degano can you prepare the updated geant4.spec ? Thanks.

ktf:
@degano we need the same fix for 62x and 62xslhc

civanch:
Thanx for adding the patch to 7_0_X. For me is working fine. The next step would be addition to 6_2_X and 6_2_XSLHC where the bug was first observed.

ghost:
The patch is merged in all the relevant releases.

PR #327, opened 2014-01-13
Add geant4 patch

ghost:
See also #310


ktf:
Thanks. Can you please also check that there are no changes since the last CMSSW_6_2_X release?

PR #330, opened 2014-01-15
Add 2 patch to ROOT fixing threading issues.

ghost:
See also https://github.com/cms-sw/cmssw/issues/1941#issuecomment-31641060


davidlt:
The patches should be copied to CMSDIST repository.

ktf:
Only if not available in some better place. uploading them will in any case store it to cmssdt. In theory we could even fork our own root.

PR #334, opened 2014-01-15
New DAS version for January upgrade

vkuznet:
New DAS version supports configuration switch for main DBS source, i.e. choose DBS2 or DBS3, as well as explicit choice of DBS instnaces.

### DIFF
--- a/das.spec
+++ b/das.spec
@@ -1,4 +1,4 @@
-### RPM cms das 2.4.2
+### RPM cms das 2.4.4
 ## INITENV +PATH PYTHONPATH %i/$PYTHON_LIB_SITE_PACKAGES
 %define wmcver 0.8.3
 %define webdoc_files %{installroot}/%{pkgrel}/doc/


amaltaro:
For this version, which is the main source of information, DBS2 or DBS3? How one choose between the DBS 2 or 3? How one choose the DBS3 instances? Maybe we should consider already DBS3 as main source, so the HTTP group does not need to play with any configuration in the production servers.

vkuznet:
Alan,
In current DAS version (which is on production server) the main source of DBS is
DBS2. Until DBS2 to DBS3 migration happens it should stay as is. When DBS team
will make the switch they want to change DBS main source from DBS2 to DBS3
because data will no longer be written into global DBS (although they want to
continue writing into DBS2 local instance for analysis jobs, see below).

So far users don't choose DBS, if user places the query, the default DBS
global instance is used and DAS sends request to both DBS2 and DBS3. This was
requested by data-ops to test DBS2 along with DBS2 during preparation phase.

If user choose local DBS instance, DAS sends request to that specific instance.
Since DBS2 is main source those request will routed to DBS2 local instances.

In order for DAS to choose DBS3 as main source and its instances, DAS needs to
load it at start-up time. The current release had hard-coded default DBS2 main
source, which I changed and put as configuration parameter. So, when we'll need
to switch DAS to use DBS3, DAS configuration will require a change and DAS will
need to be restarted. I provided example of such change. By default DAS will
use DBS instances from DBS main source. Currently DBS isntances are DBS2 ones since
DBS main source is DBS2. When DAS will start with DBS3 main source, it will use
its instances, e.g. global/phys01/phys02/phys03, and this instances will show
up on DAS web UI.

But DBS team also requested to support a "mixture" of DBS instances, since
after the switch data-ops will write data into DBS3 and some portion to one
local DBS2 instance. That's why "mixture" may be required to support data
look-up.

The exact day of DBS2/DBS3 migration is constantly changing and today I've been
told it will happen in early Feb. The exact date is still in an air.

Based on this the only choice I had is to make code configurable enough to
support this and I try to minimize impact on HTTP group to deploy various DAS
RPMs due to this transition.

I hope it clarifies your questions. But feel free to post more and/or address
them directly to DBS team.

Valentin.

amaltaro:
Thanks Valentin! However I still have one question. This 2.4.4 version will support DBS change via DAS configuration, but does it work with the current deployment scripts (the ones in production)? I assume if it cannot find those values (main_dbs and dbs_instances) in the DAS config file, then it defaults to DBS2 for the moment. Could you please confirm it?

vkuznet:
Yes this is the case. All DAS options have default values, which are defined in
DASOptions class. If some configuration option is missing it will use default
one. So, for main_dbs it is DBS2 and for dbs_instances is empty list which is
filled from DAS maps. This is what I explicitly test on my VM prior sending
official request.

PR #347, opened 2014-01-20
added data externals for CalibCalorimetry/CaloMiscalibTools/data and Con...

nclopezo:
...figuration/Generator/data


vciulli:
I am not sure how this will change our workflow.
If a new lha file is needed, we will then have to make a cmsdist pull? That would mean changing the external and therefore can not be done with a patch: I don’t think this is a suitable way of working for the production. 

I have added a comment in today’s ORP twiki.

ktf:
`Configuration/Generator/data` (the only bit affected) is unchanged since more than one year, and the update rate of updates nowadays seems to be "every year an handful of files is added". I do not think what you say above is an issue at all. In any case, if a patch is needed with a single file, it can be added ad-hoc in the package and it will be picked up from there. That package alone is 5% of our sources.

@davidlange6

vciulli:
I suspect the rate of changes can depend a lot on the data analysis period (these files are usually needed to simulate BSM decay models). But I am ok with what you propose. I just wanted to be sure this change will not bring any problem. Thanks.

PR #348, opened 2014-01-20
migrate to pythia8 180 plus some cleaning

mkirsano:


ktf:
@nclopezo can you do a test build locally and compare with the current one?

PR #351, opened 2014-01-21
add networkX package

gartung:


ktf:
@nclopezo can you build this?

nclopezo:
Hi, I built it and the external builds without errors.

ktf:
This is lacking the tool file. Can you prepare it please and then we look together whether it does the correct thing? Thanks.

gartung:
This package does not need a tool file. It is not used in a scram build rule.

ktf:
How do you get it in path?

gartung:
The script that calls statics.py calls ../../../external/py2-networkx/1.8.1/etc/profile.d/init.sh

ktf:
Sorry this is not ok. We should have it in path somehow, IMHO. 
@smuzzaffar do you have any better ideas?

nclopezo:
Hi,

Anyway, I created a tool file in this pull request https://github.com/cms-sw/cmsdist/pull/353 In case you decide to use it.

PR #354, opened 2014-01-22
Added @PYTHONV@ to py2-pygithub-toolfile.spec

nclopezo:
@davidlt made me notice that, as it was, if the python version is changed it will not work anymore.


smuzaffar:
please fix the tool name in the tool file, it should be py2-pygithub.

nclopezo:
I did it here: https://github.com/cms-sw/cmsdist/pull/355

smuzaffar:
looks good, thanks

PR #358, opened 2014-01-22
create_das_js renamed to dasdas_create_json_maps (for next release HG1402)

vidma:
this accompanies pull https://github.com/dmwm/DAS/pull/4131


amaltaro:
Hi, we were testing an automated script this week and it seems DAS doesn't build with this PR.
Could you please take a look and test privately building it?

vidma:
this should work for das 2.5.x, like here:
https://github.com/vkuznet/cmsdist/compare/cms-sw:comp...hg1403-das

we'll check this, and will provide a new pull for das spec.

PR #366, opened 2014-01-26
reqmgr, workqueue WMCore 0.9.89 version

ticoann:
reqmgr - change in tag collector url
workqueue - fixes dbs3 reader call when run white list is set.
Please review


amaltaro:
It's in testbed now.

Issue #297, opened 2013-12-12
Request for inclusion of external tool "FastJet Contrib"


jstupak:
Hello, I would like to propose inclusion of "FastJet Contrib" version 1.008 as an external tool in CMSSW_7_0_X.  This is a set of 3rd party extensions to the FastJet package, managed by the authors of FastJet.  It will be used for calculation of Njettiness, etc during jet reconstruction.  More information is available here: http://fastjet.hepforge.org/contrib/

ktf:
Instuctions to build from HN:

``` sh
wget http://fastjet.hepforge.org/contrib/downloads/fjcontrib-1.008.tar.gz
tar -xzvf fjcontrib-1.008.tar.gz
cd fjcontrib-1.008
setenv FASTJET_BASE `scramv1 tool tag fastjet FASTJET_BASE`
./configure --fastjet-config=$FASTJET_BASE/bin/fastjet-config --prefix=$PWD CXXFLAGS="-I$FASTJET_BASE/include -I$FASTJET_BASE/tools"
make
make check
make install
```

@degano can you prepare a spec next monday?

PR #369, opened 2014-01-27
Fastjet contrib

ghost:
Created fastjet-contrib.spec and fastjet-contrib-toolfile.spec.

See also #297.


ktf:
Are you sure about the `<lib name=fastjet-contrib>`? It looked to me as if it had many more. Can you also modify `cmssw-tool-conf.spec`?

ghost:
Modified cmssw-tool-conf and fixed libraries in fastjet-contrib-toolfile.

PR #371, opened 2014-01-28
GCC 4.9 support

davidlt:
The following adds bits needed for GCC pre-4.9.0 support, fixes some already broken URLs. Tested on `slc6_amd64_gcc481` with `CMSSW_7_1_X_2014-01-27-0200` and all RelVals passed. Once in IB, `IB/CMSSW_7_1_X/stable` will be branched to `IB/CMSSW_7_1_X/gcc490` with only `gcc.spec` being different.


ktf:
How about we back-port to stable:
- swig
- oracle-env
- py2-cx-oracle
- autotools
- mcdb

?

I assume you enabled the forward porting of changes in 71X/stable, 
right? If not can you do it?

davidlt:
The changes are on `stable`, now the **only difference** will be kept in `gcc.spec`.

PR #383, opened 2014-02-03
Updated fastjet-contrib.spec to build experimental unique library.

ghost:

PR #393, opened 2014-02-05
Add fastjet-contrib 1.009

ghost:
Fixes cms-sw/cmsdist#383


fastjet-contrib-archive-toolfile.spec
```diff
@@ -0,0 +1,29 @@
+### RPM external fastjet-contrib-archive-toolfile 1.0
```
ktf: No need to create a separate spec. Please put everything in `fastjet-contrib-toolfile.spec`.

PR #401, opened 2014-02-09
New DAS version for cmsweb-testbed

vkuznet:


amaltaro:
I think we are not going to use this PR, since it was meant to fix critical problems seen in production. Valentin, can you confirm please? If so, then please close it.

vkuznet:
Yes, we'll not use it, closing the ticket.

PR #406, opened 2014-02-13
Changed version of Sherpa to 2.0.0, added blackhat 0.9.9(+patch) and qd ...

thuer:
2.3.13 for this purpose


ghost:
Hi, @thuer 
I cannot compile your pull request: it fails building Blackhat.
I suspect it has something to do with the fact that the patch to Blackhat you attached it's a binary file: which is strange.

Can you please check if that file is really what you wanted to submit?

davidlange6:
in any case - the ORP needs to discuss the various new externals being proposed here… sorry I had missed this

vciulli:
Hi @davidlange6 
do you need any specific information ?
Please note that sherpa 1.4.2 is already available in all releases.
While blackhat and qd, which are needed by sherpa2.0, have been tested already in CMSSW_5_3_11_sherpa2beta2

thuer:
Hi Alessandro,

sorry, I don't know what went wrong there, the patch file should be a 
simple text file. I readded it to the git repo and updated the 
branch(now it shows as a normal file)
I tried building blackhat+sherpa and it worked.

Best
Sebastian

Am 19.02.2014 09:43, schrieb Alessandro Degano:

davidlange6:
did we already agree that cmssw should depend on blackhat and qd? Blackhat in particular already looks like it might be undermaintained. do we have a strong motivation for enabling that option of sherpa?

vciulli:
@davidlange6 
sherpa uses some external code to calculate virtual diagrams: blackHat is needed for vector boson + jets, so definitely to make sherpa useful we need it.
In general we are making an effort to include generators in CMSSW rather than running them using scripts (like MadGraph) because that has many drawbacks. But this is a more general discussion, so this is not the right forum indeed (let's discuss it in the ORP or in an Offline management meeting).

davidlange6:
just making sure we really need it, as blackhat is not an absolute requirement for sherpa.  hope that the blackhat authors eventually produce version 1.0 as suggested in their documentation...

PR #423, opened 2014-02-20
Significantly reduced size of das-cache.file 

ghost:
Restricting to maximum 2 files per entry.


ktf:
Thanks. Can you also make sure there is no empty entries (or request data for those which do not have files at `T2_CH_CERN`?

ghost:
@ktf updated, requesting files missing right now.

PR #425, opened 2014-02-21
Add CUDA 5.5 toolfile

davidlt:
CUDA toolfile will be distributed in standard CMSSW distribution. CUDA must be enabled explicitly via `scram setup cuda`. It will make `NVCC` (uses our C++ compiler) available in CMSSW environment and also allow linking to CUDA in CMSSW packages. We might want to split `cuda-toolfile` into several different ones in future.


ktf:
Thanks. Merged this for now, however we need to do something like it's done in the icc tool file with a symlink in %i pointing to the installation and use that symlink rather than hardcode /usr/local/cuda. Can you clean that up? Also can you add CUDACC=nvcc ?

PR #439, opened 2014-02-24
root: add support for C++11 mode for `linuxarm' target

davidlt:
Add missing fragments to `linuxarm' target Makefile to support C++11
mode.

Cleanup unneeded parts in SPEC.

Also filed as ROOT-6110 in Jira.


davidlt:
Merging as trivial as it only touched `linuxarm`.

PR #441, opened 2014-02-25
Add Rivet 2.0 as separate package.

ghost:
Note that it's disabled by default, enable it with "scram setup rivet2".

### DIFF
--- a/rivet2-toolfile.spec
+++ b/rivet2-toolfile.spec
@@ -1,0 +1,29 @@
+### RPM external rivet2-toolfile 1.0
+Requires: rivet2
+
+%prep
+
+%build
+
+%install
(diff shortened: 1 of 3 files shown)


rivet2-toolfile.spec
```diff
@@ -0,0 +1,29 @@
...
+
+%build
+
+%install
+mkdir -p %i/etc/scram.d
+cat << \EOF_TOOLFILE >%i/etc/scram.d/rivet2.xml
+<tool name="rivet2" version="@TOOL_VERSION@">
+<lib name="Rivet"/>
+<client>
+<environment name="RIVET_BASE" default="@TOOL_ROOT@"/>
```
ktf: RIVET2_BASE

PR #444, opened 2014-02-25
rivet2: fix configure cannot find boost headers needed by yaml-cpp

davidlt:
Ritve2 failed on armv7hl:

```
yaml-cpp API version could not be determined
```

yaml-cpp depends on Boost, but configure script executed GCC without
passing -I option for Boost includes. Hack it by adding Boost
include path in CPPFLAGS.

Previously it was taking Boost headers from standard system location and
Boost is not installed on armv7hl machines.


ghost:
Hi David, 
I included your fix to the spec file and re-enabled rivet2 in #450.
I am closing this one to avoid confusion.

PR #445, opened 2014-02-25
New DAS version

vkuznet:


amaltaro:
Hi Valentin, just to confirm, we do not need any deployment PR to go along with this one, right?

vkuznet:
Hi Alan,
yes we don't need to change deployment. I think DAS maps separation will be
postponed during this release, since it's not quite ready.
Valentin.

PR #459, opened 2014-02-28
Test DBS3 with HG1403a

yuyiguo:


amaltaro:
Yuyi, next time update your local copy before making the PR please (cmsdist was pointing to 3.2.2a). Otherwise we cannot automatically patch the code. Thanks

yuyiguo:
Hi, Alan:
I am a bit confused. What is the current tag of cmsdist for this deployment? Where you want the local copy for the PR pointing to?
Thanks,
Yuyi

amaltaro:
The current tag is HG1403a, now we are building HG1403b. In principle you can always get the
head of cmsdist (comp branch) before merging your changes and making a new pull request.
For this PR, it seems you merged your changes against HG1401j

yuyiguo:
OK, I see it. I had something strange last Friday afternoon around 3 pm FNAL time when I was trying to make the PR. The "git pull" command to my upcomp was not response for a long time.  I did "git checkout upstream/comp" to make the local branch. I notice that dbs*.spec in this newly made local area had dbs version 3.2.1c that was from last prod release. I was wondering about this.  I did not know what was going on during that time period.  I made the update to the version change needed, then made the PR.

In principle, we should always use the current tag as the based of the local copy before make a PR, right? Or I should use the HEAD?

amaltaro:
Yes, you should use HEAD for cmsdist/comp PR. Thanks

PR #461, opened 2014-03-03
Geant 4.10 updated to patch01.

ghost:
The patch geant4-10.0-dynamic-TLS.patch could not be applied anymore, however the "-ftls-model=global-dynamic" option is now by default on in Geant4MakeRules_cxx.cmake and Linux-g++.gmk.
It is instead missing for Darwing clang compiler in Darwin-clang.gmk and should be re-added if needed.


davidlt:
GEANT4 10.00 + TLS + CMSSW was never tested by me. We only run the special IBs on production architecture. It's unknown if it works on Darwin.

PR #472, opened 2014-03-05
Updated 2 weighting files.

ghost:
Needed by #2717.

### DIFF
--- a/data-RecoEgamma-ElectronIdentification.spec
+++ b/data-RecoEgamma-ElectronIdentification.spec
@@ -7,8 +7,8 @@
 
 cat << CMS_EOF >> ./sources
 RecoEgamma/ElectronIdentification/data/TMVA_BDTSimpleCat_17Feb2011.weights.xml
-RecoEgamma/ElectronIdentification/data/TMVA_BDTSoftElectrons_9Dec2013.weights.xml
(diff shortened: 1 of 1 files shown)


data-RecoEgamma-ElectronIdentification.spec
```diff
@@ -7,8 +7,8 @@
 
 cat << CMS_EOF >> ./sources
 RecoEgamma/ElectronIdentification/data/TMVA_BDTSimpleCat_17Feb2011.weights.xml
-RecoEgamma/ElectronIdentification/data/TMVA_BDTSoftElectrons_9Dec2013.weights.xml
-RecoEgamma/ElectronIdentification/data/TMVA_BDTSoftElectrons_27Jan2014.weights.xml
+RecoEgamma/ElectronIdentification/data/TMVA_BDTSoftElectrons_7Feb2013.weights.xml
```
slava77, line 10: @lgray
Hi Lindsey, are the names of the files as desired?
I see that the intended download.url in https://github.com/cms-sw/cmssw/pull/2717 adds files.
Here, 7Feb2013 doesn't exist (or should it?). 
Will we ever need the old files in the same release (for testing or what not)?

lgray, line 10: 7Feb2013 is a typo.

PR #476, opened 2014-03-06
Add sherpa2 as separate package.

ghost:
Disabled by default.
Adapted from #406.


thuer:
Do you plan to include both versions or just keep it there to switch between both?
Since both sherpa-toolfile and sherpa2-toolfile share the same libraries only one can be enabled at a time.

ktf:
The plan is to make sherpa2 => sherpa once 
GeneratorInterface/SherpaInterface builds fine.

PR #477, opened 2014-03-06
Add support for OpenCL 1.1 C and C++ API

davidlt:
By default OpenCL is configured for NVIDIA CUDA. <PKG>/include and
<PKG>/lib64 are symlinks, which can me modified to point to
alternative implementation.

`opencl` and `opencl-cpp` are optional toolfiles and must be loaded
explicitly.


ktf:
Again, fine to go after pre4 is built.

PR #485, opened 2014-03-08
Forward port new datafiles for root6.

ktf:


ktf:
@wmtan this should fix your problem in the ROOT6 build. Let me know if you have issues. Will be in next (Sunday 0200) IB.

PR #486, opened 2014-03-08
Move to new version which works with root5.34.17++

ktf:


ktf:
@DR15Jones this should fix the threaded IB.

PR #497, opened 2014-03-11
move to pythia8 183 in 6_2_X

mkirsano:


mkirsano:
Requires changes in GeneratorInterface/Pythia8Interface

Issue #483, opened 2014-03-07
Merge Geant4 10.0p01 on top of CMSSW_7_1_0_pre4


civanch:
According to the plan of Geant4 10.0p01 integration a special build needs to be created on top of CMSSW_7_1_0_pre4, which includes all modifications available at the GEANT10 branch. This build will be used for RelVal validation versus pre4. 

Practically it will be a merge of GEANT10 to 7_1_0_pre4.

For this special build will be optimal to recompile Geant4 10.0p01 in sequential mode, do not apply CMS private patch, and use Geant4 data files from GEANT10 branch.

GEANT10 branch should be left unchanged, it will be used for further development of MT variant of CMSSW simulation.

civanch:
Updated datasets (used  already in GEANT10 branch):
G4NDL.4.4.tar.gz
G4EMLOW.6.35.tar.gz
G4PhotonEvaporation.3.0.tar.gz
G4RadioactiveDecay.4.0.tar.gz
G4NEUTRONXS.1.4.tar.gz
G4ABLA.3.0.tar.gz

civanch:
To build in sequential mode one needs to set OFF for MT,
possible cmake options to build this version:

cmake -DCMAKE_INSTALL_PREFIX=${G4 build dir} \
-DGEANT4_USE_GDML=ON \
-DGEANT4_INSTALL_DATADIR=${pass to G4 data} \
-DXERCESC_ROOT_DIR=${pass to XERCESC} \
-DGEANT4_INSTALL_EXAMPLES=OFF \
-DGEANT4_ENABLE_TESTING=OFF \
-DGEANT4_USE_SYSTEM_CLHEP=OFF \
-DBUILD_SHARED_LIBS=ON \
-DBUILD_STATIC_LIBS=ON \
-DGEANT4_BUILD_MULTITHREADED=OFF \
${G4INSTALL}

civanch:
As we also agree the merge if successful should go into 7_1_X.

ktf:
@degano can you please add Vladimirs changes to geant4.10 to disable 
multithreading so that we can start building a release based on this? 
Thanks.

ktf:
@degano it looks like we need to have:

```
-DG4MULTITHREADED
```

dropped from the geant4-toolfile.spec

ktf:
Vladimir, the current GEANT4_X IB is the candidate for the release. Unless you have objections we can go ahead and build.

davidlange6:
[...]

PR #501, opened 2014-03-12
Fix single threaded Geant4

ktf:


ktf:
See also #483.

civanch:
If gmake is used to build G4 then G4MULITHREADED should not be set or may be dropped. If cmake is used the receipt is as in #483

ktf:
The problem is for CMSSW / scram, not geant4 itself.

PR #502, opened 2014-03-12
blackhat: fix compile on armv7hl; add dependency on openssl

davidlt:
On armv7hl target integers in switch(..) expressions overflow, .e.g,

```
case 0x0F02030908:
case 0x0E02030908:
```

generates:

```
warning: overflow in implicit constant conversion [-Woverflow]
```

and:

```
error: duplicate case value
error: previously used here
```

Change all 'long' to 'long long'. long is 4 bytes on armv7hl, but 8
bytes on x86_64. long long is 8 bytes on both platforms. Thus there
should be no changes on ABI level.

In addition package is using OpenSSL from the standard system location.
Force it to link to OpenSSL from CMSDIST.

### DIFF
--- a/blackhat-0.9.9-armv7hl.patch
+++ b/blackhat-0.9.9-armv7hl.patch
@@ -1,0 +1,1002 @@
+diff --git a/src/trees_eval/A0_1phi_Q_massive.cpp b/src/trees_eval/A0_1phi_Q_massive.cpp
+index a6d95c7..1dfe83b 100755
+--- a/src/trees_eval/A0_1phi_Q_massive.cpp
++++ b/src/trees_eval/A0_1phi_Q_massive.cpp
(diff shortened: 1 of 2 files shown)


ghost:
Hi @davidlt,
I fixed the openssl issue in a previous PR (which is yet to be integrated) that also re-enable sherpa2.
I think the best way to go is that we first merge my PR #496 and then you update yours to only take the arm7 changes.

davidlt:
@degano I am not against it. I will update PR later once you have merged another one. Just ping me.

ktf:
@degano you have `-ldl`, why is that?

ghost:
Just a cut and paste mistake, I'm removing it and merging, so we can move on.

davidlt:
Updated.

PR #506, opened 2014-03-19
Add threadsafety bits for boost-spirit.

ktf:


ktf:
@Dr15Jones this should fix the boost spirit MT issues.

PR #507, opened 2014-03-20
Request to add Photos++ 3.54, EvtGen 1.3.0 and move to Tauola++ 1.1.4 (currently 1.1.3)

inugent:
This request is to add Photos++ 3.54 and EvtGen 1.3.0 to CMSSW and also switch Tauola++ versions from 1.1..3 to Tauola++ 1.1.4 which includes new features including new currents.These spec files have been tested on sl5 and sl6 and have had preliminary validation insides CMSSW with new interfaces in https://github.com/inugent/cmssw/tree/PhotosppEvtGen_v3.   Best regards,
Ian

### DIFF
--- a/evtgen-toolfile.spec
+++ b/evtgen-toolfile.spec
@@ -1,0 +1,27 @@
+### RPM external evtgen-toolfile 1.3.0
+Requires: evtgen
+
+%prep
+
+%build
+
+%install
+
+mkdir -p %i/etc/scram.d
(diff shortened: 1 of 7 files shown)


inugent:
Since no action has been taken this request is being factorized. This is because the CMSSW EvtGen interface is not ready and since there is a request for Tauola++ 1.1.4 to be added in addition to Tauola++ 1.1.3 to allow for validation of the new currents. This request may be reactivated after discussion with the Generator group.

PR #508, opened 2014-03-20
to pythia8 183

mkirsano:


vciulli:
@cmsbuild please include it in the next 62X release

PR #512, opened 2014-03-22
adding tauolapp 1.1.4

inugent:
This request is to add Tauola 1.1.4 into release 6_2_X. The patch file modifies the makefiles to rename the libs. This allows for multiple versions of tauola++ in 1 release.  The full chain has been test from building the rpms to running tauola 1.1.3 and 1.1.4 in the stand release to product the genvalid_dy plots.


ghost:
As I think the correct PR is #513 I'm closing this one.

PR #513, opened 2014-03-23
adding tauola++ 1.1.4

inugent:
This request is to add Tauola 1.1.4 into release 6_2_X. The patch file modifies the makefiles to rename the libs. This allows for multiple versions of tauola++ in 1 release. The full chain has been test from building the rpms to running tauola 1.1.3 and 1.1.4 in the stand release to product the genvalid_dy plots.
Ian

### DIFF
--- a/tauolapp-114-toolfile.spec
+++ b/tauolapp-114-toolfile.spec
@@ -1,0 +1,27 @@
+### RPM external tauolapp-114-toolfile 1.1.4
+Requires: tauolapp-114
+%prep
+
+%build
+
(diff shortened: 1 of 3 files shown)


ghost:
@inugent, why would we need 2 versions of tauola++?
Is it possible to just update the existing one?
Why is it needed in 6_2_X?
Maybe this should be discussed at ORP meeting.

bendavid:
This pull request can be closed, as I don't think the backport is really needed for 62x.

PR #517, opened 2014-03-25
Move to geant4.10patch1

ktf:


ktf:
@davidlange6 @civanch geant4.10 externals should be available in `2014-03-26-0200`, I'll merge the cmssw changes later tonight (to avoid breaking cms-merge-topic for too long).

davidlt:
@ktf I see `-DBUILD_STATIC_LIBS=ON`, but there is no `%define keep_archives true`, thus these static libraries are deleted before packaging in RPM. Without static libraries you cannot compile ParFullCMS. You also have `-DGEANT4_BUILD_MULTITHREADED=OFF`, which means ParFullCMS is sequential. Remove ParFullCMS or actually keep static libraries.

This breaks IBs.

ktf:
Thanks for spotting this.

PR #534, opened 2014-04-01
wmagent 0.9.95pre1

ticoann:
Please create rpm for wmagent which takes WMCore 0.9.95pre1


amaltaro:
@ticoann , the RPM for 0.9.95pre1 is now available in comp.pre only. We'll build it against comp when we make a new testbed deployment. Thanks

PR #539, opened 2014-04-06
adding Photos++ 354

inugent:
Request to add Photos++ 3.54 only to CMSSW. This request will be associated with a pull request for the interface in CMSSW.

### DIFF
--- a/photospp.spec
+++ b/photospp.spec
@@ -1,0 +1,44 @@
+### RPM external photos 3.54
+
+Requires: hepmc
+
+Source: http://service-spi.web.cern.ch/service-spi/external/MCGenerators/distribution/photos++/photos++-3.54-src.tgz
+
+Patch0: photospp-354
+
+%if "%{?cms_cxx:set}" != "set"
+%define cms_cxx c++
+%endif
+
+%if "%{?cms_cxxflags:set}" != "set"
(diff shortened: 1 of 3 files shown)


photospp.spec
```diff
@@ -0,0 +1,44 @@
+### RPM external photos 3.54
```
ghost, line 1: The name of the package needs to be consistent: should be photospp.

photospp.spec
```diff
@@ -0,0 +1,44 @@
...
+
+%if "%{?cms_cxx:set}" != "set"
+%define cms_cxx c++
+%endif
+
+%if "%{?cms_cxxflags:set}" != "set"
+%define cms_cxxflags -std=c++0x -g -O2
+%endif
+
+%define keep_archives true
```
ghost, line 17: Is it really necessary to keep archives?

ghost:
Tested and built. (with some minor tweek)
However it will be discussed in the next ORP whether we should include a new package.

inugent:
Hello,
  The patch file is for a bug reported by the authors (It was provided by the authors of Tauola). We have recently found out that the authors are planning to put forward a new version of Photos fixing the bugs related to gcc 480+. I would request that this pull be put on hold until we hear back from the Photos authors. 
Ian

PR #545, opened 2014-04-09
Add data for CondFormats/JetMETObjects to github repository.

ghost:


slava77:
How long does it usually take to get these changes propagated to the actual distribution?
I was naively expecting the same sync time as done for IBs (1400 and 0200)

ktf:
@degano, you did not add it to the cmssw-tool-conf... I'll try to get it in tonight IB.

PR #567, opened 2014-04-13
move to pythia8 185

mkirsano:


mkirsano:
Tests showed that the change of GeneratorInterface/Pythia8Interface code is also needed: #3320

mkirsano:
The change in CMSSW #3320 is already made. Note: it is compatible with pythia8 183 only in the sense of compilation. Particle guns and may be residualDecay will work incorrectly. So, this change is to be done ASAP to restore correct work of Pythia8Interface.

PR #579, opened 2014-04-16
Changed the sherpa2.spec and toolfile to implement the newest sherpa ver...

thuer:
Update the Sherpa2 toolfile to the new version 2.1.0 of Sherpa
Added a new patchfile for the LHAPDF issue and added fastjet support in the configure step
Keep in mind, that either sherpa or sherpa2 has to be chosen, both do not work at the same time since the libraries & variables have the same name.

### DIFF
--- a/cmssw-tool-conf.spec
+++ b/cmssw-tool-conf.spec
@@ -154,7 +154,7 @@ Requires: igprof-toolfile
 %endif
 %endif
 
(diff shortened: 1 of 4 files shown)


vciulli:
sherpa2 should become the new default
Keeping sherpa has available external is an option, but we could just drop it

ktf:
I would vote for just dropping it, if you do not mind.

davidlange6:
me too..

ghost:
Done in #592

PR #580, opened 2014-04-17
Remove a bunch of root patches.

ktf:
We now track them from the git repository itself.


ktf:
@degano can you check we can actually build everything up to the tool conf with this applied? Thanks.

ghost:
Tested.
Everything builds fine and root still works.
Merging it now.

PR #582, opened 2014-04-17
root: fix compiledata.h CXX; add armv7 support

davidlt:
ROOT build scripts fails to set CXX and MAKESHAREDLIB while generating
compiledata.h header if CXX is set on make command.

Add ARMv7 support in Cintex and configure script. This uses less
invasive implementation then previous.


davidlt:
Compile-testd on slc6_amd64_gcc481.

TODO: We need to add support for OS X.

PR #596, opened 2014-04-24
Suppress warning when building against blackhat.

ghost:


ktf:
Please let's merge this after pre7.

PR #598, opened 2014-04-24
move to pythia8 185

mkirsano:


mkirsano:
Requires changes in CMSSW:  #3487

PR #599, opened 2014-04-24
updating to Tauola 1.1.4

inugent:
Switching Tauola++ 1.1.3 to 1.1.4 to include RCHL models in Tauola++ and TauSpinner and Higgs transverse spin effects for TauSpinner.


ktf:
@degano, please test this but do not include it in before pre7 is out.

ghost:
Tested.
Builds ok, GeneratorInterface/TauolaInterface is compatible.
Pending release of pre7.

PR #612, opened 2014-04-30
Fix boost herwigpp 71 x

ghost:


ktf:
@degano This broke ThePegInterface. How was this tested? @mkirsano can you look into it? Thanks.

mkirsano:
I did not understand - what is broken? The internal test of 
ThePEGInterface works, I checked.

mkirsano:
It is possible that when I tested it configured without Boost the 
system Boost of my computer was used. Anyway, now it works for me, and 
like this it is closer to the GENSER configuration. In GENSER it is also 
configured with fastjet.

ghost:
@mkirsano Don't worry, the problem was due to the timing of the IB.
It's all working now.
This PR in particular was needed to use cmsdist BOOST, but it's unrelated to ThePegInterface.

PR #662, opened 2014-05-07
Update llvm.spec

gartung:
New clang sha


Dr15Jones:
@ktf Giulio, what is the status on this? Right now the static analyzer is reporting lots of false positives since it is ignoring `[[cms::thread_safe]]`.

ktf:
Included in next IB.

PR #663, opened 2014-05-07
Tolhapdf591

mkirsano:


lhapdf.spec
```diff
@@ -66,27 +48,27 @@ CC="`which gcc` -fPIC"
...
             --enable-pyext --disable-octave --disable-doxygen \
             FC="$FC" CXX="$CXX" CC="$CC" \
-            CPPFLAGS="-I ${ZLIB_ROOT}/include" CXXFLAGS="%cms_cxxflags" LDFLAGS="-L${ZLIB_ROOT}/lib -lz"
+            CXXFLAGS="%cms_cxxflags"
 make %{makeprocesses}
 make install
 
+
+mkdir -p %{i}/share/lhapdf/PDFsets
+%{i}/bin/lhapdf-getdata --force --repo=http://www.hepforge.org/archive/lhapdf/pdfsets/%{realversion} --dest=%{i}/share/lhapdf/PDFsets cteq6l ct10 MSTW2008nlo68cl
```
ktf, line 66: What is this doing?

mkirsano, line 66: lhapdf-getdata is a part of lhapdf, it is a python script that 
downloads and installs a few PDF sets. So part of PDF sets will be real, 
not symlinks. This is made for systems where there is no cvmfs.

ktf:
How many files are we talking about? I do not like the idea that the fetching bypasses completely our source caching system. In theory the best thing would be to have them specified as "SourceXX".

Also, if I understand correctly your build procedure fails in case CVMFS is not there, which we will not guarantee for builds. Can you update things so that the list of files to link does not require CVMFS?

mkirsano:
It was agreed to have the sets cteq6_, MRST_, CT10_. It is 79 files, 
about 260 MB.
  I tested this on lxplus. If there is no cvmfs, there is no error, 
simply the links are not created.
  In principle, in most CMS generations only cteq6_ sets are used, this 
is only 6 MB, 3 files. If the list of the minimal real sets is 
reconsidered...

bendavid:
I think Giulio is asking to have a fixed list of symlinks to make as part of the build process, so that even if CVMFS is not available during the build, the symlinks are still correctly made.

mkirsano:
OK, I will prepare such script.

bendavid:
@ktf is this ok now?

ktf:
Looks like it's fine. @degano can you try building this?

ghost:
+1
Builds correctly, also tested rebuild of GeneratorInterface packages.

ktf:
> +1
> Builds correctly, also tested rebuild of GeneratorInterface packages.

Can you try running the matrix as well? How big is the package now?

## 

Ciao,
Giulio

ghost:
The usual matrix tests ran fine.
It's 276M.

ktf:
Sounds good. Please go ahead and merge.

PR #668, opened 2014-05-10
fix the generation of separate debug symbols

fwyzard:


fwyzard:
@davidlt , can you check if this fixes the problem with the debug symbols ?

ktf:
Ping @davidlt.

ktf:
Ping @davidlt.

davidlt:
@fwyzard why do we need `-fdebug-prefix-map`, they point to a path which is not relocatable, we still rely on relative-linkage between debug files and GBD look-up rules.

Do we want to use GNU debuglink or Build IDs? Darwin might require GNU debuglink, but the rest are already using Build IDs (each build binary gets a unique random hash). For Build IDs we need to change the layout of debug files to:

```
CMSSW_7_1_X_2014-05-07-0200/lib/slc6_amd64_gcc490/debug/.build-id/ce
CMSSW_7_1_X_2014-05-07-0200/lib/slc6_amd64_gcc490/debug/.build-id/ce/94e40e4ac2f57ed19c5bdd9b886e2a74b32fb3.debug
CMSSW_7_1_X_2014-05-07-0200/lib/slc6_amd64_gcc490/debug/.build-id/ce/53d02fd74d1c408b306666340ace0b4e48f50c.debug
CMSSW_7_1_X_2014-05-07-0200/lib/slc6_amd64_gcc490/debug/.build-id/ce/7df82ab8bf48bafbd4b8ef1b6ea519002a7c7a.debug
CMSSW_7_1_X_2014-05-07-0200/lib/slc6_amd64_gcc490/debug/.build-id/50
CMSSW_7_1_X_2014-05-07-0200/lib/slc6_amd64_gcc490/debug/.build-id/50/c191750f821030b6e646584ef4c82575289160.debug
CMSSW_7_1_X_2014-05-07-0200/lib/slc6_amd64_gcc490/debug/.build-id/50/2b3c8f46218e7b12e25ff2456da7f94e2e7e85.debug
CMSSW_7_1_X_2014-05-07-0200/lib/slc6_amd64_gcc490/debug/.build-id/50/5a82c674f6c213eb8be1e9ee9b1eaeec64b97f.debug
CMSSW_7_1_X_2014-05-07-0200/lib/slc6_amd64_gcc490/debug/.build-id/50/ad8b319dc01628b48b3011756025c733172f71.debug
CMSSW_7_1_X_2014-05-07-0200/lib/slc6_amd64_gcc490/debug/.build-id/50/6a929b916756792b6895d72a56e5093f82168f.debug
```

I am build testing it now.

fwyzard:
David,
as I've commented earlier, using build id is fine for gdb and tools, but it
makes it much harder to track files for human beings. Using debuglink works
pretty well both for gdb and for people.
What is "the rest" that is using build ids? (offline) we have not bee
shipping debugging symbols anywhere since quite some time; online it was
setup to use debuglink in 2011/2012.

Regarding -fprefix-map, it works together with the gdb's substitute-path :
at build time we make the source links point to a standard location (which
is actually the one used by the RPMs if they are installed without
relocation); at run time, substitute-path translates that into the actual
location of the source code.
I've had a look into the possibility of relocating the links in the debug
information to point to the actual source in a post-install, but it seems
additional work which we don't really need to do.

.A

ktf:
I'll merge this for now. @davidlt please do a separate pull request if you have a better solution.

scram-project-build.file
```diff
@@ -283,33 +279,32 @@ if [ "%{n}" == "coral" ]; then
...
+    dwz -m .debug/bin-common.debug -r $ELF_BINS
+    echo "$ELF_BINS" | xargs -I% sh -c 'objcopy --compress-debug-sections --only-keep-debug % .debug/%.debug %; objcopy --strip-debug --add-gnu-debuglink=.debug/%.debug % %'
   fi
 popd
 
 pushd $BASE_TEST_PATH
+  mkdir -p .debug
   # ELF binaries
-  ELF_BINS=$(find . -type f -exec file {} \; | grep -i elf | cut -d':' -f1)
+  ELF_BINS=$(ls | xargs -I% file % | grep ELF | cut -d':' -f1)
```
ktf, line 304: Why didn't you like the find part? Looks better to me.

fwyzard, line 304: I think that `find` risks finding binary files in a subdirectory, which would break the objcopy part, as we would not have the corresponding subdirectory under .debug

davidlt, line 304: `find` is more reliable. `ls` would be fine if it would produce null-terminated strings.
I was doing some small testings and you can find your initial patch + some my modifications: `~davidlt/public/debug_ib.patch`

Mainly, I would call a common part as `common.debug` in each `.debug` directory. I would call `xargs` as `xargs -t -n1 --max-procs=%{compiling_processes}` (that will print the command to be executed and run in concurrent way ==> speedup).

I also had problems running `--compress-debug-sections` and `--only-keep-debug` on the same `objcopy` invocation, thus I split those in two invocations.

fwyzard, line 304: On 27 May 2014 13:17, davidlt notifications@github.com wrote:

> In scram-project-build.file:
> 
> >    # ELF binaries
> > -  ELF_BINS=$(find . -type f -exec file {} \; | grep -i elf | cut -d':' -f1)
> > -  ELF_BINS=$(ls | xargs -I% file % | grep ELF | cut -d':' -f1)
> 
> find is more reliable. ls would be fine if it would produce
> null-terminated strings.
> I was doing some small testings and you can find your initial patch + some
> my modifications: ~davidlt/public/debug_ib.patch

I'll have a look.

Actually, using "file *" would work, too :-)

> Mainly, I would call a common part as common.debug in each .debugdirectory.
> 
> Sounds good.

I would call xargs as xargs -t -n1 --max-procs=%{compiling_processes} (that

> will print the command to be executed and run in concurrent way ==>
> speedup).
> 
> Sounds good.
> 
> I also had problems running --compress-debug-sections and
> --only-keep-debug on the same objcopy invocation, thus I split those in
> two invocations.

Mhm, that has been working fine in the last few IBs. Do you remember what
problems you were running into ?

.A

davidlt, line 304: `file *` again is unsafe. It depends on kernel buffer size and files count. It might break if we have too many files.
Your original patch was throwing `objcopy` usage documentation. I think, splitting the command helped in addition to removing unneeded `%`. As you said it works now, thus no modification needed.

fwyzard, line 304: On 27 May 2014 15:12, davidlt notifications@github.com wrote:

> Your original patch was throwing objcopy usage documentation.
> 
> Yes, the issue was an extra "%" at the end of the line (no idea why it got
> there).
> 
> I think, splitting the command helped in addition to removing unneeded %.
> As you said it works now, thus no modification needed.
> 
> OK, I'm looking at your other changes.

.A

fwyzard, line 304: On 27 May 2014 15:12, davidlt notifications@github.com wrote:

> file \* again is unsafe. It depends on kernel buffer size and files count.
> It might break if we have too many files.

True - but the number of files under $CMSSW_RELEASE_BASE/lib/$SCRAM_ARCH is
safe enough, I need to run over 20 times as many in a single command to get
"Argument list too long" (19 times as many is still good :-) ).

Anyway, if we use the same name for the common symbols in each directory I
can simplify the logic a bit, and go back to "find".

.A

PR #694, opened 2014-05-14
Lhapdf591 fix1

mkirsano:


mkirsano:
This fix is needed to make .../full/bin/lhapdf-config correctly work

PR #715, opened 2014-05-20
Drop fastjet patches

ktf:


ktf:
New fastjet build seems to work fine. Dropping these.

PR #716, opened 2014-05-20
Fix python relocation

ktf:


ktf:
Not sure how long this has been there, but rebuilding python on a VM which did not have python-devel installed caused a problem.

PR #724, opened 2014-05-23
adding runtime var to mctester-toolfile

inugent:
This modification adds a runtime variable that enables the mctester analysis stage to be run in the default CMSSW environment.

### DIFF
--- a/mctester-toolfile.spec
+++ b/mctester-toolfile.spec
@@ -17,6 +17,7 @@ cat << \EOF_TOOLFILE >%i/etc/scram.d/mctester.xml
     <environment name="LIBDIR" default="$MCTESTER_BASE/lib"/>
     <environment name="INCLUDE" default="$MCTESTER_BASE/include"/>
   </client>
+  <runtime name="MCTESTERLOCATION" value="$MCTESTER_BASE/lib/"/>
   <use name="root"/>
   <use name="HepMC"/>
 </tool>


ktf:
Why not simply using LD_LIBRARY_PATH?

smuzaffar:
scram does not add tools' lib directories to LD_LIBRARY_PATH.

ktf:
> scram does not add tools' lib directories to LD_LIBRARY_PATH.

True, but the contents of that directory are in the global 
LD_LIBRARY_PATH. @inugent can you elaborate a bit more what you are 
trying to do here?

davidlt:
It's in `LD_LIBRARY_PATH` by default:

```
$ ls /afs/cern.ch/cms/sw/ReleaseCandidates/vol1/slc6_amd64_gcc481/cms/cmssw/CMSSW_7_1_X_2014-05-23-0200/e
xternal/slc6_amd64_gcc481/lib | grep -E 'HEPEvent|HepMCEvent|MCTester'
libHEPEvent.a
libHEPEvent.so
libHepMCEvent.a
libHepMCEvent.so
libMCTester.a
libMCTester.so
```

@inugent do you plan to use `MCTESTERLOCATION` to find them and then `dlopen()`?

inugent:
Hello,
 MC-Tester runs in two stages. One  is run as a plugin, while the other is run in root, but requires the libraries from the external rpm (ie https://github.com/inugent/cmssw/blob/Tauola114_7_1_X/GeneratorInterface/MCTESTER/test/analyze/MERGE.C or ). The idea was to setup a variable such that the correct libraries will automatically be load in the CMS environment. Otherwise the user has to add the path by hand or use a hack. Best regards,
Ian

smuzaffar:
In CMS environment libHEPEvent.so (and other libs) should already be 
available under LD_LIBRARY_PATH. So just doing 
gSystem.Load("libHEPEvent.so"); should work ... is it not working in CMS 
environment?

ktf:
Which library are we talking about? AFAICT all the mctester libraries 
are in LD_LIBRARY_PATH so you can simply pick them up from there.

inugent:
You are right. It will be automatically found from the LD_LIBRARY_PATH. This request is not needed. I have modified the scripts in GeneratorInterface/MCTESTER accordingly. I am closing this request. Thank you and best regards,
Ian

PR #726, opened 2014-05-24
scram-project-build.file: fix processing of debug symbols

fwyzard:
- fix objcopy syntax for debug symbols
- check that a directory exists before processing it for debug symbols


fwyzard:
CORAL builds successfully with this patch. I'm not able to test CMSSW on an online machine (cannot access GitHub over https), and I don't have any build machines at hand.

PR #728, opened 2014-05-26
CRAB3 rc9 (server) and rc4 (client) releases

mmascher:


geneguvo:
-1 to e98950c23cb4aa6fc48bde2415bdeb45b999498b on slc5_amd64_gcc461.

mmascher:
Should I worry about the build failure reported here?

geneguvo:
Usually yes, but not this specific one.

I'm investigating an environment issue which is making the build-agent build of the bigcouch rpm to fail. Since we had to push the changes to bigcouch.spec before releasing the RPMs to comp.pre, the build test for your pull request is triggering the bigcouch build together.

PR #739, opened 2014-05-28
New spec/patches for CouchDB 1.5

vkuznet:
Please take separate patches and spec file for couchdb 1.5 which we can use independently from default couchdb. This will allow to test CDB1.5 within wmagent environment.

### DIFF
--- a/autotools.spec
+++ b/autotools.spec
@@ -1,6 +1,6 @@
 ### RPM external autotools 1.0
 # We keep all of them together to simplify the "requires" statements.
-%define autoconf_version 2.68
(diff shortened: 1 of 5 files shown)


geneguvo:
+1 to a1d606380e139e843d93b629ee08f64edb6c62ab on slc5_amd64_gcc461.

ticoann:
@geneguvo, Could you review the code? This doesn't have to be commissioned for cmsweb application. For now, this will be only used for agent. I tested the agent with couch1.5 rpm and made the necessary changes in agent part.

PR #741, opened 2014-05-29
DBS_3_2_4a

yuyiguo:
This PA include:
1. DBS client update: \* user configurable  ssl_verifypeer when creating DBS client object.
                                   \* support long list of block for block summary API.
2. Add new EventsPerDay script for Dataops.
3. Bug fix for DBS3 server when auto generate sql with input as a list.


geneguvo:
+1 to a3be85e92daed372cfcd5823311ca88239e60e7e on slc5_amd64_gcc461.

PR #751, opened 2014-06-03
Lhapdf6

mkirsano:


mkirsano:
This is the attempt of the first integration of lhapdf6. In order to use it in CMSSW one has to substitute lhapdf by lhapdf6 in the file(s) BuildFile. For example, if one wants that pythia8 uses lhapdf6, the file to change is GeneratorInterface/Core/BuildFile.

PR #779, opened 2014-06-06
Upgrade HTCondor version to 8.1.6

bbockelm:
This allows us to drop backported patches and pickup a few new method calls in the API (necessary for supporting an 8.1.6 schedd)

### DIFF
--- a/cms-htcondor-build.patch
+++ b/cms-htcondor-build.patch
@@ -20,16 +20,18 @@ index 99b0584..4f2a45f 100644
  		if ( "${CONDOR_PLATFORM}" STREQUAL "x86_64_Ubuntu12")
  			set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,--no-as-needed")
  		endif()
---- a/build/cmake/CondorPackageConfig.cmake.orig	2013-11-07 23:54:26.000000000 +0100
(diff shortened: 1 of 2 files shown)


bbockelm:
What needs to be done to make sure the next build of CRABServer picks this up?

Thanks,

geneguvo:
+1 to 83785ab044e0809c84f66bd6b49eca8ac8fe1c22 on slc5_amd64_gcc461.

geneguvo:
Hi Brian,

You meant to not only get RPMs for it, but also have it deployed on cmsweb servers together with Crab, right?

At every ~3 months or so, we announce the release schedule and corresponding tickets for requests  for the next releases (see the message sent today: https://hypernews.cern.ch/HyperNews/CMS/get/webInterfaces/1225.html). You'd usually point us the pull requests (to cmsdist and/or deployment) you need to include, with a short summary of what is changing. If you only have one pull request and it is self-explanatory or we've been made aware of this some other way, it is ok too. However, it is not always the case we are 100% we should pick it or wait..

We've agreed on the release planning meeting last friday to have an special update to the HG1406 release to be pushed towards the end of this month. So my question is: shall we include it in this special release or just pick it for the next release cycle targeting the production deployment in July?

Cheers,
Diego.

bbockelm:
Hi Diego,

This PR is just for the HTCondor build; we'll need it for dev servers.

As with previous months, Marco will handle the corresponding update to CRAB*.  If Marco wants this for HG1406 cmsweb update, he'll let you know in his corresponding ticket.

Brian

PR #799, opened 2014-06-18
Cmssw 7 2 x sherpa 2 1 1

pmillet:
Update to sherpa 2.1.1
Fixes needed for interface to work properly

### DIFF
--- a/sherpa-toolfile.spec
+++ b/sherpa-toolfile.spec
@@ -9,101 +9,9 @@ Requires: sherpa
 mkdir -p %i/etc/scram.d
 cat << \EOF_TOOLFILE >%i/etc/scram.d/sherpa.xml
 <tool name="sherpa" version="@TOOL_VERSION@">
-<lib name="AhadicDecays"/>
-<lib name="AhadicFormation"/>
-<lib name="AhadicMain"/>
-<lib name="AhadicTools"/>
-<lib name="AmegicCluster"/>
-<lib name="AmegicPSGen"/>
-<lib name="Amegic"/>
-<lib name="AmisicModel"/>
-<lib name="Amisic"/>
-<lib name="AmisicTools"/>
-<lib name="Amplitude"/>
-<lib name="Beam"/>
-<lib name="ComixAmplitude"/>
-<lib name="ComixCluster"/>
-<lib name="ComixPhasespace"/>
-<lib name="Comix"/>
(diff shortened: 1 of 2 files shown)


sherpa-toolfile.spec
```diff
@@ -113,11 +21,13 @@ cat << \EOF_TOOLFILE >%i/etc/scram.d/sherpa.xml
 <runtime name="CMSSW_FWLITE_INCLUDE_PATH" value="$SHERPA_BASE/include" type="path"/>
 <runtime name="SHERPA_SHARE_PATH" value="$SHERPA_BASE/share/SHERPA-MC" type="path"/>
 <runtime name="SHERPA_INCLUDE_PATH" value="$SHERPA_BASE/include/SHERPA-MC" type="path"/>
+<runtime name="SHERPA_LIBRARY_PATH" value="$SHERPA_BASE/lib/SHERPA-MC" type="path"/>
```
ktf, line 24: Do you really need this? Why not picking it up from LD_LIBRARY_PATH?

sherpa-toolfile.spec
```diff
@@ -9,101 +9,9 @@ Requires: sherpa
 mkdir -p %i/etc/scram.d
 cat << \EOF_TOOLFILE >%i/etc/scram.d/sherpa.xml
 <tool name="sherpa" version="@TOOL_VERSION@">
-<lib name="AhadicDecays"/>
```
ktf, line 12: Out of curiosity, what happened to those?

sherpa-toolfile.spec
pmillet, line 24: Sherpa dynamically loads libraries which are needed at runtime and  looks for them in the location specified in SHERPA_LIBRARY_PATH. Previously we used all libraries in the compilation, thus this variable was not needed.
However, this approach does not work anymore and the authors suggest to only use the SherpaMain library for compilation since (nearly) all needed function are  available from there. The two other remaining libraries (ToolsMath, ToolsOrg) are needed for the external RNG.

sherpa-toolfile.spec
pmillet, line 12: see comment to line 115

ghost:
I have tested it building cmssw-tool-conf. Everything seems to compile fine.
Also tested compilation of GeneratorInterface/SherpaInterface and that's also fine.

PR #802, opened 2014-06-19
CRAB3 3.3.7.rc1 versions

mmascher:


geneguvo:
+1 to cd9a01e95689db111b6bf43a7321b78a71b21ce2 on slc5_amd64_gcc461.

amaltaro:
Marco, I don't want to be picky. But it's always good to provide a small description for the cmsdist and deployment PRs. Specially this one which is still for June cycle (critical fixes). Thanks

amaltaro:
BTW, I noticed only now that we have an update for crab cache as well. Do we need to push a new crab cache before CSA14 starts or you did it just to keep all crab services in sync?

mmascher:
Hi Alan,

the pull include two missing features for CSA14, support for /store/group stageout and DBS based crab report. I commented on the HG1406 ticket but not here, sorry for that.

About the crabcache I change the version to keep the services in sync, but there is no reason why the last deployed version would not work. I don't actually think there were updates since 3.3.6.rc11..

amaltaro:
Thanks Marco! I won't merge the crabcache change then. For the record, these changes are going to comp_hg1406 branch. If you agree, I'll also pick these changes (including the crabcache one) and push then to the comp branch for the upcoming July release. Unless you'll have new material by next Monday?

mmascher:
I don't think there will be changes for Monday (unless we discover critical bugs during the week end, bu hopefully we will not..). I can include the monitor API you requested if you want, I worked on it yesterday afternoon.

amaltaro:
The monitor API would be great! What about the following: you make a new PR by Monday evening based on comp HG1407b tag (or HEAD). Then I merge this one only in the special branch and we go with your new PR for July's cycle. Is that ok for you?

mmascher:
Sounds good!

amaltaro:
I noticed you made this PR based in comp HEAD instead of the latest HG1406 tag. I've just deployed testbed and crabserver is crashing, my wild guess is that it requires a new condor, as in #779 . Can you add me in gtalk or come over to my office please so we can debug this issue. Thanks

amaltaro:
Merged in comp_hg1406. Don't forget to create a new one to the main comp branch (July's cycle). Thanks

PR #822, opened 2014-07-02
lhapdf: disable $(srcdir) in TESTS

davidlt:
This is not supported by autotools. Only affects LHAPDF version 5.

### DIFF
--- a/lhapdf-5.9.1-tests-no-srcdir.patch
+++ b/lhapdf-5.9.1-tests-no-srcdir.patch
@@ -1,0 +1,10 @@
+diff --git a/5.8.5/octave/Makefile.am b/5.8.5/octave/Makefile.am
+index 7ef2eac..079c5d5 100644
+--- a/5.8.5/octave/Makefile.am
++++ b/5.8.5/octave/Makefile.am
+@@ -6,4 +6,4 @@ lhapdf.oct: wrapoctave.cc
+ 	$(MKOCTFILE) -I$(top_srcdir)/include -v -g -o lhapdf.oct wrapoctave.cc -L$(top_builddir)/lib -L$(top_builddir)/lib/.libs -lLHAPDF
+ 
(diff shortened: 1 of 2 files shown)


ktf:
too late. I've already fixed this. do not merge.

lhapdf.spec
```diff
@@ -33,11 +35,7 @@ rm ./pyext/lhapdf_wrap.cc
 %build
 # We do everything in install because we need to do it twice.
 %install
-libtoolize --force --copy
-autoupdate
-aclocal -I m4
-autoconf
-automake --add-missing
+autoreconf -fiv
```
ktf, line 38: This one I actually like better.

lhapdf-5.9.1-tests-no-srcdir.patch
```diff
@@ -0,0 +1,10 @@
+diff --git a/5.8.5/octave/Makefile.am b/5.8.5/octave/Makefile.am
+index 7ef2eac..079c5d5 100644
+--- a/5.8.5/octave/Makefile.am
++++ b/5.8.5/octave/Makefile.am
+@@ -6,4 +6,4 @@ lhapdf.oct: wrapoctave.cc
+ 	$(MKOCTFILE) -I$(top_srcdir)/include -v -g -o lhapdf.oct wrapoctave.cc -L$(top_builddir)/lib -L$(top_builddir)/lib/.libs -lLHAPDF
+ 
+ TESTS_ENVIRONMENT = LHAPATH=$(top_srcdir)/PDFsets $(OCTAVE)
+-TESTS = $(srcdir)/lhapdf-octave-example1.m
++TESTS = lhapdf-octave-example1.m
```
ktf, line 10: And also this one is probably better than my fix (which removes octave completely). However I'll implement it in the cms-externals/lhapdf repository rather than as a patch.

PR #825, opened 2014-07-03
Push crabserver 3.3.7.rc6

mmascher:
We have been asked to change how we validate the outlfn parameter and do not check that users write in they space (e.g.: we were enforcing `/store/user/mmascher` but sites like LPC have group space under `/store/user`).

The discussion might not be over (https://github.com/dmwm/CRABServer/issues/4344), but I guess for CSA14 we need to loose this up.


geneguvo:
+1 to 215d4ddb46ae8cd7ff2c6cc4cad020bf7af83740 on slc5_amd64_gcc461.

PR #831, opened 2014-07-04
Geant4: updated to version 10.0.p02 

nclopezo:
Removed geant4-10.0.p01-dynamic-tls.patch which was not needed anymore.


nclopezo:
I tested it building cmssw-tool-conf

ktf:
Great, thanks. Can you prepare a back port to 71X as well (we need to wait for ORP before merging, though).

PR #836, opened 2014-07-07
Rename LHAPDF6 libraries and python modules.

ktf:
This is needed to have lhapdf5 and lhapdf6 coexisting in one given scram environment.


ktf:
@bendavid @vciulli

PR #839, opened 2014-07-08
Post-install step for ELF patching on SCRAM projects

davidlt:
The following adds `patchelf` new external package, modifies `glibc` to match SLC6 in terms on source code and compilation flags, modifies LDFLAGS for C/C++ to support custom dynamic loader/interpreter.

Enables modifying dynamic loader on CMSSW. ELF objects are modified in `bin/$SCRAM_ARCH` and `test/$SCRAM_ARCH`, though we don't ship tests with CMSSW by default.

We might also want to remove `glibc-toolfile.spec` it was mainly needed for bash wrappers.


smuzaffar:
Please add the following in cmssw-patch.spec too.
%define changeDynLoader yes

smuzaffar:
@davidlt,iIf we only do the changeDynLoader stuff for slc6_amd64 then shouldn't we only add the following for slc6_amd64?
Requires: patchelf
Requires: glibc
Requires: file

PR #843, opened 2014-07-09
scram-project-build: use $CMS_INSTALL_PREFIX for glibc path

davidlt:
`$RPM_INSTALL_PREFIX` is not the final destination for CMSSW, we
might install _CMSSW_ locally and then move to AFS. `$CMS_INSTALL_PREFIX`
holds the final destination for software.


bbockelm:
Does this make it possible to do runtime relocations of the software?

davidlt:
This is intended for post-install step of RPM. We install CMSSW and then we change interpreter (dynamic loader) for ELF executable files to a different one (in CMSDIST). I am not sure what you mean by "run-time relocation of software".

bbockelm:
Ok, nevermind - I just thought you had meant something else by the original message.

PR #851, opened 2014-07-13
make lhapdf6 inactive and remove library renaming

mkirsano:


mkirsano:
To be checked that generators use lhapdf, not lhapdf6 (although it worked like this with rivet2)

lhapdf6.spec
```diff
@@ -10,7 +10,6 @@ Source3: http://www.hepforge.org/archive/lhapdf/pdfsets/6.1/CT10.tar.gz
 Source4: http://www.hepforge.org/archive/lhapdf/pdfsets/6.1/MSTW2008nlo68cl.tar.gz
 
 Requires: boost yaml-cpp python cython
-BuildRequires: autotools
```
ktf, line 13: Why did you remove autotools?

mkirsano, line 13: I removed autotools because they are not used in the build and also because after the upgrade of autotools in cmsdist in June I don't manage to build anything depending on it.

ktf, line 13: Ok, but why did you need it in the first place?

mkirsano, line 13: Historical: it is used for lhapdf 5 series, came from there

PR #866, opened 2014-07-16
Add support for a custom built dynamic loader

davidlt:
- Add _PatchELF_ utility for patching ELF binaries in post-install
  step.
- Do ELF relocation for `bin/$SCRAM_ARCH` and `test/$SCRAM_ARCH` using
  `CMS_INSTALL_PREFIX`.
- Relocation is only done for slc6_amd64_\* (current production
  architecture for CMSSW_7_2_X).

This makes `IB/CMSSW_7_2_X/stable` and `IB/CMSSW_7_2_X/next` identical. 


ktf:
This will not work on mac. Can you please add the relevant checks?

smuzaffar:
@davidlt , can you put all the new
Requires: xxx

in %islinux?

smuzaffar:
also please fix these:
- you do not need LDFLAGS in gcc-ccompiler.xml. Having it in gcc-cxxcompiler is enough.
- In cmssw-tool-conf.spec we have multiple Requires: glibc-toolfile (one within %islinux and one outside)

davidlt:
Updated. That should do it.

smuzaffar:
Looks much better now. Just a minor fix 
- cmssw-tool-conf.spec: glibc-toolfile dependency is mentioned twice. Please clean it up and may be put it in %islinux %isamd64 (just to make it consistent with rest of the changes)

By the way what will happen if changeDynLoader is not set for a project? gcc-toolfile will set LDFLAGS=-Wl,-dynamic-linker,@GLIBC_ROOT@/lib64/ld.so , so if changeDynLoader is not set this mean there will be no relocation at install time .... right?

davidlt:
Must be blind. Cleaned `cmssw-tool-conf.spec`.

You are correct `changeDynLoader` activates post-install relocation for SCRAM based project. We could replace `changeDynLoader` with some file in CMSDIST. We could read the content of this file (yes/no) in `cmssw-tool-conf.spec`, `gcc-toolfile.spec`, `cmssw.spec` and maybe `coral.spec`. This would complete allow to control it ON/OFF across multiple SPECs.

PR #873, opened 2014-07-18
Fix a numer of issues with glibc post-install step

davidlt:
- Only add dependencies on file, glibc and patchelf if we are building
  CMSSW;
- Only execute post-install relocation for CMSSW. There are no binaries
  in CORAL;
- Remove GLIBC_ROOT environment variable. It was used for bash wrappers
  around cmsRun*. Not needed anymore;
- Do not source init.sh for file and patchelf as we install in
  /afs/.cern.ch (write/read) and then relocate to /afs/cern.ch
  (read-only). The paths in init.sh script are relocated, but files are
  not yet present on /afs/cern.ch. Use binaries directly by full path.
- If grep fails (does not find) the post-install scriplet fails. Ignore
  if grep finds nothing.

**Not tested**

### DIFF
--- a/scram-project-build.file
+++ b/scram-project-build.file
@@ -10,7 +10,7 @@ Requires: SCRAMV1
 
 %if %islinux
 %if %isamd64
-%if "%{?n}" == "cmssw"
+%if "%{?n}" != "coral"
 Requires: patchelf
 Requires: glibc
 Requires: file
@@ -329,7 +329,7 @@ done
 %{?PartialReleaseFilesRelocate:%PartialReleaseFilesRelocate}
 [ -f $RPM_INSTALL_PREFIX/etc/scramrc/%{pkgname}.map ] || (mkdir -p $RPM_INSTALL_PREFIX/etc/scramrc && echo '%{ucprojtype}=$SCRAM_ARCH/%{pkgcategory}/%{pkgname}/%{ucprojtype}_*' > $RPM_INSTALL_PREFIX/etc/scramrc/%{pkgname}.map)
 
-%if "%{?n}" == "cmssw"
+%if "%{?n}" != "coral"
 case %cmsplatf in
   slc6_amd64_*)
     FILE_PKG=$(echo "%{directpkgreqs}" | tr ' ' '\n' | grep 'external/file/')
--- a/scram-project-build.file
+++ b/scram-project-build.file
@@ -10,9 +10,11 @@ Requires: SCRAMV1
 
 %if %islinux
 %if %isamd64
+%if "%{?n}" == "cmssw"
 Requires: patchelf
 Requires: glibc
 Requires: file
+%endif # cmssw
 %endif # isamd64
 %endif # islinux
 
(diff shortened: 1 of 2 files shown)


smuzaffar:
why only cmssw? I thought we said for all non-coral projects?

davidlt:
Same result. There are only two SCRAM projects and we only intend to use it for CMSSW.

smuzaffar:
we also have cmssw-patch (and who knows if we have to build fwlite releases too)

scram-project-build.file
```diff
@@ -327,20 +329,20 @@ done
...
 
+%if "%{?n}" == "cmssw"
 case %cmsplatf in
   slc6_amd64_*)
     FILE_PKG=$(echo "%{directpkgreqs}" | tr ' ' '\n' | grep 'external/file/')
     FILE_PATH=$RPM_INSTALL_PREFIX/%cmsplatf/$FILE_PKG
     PATCHELF_PKG=$(echo "%{directpkgreqs}" | tr ' ' '\n' | grep 'external/patchelf/')
     PATCHELF_PATH=$RPM_INSTALL_PREFIX/%cmsplatf/$PATCHELF_PKG
-    . $PATCHELF_PATH/etc/profile.d/init.sh
-    . $FILE_PATH/etc/profile.d/init.sh
```
ktf, line 337: Mmm… actually this is not ok, IIRC. file has a magic file which will not be there (or the system one will be used). Can't we just use the system file? We do not care about this not returning the same architecture on linux and macos, since we grep for ELF.

davidlt:
@smuzaffar done. Forgot about it by the time I got to office.

scram-project-build.file
davidlt, line 337: That's why I need file. The magic file is in bootstrap, but not the `file` binary. Magic files we distribute are not compatible with system `file` binary and it cannot read magic file, and produces tons of warnings.

ktf, line 337: Ok. sounds good.

smuzaffar:
Please do check if %{?n} works? If I remember correctly we do some post procssing  in cmsBuild e.g.
%n -> %{pkgname}
%{n} -> %{pkgname}

so I am not sure if we do any processing for %{?n}.

ktf:
correct %{?n} will not work.

davidlt:
Testing the whole thing (incl. `%{pkgname}`) with CORAL now.

davidlt:
CORAL built clean. No patchelf, glibc, and file dependencies on RPM. No ELF relocation fragment in post-install scriplet. Looks good to me.

smuzaffar:
sounds good

davidlt:
Only thing left is to move _glibc_ sources to _cms-externals_. We could probably re-do it based on official _glibc_ repository. Current base in my repository is 2.12 + 150 patches from Red Hat.

ktf:
Mmm… I would say we simply clone your repo and continue from there. If we want to do a full mirror we can always prefix it with "upstream/".

PR #877, opened 2014-07-21
PhEDEx web updates for 2014-07

TonyWildish:
This brings in a few small fixes for things like default DBS urls, support URLs etc.


geneguvo:
+1 to b8979c569568b7347c6eba848a1b4502e049779f on slc5_amd64_gcc461.

PR #884, opened 2014-07-23
First release candidate of CRABServer 3.3.8 series

mmascher:
As usual listing only the REST changes here:

-added a first prototype of the monitor for operator (needed to modify the RPM packaging)
-added RESTTask API used by the monitor
-add a per-user throttle as a mechanism to manage excess load
-use always update_after in couch queries
-detect if collector is down and print a proper message
-fix handling of default values of ignorelocality and savelogs flags


geneguvo:
+1 to ff6cf7c3d898dd44bb612c6603dbd36aff3323db on slc5_amd64_gcc461.

PR #885, opened 2014-07-23
dbs3 tag 3.2.5 and WMCore 0.9.96

yuyiguo:
This release include: 
1. Let user specify if only valid files counted when call listFileSummaries API.
2. fixed a DBS client bug that prevent update file status when the file list is very long.
3. fixed bug in DBS2SetFileStatus.py script.
4. Extend DBSUser validation in Lexicon.py based user reported DNs.


geneguvo:
+1 to a8fc7ee7ecd6ee2b90991c27e556e9ba0a6ba4f2 on slc5_amd64_gcc461.

PR #891, opened 2014-07-23
First import of fasthadd spec file

ktf:
Compared to @deguio version in #883 this has the following fixes:
- Actually compiles fine.
- init.sh / init.csh files provided. Source those to have a
  consistent, standalone, environment.
- Include in cmssw-tool-conf to make sure this gets built with
  every IB (can still be deployed separately).


deguio:
thanks!

ktf:
For the record. I uploaded a first version to the repository, so if you ask the DAQ people to add:

```
package {"external+fasthadd+1.0":
  ensure => present,
  provider => cmsdist
}
```

to their puppet configuration they should know what to do.

PR #893, opened 2014-07-24
Backports from 72X.

ktf:
- Use own loader to prevent TLS issues.
- Move config to the one which support big plugins. Disable as there are
  no big plugins in 71X.
- Cleanup gcc / llvm toolfiles.
- Advance dcap to latest version.
- Disable a few things (documentation / rpath) in xz. Not really needed
  but since we build root due to dcap in any case.


ktf:
Tested locally up to cmssw-tool-conf.

PR #897, opened 2014-07-25
add enable-pythia to the sherpa configure arguments

mkirsano:


bendavid:
+1

ghost:
+1
Tested the compilation itself and the compilation of GeneratorInterface/SherpaInterface against it as well.

ghost:
@mkirsano can you please add the explicit dependency of pythia in the requirements?
That way in case it gets dropped in some part of the dependency chain we are sure it's there.
Also please bring this up next ORP (02/08).

davidlange6:
@mkirsano - so then how are the potentially conflicting pythia6 versions handled between the internals of sherpa and CMSSW (e.g., the SherpaInterface package)?

mkirsano:
I don't know how this is handled. Here is the information that I have:
1. I checked that Pythia6Interface works.
2. sherpa installed in GENSER is compiled with this option, and it is 
   used by ATLAS, in the software that uses also pythia6.
   
   If there are doubts, I think we can ask the sherpa authors about 
   handling this.

bendavid:
Since we do not have a better solution or understanding about possible symbol conflicts with the default pythia 6 library, suggest closing this for now.

PR #904, opened 2014-07-30
CRAB 3.3.8.rc5 series

mmascher:
As usual only listing changes for the CRAB3 REST interface:
- Catch error when contacting schedd during the status
- reverse saveoutput and ignorelocality boolean conditions
- Cooloff states are no longer possible for nodes in STATUS_ERROR
- New version of the monitor: add quicksearch and list users for tasks


geneguvo:
+1 to b177c5695d4ee05dd99bb0bbaedf131994444856 on slc5_amd64_gcc461.

PR #907, opened 2014-08-04
Pushes 3.3.8.rc10. Fix problem in the status

mmascher:


geneguvo:
+1 to 32c963369d43d27435f2208990872ea4a6f96340 on slc5_amd64_gcc461.

geneguvo:
To be clear: you pushing crabserver to rc10, but crabclient to rc5. Is that right?

mmascher:
Yes please, I have got more updates on the server than the client and they got out of sync

PR #920, opened 2014-08-07
Revert versions of autoconf and automake which broke lhapdf compilation.

ghost:


davidlt:
@degano what exactly was broken?

davidlt:
ping

Issue #928, opened 2014-08-08
CMS private patch for Geant4 10.0p02


civanch:
CMS private patch on top of Geant4 10.0p02 has been prepared:

http:://cern.ch/vnivanch/verification/verification/data/ geant4-10.0p02-cms01.patch

This patch include following modifications:
- several fixes of data race (two identified by CMS, other by Geant4 developers);
- fix of integer overflow in pre-compound model
- fix of acces out of array boundary in Binary cascade
- fix of electromagnetic/highenergy library - models describing very rare processes not used by default

These fixes should not affect mainstream production because all are technical. Should be applided both on 7_2_X and 7_2_GEANT10

ktf:
@degano can you take care of this?

PR #929, opened 2014-08-11
Add cms private patch to geant 10.0p02.

ghost:
See also cms-sw/cmsdist#928.


ktf:
Given we have a cms-externals/geant4, why adding an extra patch?

PR #944, opened 2014-08-19
HG1409 testbed

ticoann:
fixes wq data location when se name is used instead of PhEDEx node name


geneguvo:
+1 to 2b0095234e345caa79cffbfad6577924878f75e0 on slc5_amd64_gcc461.

PR #945, opened 2014-08-23
move to lhapdf6 6.1.4 and make a few fixes

mkirsano:


bendavid:
@mkirsano,
I see an explicit list of pdfs in
https://github.com/cms-sw/cmsdist/pull/945/files#diff-3aca08d05fafad4c82458388862b75c7L4

Does this need to be updated for the additional pdfs available in 613/614? (in particular nnpdf3.0)

mkirsano:
You are right, Josh. I also added some automatization.

bendavid:
Thanks,
If  lhapdf6_makeLinks.file is now generated automatically, does there still need to be a copy in the repository?

mkirsano:
It is to be generated semiautomatically when lhapdf6 version is upgraded, on a computer with cvmfs. It still needs to be in the repository.

bendavid:
+1

bendavid:
Can this one be moved forward?
(We may need to backport this to 53x on short notice, so having it in 72x ASAP would be helpful)

lhapdf6_makeScript.sh
```diff
@@ -0,0 +1,26 @@
+#TODO: check hat versions coincide
+
+rm -f lhapdf6_makeLinks.file
```
ktf, line 3: Can you elaborate on what is happening here? Is this generating the lhapdf6_makeLinks.file file on the fly?

bendavid, line 3: My understanding is that this script is run manually when preparing a new pull request.  Then the version of lhapdf6_makeLinks which is actually used finally is the one in the repository/pull request.

ktf, line 3: Yeah, I now see it's listing /cvmfs… sigh… I personally do not like the idea of having in CMSDIST a random script running on a random machine using an hardcoded path, but I guess I'll live with it.. @degano can you test this out and merge it?

ghost:
+1
Tested compilation. Merging.

PR #948, opened 2014-08-25
0.9.97.pre8

ticoann:
Add wmdatamining view and code for different polling cycle.
wmstats delete library fix


geneguvo:
+1 to 23dbeeacde8eb99211e98d6bbf16b87c55e11a99 on slc5_amd64_gcc461.

PR #949, opened 2014-08-26
Move to cms-sw/root@430b659dc28

ktf:
Move to cms-sw/root@430b659dc28


ktf:
@wmtan FYI. Will be in next IB.

PR #950, opened 2014-08-26
CRAB3 3.3.9 release candidate 1

mmascher:
News are a new version of the monitor plus three minor fixes:

Use lfnPartsfrom lexicon for the outfile reg exp
Report an error if usedbs is used in the report and no outputdataset is found


geneguvo:
+1 to f724e24c1fcbe6a3f064c522a45a6499a03fd18e on slc5_amd64_gcc461.

PR #960, opened 2014-08-27
p5-time-hires is required for taskworker

juztas:
Requires to include this, on slc6 it fails to deploy because of missing it.


geneguvo:
+1 to f7141c2c22e48cd4f1f2f28d2789c26a273b67c9 on slc5_amd64_gcc461.

PR #963, opened 2014-08-28
use 0.0.97.pre10 for reqmgr2

ticoann:
Fixes thread crashing, and add the log message.


geneguvo:
+1 to a7ba5a5afbba85e8b89f3d55fc4bea9ac9100873 on slc5_amd64_gcc461.

PR #974, opened 2014-09-01
Adding crab3 to cms-common

juztas:
@mmascher @davidlt @ktf 
For testing purpose I used my private git repository. Whenever it will be merged to crabclient repo, I will use tag name and not master for source ... Let me know if I am missing something.

Signed-off-by: Justas Balcas juztas@gmail.com

### DIFF
--- a/cms-common.spec
+++ b/cms-common.spec
@@ -1,5 +1,5 @@
 ### RPM cms cms-common 1.0
-## REVISION 1119
+## REVISION 1120
 ## NOCOMPILER
 
 %define online %(case %cmsplatf in (*onl_*_*) echo true;; (*) echo false;; esac)
@@ -10,6 +10,8 @@
 Source: cmsos
 Source1: migrate-cvsroot
 Source2: cmspm
+Source3: https://raw.githubusercontent.com/juztas/CRABClient/master/bin/crab-client
+Source4: https://raw.githubusercontent.com/juztas/CRABClient/master/bin/crab2to3-client
 
 %prep
 #Make sure that we always build cms-common with a different revision and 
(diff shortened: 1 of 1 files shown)


davidlt:
I would suggest testing this on AFS first before pushing to CVMFS. `cms-common` is built and installed separately from `cmssw`.

ktf:
This still sources crab provided dependencies, which IMHO is a bad idea. IMHO you should not source anything if $CMSSW_BASE is set, so that you can get consistent dependencies.

davidlt:
There is useful bash snippet 

```
[ -z "$CLEANED" ] && exec /usr/bin/env -i CLEANED=1 /bin/sh "$0" "$@"
```

Which would re-execute the wrapper in "a clean" environment (no more CMSSW changes).

smuzaffar:
why crab-client and crab2to3-client has hard coded path to /cvmfs/cms.cern.ch ? May be we should generate these scripts from cms-common itself. Also these scripts source /cvmfs/cms.cern.ch/crab3/crab_$os.sh file which sources init.sh for various binary packages (gcc, zlib, etc.). I would suggest that add a guard in /crab3/crab.sh  and source /cvmfs/cms.cern.ch/crab3/slc5_amd64_gcc461/cms/crabclient/3.3.8/etc/profile.d/dependencies-setup.sh only is scram env (SCRAMRT_SET env variable) is not set

davidlt:
@smuzaffar CRAB3 is only installed on CVMFS, that's the only place where you can find it AFAIK.

ktf:
what's the status of this? can we close it?

juztas:
Closing. @mmascher is working on another solution.

PR #978, opened 2014-09-01
fasthadd file permissions

diguida:
The new fastHadd external deals correctly with file permissions, by setting them to `0664`: this allows for testing/debugging purposes (files are accessible by all users), and for correctly handling files in `DAQ2` (while in the same step, files are handled by the same user, when moved in the data flow chain, files should be handled by users in the same group).

### DIFF
--- a/fasthadd.spec
+++ b/fasthadd.spec
@@ -1,6 +1,8 @@
 ### RPM external fasthadd 1.1
 
-%define commit e193ee35083c2c1583e186130891087d6a4c7ac1 #commit mapped to CMSSW_7_1_6
+#Change the commit hash every time a new version is needed.
+#Commit mapped to CMSSW_7_1_6
+%define commit e193ee35083c2c1583e186130891087d6a4c7ac1
 Source0: https://raw.githubusercontent.com/cms-sw/cmssw/%commit/DQMServices/Components/bin/fastHadd.cc
 Source1: https://raw.githubusercontent.com/cms-sw/cmssw/%commit/DQMServices/Core/src/ROOTFilePB.proto
 Requires: protobuf root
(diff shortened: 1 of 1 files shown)


diguida:
@danduggan @deguio as discussed today and reported in Asana.

fasthadd.spec
```diff
@@ -1,8 +1,8 @@
-### RPM external fasthadd 1.0
+### RPM external fasthadd 1.1
 
-%define tag 743bde9939bbc4f34c2b4fd022cfb980314f912a
-Source0: https://raw.githubusercontent.com/cms-sw/cmssw/%tag/DQMServices/Components/bin/fastHadd.cc
-Source1: https://raw.githubusercontent.com/cms-sw/cmssw/%tag/DQMServices/Core/src/ROOTFilePB.proto
+%define commit e193ee35083c2c1583e186130891087d6a4c7ac1 #commit mapped to CMSSW_7_1_6
```
ktf: Did you actually try this out? # is not a commend character for rpm keywords (like %define). You might want to move the comment above the define.

diguida: @ktf 
I checked it with `rpmlint`, it was ok: mmhh :-(
Fixing it right now.
BTW, can you point me to the `cmsbuild` documentation? I found only sparse stuff in the Twiki.

ktf: all we have is in:

https://github.com/cms-sw/pkgtools/tree/V00-22-XX/docs

and it's admittedly sparse. Will try to improve it as I find time.

diguida: Oh no!
So dummy, I have bookmarked it... Shame on me :-(

ghost:
+1
Tested compilation till cmssw-tool-conf.

PR #986, opened 2014-09-03
Fix location of js files for couch configure

vkuznet:
Please take this changes which fixes issue with locating jsapi.h using spidermonkey 1.8.5 (current version in cmsweb).


ticoann:
@geneguvo, Diego, please review. I tested it fixes the problem I had.

geneguvo:
+1 to 390056b6533da98aa0b7cc739d2009205db98182 on slc5_amd64_gcc461.

geneguvo:
Hi,
We noticed an issue with the couchdb15 package while testing it. The pkg name on the header of the file was still "couchdb", not "couchdb15". We've fixed it (don't need to update this PR) and will soon push the changes.

PR #980, opened 2014-09-02
Replace old HF shower data in SimG4CMS/Calo.

ghost:
According to issue #795.

PR #990, opened 2014-09-08
Fix correct HF file.

ghost:
See also #980.


civanch:
@degano, thanks, file will be available in the afternoon IB?

ghost:
@civanch yes, now that I've merged them.
If it's really urgent I can trigger another IB right now, though.

civanch:
@degano, we need to verify that it is working before pre6 is done.

PR #1003, opened 2014-09-10
Backport of DNS resolution patch for Xrootd client.

bbockelm:
Previously, we tried to update to 3.3.3 to get this patch, then to 4.0.3, then hit
a ROOT issue.  Meanwhile, Ops didn't get the fix they were waiting
for.  So, while we wait on ROOT 5.34.20, here's the originally-requested
backport.

Original patch here:
https://github.com/xrootd/xrootd/commit/6184cd0f24145ebae66bc782d929444d19700fdc

I think this is the correct branch to pull to - if not, let me know.  Tested the build on cmsdev6.cern.ch.


davidlange6:
@ktf, @nclopezo @degano  - Could you have a look at this one?

davidlange6:
(but for 73x instead of 72x)

PR #1011, opened 2014-09-13
Tauola115 7 2 x

inugent:
Update to Tauola 1.1.5 to add Transverse spin effects for Z in TauSpinner. Also upgrade to include the default SANC file for Tauola and TauSpinner at run time.


bendavid:
@inugent can you open another pull request for 73x?  (which is now the new development release)

inugent:
replaced by 1040 which is for 7_3_X.

PR #1028, opened 2014-09-18
Use a private pythia8 200pre1 tarball in cmsdist 73x with some other changes in the pythia8 spec files

mkirsano:


mkirsano:
To be synchronized with CMSSW_7_3_X GeneratorInterface changes, PR CMSSW #5422

bendavid:
Just for book-keeping sake can you rename this pull request to something more descriptive?

bendavid:
+1

mkirsano:
Josh, do you mean that the branch name should be longer?
I have never renamed a branch that is already pushed, can try now.

bendavid:
No no I just meant the name of the pull request in the web interface.

bendavid:
The "my log message of changes" that appears at the top of the page.  I think you should have an edit button to change it right from the web interface.

bendavid:
perfect, thanks

bendavid:
@degano, this is going to need another iteration.  Shouldn't break anything as is, but the new jet matching code is not quite usable.  Might be useful to anyways include this and see if we break existing tests in the 73x IB.
(The forthcoming fix concerns namespace/inlining issues in headers which are curently not used in CMSSW)

PR #1033, opened 2014-09-19
Debug 73x debug

ghost:


smuzaffar:
we also need to update cms-bot/auto-update-git-branches to do something like 
updateBranch IB/CMSSW_7_3_X/stable IB/CMSSW_7_3_X/debug 
right?

PR #1046, opened 2014-09-22
Revert "Use a private pythia8 200pre1 tarball in cmsdist 73x with some o...

ghost:
Reverts cms-sw/cmsdist#1028

I missed it the first time but this PR actually breaks evtgen which requires to update its use of pythia API.
Example error:

```
EvtGenExternal/EvtPythiaEngine.cpp:240:40: error: 'class Pythia8::Event' has no member named 'daughterList'
std::vector<int> daugList = theEvent.daughterList(startInt);
```

### DIFF
--- a/pythia8-toolfile.spec
+++ b/pythia8-toolfile.spec
@@ -10,12 +10,13 @@ mkdir -p %i/etc/scram.d
 cat << \EOF_TOOLFILE >%i/etc/scram.d/pythia8.xml
 <tool name="pythia8" version="@TOOL_VERSION@">
   <lib name="pythia8"/>
+  <lib name="pythia8tohepmc"/>
   <client>
     <environment name="PYTHIA8_BASE" default="@TOOL_ROOT@"/>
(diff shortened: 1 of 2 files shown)


bendavid:
@degano, is this problem in the CMSSW interface or in evtgen itself?

ghost:
@bendavid with evtgen itself I would say as it appear when compiling it as external.

bendavid:
There are multiple evtgen specs in cmsdist, is it the one associated with
https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_7_3_X/stable/evtgen.spec
?

ghost:
Yes, the issue appear when explicitly building that one.

bendavid:
Ok thanks, can you confirm whether there is any issue with tauolapp, which also declares pythia8 as a dependency?

ghost:
tauolapp looks fine.

bendavid:
Thanks,
This should be straightforward to solves on the pythia side.  Will try to get it taken care of together with the other needed header changes.  (ie 8.200pre2 coming soon with this addressed)

PR #1049, opened 2014-09-22
Changes for Makefile.in to include couch_cms_auth.erl

vkuznet:
Original patch only included changes to Makefile.am. Turns out Makefile.in also need to be modified that couch_cms_auth.erl file will be compiled and included into RPM.


geneguvo:
+1 to 5b77f25b23a5abc093ca1db5b2bb1cdf2240f153 on slc5_amd64_gcc461.

PR #1051, opened 2014-09-23
Create web area to access das-client

vkuznet:
Upon Guilio's suggested I copied das-client files into cmsrep WEB area.


geneguvo:
+1 to 5fa7e5d4a4d2c96514d432b206a17aee8185011b on slc5_amd64_gcc461.

PR #1059, opened 2014-09-26
Fix directory structure for lhapdf6 and some improvements

mkirsano:


mkirsano:
Should have no effect on CMSSW as lhapdf6 is kept inactive

ghost:
+1
Tested compilation till cmssw-tool-conf.

PR #1061, opened 2014-09-27
update to WMCore 1.0.0.pre3

ticoann:
fixes reqmgr update and workque error when force-complete is set


geneguvo:
+1 to 7245452d8822679a29aa37a291feb89f7528a524 on slc5_amd64_gcc461.

PR #1071, opened 2014-09-29
Setup dxr as an external python package and all of its python dependencies

gartung:
This allows dxr-build.py to be run in the CMSSW environment. This script sets environment variables that can be used to override the default CC and CXX.


gartung:
Need to include pygments.

PR #1072, opened 2014-09-29
setup dxr as an external python package and all of its python dependencies

gartung:
This allows dxr-build.py to be run in the CMSSW environment. This script sets environment variables that can be used to override the default CC and CXX.

### DIFF
--- a/py2-dxr.spec
+++ b/py2-dxr.spec
@@ -1,4 +1,4 @@
-### RPM external py2-dxr master
+### RPM external py2-dxr 1.0
 ## INITENV +PATH PYTHONPATH %i/$PYTHON_LIB_SITE_PACKAGES
 BuildRequires: llvm sqlite
 Requires: python py2-setuptools py2-futures py2-jinja py2-markupsafe py2-ordereddict py2-parsimonious
@@ -7,33 +7,31 @@ Requires: python py2-setuptools py2-futures py2-jinja py2-markupsafe py2-ordered
 %define triliteCommit e64a2a1 
 %define re2Version 20140304
 %define branch master
-%define github_user mozilla
 
-Source0: git+https://github.com/%github_user/dxr.git?obj=%{branch}/%{dxrCommit}&export=dxr-%{realversion}-%{dxrCommit}&module=dxr-%realversion-%dxrCommit&output=/dxr-%{realversion}-%{dxrCommit}.tgz
-Source1: git+https://github.com/jonasfj/trilite.git?obj=%{branch}/%{triliteCommit}&export=trilite-%{realversion}-%{triliteCommit}&module=trilite-%realversion-%triliteCommit&output=/trilite-%{realversion}-%{triliteCommit}.tgz
+Source0: git+https://github.com/mozilla/dxr.git?obj=%{branch}/%{dxrCommit}&export=dxr-%{dxrCommit}&module=dxr-%dxrCommit&output=/dxr-%{dxrCommit}.tgz
+Source1: git+https://github.com/jonasfj/trilite.git?obj=%{branch}/%{triliteCommit}&export=trilite-%{triliteCommit}&module=trilite-%triliteCommit&output=/trilite-%{triliteCommit}.tgz
 Source2: https://re2.googlecode.com/files/re2-%re2Version.tgz
 Patch0: py2-dxr
 Patch1: trilite
 %define keep_archives true
 
 %prep
-%setup -T -b0 -n dxr-%realversion-%dxrCommit
-%setup -T -D -a1 -c -n dxr-%realversion-%dxrCommit
-%setup -T -D -a2 -n dxr-%realversion-%dxrCommit/trilite-%realversion-%triliteCommit
+%setup -T -b0 -n dxr-%dxrCommit
+%setup -T -D -a1 -c -n dxr-%dxrCommit
+%setup -T -D -a2 -n dxr-%dxrCommit/trilite-%triliteCommit
 %patch -P 1 -p0
 cd ..
 %patch -P 0 -p1
-mv trilite-%realversion-%triliteCommit/* trilite
-%setup -T -D -n dxr-%realversion-%dxrCommit
+mv trilite-%triliteCommit/* trilite
+%setup -T -D -n dxr-%dxrCommit
 
 
 %build
-make build-plugin-clang build-plugin-pygmentize;cd  trilite; make ; cd re2; make ; cd ../../; python setup.py build
+export SQLITE_ROOT; make build-plugin-clang build-plugin-pygmentize;cd  trilite; make release; cd re2; make ; cd ../../; python setup.py build
 
 
 %install
 mkdir %i/lib
 cp -p trilite/libtrilite.so %i/lib
-cp -p trilite/re2/obj/so/libre2.so.* %i/lib
 cp -p trilite/re2/obj/libre2.a %i/lib
 python setup.py install --prefix=%i  --single-version-externally-managed --record=/dev/null
--- a/py2-dxr.spec
+++ b/py2-dxr.spec
@@ -35,3 +35,4 @@ mkdir %i/lib
 cp -p trilite/libtrilite.so %i/lib
 cp -p trilite/re2/obj/libre2.a %i/lib
 python setup.py install --prefix=%i  --single-version-externally-managed --record=/dev/null
+perl -p -i -e "s|^#!%{cmsroot}/.*|#!/usr/bin/env python|" %{i}/bin/*.py
--- a/py2-dxr.spec
+++ b/py2-dxr.spec
@@ -1,0 +1,39 @@
+### RPM external py2-dxr master
+## INITENV +PATH PYTHONPATH %i/$PYTHON_LIB_SITE_PACKAGES
+BuildRequires: llvm sqlite
+Requires: python py2-setuptools py2-futures py2-jinja py2-markupsafe py2-ordereddict py2-parsimonious
+
+%define dxrCommit 6ea764102a
+%define triliteCommit e64a2a1 
+%define re2Version 20140304
+%define branch master
+%define github_user mozilla
+
+Source0: git+https://github.com/%github_user/dxr.git?obj=%{branch}/%{dxrCommit}&export=dxr-%{realversion}-%{dxrCommit}&module=dxr-%realversion-%dxrCommit&output=/dxr-%{realversion}-%{dxrCommit}.tgz
+Source1: git+https://github.com/jonasfj/trilite.git?obj=%{branch}/%{triliteCommit}&export=trilite-%{realversion}-%{triliteCommit}&module=trilite-%realversion-%triliteCommit&output=/trilite-%{realversion}-%{triliteCommit}.tgz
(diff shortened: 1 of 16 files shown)


gartung:
The generation of the DXR indicies is done by 

dxr-build.py -f dxr.config

where dxr.config contains 

[DXR]
enabled_plugins     = clang pygmentize
temp_folder         = /storage/local/data1/gartung/temp
target_folder       = /storage/local/data1/gartung/target
nb_jobs             = 16
wwwroot             = /dxr/

[root_v5-34-18-Sept25]
source_folder       = /storage/local/data1/gartung/root-dxr-build
object_folder       = /storage/local/data1/gartung/root-dxr-build
build_command       = make -e -k -j $jobs

The -e option to make allows dxr-build.py to set DXR_CC and DXR_CXX. In the root example above I modified the Makefile template to include a call to DXR_CC or DXR_CXX everywhere there was a call to $CC or $CXX.

I did a similar modification to the analyze_cxx_common and analyze_c_common targets in config/SCRAM/GMake/Makefile.rules and changing  llvm-analyzer-cxxcompiler.xml and llvm-analyzer-ccompiler.xml  to use CXX=$DXR_CXX and CC=$DXR_CC. This probably needs a new tool configuration.

gartung:
@ktf

ghost:
@gartung this PR is not merge-able, could you please rebase it?

gartung:
Which branch should I rebase to?

gartung:
Rebased.

ghost:
@gartung The test failed because tirlite makefile (inside dxr) wants to download the code using mercurial which is not part of cmssw externals.
2 possible solutions:
1) If trilite isn't necessary we could just drop it.
2) If it's necessary we should patch the makefile to pick the code in another way (perhaps hosting trilite in our github repository)

gartung:
It needs mercurial to check out re2. The command is hg.

-------- Original message --------
From: Alessandro Degano
Date:09/30/2014 8:34 AM (GMT-06:00)
To: cms-sw/cmsdist
Cc: Patrick E Gartung
Subject: Re: [cmsdist] setup dxr as an external python package and all of its python dependencies (#1072)

@gartunghttps://github.com/gartung The test failed because tirlite makefile (inside dxr) wants to download the code using mercurial which is not part of cmssw externals.
2 possible solutions:
1) If trilite isn't necessary we could just drop it.
2) If it's necessary we should patch the makefile to pick the code in another way (perhaps hosting trilite in our github repository)

ghost:
@gartung Sorry, I probably missed the point. I'm aware that it needs mercurial to checkout re2. The point is that we don't ship it with cmsdist and we are not going to add it for this specific need only. Moreover it cannot be installed separately because it should always be possible to build full cmsdist from scratch. Therefore the 2 options I was suggesting earlier.

gartung:
Sorry. I was reading email on my phone and didn't read the full body of 
the email. I will patch the make file and change the spec file to get 
the source tar file for re2.

gartung:
To make this work I created a file src/dxr.config 

[DXR]
enabled_plugins     = clang pygmentize
temp_folder         = /storage/local/data1/gartung/CMSSW_7_3_X_2014-09-24-0200/dxr/temp
target_folder       = /storage/local/data1/gartung/CMSSW_7_3_X_2014-09-24-0200/dxr/target
nb_jobs             = 16
wwwroot             = /dxr

[CMSSW]
source_folder       = /storage/local/data1/gartung/CMSSW_7_3_X_2014-09-24-0200/src
object_folder       = /storage/local/data1/gartung/CMSSW_7_3_X_2014-09-24-0200/src
build_command       = scram b distclean; scram b -f -v -k -e -j $jobs compile COMPILER=llvm-analyzer SCRAM_IGNORE_MISSING_COMPILERS=yes; exit 0

changed the config/SCRAM/Makefile.rules  to remove the -### from the analyze_*_common defines

#LLVM Analyzer Rules
define run_analyze_command
  @$(startlog_$(2)) [ -d $(@D) ] ||  $(CMD_mkdir) -p $(@D) &&\
  $(CMD_echo) ">> Analyzing $(4) $< " &&\
  $(VERB_ECHO) $5 $< -o /dev/null &&\
               $5 $< -o /dev/null $(endlog_$(2))
endef
define analyze_cxx_common
  $(call run_analyze_command,$1,$2,$3,$4,$(CXX) -c $(3) $(CXXOPTIMISEDFLAGS) $(CXXSHAREDOBJECTFLAGS))
endef
define analyze_c_common
  $(call run_analyze_command,$1,$2,$3,$4,$(CC) -c $(3) $(COPTIMISEDFLAGS) $(CSHAREDOBJECTFLAGS))
endef

Once dxr and its dependencies are added to the environment for CMSSW the build can be run with 
cmsenv; dxr-build.py $LOCALRT/src/

The output is put in $LOCALRT/dxr/

gartung:
@degano @ktf 
The vpp_generated.cpp files are causing the plugin to run up to 8GB RES memory. It might be best to skip those file. Another option is to change the compiler option to  -fsyntax-only -xc++

ktf:
A workaround I've is to use -O0 to compile it.

gartung:
Hold off until the next commit. I need to include libre2.so as well.

gartung:
@degano @ktf This is ready to try again.

smuzaffar:
@gartung and @ktf , I will add a new build rule "scram build dxr" which should create the dxr.config file mentioned above and run the dxr-build.py. The build_command will be
build_command = scram b -f -v -k -e -j $jobs compile COMPILER=dxr SCRAM_IGNORE_MISSING_COMPILERS=yes;
We should also update py2-dxr-toolfile.spec to dump dxr-cxxcompiler.xml and dxr-ccompiler.xml to be used by the above rule.

Does this sound reasonable?

ktf:
Sounds good to me.

py2-dxr.spec
```diff
@@ -0,0 +1,39 @@
+### RPM external py2-dxr master
```
smuzaffar: do we want "master" as version of this package?

smuzaffar:
@gartung , I tried to build it and it failed with following error

```
kmp.c:3:24: fatal error: sqlite3ext.h: No such file or directory
 #include <sqlite3ext.h>

```

py2-dxr.spec
gartung: dxr only has one branch and no versions. I could change it to 1.

gartung:
dxr uses sqlite3. The compilation worked on my system and Alex got it to compile as well. How do I modify the spec file to ensure that the sqlite3 include path is included.

smuzaffar:
for sqlite I have fixed it by adding the following patch for trilite

```
--- makefile.orig   2014-09-29 18:48:21.000000000 -0500
+++ makefile    2014-09-30 13:26:00.737812661 -0500
@@ -1,5 +1,5 @@
-CFLAGS := -Ire2/ $(shell pkg-config --cflags sqlite3) -Wall -fPIC -ansi
-LDFLAGS := -Lre2/obj -lre2 $(shell pkg-config --libs sqlite3) -shared
+CFLAGS := -Ire2/ -I$(SQLITE_ROOT)/include -Wall -fPIC -ansi
+LDFLAGS := -Lre2/obj -lre2 -L$(SQLITE_ROOT)/lib -lsqlite3 -shared
 SOURCES := kmp.c scanstr.c varint.c hash.c expr.c match.c regexp.cpp cursor.c vtable.c trilite.c
 OBJECTS := $(patsubst %.cpp,%.o,$(patsubst %.c,%.o,$(SOURCES))) 
 all: debug
```

and then exporting SQLITE_ROOT in the py2-dxr.spec file.

smuzaffar:
can you please send me the changes you did in config/SCRAM/GMake? Also just have CXX=$(DXR_CXX) in the llvm-analyzer-cxxcompiler.xml was enough?

gartung:
I didn't change llvm-analyzer-compiler.xml. I had to modify Makefile.rules to remove the -### options. The compilation with just -c was causing large memory using on vpp_generated.cc files, so I switched to -fsyntax-only -xc++. This does not work with scan-build, so maybe the common rule should use -c -O0. I am testing this now.

define analyze_cxx_common
  $(call run_analyze_command,$1,$2,$3,$4,$(CXX) -c -O0 $(3) $(CXXOPTIMISEDFLAGS) $(CXXSHAREDOBJECTFLAGS))
endef
define analyze_c_common
  $(call run_analyze_command,$1,$2,$3,$4,$(CC) -c -O0 $(3) $(COPTIMISEDFLAGS) $(CSHAREDOBJECTFLAGS))
endef

gartung:
@smuzaffar Your propocal looks fine. The generation of the dxr.config could also be handled outside the makefile using the compiler=dxr rule.

ktf:
I agree -O0 should be default for DXR, or even better "-fsyntax-only", 
if it works as expected.

smuzaffar:
@gartung , ok, so basically we do not need dxr-*compiler.xml files. I will add something like this

```
dxr:    $(SCRAM_SOURCEDIR)/dxr.config
    @mkdir -p dxr &&\
    SCRAM_DXR_RUN=yes $(DXR_BASE)/bin/dxr-build.py -v -f $(SCRAM_SOURCEDIR)/dxr.config
```

and then based on the SCRAM_DXR_RUN we can configure flags to be used by the analyze_*_common rules e.g

```
define analyze_cxx_common
  $(call run_analyze_command,$1,$2,$3,$4,$(CXX) -c $(if $(strip $(SCRAM_DXR_RUN)),FLAGS FOR DXR,-###) $(3) $(CXXOPTIMISEDFLAGS) $(CXXSHAREDOBJECTFLAGS))
endef
```

should this be enough?

gartung:
Only scan-build needs the -c to work correctly and ccc-analyzer now 
skips calling the compiler at all. I would modify the analyze_cxx_common 
to read

$(call run_analyze_command,$1,$2,$3,$4,$(CXX) $(if $(strip $(SCRAM_DXR_RUN)),-fsyntax-only -xc++,-c) $(3) $(CXXOPTIMISEDFLAGS) $(CXXSHAREDOBJECTFLAGS))|

and analyze_c_common

$(call run_analyze_command,$1,$2,$3,$4,$(CC) $(if $(strip $(SCRAM_DXR_RUN)),-fsyntax-only -xc,-c) $(3) $(COPTIMISEDFLAGS) $(CSHAREDOBJECTFLAGS))|

gartung:
I tried adding the dxr target to the end of Makefile.rules but nothing happens. I have to set up the environment for dxr-build.py anyway because of the other python packages it depends on. A dxr target is probably not needed. The modifications to the analyze_c*_command are needed.

gartung:
@smuzaffar Should I also include the patch you make to the trilite makefile in this pull request?

smuzaffar:
@gartung , can you please include 
https://github.com/smuzaffar/cmsdist/commit/97c3646191813a6cebfbcbda10f3a4f8692db026 
It fixes shebang statement for couple of scripts in py2-dxr/bin directory. Also I replaced PY2_DXR with PY2_DXR_BASe in the tool file.

smuzaffar:
@gartung , can you please provide a pull request for IB/CMSSW_7_3_X/next?

gartung:
Moved to CMSSW_7_3_X/next

PR #1073, opened 2014-09-30
Resolve Darwin issues with cmssw-tool-conf

davidlt:
The following fixes several minor issues and brings backs `cmssw-tool-conf` for Darwin.


ktf:
Don't we need the bindings for CondDB2???

davidlt:
No. This is LLVM option and Cond uses Clang Python bindings, which are always installed.

PR #1074, opened 2014-10-01
Moving to root v6.2.0 + patches + cms specific patches.

ktf:
Using cms-sw/root@08c9ffaf2.


ktf:
@wmtan @davidlange6 

@nclopezo please kill the current root6 IB and start a new one which picks this up.

PR #1084, opened 2014-10-03
Dxr python and dependencies 73n

gartung:


gartung:
@smuzaffar

smuzaffar:
thanks @gartung .
can you please update all *-toolfile.spec and use <toolname>_BASE variable e.g. change PY2_FUTURES to PY2_FUTURES_BASE and then use PY2_FUTURES_BASE in the toolfile.

gartung:
Done

smuzaffar:
looks good, I am including it for next 73X

smuzaffar:
This is now available in CMSSW_7_3_DEVEL_X IBs. You can run it by doing 
`scram build dxr`
there is dxr.config template available in <devarea>/config directory which you can configure to configure dxr.

@gartung , Let me know if there are other changes needed for this

gartung:
@smuzzaffar build_command in dxr.config template needs the -e option added to the scram command so dxr can set the compile command.

gartung:
@smuzaffar 
build_command   = scram b -f -v -e -j $jobs compile COMPILER=llvm-analyzer SCRAM_IGNORE_MISSING_COMPILERS=yes

smuzaffar:
build rule uses DXR_CXX and DXR_CC, so do we still need -e option?

```
define analyze_cxx_common
  $(call run_analyze_command,$1,$2,$3,$4,$(if $(strip $(SCRAM_DXR_RUN)),$(DXR_CXX) -fsyntax-only -xc++,$(CXX) -c) $(3) $(CXXOPTIMISEDFLAGS) $(CXXSHAREDOBJECTFLAGS))
endef
```

gartung:
Let me try without the -e. Also the number of jobs should be change from 16 to 4 in the template.

smuzaffar:
I can change the jobs to be variable and use the same -j n we pass to "scram build -j n dxr" command

gartung:
That would be great.

gartung:
OK. The build command does not need the -e option but it does need the -k option. Clang dies on some boost header errors which it is safe to skip past.

davidlt:
And we need to move away from SQLite 3.7.17 to 3.8.7 (alpha currently). It's up to 50% faster.

```
The latest SQLite 3.8.7 alpha version (available on the download page
http://www.sqlite.org/download.html) is 50% faster than the 3.7.17 release
from 16 months ago.  That is to say, it does 50% more work using the same
number of CPU cycles.
```

Probably we can move to 3.8.6 now.

ktf:
Obviously we should not move to alpha versions of anything. If there is a better, stable, sqlite, please do (separately) a PR with it.

gartung:
@smuzaffar  The build_command needs an exit 0 appended to the end so dxr will continue even if gmake reports errors. This is in addition to the -k to continue the build past errors.

build_command   = scram b -f -v -k -j $jobs compile COMPILER=llvm-analyzer SCRAM_IGNORE_MISSING_COMPILERS=yes;exit 0

smuzaffar:
@gartung , the above changes are now part of V05-03-27 cmssw-config tag. Nex 73X IB (CMSSW_7_3_X_2014-10-14-0200) should have it.

gartung:
@smuzzaffr The changes to dxr.config are in but the python packages have not been added to the CMSSW environment config yet.

gartung:
Also are you going to add the option to put the number from scram b -j # in dxr.config.

smuzaffar:
We only added python packages in IB/CMSSW_73_X/next branch (i.e. 7_3_DEVEL_X Ibs). I do see python packages in the PYTHONPATH. If things look ok then may be we should port it back to stable branch?

For -j N option, there is no easy way to find out what was the original value of -j passed scram. I thought MAKE might have some internal flag which can be used but could not find any thing. Should I change it to 4 in the dxr.config file?

gartung:
Can you please merge the pull request with CMSSW_7_3_X/stable.

Also set number of jobs to 4. I can change it when I configure a dxr build job.

PR #1103, opened 2014-10-10
thepeg: build shared version of LesHouches; misc. cleanup

davidlt:
Build two versions of LesHouches: module and shared. These two
have different meaning on Darwin/OSX. You are not allowed to link to
module (MH_BUNDLE), only load it with dlopen(). CMSSW and ThePEG
interface is implemented against ThePEG design. This is hack to keep
the same workaround to work on Darwin, yet the interface should be
reimplemented.

Remove all ancient Darwin/OSX bits.

Remove line for libLHAPDF, such doesn't exist in ThePEG.

Fix relocation list, new files added and non-existing removed.

### DIFF
--- a/thepeg.spec
+++ b/thepeg.spec
@@ -1,8 +1,8 @@
 ### RPM external thepeg 1.9.2
-## INITENV +PATH LD_LIBRARY_PATH %i/lib/ThePEG
-## INITENV +PATH DYLD_LIBRARY_PATH %i/lib/ThePEG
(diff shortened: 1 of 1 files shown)


davidlt:
I tested in on Darwin. Now I will double-check it on Linux before merging, hopefully before 1400 IB.

davidlt:
Works as expected on Linux. Adding to IBs.

PR #1106, opened 2014-10-13
update to pythia 8.200 (official release) (73x)

bendavid:
tarball source reverted to Pythia authors website (since the new version is not available yet from http://cern.ch/service-spi/external/MCGenerators)

@mkirsano @vciulli @covarell


bendavid:
No interface changes are needed for things to technically compile/run, but desired physics behaviour requires also a small set of interface changes in 
https://github.com/cms-sw/cmssw/pull/5792

bendavid:
+1

PR #1122, opened 2014-10-16
Added Openloops as a new external, added OpenLoops to the sherpa configu...

thuer:
Added Openloops as a new external package. Downloads all existing processes and compiles them. openloops.cfg contains gFortran flag needed for long directory names. Total size of all *.so files 470MB

### DIFF
--- a/openloops.spec
+++ b/openloops.spec
@@ -5,15 +5,14 @@ Source: http://www.hepforge.org/archive/openloops/OpenLoops-%{realversion}.tar.g
 
 %setup -n OpenLoops-%{realversion}
 
-cat \EOF >>  openloops.cfg
+
+cat << \EOF >> openloops.cfg
 [OpenLoops]
 fortran_compiler = gfortran
 gfortran_f90_flags = -ffixed-line-length-0 -ffree-line-length-0
 EOF
 
 cat openloops.cfg
-
-./scons
 
 ./scons auto=lhc/ 
 
--- a/openloops.spec
+++ b/openloops.spec
@@ -1,0 +1,29 @@
+### RPM external openloops 1.0.0
+Source: http://www.hepforge.org/archive/openloops/OpenLoops-%{realversion}.tar.gz
+
+%prep
+
+%setup -n OpenLoops-%{realversion}
+
+touch openloops.cfg
+echo "[OpenLoops]" >> openloops.cfg
+echo "fortran_compiler = gfortran" >> openloops.cfg
+echo "gfortran_f90_flags = -ffixed-line-length-0 -ffree-line-length-0" >> openloops.cfg
+echo "openloops.cfg"
+cat openloops.cfg
+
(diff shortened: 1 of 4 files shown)


openloops.spec
```diff
@@ -0,0 +1,29 @@
...
+echo "fortran_compiler = gfortran" >> openloops.cfg
+echo "gfortran_f90_flags = -ffixed-line-length-0 -ffree-line-length-0" >> openloops.cfg
+echo "openloops.cfg"
+cat openloops.cfg
+
+./scons
+
+echo "test"
+
+./scons auto=all/
```
ktf, line 17: I think it was asked elsewhere if we could use lhc rather than all. Is that possible? Can you change that?

openloops.spec
```diff
@@ -0,0 +1,29 @@
+### RPM external openloops 1.0.0
+Source: http://www.hepforge.org/archive/openloops/OpenLoops-%{realversion}.tar.gz
+
+%prep
+
+%setup -n OpenLoops-%{realversion}
+
+touch openloops.cfg
```
ktf: You want:

```
cat << \EOF >  openloops.cfg
[OpenLoops]
fortran_compiler = gfortran
gfortran_f90_flags = -ffixed-line-length-0 -ffree-line-length-0
EOF
```

openloops.spec
ktf, line 17: Why twice?

thuer:
the two calls to scons are copied from the instructions on the openloops hepforge site

ktf:
Can you check if they are they actually needed (i.e. if you get the same result with one?)?

thuer:
one call to scons is enough
Regarding the libraries, with auto=lhc/ only the following processes are downloaded and compiled
ppll ppllj pplljj 
ppaj ppajj 
ppln  pplnj  pplnjj
so neither all jet multiplicities available nor processes like ttbar or diboson, just to make sure this is what is wanted

bendavid:
That rather sounds like we want auto=all...

thuer:
Changed it back to all libraries, alternatively we can specify individually which processes we want from the list here http://openloops.hepforge.org/process_library.php

ghost:
@thuer I am going to test this now. Should we also include the toolfile for openloops in cmssw-tool-conf? Otherwise we won't be able to pick it from cmssw.

thuer:
I think it is not needed in order to work within Sherpa. But it would help to see the Installation directory within cmssw to see which libraries are there, so yes it should be included

ghost:
+1
Tested compilation. @thuer please bring this up at the next ORP meeting (tue 21/10/2014) and also please include openloops-toolfile in cmssw-tool-conf.

bendavid:
Can this be merged now (or are we still waiting for something?)

ktf:
Superseded by #1178.

PR #1124, opened 2014-10-17
Move xrootd to cms-externals/xrootd@3e34e7273

ktf:


nclopezo:
Tested building root.

ktf:
There is an extra bit which we need from @Dr15Jones, see cms-externals/xrootd#1, can either you or @degano merge it, update cmsdist and restart the CMSSW_7_3_DEVEL_X IB? Thanks.

PR #1126, opened 2014-10-17
Spec/patch files for frontend+apache24

vkuznet:


geneguvo:
+1 to 35ab160b5cdbe8a9ad31efd08cb9715322f7e161 on slc6_amd64_gcc481.

ktf:
What's the status of this? Can it be closed or merged?

vkuznet:
Giulio, I don't know status of HTTP group who suppose to test/validate these changes. But we do need these changes to move forward with apach+new couchdb deployment.

geneguvo:
Please keep it open. This is not a simple change to be pushed. Needs careful validation and is scheduled for February, but problably we'll have to postpone it to March.

geneguvo:
This looks very good. Thanks Valentin!

I only had to work on porting the patch Lassi did for reporting SSL information, which you had dropped on the move to the new apache.

There is only one patch remaining, apache2-verify-error, that reports the error reason when client authentication fails. This is important for debugging client authentication issues. I'm not sure, however, this is needed as it might be this is already the case on apache24, or it has some other mechanism to point client authentication errors.

In any case, I've pushed everything. We can proceed and later port apache2-verify-error if still needed.

Cheers, Diego.

PR #1130, opened 2014-10-19
upgrade t0_reqmon to 1.0.1.pre1

ticoann:
Hi Diego, please review, Not completely sure what following patch does. But it seems reqmon is updated with that.
Patch0: reqmon-erl-views
Also I was wondering this can be updated as soon as possible on testbed. Global run starts 27th and this need to be validated before that. It doesn't have to be in the production on that day though. Could you also delete all the data in tier0_wmstats in the testbed and production right before Global run starts?
@hufnagel, Dirk, please confirm that decision (deleting all the data from tier0_wmstats)

Thanks, 
Seangchan

### DIFF
--- a/t0_reqmon.spec
+++ b/t0_reqmon.spec
@@ -1,4 +1,4 @@
-### RPM cms t0_reqmon 0.9.39
+### RPM cms t0_reqmon 1.0.1.pre1
 ## INITENV +PATH PYTHONPATH %i/${PYTHON_LIB_SITE_PACKAGES}
 
 Source0: git://github.com/dmwm/WMCore?obj=master/%realversion&export=%n&output=/%n.tar.gz


ticoann:
Sorry comments in the commit is wrong. The correct version is 1.0.1.pre1

geneguvo:
+1 to 9e2aa2d0014776445eedcf208bd53c2bf5edc1b3 on slc6_amd64_gcc481.

hufnagel:
All data can be deleted from tier0_wmstats, there is nothing there we still need.

ticoann:
Deployment script need to be updated as well.
https://github.com/dmwm/deployment/pull/147

geneguvo:
Hi Seangchan,

Yes, that's fine to include it in the testbed since  t0_reqmon was never production ready anyway.

The reqmon-erl-views patch rewrites half of the wmstats views natively in erlang. And those views are what we use in production wmstats. We wrote them in a somewhat critical situation where wmstats view operations were taking ages to finish. This boosted considerably the performance.

Would you please push those WMStatsErl views upstream to wmcore? then we can drop that patch on next wmstats version.

I actually have 4 more wmstats views written in Erlang to include there. That means only 4 are remaining to be translated (of a total of 16 views). It would be nice this task eventually gets done by someone.

Diego.

ticoann:
Thanks Diego, I will update the WCore for next release. And I will remove the patch from t0-reqmon spec.
I will try to convert to Erlang view as well. (Maybe Valentin. :smile: )

geneguvo:
+1 to f09322abef2b9bbb47795b53219f663dfac88150 on slc6_amd64_gcc481.

ticoann:
I removed reqmon patch the patch since those view are not used from tier0. I can cleanup views for next release when wmstats db separation gets tested thoroughly. thanks

PR #1143, opened 2014-10-27
Xrootd: updated to version cms/v4.0.4

nclopezo:


ktf:
Please put a proper pointer to xrootd in the commit, so that we can click on it and see the sources directly. E.g. cms-externals/xrootd@333bc986604f0e127ffd705be2abb491a1b443b7

PR #1154, opened 2014-10-29
SCRAMV1: add BuildRequires on gmake

davidlt:
`gmake` must be used from CMSDIST as it's not required or/and provided
on the target system.


smuzaffar:
thanks

davidlt:
We probably should consider making `Requires:` as SCRAM actually calls `gmake`.

PR #1178, opened 2014-11-11
Add openloops external.

ktf:
This is just the initial attempt to see if we can get it smaller.


ktf:
Looks like we need to bite the bullet and build 1.0.1.

PR #1200, opened 2014-11-20
update to pythia 8.201 with patch to fix problem with multiple initializ...

nclopezo:
...ations for diffractive processes (ie minbias).


nclopezo:
The files pythia8-201-fix-init.patch and pythia8-201-fix-gcc-options.patch are already in the branch IB/CMSSW_7_3_X/stable brought by the automatic merges:
https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_7_3_X/stable/pythia8-201-fix-gcc-options.patch
https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_7_3_X/stable/pythia8-201-fix-init.patch

PR #1182, opened 2014-11-13
Inclusion of Latest Pede in CMSSW_7_3_X for PCL needs.

arunhep:
Please check with these people about the pull request.
@mmusich @frmeier @diguida @rcastello @tlampen

PR #1203, opened 2014-11-21
Update millepede to V04-02-02.

ghost:
Purge old unused patches.
Substitute #1182.


diguida:
@degano Thanks a lot!
The next IB will have it: @arunhep @frmeier @mmusich @tlampen let me know if you need something more

frmeier:
Nice to have this done - thanks!
Next thing would be to backport it to earlier releases where we might use it. 7_X_Y and 5_3_X would be candidates:
- 7_X because there might be some data taking with them where we have to deploy the PCL. Having pede in the release will likely be the only way to make it avaiable to the PCL.
- 5_3_X because we are currently running the post BPix repair studies in there (only release where we have full access to essential data)
  Who could take care of this?

PR #1224, opened 2014-11-29
use gcc realversion

smuzaffar:
@davidlt , does this looks correct to you? 74ROOT6 are failing due to wrong gcc paths

### DIFF
--- a/root-toolfile.spec
+++ b/root-toolfile.spec
@@ -9,6 +9,7 @@ Requires: gcc
 
 export GCC_ROOT
 export GCC_VERSION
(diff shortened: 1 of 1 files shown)


smuzaffar:
approving and starting a root6 ib

davidlt:
Maybe this is because of having gcc+4.9.1-cms?  (-cms part). What we should is similar to this:

```
TARGET_TRIPLET=$(gcc -dumpmachine)
```

```
davids-macbook:~ davidlt$ gcc -dumpversion
4.2.1
```

smuzaffar:
feel free to fix and merge it for next IB :-)

PR #1247, opened 2014-12-10
Drop openloops

ktf:


vciulli:
why it has been finally dropped? compilation takes too long?

davidlange6:
yes - we did not want to delay 730 further.

bendavid:
Ok, at this point we are happy if it enters 74x clearly (the pull request was only for 73x because it had been hanging around so long)

davidlange6:
yes, I believe after a few days we've managed to get an IB in 74x (or almost..)

vciulli:
next IB for 74X will have openloops?

PR #1250, opened 2014-12-11
Remove openloops from sherpa.

nclopezo:


ktf:
For the record, this is needed because openloops does not actually compile on mac.

PR #1289, opened 2015-01-14
Compile roofit inside root. No longer separate package.

ghost:
Tested locally.
Compilation of root.
Tested sample roofit fitting against the relocated library.


ktf:
@dpiparo @wmtan

PR #1291, opened 2015-01-15
DCAFPilot for DMWM analytics and its dependencies

vkuznet:
Please take spec files for the following set of packages:
- DCAFPilot is DMWMAnalytics machine learning package
- pandas is python data analysis library
- SciPy is python libraries for scientific computing
- sckikit-learn is python Machine Learning libraries

### DIFF
--- a/DCAFPilot.spec
+++ b/DCAFPilot.spec
@@ -1,0 +1,50 @@
+### RPM cms DCAFPilot 0.0.4
+## INITENV +PATH PYTHONPATH %i/${PYTHON_LIB_SITE_PACKAGES}
+%define pkg DCAFPilot
+#Source: git://github.com/dmwm/DMWMAnalytics.git?obj=master/%realversion&export=%pkg-%realversion&output=/%pkg-%{realversion}.tar.gz
+Source: git://github.com/dmwm/DMWMAnalytics.git?obj=master/%realversion&export=%pkg&output=/%pkg.tar.gz
+Requires: python py2-numpy py2-scipy py2-scikit-learn
+BuildRequires: py2-sphinx
+
+# RPM macros documentation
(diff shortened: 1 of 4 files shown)


geneguvo:
+1 to 911d7c35bfe198521c5e7e9105b2a110af426eaa on slc6_amd64_gcc481.

geneguvo:
This is a new project that it looks like I completely missed from discussions? In any case, as long as these bunch of new dependencies don't need something fancy that makes the builds challenging when migrating to new software stacks (new SLC, new gcc, new python, new libraries, etc), I'm fine to include them.

If this is planned to be hosted on cmsweb, as usual we better talk first before making technology choices. Otherwise, it might later happen it cannot fit the cmsweb framework nor the framework adapted to be able to host it.

vkuznet:
Diego,
thanks for reviewing the changes. The new project was started under
DMWM-Analytics umbrella. We don't have yet concrete plans how to run it, but
rather need RPMs in place that we can test the procedure and being able
to deploy it via standard procedure. Moreover it is not a web-service and
neither data-service, therefore its hosting will likely be elsewhere (not part
of cmsweb). I also do not require its inclusion into cmsweb release schedule
and if it happens will certainly inform you and other through standard
channels. But I want to have spec in place to build RPMs. All packages I
requested are standard Machine Learning libraries written in python. Therefore
I don't anticipate any problem with them in a future. If changes will require
they may be on the similar ground as any other python packages we use, e.g.
cherrypy. I also was able to built them successfully in my own build area and
upload them to my own RPM repository, as such I did their testing as well.

I'll give a talk either next week or week after that on Project meeting and you
can get the idea what it is.

Best,
Valentin.

geneguvo:
Thanks Valentin for the explanation! Changes pushed.

PR #1295, opened 2015-01-19
Enable openloops and its sherpa plugins for amd64 archs.

ghost:


cmssw-tool-conf.spec
```diff
@@ -164,6 +163,7 @@ Requires: freetype-toolfile
 Requires: tkonlinesw-toolfile
 Requires: py2-cx-oracle-toolfile
 Requires: oracle-toolfile
+Requires: openloops-toolfile
```
smuzaffar, line 166: this is not correct , we should not add openloop for amd64 archs. better to define
%define isslc %(case %{cmsos} in (slc_) echo 1 ;; (_) echo 0 ;; esac)

on top and then use to to add openloops requirement.

sherpa.spec
```diff
@@ -5,6 +5,10 @@
 Source: git+https://github.com/%github_user/%{n}.git?obj=%{branch}/%{tag}&export=%{n}-%{realversion}&output=/%{n}-%{realversion}-%{tag}.tgz
 Requires: hepmc lhapdf blackhat sqlite fastjet openssl
 
+%if "%(case %cmsplatf in (*amd64*) echo true ;; (*) echo false ;; esac)" == "true"
```
smuzaffar, line 8: same here, , it should be (slc_) instead of (_amd64*)

davidlt, line 8: `*_amd64_*` is more correct approach IMHO. This will not work on AArch64 until compiler and linker improve. This should build on any amd64 target, doesn't matter what linux distribution (for example at some point I was building CMSSW on Fedora 20, ie f20_amd64_*). There is nothing SLC specific here.

ktf, line 8: OS X is also _amd64_ . It needs to be slc*.

## 

Ciao,
Giulio

davidlt, line 8: Ah, I see. Our triplets aren't that good, missing "linux", "darwin", "freebsd", etc. We don't have a type of OS.

ktf, line 8: Indeed. We should really have something like `linux_x86_64_glibcXX` 
because in the end that's what we really need to know for compatibility. 
This is a separate issue, though.

PR #1296, opened 2015-01-19
Include openloops on slc* rather than *amd64*.

ghost:


cmssw-tool-conf.spec
```diff
@@ -163,6 +164,9 @@ Requires: freetype-toolfile
 Requires: tkonlinesw-toolfile
 Requires: py2-cx-oracle-toolfile
 Requires: oracle-toolfile
+
+# Only for slc platforms.
+%if %isslc
 Requires: openloops-toolfile
```
smuzaffar, line 172: are you missing a %endif here?

ghost, line 172: yes, fixing now.

PR #1297, opened 2015-01-19
First release candidate for CRAB 3.3.13

mmascher:
As usual posting a summary of the changes concerning just the REST interface here:
- Return information to spot when publication is disabled
- Report: reutrn the input file lfns to correctly count read files
- Refactoring: added a ServerUtilities with the checkOutLFN function (to be shared with the client)
- Refactoring: remove transformation parameter from the task database
- Changed getoutput and getlog to align with new local stageout policy
- Do not use merge but insert/update in filemetadata upload
- Validate the publication dataset name and the primary dataset name
- Add the possibility to specify the collector
- Add parameter for Hook WMCore event aware lumi splitting

### DIFF
--- a/crabclient.spec
+++ b/crabclient.spec
@@ -1,11 +1,11 @@
-### RPM cms crabclient 3.3.12.patch1
+### RPM cms crabclient 3.3.13.rc1
 ## INITENV +PATH PATH %i/xbin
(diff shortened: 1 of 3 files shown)


geneguvo:
+1 to 379a0cb1ced8955c9190da436b5cf3ee26d5ec5b on slc6_amd64_gcc481.

PR #1320, opened 2015-01-22
Update LLVM/Clang to pre-3.6

davidlt:
Updates LLVM and Clang to pre-3.6, which adds additional features used in CMSSW. In addition to that resolves some asserts in Clang while compiling CMSSW. Same PR (#1220) was merged into DEVEL IBs ~2 months ago. Plain CMSSW now fully compiles and links with Clang.

We will probably again update Clang soonish as 3.6 RC1 is now released, but not tested with CMSSW.

@ktf @davidlange6


davidlt:
ping^1

PR #1330, opened 2015-01-27
Update HF shower library to version 3, backported from 74X.

ghost:


ghost:
+1
Tested compilation. Needs corresponding cmssw PR before going in.

PR #1331, opened 2015-01-28
fix workqueue resubmission and reqmgr modify

ticoann:
HG1502


geneguvo:
+1 to 91581d6a566121f1269aaec03d7899ac78f52bee on slc6_amd64_gcc481.

PR #1334, opened 2015-01-28
Pythia 8.205 (71x)

bendavid:
Must be taken together with corresponding interface change
https://github.com/cms-sw/cmssw/pull/7428


davidlange6:
Will merge and see how the IB goes overnight.

PR #1340, opened 2015-01-29
Pythia 8.205 (74x)

bendavid:
Must be taken together with corresponding cmssw interface changes.


bendavid:
cmssw interface changes in
https://github.com/cms-sw/cmssw/pull/7447

PR #1342, opened 2015-01-29
Version 7.5.1. Build on SLC6 using ROOT6

rovere:
It does not require ROOT6, but it will build against it.


geneguvo:
+1 to bd2b109b57d4d4bdb0f76a314d11641f907cf39a on slc6_amd64_gcc481.

PR #1343, opened 2015-01-30
Pythia 8.205 (53x)

bendavid:
Needs to be merged together with corresponding CMSSW interface change.


bendavid:
CMSSW interface changes in https://github.com/cms-sw/cmssw/pull/7464

PR #1344, opened 2015-01-30
add forgotten boost location

mkirsano:


bendavid:
+1

davidlange6:
if pythia8 needs boost, such a dependency needs to be added to the spec file.

From what I can gather from the Pythia8 readme - this is used for --enable-gzip support (and only that) - in which case there are other issues - could you clarify the usage here (growing boost dependencies is not ideal)

bendavid:
As far as I understand the boost dependency is brought in through the lhapdf6 headers in this case.

mkirsano:
This is used to build lhapdf interface for the lhapdf versions > 6. There is such dependency through lhapdf - is it OK?

davidlange6:
@ktf or @smuzaffar should clarify - I would think the right fix is to add boost to lhapdf-toolfile.spec as a 
<USE name=boost>

mkirsano:
OK, please note that only boost includes are used, so boost libraries are not loaded for lhapdf.

ktf:
This is missing:

```
Requires: boost
```

in the pythia8 spec and then, as you mention we should probably have a "boost" dependency also in the toolfile. Needless to say that depending on boost in one of the generator is a pain.

mkirsano:
Giulio, sorry for asking again: pythia8 depends on lhapdf, lhapdf depends on boost. It is insufficient to have this dependency in lhapdf.spec, it should be added explicitly to pythia8? The same for toolfiles (boost is to be added to lhapdf toolfile - and also to pythia8 toolfile?)

mkirsano:
I found that actually the dependency lhapdf <- boost exists in the toolfiles, it is like this: 
lhapdf <- yaml-cpp <- boost.

ktf:
ah, ok, yes what you have done is sufficient.

davidlange6:
lhapdf does use boost directly.

bendavid:
I think dependencies on common libraries like boost may be the tradeoff for the migration away from fortran in the generator tools...

mkirsano:
If pythia8 finds a problem with lhapdf or boost, it does not exit with error.

ktf:
Depending on boost to be able to parse:

```
Dictionary:
- List Item 1
- List Item 2
```

it's even worse than fortran, IMHO. Anyways, clearly what we have now 
it's wrong so I'll simply merge it.

PR #1347, opened 2015-02-02
Revert "boost,boost-toolfile: update to 1.57.0"

davidlange6:
AlCa needs the new version of boost in 74x reverted… I'm creating a PR by reverting David's original one - but could someone test and see how a CMSSW build goes - I don't see that we had fallout issues from this change, but maybe I've missed something.

Reverts cms-sw/cmsdist#1184


ktf:
Ciao,

why not simply reverting boost serialisation to the one version?

## 

Ciao,
Giulio

davidlange6:
if that is "easy", that would be fine.

davidlt:
You mention 74X, but this is being done for 73X. Older versions of Boost will break condDBv2 conditions XML output.

What is exact reasoning for reverting Boost?

apfeiffer1:
Hi David,

You mention 74X, but this is being done for 73X. Older versions of Boost

> will break condDBv2 conditions XML output.
> 
> What is exact reasoning for reverting Boost?

the reason is that releases with boost 1.51 (like 7.2 and 7.3) can not read
the objects written by 74x with boost 1.57, which is critical.

Thanks,
       cheers, andreas

davidlt:
How do plan to solve this in a long term? I hope you don't expect CMSSW to use already old Boost for N years further.

Have you tried debugging the issue?

This might be another good example of downfall of using ROOT+data, Boost+data. Data must be in a different realm and not connected to the code.

apfeiffer1:
Hi David,

How do plan to solve this in a long term? I hope you don't expect CMSSW to

> use already old Boost for N years further.
> 
> Have you tried debugging the issue?
> 
> This might be another good example of downfall of using ROOT+data,
> Boost+data. Data must be in a different realm and not connected to the code.
> 
> this was discussed at the AlCaDB meeting yesterday. We will be looking into
> this for 75x and try to find a solution for the future which takes the
> evolution into account.

Thanks,
       cheers, andreas

davidlange6:
Hi Andreas,

@fwyzard, fyi
From the 74x devel IB, it seems that CMSSW has grown some boost 1.57 dependencies

#include <boost/core/demangle.hpp>

from

https://github.com/cms-sw/cmssw/blob/CMSSW_7_4_X/HLTrigger/HLTcore/interface/defaultModuleLabel.h

Could you suggest a workaround?

thx-
David

fwyzard:
I guess we can check the boost version, and make

```
 std::string name = boost::core::demangle(typeid(T).name());
```

fall back to

```
 std::string name = boost::units::detail::demangle(typeid(T).name());
```

for older versions of boost.

davidlange6:
i'm ok with that (as hopefully a short term solution)

ghost:
@davidlange6 is this still wanted? Or can we close it?

davidlange6:
this PR is superseded

PR #1357, opened 2015-02-05
Add data-RecoBTag-SoftLepton to cmsswdata.

ghost:


slava77:
nice
this made it in 05-1400 IB, right?

ghost:
As I merged it at 13.46 it should be picked up, yes.

PR #1360, opened 2015-02-06
add patch to fix pythia8 matching for events with top quarks (53x)

bendavid:
Fixes remaining problem with LO ttbar samples in gen validation


bendavid:
+1

PR #1375, opened 2015-02-11
Updated to cms-externals/libpng@88dd30

ktf:


ghost:
As it is didn't compile because of an extra source. 
Fixed, tested and merged at #1379

PR #1376, opened 2015-02-11
Updated to cms-externals/libpng@88dd30

ktf:


ghost:
Fixed, tested and merged in #1380.

PR #1395, opened 2015-02-16
CRAB3 3.3.14.rc1 version

mmascher:
The main changes worth mentioning is that schedd names are put in the DB (will be used for the late binding), and that the state of the jobs will be reported when the tasks are QUEUED and have been resubmitted


geneguvo:
+1 to af142579feb2ba2f9d7989e679d2b7ba26bfbfc8 on slc6_amd64_gcc481.

PR #1398, opened 2015-02-18
Update geant4 to p03.

ghost:
https://github.com/cms-externals/geant4/commit/86c476e297a2f1e3461929714240b0c6709012e3.

Fix https://github.com/cms-sw/cmsdist/issues/1359.

PR #1403, opened 2015-02-19
Revert "Update geant4 to p03."

ktf:
Reverts cms-sw/cmsdist#1398. This seems to break big libraries. Reverting for the moment.

### DIFF
--- a/geant4.spec
+++ b/geant4.spec
@@ -1,5 +1,5 @@
-### RPM external geant4 10.00.p03
-%define tag 86c476e297a2f1e3461929714240b0c6709012e3
+### RPM external geant4 10.00.p02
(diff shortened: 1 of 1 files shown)


civanch:
It is seems not to have cms private patch on top. I do not find out in cmdist branches the commit of the patch because it was tested at THREADED branch which is not listed . In the newest  CMSSW_7_4_THREADED_X_2015-02-19-1400 both biglib and Geant4 10.0p03+cms are working fine.
If needed for the regular 7_5_X redo cms private patch it is at the url: 
http://vnivanch.web.cern.ch/vnivanch/verification/verification/data/geant4-10.0p03-cms.patch

PR #1419, opened 2015-02-25
Geant4p03 tf fix 75 x

ghost:
+1
Tested compilation.
Tested compilation of SimG4Core/ against it.
Tested matrix tests against it.

### DIFF
--- a/geant4-toolfile.spec
+++ b/geant4-toolfile.spec
@@ -36,7 +36,7 @@ cat << \EOF_TOOLFILE >%i/etc/scram.d/geant4core.xml
   <lib name="G4tracking"/>
   <lib name="G4track"/>
   <lib name="G4analysis"/>
-  <flags CXXFLAGS="-ftls-model=global-dynamic -pthread"/>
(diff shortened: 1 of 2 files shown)


civanch:
@degano , Great! do not forget the Geant4e patch on top of 10.0p03. For any case I put a link for any case: http://vnivanch.web.cern.ch/vnivanch/verification/verification/data/geant4-10.0p03-cms.patch

ghost:
@civanch sorry, I don't understand: so https://github.com/cms-externals/geant4/tree/cms/4.10.00.p03  never had that patch?
But it does have other cms patches?

ghost:
@civanch Here you can see what we already have on the externals on top of 4.10.00.p03:
https://github.com/cms-externals/geant4/compare/4.10.00.p03...cms/4.10.00.p03 it seems that patch is already merged in.
Can you confirm please?

civanch:
@degano , Thanks, Geant4e is there. This is what we need.

PR #1435, opened 2015-03-09
Move to cms-sw/root@eedcfdc9

ktf:


ktf:
Includes cms-sw/root#68

PR #1447, opened 2015-03-16
Move to cms-sw/root@0c83a33

ktf:


ktf:
@Dr15Jones @davidlange6 

This includes the patches branch until today.

PR #1476, opened 2015-04-09
Updated config tag to V05-04-06

nclopezo:


smuzaffar:
thanks, this new tag contains a bug fix for REM_<flag> flags specially used for icc and clang/llvm builds

PR #1490, opened 2015-04-14
Add MAGIC environment variable to file toolfile to make it work properly

ghost:
+1


ghost:
Note: 7_1_X will merge this commit from here automatically.

PR #1499, opened 2015-04-21
Changes to build MongoDB 3.X

vkuznet:


vkuznet:
These are the changes required to build MongoDB version 3.X. So far we can't build it within comp_gcc481 since it requires gcc 491 or above. But I wanted to have this PR in place once we move to gcc 491 in comp.

geneguvo:
-1 to 83400724cafe81627f147bda3ab2bdb59e7a9e2e on slc6_amd64_gcc481.

vkuznet:
@geneguvo, yes it will fail on slc6_amd64_gcc481. As I wrote in comments the new build requires gcc491 or above.

geneguvo:
Yeah, but the build-agent does not interpret your comments ;-) It will try build testing on gcc481 everything submitted against comp_gcc481 branch.

geneguvo:
I've pushed this to the comp_gcc493 branch instead.

PR #1500, opened 2015-04-21
patch to sqlalchemy so it can read Oracle timestamps 

apfeiffer1:
the c++ tools write timestamps into sqlite files which are formatted like in oracle (nano-seconds since the unix epoch as integers), which are non-standard for sqlite (which expects date-time formatted values). 
This patch allows to read in these time-stamps and converts them to the expected format.


ghost:
+1
Tested compilation. Merging now.

PR #1511, opened 2015-04-23
patch to coral to read (and write) sqlite formatted timestamps 

apfeiffer1:
With this patch, coral will write "sqlite-formatted" ("datetime") timestamps into sqlite files for conddb V2, such that sqlite can read them w/o changes (we still need the patch from PR #1500 so that the V2 tools can read files created with older versions of the V2 tools, though older versions of the V2 tools will not be able to read the new files). 
Reading/writing from/to Oracle is not changed. 

The change in the scram-project-build.file was needed in order to build the project; the build w/o that change was failing with "invalid number for -P option" (using the "minimal" cmsBuild command), so it may be possible to "ignore" that.

### DIFF
--- a/scram-project-build.file
+++ b/scram-project-build.file
@@ -285,7 +285,7 @@ for DIR in $ELF_DIRS $DROP_SYMBOLS_DIRS; do
     if [ $(echo $ELF_BINS | wc -w) -gt 1 ] ; then
       dwz -m .debug/common-symbols.debug -M common-symbols.debug $ELF_BINS
     fi
-    echo "$ELF_BINS" | xargs -t -n1 -P%{compiling_processes} -I% sh -c 'objcopy --compress-debug-sections --only-keep-debug % .debug/%.debug; objcopy --strip-debug --add-gnu-debuglink=.debug/%.debug %'
+    echo "$ELF_BINS" | xargs -t -n1 -I% sh -c 'objcopy --compress-debug-sections --only-keep-debug % .debug/%.debug; objcopy --strip-debug --add-gnu-debuglink=.debug/%.debug %'
   fi
   popd
 done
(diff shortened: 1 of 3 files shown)


scram-project-build.file
```diff
@@ -285,7 +285,7 @@ for DIR in $ELF_DIRS $DROP_SYMBOLS_DIRS; do
     if [ $(echo $ELF_BINS | wc -w) -gt 1 ] ; then
       dwz -m .debug/common-symbols.debug -M common-symbols.debug $ELF_BINS
     fi
-    echo "$ELF_BINS" | xargs -t -n1 -P%{compiling_processes} -I% sh -c 'objcopy --compress-debug-sections --only-keep-debug % .debug/%.debug; objcopy --strip-debug --add-gnu-debuglink=.debug/%.debug %'
+    echo "$ELF_BINS" | xargs -t -n1 -I% sh -c 'objcopy --compress-debug-sections --only-keep-debug % .debug/%.debug; objcopy --strip-debug --add-gnu-debuglink=.debug/%.debug %'
```
smuzaffar, line 288: @apfeiffer1 , on which system did you test it? this section should only run for slc_|fc_ archs and I think -P option works there

apfeiffer1, line 288: @smuzaffar this was on cmsdev07 - slc6 standard, as far as I can tell. I did not explicitly build with multiple threads/streams/builders, so I guess %{compiling_processes} was not set and this caused the problem. Though that's just a guess from my side ...

apfeiffer1:
any news on this ??? we'd like to get this in ... :)

scram-project-build.file
smuzaffar, line 288: @apfeiffer1 , humm, I guess if we run cmsBuild without -j then compiling_processes is not defined that is why you saw this error. I would suggest that we keep the -P%{compiling_processes} option in the spec and fix cmsBuild to always set compiling_processes (default would be 1).

smuzaffar:
we will integrate it in the next IB (11h00 today)

PR #1516, opened 2015-04-28
Prepare the branch to compile on SLC7 arch.

ghost:
glibc is now a meta package picking the correct version according to the OS.
Advanced xz version, added slc7 keyword alongside several slc6 ones.

### DIFF
--- a/curl.spec
+++ b/curl.spec
@@ -11,7 +11,7 @@ Requires: zlib
 export OPENSSL_ROOT
 export ZLIB_ROOT
 case %cmsplatf in
-  slc6*) KERBEROS_ROOT=/usr ;;
+  slc7*|slc6*) KERBEROS_ROOT=/usr ;;
   slc5*) KERBEROS_ROOT=/usr/kerberos ;;
   osx*) KERBEROS_ROOT=/usr/heimdal ;;
 esac
--- a/curl.spec
+++ b/curl.spec
@@ -1,19 +1,19 @@
 ### RPM external curl 7.28.0
 Source: http://curl.haxx.se/download/%n-%realversion.tar.gz
-Provides: libcurl.so.3()(64bit) 
+Provides: libcurl.so.3()(64bit)
 Requires: openssl
 Requires: zlib
-   
+
 %prep
 %setup -n %n-%{realversion}
 
 %build
 export OPENSSL_ROOT
 export ZLIB_ROOT
+
 case %cmsplatf in
-  slc7*|slc6*) KERBEROS_ROOT=/usr ;;
-  slc5*) KERBEROS_ROOT=/usr/kerberos ;;
   osx*) KERBEROS_ROOT=/usr/heimdal ;;
+  *) KERBEROS_ROOT=/usr ;;
 esac
 ./configure --prefix=%i --disable-static --without-libidn --disable-ldap --with-ssl=${OPENSSL_ROOT} --with-zlib=${ZLIB_ROOT} --with-gssapi=$KERBEROS_ROOT
 # This should change link from "-lz" to "-lrt -lz", needed by gold linker
@@ -23,7 +23,7 @@ make %makeprocesses
 
 %install
 make install
-case %cmsos in 
+case %cmsos in
   osx*) SONAME=dylib ;;
   *) SONAME=so ;;
 esac
--- a/git-toolfile.spec
+++ b/git-toolfile.spec
@@ -7,10 +7,10 @@ Requires: git
 %install
 
 case "%{cmsplatf}" in
-  slc7*|slc6*|fc*)
+  *)
     PERL5LIB_PATH=/share/perl5
     ;;
-  *)
+  osx*)
     PERL5LIB_PATH=/lib/perl5/site_perl
     ;;
 esac
--- a/git-toolfile.spec
+++ b/git-toolfile.spec
@@ -7,7 +7,7 @@ Requires: git
 %install
 
 case "%{cmsplatf}" in
-  slc6*|fc*)
+  slc7*|slc6*|fc*)
     PERL5LIB_PATH=/share/perl5
     ;;
   *)
--- a/glibc-toolfile.spec
+++ b/glibc-toolfile.spec
@@ -1,7 +1,6 @@
 ### RPM external glibc-toolfile 1.0
 Requires: glibc
 
-
 %prep
 %build
 %install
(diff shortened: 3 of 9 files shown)


xz.spec
```diff
@@ -1,13 +1,14 @@
-### RPM external xz 5.0.3__5.1.2alpha
-%define generic_version 5.0.3
+### RPM external xz 5.2.1__5.1.2alpha
+%define generic_version 5.2.1
 %define fcarm_version 5.1.2alpha
-%define tag 931d2d5
+%define tag dec11497a71518423b5ff0e759100cf8aadf6c7b
 %define branch cms/v%generic_version
-%define github_user cms-externals
+%define github_user degano
```
davidlt: This should be **cms-externals**. Also there is no need for `5.1.2alpha`. Just build `5.2.1` on all archs and remove ARMv7 bits (not needed).

glibc.spec
```diff
@@ -1,10 +1,22 @@
-### RPM external glibc 2.12-1.149.el6
+### RPM external glibc 2.17-78.el7_2.12-1.149.el6
 ## NOCOMPILER
 
-%global official_version %(echo "%{realversion}" | cut -d'-' -f1)
+%define isslc6 %(case %{cmsplatf} in (slc6*) echo 1 ;; (*) echo 0 ;; esac)
+%define isslc7 %(case %{cmsplatf} in (slc7*) echo 1 ;; (*) echo 0 ;; esac)
+
+%if %isslc7
```
davidlt, line 7: Does this include our patch or for now it's just plain system glibc?

glibc.spec
```diff
@@ -13,15 +25,15 @@ Source: git+https://github.com/%{github_user}/glibc.git?obj=%{branch}/%{tag}&exp
...
-../glibc-%{realversion}/configure \
+rm -rf ../%{n}-build
+mkdir ../%{n}-build
+cd ../%{n}-build
+../%{n}-%{realversion}/configure \
   CC=gcc \
   CXX=g++ \
   CFLAGS='-mtune=generic -fasynchronous-unwind-tables -DNDEBUG -g -O3' \
   --prefix=/usr \
-  --enable-add-ons=nptl,rtkaio,c_stubs,libidn \
```
davidlt, line 24: These flags were picked by rebuilding glibc with _rpmbuild_ or _mock_ to match exactly what was used in distro. Have you cross-checked this by building glibc on CC7? This is to make sure that we are not breaking system is some way.

glibc-toolfile.spec
```diff
@@ -1,6 +1,7 @@
 ### RPM external glibc-toolfile 1.0
 Requires: glibc
 
+
```
davidlt: Why empty line added?

xz.spec
ktf: Do not use "degano" as a repository.

xz.spec
```diff
@@ -1,13 +1,14 @@
...
 %define armtag 5cc6656
 %define armbranch cms/v%fcarm_version
 %define isfcarm %(case %{cmsplatf} in (fc*_arm*) echo 1 ;; (*) echo 0 ;; esac)
-Source0: git+https://github.com/%github_user/xz.git?obj=%{branch}/%{tag}&export=%{n}-%{generic_version}&output=/%{n}-%{generic_version}.tgz
+#Source0: git+https://github.com/%github_user/xz.git?obj=%{branch}/%{tag}&export=%{n}-%{generic_version}&output=/%{n}-%{generic_version}.tgz
```
ktf: You should pick up things from cms-externals.

glibc.spec
```diff
@@ -32,12 +44,12 @@ cd ../glibc-build
   --host=x86_64-redhat-linux \
   --enable-multi-arch \
   --disable-profile \
-  --enable-experimental-malloc
+  --enable-experimental-malloc \
```
ktf: why the final `\`?

curl.spec
```diff
@@ -11,7 +11,7 @@ Requires: zlib
 export OPENSSL_ROOT
 export ZLIB_ROOT
 case %cmsplatf in
-  slc6*) KERBEROS_ROOT=/usr ;;
+  slc7*|slc6*) KERBEROS_ROOT=/usr ;;
```
ktf: Since we do not support slc5 anymore, I would drop its stanza and make the slc6 / slc7 one default.

git-toolfile.spec
```diff
@@ -7,7 +7,7 @@ Requires: git
 %install
 
 case "%{cmsplatf}" in
-  slc6*|fc*)
+  slc7*|slc6*|fc*)
```
ktf: Same here. Make Linux default and keep the current default for OS X, if it actually applies.

python.spec
```diff
@@ -74,7 +74,7 @@ sed -ibak "s/ndbm_libs = \[\]/ndbm_libs = ['gdbm', 'gdbm_compat']/" setup.py
 # macros on Linux. The following problem does not exists on BSD machines as
 # cdefs.h does not define these macros.
 case %cmsplatf in
-  slc6*|fc*)
+  slc7*|slc6*|fc*)
```
ktf: Again, I would make an empty `osx*` stanza and keep the default for linux. I guess we have better job security options than adding `slc8` when it comes out... ;-)

scram-project-build.file
```diff
@@ -348,7 +348,7 @@ done
 
 %if "%{?pkgname}" != "coral"
 case %cmsplatf in
-  slc6_amd64_*)
+  slc7_amd64_*|slc6_amd64_*)
```
ktf: This should really be `slc*_amd64_*`. No need to support slc5 at this point.

xz.spec
```diff
@@ -20,7 +21,6 @@ BuildRequires: autotools
 %endif
 
 %build
-./autogen.sh
```
ktf, line 23: This should not be dropped, and you should use the cms-externals version, as @davidlt mentioned.

glibc.spec
ghost, line 7: It does. As extracted from comparing 2.12 cms wrt 2.12 provided by rhel.

glibc.spec
ghost, line 24: I compiled glibc on both slc6 and CC7. The reason I removed the add-ons Real Time Kernel Asynchronous I/O is that 1) Probably we don't use it in the limited scope of the glibc we ship.
2) Its library is not installed in CC7 system thus having us to ship it alongside our externals.

PR #1517, opened 2015-04-29
CRAB3 May bugfix (3.3.16.rc2)

mmascher:
This fixes a bug with resubmit which do not work anymore when an early task resubmission already failed. As usual there are other fixes in the client and taskworker I am not mentioning here.


geneguvo:
+1 to bbadfff64c5d7a27c8682d56b73d9e5d73036921 on slc6_amd64_gcc481.

PR #1521, opened 2015-05-01
Generate dependencies-setup.(c)sh files

vkuznet:


geneguvo:
-1 to 22bebbfdc217954d1b8272bc0531061a7dfe59b7 on slc6_amd64_gcc481.

vkuznet:
Where I can look for build error? My build was fine with requested changes.

geneguvo:
Hi Valentin,
Log is on vocms22:/build/dmwmbld/srv/state/dmwmbld/webarea/1521/Log.txt and build area /build/dmwmbld/srv/state/dmwmbld/webarea/1521

It failed re-building bigcouch, which has dependency on it. However,  this is an unrelated bigcouch issue we can ignore.

Anyhow, we don't setup the RPM dependencies for non-top level projects nor libraries like curl. If you want to use curl (the command line), just source the init.sh of das-client or any other client project that needs it. So I'll refuse this PR, ok?

Although it would probably be ok for curl, we already had issues with unwanted dependencies and/or environment set by underlying libraries. In order to give top-level projects full control of the environment and dependencies, we do not set dependencies.sh to other packages.

Alternatively, you could create a etc/profile.d/env.(c)sh file to source the curl (and only curl) environment, so you can use it standalone and don't need any top level project installed.

Cheers,
Diego.

geneguvo:
I meant, put an env.sh script in the RPM that would sources the init.sh of that RPM and also the dependency stuff you've put into dependencies.sh.

geneguvo:
Closing this PR as it was refused. If you want to do the suggested env.sh sotution, please open a new PR.

PR #1529, opened 2015-05-06
DQMGUI 8.0.1

rovere:
Include new developments from Broen for Castor agents. Use long URL when shorten one fail.


geneguvo:
+1 to 060b473aec3994334d6d46521eee745387570606 on slc6_amd64_gcc481.

PR #1532, opened 2015-05-07
Python2.7 for COMP environment

juztas:


juztas:
Merged

geneguvo:
Thanks Justas for taking care of this!

I have a few comments, though.
- The change made here https://github.com/cms-sw/cmsdist/pull/1532/files#diff-3557a42f80fd8c508f2c10858fc87c9cR183 is wrong. It would evaluate to a syntax error in csh when the variable would not be set. The correct syntax is what we're using before, that is, `test X\$?$root = X1 || source ...`.
- The SSL patch I mentioned in the meeting is python-2.6.8-ssl-fragment. We need to double check it is being included in the py27 sources taken from cmssw externals, or at least make sure it is not anymore needed.
- I remember there was an issue with tkinter, but unfortunately I cannot remember what. On this build for py27, however,  I see the "--without-tkinter --disable-tkinter" options were removed. Is tkinter disabled by default on py27?

None of the comments above shall be a show stopper for the purpose of this branch, so no worries on fixing them now.

Cheers,
Diego.

juztas:
Hi,
First bullet, sure, will fix it.
Second bullet, I saw that a lot of changes are done and some are included already. We should make sure it is not needed as you say.
Third - its not removed from python2.7 and would like to know what was the problem. And we can re-enable it if needed

geneguvo:
Hi,
For the tkinter, I don't see where in the py27 build it is being disabled. And if it is disabled, why do we still need this hack https://github.com/cms-sw/cmsdist/blob/d67e2867d29baf8e88783edcef1b4f6ec49363d3/python.spec#L159 ? 

Anyhow, we don't need tkinter anywhere and therefore I don't see a point in re-enabling it, even if it would not be a troublemaker in py27.

On another issue, I just noticed `$additionalConfigureOptions` on this line https://github.com/cms-sw/cmsdist/pull/1532/files#diff-3557a42f80fd8c508f2c10858fc87c9cR77 is not set anywhere, so we should remove it or set it before the `./configure` with the appropriate value, if any.

PR #1533, opened 2015-05-07
python future package

juztas:
@geneguvo build succeeded, am I missing something?


geneguvo:
+1 to e87026d1114fb7b3dc93e143ca68fdfa5065226f on slc6_amd64_gcc481.

geneguvo:
Hi,
I think the build-agent haven't even tried it because nothing else is depending on it yet. We shall put it on comp.spec.

On a separate note, shouldn't we be including it to the py27 branch instead?

Cheers,
Diego.

juztas:
I talked about this with @amaltaro and he also thinks that we need it in py26, we will be doing changes in current code and might be using future package

PR #1534, opened 2015-05-07
Fixes for python27

juztas:
@geneguvo please review

Signed-off-by: Justas Balcas justas.balcas@cern.ch


geneguvo:
This has been pushed to the comp_gcc493 branch instead.

PR #1548, opened 2015-05-12
Gcc491 build on osx fix for gmp

gartung:


ktf:
@gartung is this ok also on linux?
@degano we should probably have a repository for gmp (and similars) in cms-externals.

gartung:
@ktf I have not tested on linux.

gartung:
@ktf @degano How about moving to gcc492. I know gcc492 builds on OSX. I build gcc first in order to build rpm480.

gartung:
Will submit in combination with rpm patch.

PR #1555, opened 2015-05-14
fix paths in rivet scripts and add variable TEXMFHOME

mkirsano:


nclopezo:
+1

PR #1572, opened 2015-05-22
fix for timestamp handling in coral frontier and sqlite plugins

ggovi:
This PR contains:
- a fix for the timestamp format in insert/update with SQLite coral plugin
- a fix for the timestamp format in selecting with Frontier coral plugin


apfeiffer1:
please test

nclopezo:
+1

ggovi:
Hi, 
I'm not sure how, but this patch has broken the IB.
Some previously added frontier fixes have been reverted...

Currently investigating.

davidlt:
Already solved for 1100 IB. `patchsrc9` was used twice, thus removing one of the patches. We will put CORAL source code into git as the next step, because we used all `patchsrc*` commands.

PR #1579, opened 2015-05-26
upgrade to pythia8 209

mkirsano:


bendavid:
Did the the jet matching patch not need to be updated?

mkirsano:
Matching patch worked as is.

mkirsano:
Question to SW people: what was the reason to introduce the compiler option patch adapted here?

bendavid:
What's the reason we don't need --with-boost anymore?

mkirsano:
pythia8 uses lhapdf compatibility/access functions. Boost is not used in the compilation.
Actually it turned out that it was so also in the previous version, but the configure script checked for Boost and did not build lhapdf interface in absence of Boost. This is fixed now.

bendavid:
Ah, ok, I was confused by the release notes, which say 
"Linkage to LHAPDF6 so far has been based on the LHAPDF5 compatibility mode, which requires no Boost headers. The new configure option --with-lhapdf6-plugin=LHAPDF6.h uses native mode, and then requires Boost headers."

So we are still using the compatability mode then?  Is there any advantage moving to native mode?

mkirsano:
In principle there could be some timing advantages. I don't know about such studies.

nclopezo:
+1

nclopezo:
@bendavid is this PR ok for you? I would like to confirm before merging it.

bendavid:
Basic tests are ok (identical results on 1000 minbias events)

bendavid:
+1

PR #1581, opened 2015-05-26
gccxml: remove obsolete package

davidlt:
The package is no more used by ROOT5 or Boost. ROOT6 is based on
Clang/LLVM stack and does not need GCCXML.


ktf:
Die! Die! Die!

PR #1582, opened 2015-05-26
Changes needed to build bootstrap bundle on OSX10.10

gartung:


rpm.spec
```diff
@@ -73,6 +38,10 @@ case %cmsplatf in
     export CFLAGS_PLATF="-arch x86_64 -fPIC"
     export LIBS_PLATF="-liconv"
   ;;
+  osx10A_amd64_gcc49*)
```
ktf: I think it's safe to simply have `osx*)`.

apt.spec
```diff
@@ -4,7 +4,8 @@
 ## INITENV CMD_CSH if ( -f %{instroot}/common/apt-site-env.csh )  source %{instroot}/common/apt-site-env.csh; endif
 ## NOCOMPILER
 
-Source0: git://github.com/cms-sw/apt-rpm.git?obj=master/%{realversion}&export=apt-rpm-%{realversion}&output=/apt-rpm-%{realversion}.tgz
+Source0: https://github.com/cms-sw/apt-rpm/archive/%{realversion}.tar.gz 
+#git://github.com/cms-sw/apt-rpm.git?obj=master/%{realversion}&export=apt-rpm-%{realversion}&output=/apt-rpm-%{realversion}.tgz
```
ktf: You should really use the git repository. If it does not work, please provide the error message. No need to change this.

gcc.spec
```diff
@@ -1,10 +1,10 @@
-### RPM external gcc 4.9.1
```
ktf, line 1: Moving to gcc 4.9.2 requires building the bootstrap for `slc6` as well, as this branch is also used for other stuff (e.g. the CMSSW_7_5_DEVEL_X IBs, for root6).

python.spec
```diff
@@ -1,4 +1,4 @@
-### RPM external python 2.7.6
+### RPM external python 2.7.10
```
ktf, line 1: Moving to the new python should be checked with computing. You should also pick up sources from the `cms-externals/cpython` repository, not the official one.

ktf:
I've added my "two cents" on this PR. I'll leave to @smuzaffar @davidlange6 decide what to do.

gcc.spec
gartung, line 1: 491 will not build on osx10.10

python.spec
gartung, line 1: 276 will not build on osx10.10

cms-externals/cpython needs to be updated to get v2.7.10

apt.spec
gartung: Had this worked at Fermilab I WOULD HAVE USED IT!!!!!
I think is only works from account with the cmsbuild credentials.

rpm.spec
gartung: Had I made that change you would have said to make a special case. I can't do anything right.

gartung:
@ktf @smuzaffar @davidlt @davidlange6 @Dr15Jones
OSX10.9 comes with python 2.7.5 and
OSX10.10 comes with python 2.7.6
Maybe python should be added to the systemtools package for OSX. 
I still cannot get gcc 491 to compile on OSX10.10 so I went to gcc492
I could not get python 2.7.6 to compile on OSX10.10. Perhaps Apple has some special recipe for building it. I did git python 2.7.10 to build, but only after I disabled toolboxglue (whatever that is)

python.spec
smuzaffar, line 1: Looks like python 2.7.9 and above has enabled SSL certificate verification by default for stdlib http clients.
This breaks at least the das_client.py 
`urllib2.URLError: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:590)>`
I am sure it will break many of the computing workflows :-(

gartung, line 1: OK. I think OSX10.10 comes with python 2.7.6. Maybe there is a way to use the system python.

gartung, line 1: With the bootstrap bundle already built for OSX10.10 maybe we can leave this unmerged.

davidlt, line 1: If this issue is triggered we should carefully investigate why. 2.7.9 enabled it by default, PEP 0476 [1]. I am a bit amazed that before it actually never validated certificate thus leaving an attack vector.

I guess DAS certificate was signed with CERN ROOT CA, and those certificates are installed on the system by default.

[1] https://www.python.org/dev/peps/pep-0476/

bbockelm, line 1: For what it's worth, many computing workflows utilize the curl module instead (which does enable validation) precisely because of the validation issue.  I'm a bit surprised to see that this is not true for the DAS client - maybe it's missing curl and falling back?

Anyhow, CERN ROOT CA tends to _not_ be available on the average system bundle (if it is, it's likely somewhat unique to the CERN site).

ghost:
@gartung If the bootstrap on OSX 10.10 is still desired could you please check with the latest 76X externals branch where we ship already gcc493? And maybe we can proceed from there.

PR #1583, opened 2015-05-27
CRABClient 3.3.1506.rc3

mmascher:
Make sure firstEvent is not negative in dryrun
Improve splitting algorithms check in client for PrivateMC and Analysis workflows
Use proxy for gfal commands if user use --proxy parameter or setting env for it


geneguvo:
+1 to 8143ca6860da69b9fc30d194d83838c38d0e857a on slc6_amd64_gcc481.

PR #1585, opened 2015-05-28
coral: fix patchsrc* numbering; remove ARMv7 support

davidlt:
We are out of patchsrc\* statements, thus have to remove ARMv7 patch
string. patchsrc9 was used twice and this caused
coral-CORAL_2_3_21-forever-ttl.patch being dropped.


smuzaffar:
thanks.
better would be to have coral sources in cms-externals/coral and apply 
our patches there.

davidlt:
That's next step.

PR #1602, opened 2015-06-04
Update popdb web service to 0.3.3

nikmagini:
This update fixes https://github.com/dmwm/DDM/issues/64

It is required to remove a call to the old popularity webservice cms-popularity.cern.ch replacing it with a direct query on the DB

Since the retirement of cms-popularity.cern.ch is not urgent this change is not critical

Deployed and tested in private dev instance popularity-web-dev-magini.cern.ch


geneguvo:
+1 to 75c39fb8934d5b29a6f91039b95c16ba7e9ebaef on slc6_amd64_gcc481.

PR #1607, opened 2015-06-05
Remove patch for symlink and force them through makefile options instead

ghost:
+1


smuzaffar:
@degano, can you please revert this change for 74X and above.
Looks like symlinks in git-core directory are causing the problem https://hypernews.cern.ch/HyperNews/CMS/get/git/211.html

PR #1614, opened 2015-06-08
Add MadGraph5_aMC@NLO as external package

mharrend:
The source code of MadGraph5_aMC@NLO is needed for the upcoming version of Herwig++ 3 since Herwig++ 3 will directly use the LO ME from MG5_aMC@NLO via its own interface.

After consultation with Josh I have provided a MG5_aMC@NLO package which can also be used standalone. For this purpose, I have adjusted the MG5_aMC@NLO standard configuration file. However, this standalone mode will not be used by default, it is only a by-product.


bendavid:
I thought you wanted to include 2.3.0beta directly?

mharrend:
Yes, thanks for noting. I was just copy and pasting without thougt. I will update the pull request in a few minutes since the new Herwig++ release will need MG5_aMC@NLO Version 2.3.0.beta

mharrend:
BTW Is there a good way to test a locally built CMSDIST package inside a CMSSW release like CMSSW 7_5_X, so that I can use cmsRun with the updated package?

smuzaffar:
you need to build package-toolfile and then run "scram setup for each xml file found in buildpath/arch/externals/package/version/etc/scram.d e.g

`scram setup <buildpath>/slc6_amd64_gcc491/externals/madgraph5amcatnlo/2.3.0beta/etc/scram.d/madgraph5amcatnlo.xml`

mharrend:
Okay, thanks for this information.

I have now updated the package to the newer MG5_aMC@NLO 2.3.0 beta release as required by the upcoming Herwig++ 3 release

mharrend:
Hi, 

the package is working fine. What steps do I / we need to take that this pull request is merged?
Do I have to start any tests manually?

Thanks,
Marco

madgraph5amcatnlo-toolfile.spec
```diff
@@ -0,0 +1,19 @@
...
+
+%install
+
+mkdir -p %i/etc/scram.d
+cat << \EOF_TOOLFILE >%i/etc/scram.d/madgraph5amcatnlo.xml
+<tool name="madgraph5amcatnlo" version="@TOOL_VERSION@">
+  <client>
+    <environment name="MADGRAPH5AMCATNLO_BASE" default="@TOOL_ROOT@"/>
+  </client>
+  <runtime name="PATH" value="$MADGRAPH5AMCATNLO_BASE/bin" type="path"/>
```
bendavid: Is it really needed to add this to the PATH?  I'm concerned this could interfere with standalone usage of another madgraph folder within a cmssw environment (as used for gridpacks).

smuzaffar: @mharrend, please also fix the tool name in first line, it should be `### RPM external madgraph5amcatnlo-toolfile 1.0` instead of alpgen

smuzaffar:
@mharrend , looks like madgraph5amcatnlo.spec is not installing any thing. It only contains following two files generated by cmsBuild

```
slc6_amd64_gcc491/external/madgraph5_amcatnlo/2.3.0.beta/etc/profile.d/init.sh
slc6_amd64_gcc491/external/madgraph5_amcatnlo/2.3.0.beta/etc/profile.d/init.csh
```

mharrend:
Hello smuzaffar,

since the package does not need to be build it is only required that the package is unpacked and the default configuration is patched so that e.g. the lsf queue can be used.

So from my understanding the current spec file should download the MG5_aMC@NLO 2.3 version. Adopt the mg5configuration file and then create a rpm package out of it, so that the MG5_aMC folder can be found by using $MADGRAPH5AMCATNLO_ROOT.
Am I right?

davidlange6:
I’m in the process of getting some comments together. A partial list
- why do we need a beta release? Would rather not have one
- I don’t think the dependencies listed in the spec file are real. Why are they needed?
- software authors tend to define an ‘install’ directorive to trim off all of the documentation/build time stuff that is not needed to use their package. This is what we distribute. Given the distribution as a whole is 200MB, we need something like that here.
- As Josh queried- how does this interact with mad graph ? The idea is that the generator group will all use this version once intergrated?

bendavid:
Hi David,
For now as you know we use a fully standalone madgraph_aMC@NLO version which is self-contained in the gridpack tarballs, and no madgprah_aMC@NLO installation is currently shipped with CMSSW.   We will continue to do lhe production in this way for the foreseeable future, ignoring the madgraph_aMC@NLO version shipped with CMSSW.  The CMSSW version will only be used where it needs to be called directly from other externals in the CMSSW installation (starting with Herwig++)

Using the CMSSW installation directly for gridpack production will likely create a version dependency mess.  (Ie old gridpacks would stop working as soon as madgraph_aMC@NLO was updated in CMSSW)

The beta version is needed because only this version contains the needed functionality for Herwig++ usage.  Since this is a completely new feature, I think it's ok to integrate this in the development release to get everything tested/in place on the herwig side.  The non-beta 2.3.1 madgprah_aMC@NLO release will likely be available before we want/need to backport any of this to 71x.

mharrend:
I will try to answer your comments:
- This madgraph5_aMC@NLO package is only needed for the upcoming Herwig++ 3 release, in which Fortran functions from MG5 are directly compiled with Herwig++ 3, so that MG5 can be used as a LO ME provider.
- Since Herwig++ 3 relies on some new colour flow information which is only implemented in the MG5_aMC@NLO 2.3 Version we have to use the 2.3 beta version until the official 2.3 version is released.
- In general I would need no dependency at all in the case of using it with Herwig++ 3, so if you want I can remove all dependencies. However, after discussion with  Josh my understandig was that it would be nice to have also the possibility to use it standalone for a cross check. For this reason I also patched the mg5 configuration, so that one could use MG5_aMC@NLO also on the lsf.

bendavid:
The dependencies on pythia, lhapdf, fastjet, etc  is only at the level of setting the corresponding paths in the config file.

davidlange6:
> On Jun 12, 2015, at 12:35 PM, Josh Bendavid notifications@github.com wrote:
> 
> Hi David,
> For now as you know we use a fully standalone madgraph_aMC@NLO version which is self-contained in the gridpack tarballs, and no madgprah_aMC@NLO installation is currently shipped with CMSSW. We will continue to do lhe production in this way for the foreseeable future, ignoring the madgraph_aMC@NLO version shipped with CMSSW. The CMSSW version will only be used where it needs to be called directly from other externals in the CMSSW installation (starting with Herwig++)
> 
> Using the CMSSW installation directly for gridpack production will likely create a version dependency mess. (Ie old gridpacks would stop working as soon as madgraph_aMC@NLO was updated in CMSSW)
> 
> The beta version is needed because only this version contains the needed functionality for Herwig++ usage. Since this is a completely new feature, I think it's ok to integrate this in the development release to get everything tested/in place on the herwig side. The non-beta 2.3.1 madgprah_aMC@NLO release will likely be available before we want/need to backport any of this to 71x.

Seems we can consider this for 76x for the moment once issues are sorted out. But this sounds like it has a big a potential to create a mess of inconsistencies to me. Is there already some prototype of the Herwig++ interface to mad graph that we can see?

mharrend:
Hi David,

I am working on it and I could upload the changed HerwigPP and ThePEG spec files and so on. 
However, the Herwig++ 3 version is not officially released yet, so I am working with an unofficial beta release.
I have to ask the Herwig++ authors first if I can provide you with some beta versions now.

bendavid:
Hi David,
If 76x is open then no problem for this and herwig++ 3.0 to go there instead.

smuzaffar:
it does download the MG5_aMC@NLO 2.3  but in the %install section you still need to copy it from BUILD directory to install directory e.g. something like this in %install section should do the copy
rsync -avh %{_builddir}/MG5_aMC_v%{versiontag}/ %{i}/

smuzaffar:
@mharrend , look like you are only patching/adding `$<tool>_ROOT` and not the actual path of `<tool>`. So basically you do not need the dependency.

mharrend:
@smuzaffar Thanks. I have added your recommended changes.

mkirsano:
Hello, Marco,
 I don't know exactly how it will be used, so I just give you some information without knowing if it will be useful for you:
 Some time ago I worked at the installation of this package in GENSER. It required some commands at the build step:
  make -j1 -C vendor/CutTools
  make -j1 -C vendor/StdHEP

The point is that if it is installed in some read-only location, it will not work because it will not be able to do this compilation. While your test installation can work because it is all in your writable locations.

mharrend:
Hello Mikhail, 
thank you for pointing that out.
I will contact the MG5_aMC@NLO authors and ask them which subpackages I have to compile in advance.

Update: Question to MG5_aMC@NLO authors can be found here
https://answers.launchpad.net/mg5amcnlo/+question/268231

bendavid:
Sorry @mharrend can you open a new pull request instead for 76x?

mharrend:
Hello Josh, dear all,

am I wrong or is there no CMSSW  76X branch up to now to which I could
submit my pull request?

Best regards,

Marco

Marco A. Harrendorf
KIT - IEKP

bendavid:
Should be
https://github.com/cms-sw/cmsdist/tree/IB/CMSSW_7_6_X/stable

mharrend:
I am closing this pull request since I have opened a CMSSW_7_6_X pull request #1633

PR #1629, opened 2015-06-22
Lhapdf6 fullmigration

mkirsano:


bendavid:
This is a precursor to herwig++ backports to 71x.  Effect on standard pythia workflows should be ~zero (since default tune uses internal pdf's), but to be confirmed once an IB is available.

ghost:
+1
Tested compilation, tested copilation of GeneratorInterface package vs it, tested workflow 1.0.
Merging now.

PR #1630, opened 2015-06-23
HG1507 using WMCore 1.0.8.pre5

ticoann:
Include logdb interface and reqmgr propery in couch db and few other bug fixes


geneguvo:
+1 to 33af3bbc731643aade3e968419495258d48f2531 on slc6_amd64_gcc481.

PR #1633, opened 2015-06-25
MadGraph5_aMC@NLO: 

mharrend:
- Standalone package needed by the upcoming Herwig++ 3 version
- MadGraph5_aMC@NLO is now also standalone fully usable.

Follow-Up from CMSSW_7_5_X pull request  #1614

### DIFF
--- a/madgraph5amcatnlo-toolfile.spec
+++ b/madgraph5amcatnlo-toolfile.spec
@@ -1,0 +1,22 @@
+### RPM external madgraph5amcatnlo-toolfile 1.0
+Requires: madgraph5amcatnlo
+%prep
+
+%build
+
+%install
+
+mkdir -p %i/etc/scram.d
+cat << \EOF_TOOLFILE >%i/etc/scram.d/madgraph5amcatnlo.xml
+<tool name="madgraph5amcatnlo" version="@TOOL_VERSION@">
+  <lib name="MadGraph5aMCatNLO"/>
+  <client>
+    <environment name="MADGRAPH5AMCATNLO_BASE" default="@TOOL_ROOT@"/>
+  </client>
+  <runtime name="ROOT_INCLUDE_PATH" value="$MADGRAPH5AMCATNLO_BASE" type="path"/>
+  <runtime name="ROOT_PATH" value="$MADGRAPH5AMCATNLO_BASE" type="path"/>
+  <use name="root_cxxdefaults"/>
+</tool>
+EOF_TOOLFILE
+
+## IMPORT scram-tools-post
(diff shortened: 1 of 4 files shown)


madgraph5amcatnlo.spec
```diff
@@ -0,0 +1,32 @@
+### RPM external madgraph5_amcatnlo 2.3.0.beta
```
ghost: Please change the name of the package to madgraph5amcatnlo (no underscore) as the spec and as it's Required in the toolfile.

madgraph5amcatnlo-toolfile.spec
```diff
@@ -0,0 +1,22 @@
...
+
+mkdir -p %i/etc/scram.d
+cat << \EOF_TOOLFILE >%i/etc/scram.d/madgraph5amcatnlo.xml
+<tool name="madgraph5amcatnlo" version="@TOOL_VERSION@">
+  <lib name="MadGraph5aMCatNLO"/>
+  <client>
+    <environment name="MADGRAPH5AMCATNLO_BASE" default="@TOOL_ROOT@"/>
+  </client>
+  <runtime name="ROOT_INCLUDE_PATH" value="$MADGRAPH5AMCATNLO_BASE" type="path"/>
+  <runtime name="ROOT_PATH" value="$MADGRAPH5AMCATNLO_BASE" type="path"/>
```
ghost, line 17: Please add here 
`<runtime name="PATH"      value="$MADGRAPH5AMCATNLO_BASE/bin" type="path"/>`
In order to add the binaries of this package to the PATH when the toolfile is sourced

bendavid, line 17: No, we don't want this in the path, because we still commonly use standalone madgraph in a cmssw environment, and want to avoid conflicts.

ghost, line 17: There are 2 python scripts there (3 including the one used to install which should be removed). `mg5  mg5_aMC`
Maybe the second should be in path?

ghost:
Also the installation path still have the .tgz that are then unpacked, can those be removed after installation to reduce the size of the final rpm?

madgraph5amcatnlo-toolfile.spec
bendavid, line 17: No, neither should be in the path, since both are used in the standalone usage.

mharrend:
I have now added two lines of code which remove the compile.py script and all downloaded .tgz files to reduce the size of the final rpm.

Anything else which I should take care of?

Without the new changes I could generate a nice working standalone version of MG5_aMC@NLO which could directly be used to produce LHE files, decay them with MadSpin and use Herwig6 to shower them. I will now try it again with the changed settings. However, I do not expect any problems.

mharrend:
Dear all,

I have now compiled the MG5_aMC@NLO package again with the two further changes and tested it. The standalone version of MG5_aMC@NLO is working fine, so I would kindly ask you to merge this pull request.

ghost:
+1
Merging now.

PR #1634, opened 2015-06-26
Enable the sherpa interface to MCFM (with inbuild mcfm version)

thuer:
Enabled the sherpa interface to MCFM (previously a patched version of MCFM 6.3 is built which is then used for sherpa)

### DIFF
--- a/sherpa.spec
+++ b/sherpa.spec
@@ -35,9 +35,12 @@ case %cmsplatf in
   ;;
 esac
 
+./AddOns/MCFM/install_mcfm.sh
+
 %build
 ./configure --prefix=%i --enable-analysis --disable-silent-rules \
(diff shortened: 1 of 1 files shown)


ghost:
@thuer Could you please do this PR for 7_6_X? Then we can backport it if needed once that'll be integrated.

bendavid:
Sorry, the development release is now 76x.  Can you close this one and open it for 76x instead?

If I understand correctly this is installing MCFM as part of the sherpa external.  Does this make sense, or should we add mcfm as a separate external instead?

thuer:
new pull request #1635
It is not installing MCFM as a new external, it just builds a temporary mcfm where the needed library is then incorporated in the libSherpaMCFM.so and from then on the MCFM installation is not needed anymore

PR #1635, opened 2015-06-26
Enable the sherpa interface to MCFM (with inbuild mcfm version)

thuer:


thuer:
It is not installing MCFM as a new external, it just builds a temporary MCFM version where the needed library is then incorporated in the libSherpaMCFM.so and from then on the MCFM installation is not needed anymore

bendavid:
+1
Ok sounds good.

vciulli:
It is needed for H+2 jets NLO 
Thx @thuer

ghost:
+1
Tested compilation, tested compilation of GeneratorInterface/SherpaInterface against it.
Merging now.

PR #1640, opened 2015-06-29
crabserver 3.3.1507.rc2 with fix for memory leak (?)

mmascher:
This is an attempt to fix the memory leak reported by Diego
on the cms-service-webtools. See thread that starts with
"crabserver not releasing memory over time"

Even though it is possible that the memory leak is in some external
library, an examination of the code revealed that there are StringIO
objects and TemporaryFile objects that were created and not closed.

I am not sure if the python garbage collector is taking care of
closing these objects where the corresponding references are not used
anymore (hence the leakage is somewhere else), but surely this commits
is something that needed to be done.


geneguvo:
+1 to b3aa70045374b5b25bb2e80bb9994598bef78722 on slc6_amd64_gcc481.

PR #1666, opened 2015-07-08
Additional fix for professor and cython

mkirsano:


davidlange6:
@mkirsano - I don't understand the reason behind these changes. Could you elaborate?

mkirsano:
This should do the same job that is done for example in cvs2git.spec by the last perl command. Without it, in the header of professor scripts there is some jenkins path, and they don't not work.

mkirsano:
@davidlange6, is this explanation insufficient? professor in 76X still does not work...
When I make a test local build there is no problem with incorrect python path inserted, for this reason I had to do two additional fixes.

davidlange6:
typically we build python tools with simply

python setup.py build

but now I see what we do some set of query-replaces different from but with the same effect as what you've done. I'll ask the experts

mkirsano:
Essential is this argument -e. We could test it on professor and then eventually use it in other spec files.

There is also cython with a similar problem, but somehow incorrect headers do not cause problems (may be it is used only at build time).

PR #1685, opened 2015-07-22
Revert "Remove patch for symlink and force them through makefile opti…

ghost:
…ons instead."

This reverts commit c1077a4a753e7ea9a406fd3c5359dd4588feee21.

### DIFF
--- a/git-1.8.3.1-no-symlink.patch
+++ b/git-1.8.3.1-no-symlink.patch
@@ -1,0 +1,53 @@
+diff --git a/Makefile b/Makefile
+index 0f931a2..db0d69e 100644
+--- a/Makefile
++++ b/Makefile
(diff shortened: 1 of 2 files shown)


davidlt:
Could you provide a reason the change?

smuzaffar:
Looks like symlinks in libexec/git-core directory are causing the problem mentioned here https://hypernews.cern.ch/HyperNews/CMS/get/git/211.html

In CMSSW_7_5_4 runtime env, if we set GIT_EXEC_PATH to one of the git versions where we did not create symlinks (e.g. the one in 7.4.4 release) then above mentioned issue is gone.

while we are trying to understand why git is not happy with symlinks, we decided to revert the change.

PR #1691, opened 2015-07-23
Update RecoEgamma/ElectronIdentification data to V01-00-04

ghost:


ghost:
@lgray I'm including it in 76X, if tomorrow everything is fine I'll include it also for 7[4-5]X

PR #1726, opened 2015-08-07
Update geant4 to version 10.01.p02.

ghost:
Update geant4 data files accordingly.
https://github.com/cms-externals/geant4/commit/ba06375bc3f709e90555746c17f2daae663e2924


davidlange6:
looks like we have problems in the IB that I don't really understand (plus we have a lot of archive libraries that we don't want)

PR #1770, opened 2015-08-25
Upgrade to evtgen 140p1 and clean out evtgenlhc

mkirsano:


covarell:
Since we do not have yet a complete validation of neither the Pythia8-EvtGen CMSSW chain, nor of the new Pythia8 internal implementation, maybe I would not drop for the moment the only version of the generator that has worked for sure...

mkirsano:
This PR requires removal of the evtgenlhc plugin in GeneratorInterface/EvtgenInterface.
I created also the variant with only upgrade.

covarell:
-1

PR #1784, opened 2015-08-31
upgrade t0wmadatasvc to 0.0.3

hufnagel:
Updating the t0wmadatasvc version. The new version contains an expanded schema having more monitoring information. The Oracle databases were already updated as the new schema is backward compatible with both the old insert code and retrieval code.

The new insert and retrieval codes are also independent of each other, so this can just go in whenever and 0.0.3 should then be deployed in the next cmsweb upgrade.

Also updated the WMCore version (no reason really, just to stay up-to-date with the latest version).

### DIFF
--- a/t0wmadatasvc.spec
+++ b/t0wmadatasvc.spec
@@ -1,10 +1,10 @@
-### RPM cms t0wmadatasvc 0.0.2
+### RPM cms t0wmadatasvc 0.0.3
 ## INITENV +PATH PATH %i/xbin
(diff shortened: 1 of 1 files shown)


geneguvo:
+1 to 8919fea394f54f07e0f50b4a35a99de267dd69c6 on slc6_amd64_gcc481.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/1784/test_8919fea-20150831-154503.log" target="_blank">here</a>.

PR #1789, opened 2015-09-03
fix memory leak in wmstats

ticoann:
It seems previous code gets the request data for each cherry thread (when the called is made) in worst case instead of getting the data from server side cache (one per node). Not completely sure how much this will reduce the memory usage since whether current status hits the worst case.  (In best scenario, it will just get the data from the server side cache (one per node). Please deploy this in testbed for further test.  Sorry for all the trouble.


geneguvo:
+1 to 6739cf3ccba04ad939ee851bd7e3f4daba0c03d0 on slc6_amd64_gcc481.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/1789/test_6739cf3-20150903-041503.log" target="_blank">here</a>.

PR #1802, opened 2015-09-09
move to latest config tag V05-05-03

smuzaffar:


davidlt:
This include fixes for _rootcint_ dictionaries and compiler flags order?

smuzaffar:
yes. It is already being used in devel/next branch.

smuzaffar:
@davidlt, by the way this only contains one fix for rootcint i.e. it drops the starting src/ from the path of LinkDef file. The other fix for finding out the headers directly included by LinkDef file and passing those on command-line is not yet done.

davidlt:
@smuzaffar we don't need the second part, because decision was made to move away from _rootcint_ dictionaries at the end. We just need to actually do it and not keep it dragging for a long time.

PR #1807, opened 2015-09-11
Add dbsenv and install it in common area

vkuznet:


geneguvo:
+1 to d87eaff2ee62ce8022164ac903c484b9c30a9372 on slc6_amd64_gcc481.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/1807/test_d87eaff-20150911-163504.log" target="_blank">here</a>.

BrunoCoimbra:
Hi @vkuznet, should I pick this pull request now?

vkuznet:
Bruno, not yet please. I sent it to Shahzad for evaluation and we need to decide how we should proceed with DBS environment within CMSSW.

smuzaffar:
as I mentioned in the private message. It is not the right way to set the environment. If it is just the environment you want to set then dbs spec file should just create simple env scripts (e.g. Sdbsenv.csh and Sdbsenv.sh) in %{instroot}/%{cmsplatf}/etc/profile.d and then these will be sources when we do source %{instroot}/cmsset_default.[c]sh script.

vkuznet:
Shahzad,
that's why I didn't understand Eric's request in a first place.
EVERY cmsweb RPM provides %{instroot}/%{cmsplatf}/etc/profile.d/init.(c)sh
script which set full environment. Why we can't use them?

I thought that Eric wanted to always have latest DBS environment and this is
why I followed your suggestion of das_client setup and adopted it to dbs.
V.

smuzaffar:
No, you are confusing 
%{instroot}/%{cmsplatf}/etc/profile.d/init.(c)sh 
with 
%{instroot}/%{cmsplatf}/%{pkggroup}/%{pkgname}/%{pkgversion}/etc/profile.d/init.(c)sh

`<package>/etc/profile.d/init.(c)sh` scripts are not sources by cmsset_default.csh but if you copy some script in %{instroot}/%{cmsplatf}/etc/profile.d then it will be sources by cmsset_default.[c]sh

vkuznet:
Ahh, I see. Yes, then copying these init script (probably renaming them) under
%{instroot}/%{cmsplatf}/etc/profile.d/ is a way to go.

smuzaffar:
If you let me know what env variable should be set then I can update the dbs3 spec file. Note that the dbs3 env setup script should not call any init.c[s] or dependencies.[c]sh scripts otherwise it can change the default python version.

vkuznet:
Shahzad,
it's seems we round a circle.
Everything what needs to be set comes from init.(c)sh file.
We need to setup PYTHONPATH to location of DBS installed files and we need to
setup all dependencies, e.g. pycurl, etc. which themselves can depend on other
stuff. Assuming that CMSSW can provides all dependencies (I'm not sure about
that) what needs to be set is the following:

```
PVER=<WHATEVER_PYTHON_VERSION>
DBS3_CLIENT_ROOT=<INSTALL_PATH_OF_DBS3_CLIENT>
export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$DBS3_CLIENT_ROOT/lib
export PYTHONPATH=${PYTHONPATH:$DBS3_CLIENT_ROOT/lib/python-$PVER/site-packages
```

In this regard DBS3 follows standard python install structure.
V.

smuzaffar:
If DBS3 is API only then why the dbsenv script needs to set LD_LIBRARY_PATH or PYTHONPATH? why it needs to source dependencies.[c]sh?

vkuznet:
Well, API resides somewhere, so PYTHONPATH should know about that, since user
just need to do
python

> > > import dbs

and DBS APIs relies on 3d party tools too, e.g. pycurl so they should be
in a PYTHONPATH as well.

This behavior is what Eric wants, he said that once CMSSW is set user should be
able to use DBS APIs right away without additional tweaks, i.e. import dbs
should work out of the box.

davidlange6:
but this requirement means using a common python, no?

vkuznet:
Not really, all cmsweb tools are built entire chain of dependencies.
If you meant common python as supplied by SL6 it may not work, since we built
our dependencies, e.g. ssl. If you meant common python as supplied from CMSSW
we may be in a better shape. But as I mention DBS client relies on pycurl
library and I have no clue if it is built in CMSSW.

That's why I said that everything DBS needs can be set by sourcing init.(c)sh file.
If DBS will change its dependency one day we still will face the same problem.

smuzaffar:
if the goal is that we always pick up the latest dbs api then I would suggest that dbsevn script should just set the PYTHONPATH pointing to the latest version of dbs client.

The dependency needed by the dbs should be either provided by cmssw or comp env.

davidlange6:
I think what you say plus having always the latest Dbs version regardless of cmssw release version implies that the Python in all cmssw versions shall never change and should be the same as needed by dbs. That can't work.

On Sep 23, 2015, at 9:02 PM, Valentin Kuznetsov <notifications@github.com<mailto:notifications@github.com>> wrote:

Not really, all cmsweb tools are built entire chain of dependencies.
If you meant common python as supplied by SL6 it may not work, since we built
our dependencies, e.g. ssl. If you meant common python as supplied from CMSSW
we may be in a better shape. But as I mention DBS client relies on pycurl
library and I have no clue if it is built in CMSSW.

That's why I said that everything DBS needs can be set by sourcing init.(c)sh file.
If DBS will change its dependency one day we still will face the same problem.

geneguvo:
Hi Valentin, Eric,

I confess I'm a bit confused by this request here. I thought the idea was to decouple both environments as much as possible. We'd do for dbs-client the same we did for crab-client. But if we want to support "import dbs" out from CMSSW code/environment, then like David said, we have to make COMP and CMSSW in sync about the software stack they use behind these clients. 

This thing was maybe possible for the das-client as it had a simpler stack behind it, and atm it does not yet require authentication (so less problematic to openssl/proxy problems).

To remind about the crab-client solution:
- User runs "cmsenv", which sets the "crab" command as an alias to "/cvmfs/somewhere/current/crab-client/crabwrapper"
- Whenever "crab" is called from inside this CMSSW env, it runs the "crabwrapper" from CVMFS (always the latest crab-client version)
- The crabwrapper script then unloads the CMSSW environment on a subshell, loads the COMP init.sh for crab-client, runs the "crab" command
- Whenever the crab command finishes, the subshell exits and the user is back in the cmssw environment with no changes to it.

If the user would not have CVMFS (or wants to run a local crab-client), it is enough for him to set the "crab" alias to wherever path the crab-client was installed locally.

Cheers,
Diego.

vkuznet:
Diego,
I'll let Eric explain reasoning about this request.

Regarding differences in dbs and das/crab. The former is framework which can be
used only from python code. The later are tools which we may wrap and run in
their own environment. Yes, this creates the issue, it is clear to me know.

Valentin.

davidlange6:
the reason for the desire is clear. One wants to be able to do

import dbs 

without extra work (and without worry of mixing environments). But actually satisfying the desire is not clearly achievable.

geneguvo:
If the dbs-client would provide a command-line tool like crab/das we could then wrap it properly decoupling the environements. That's not as convenient as "import dbs", but maybe that's enough for the use cases (?) of calling DBS from inside CMSSW environment.

ericvaandering:
Why can't scram find the latest version of DBS and add that to python path? If we can't do it, we can't do it. It comes with CRAB2 and CRAB3 now and people use the latest versions of those, so fine.

Aesthetically I don't like that CRAB3, DAS, and DBS will all use different solutions to solve similar problems.

davidlange6:
> On Sep 25, 2015, at 5:48 PM, Eric Vaandering notifications@github.com wrote:
> 
> Why can't scram find the latest version of DBS and add that to python path? If we can't do it, we can't do it. It comes with CRAB2 and CRAB3 now and people use the latest versions of those, so fine.

that is likely doable.  But are you happy with the implied compatibility issues that puts on dbs (e.g., to operate with the oldest supported CMSSW release)

ericvaandering:
Good question. Or we tell people with really old CMSSW that they can’t use DBS in the CMSSW environment and have to use it from CRAB instead.

I don’t even know if this is a problem, but it seems to me it is better than the alternative proposed here: no ability to run DBS in CMSSW environment at all.

davidlange6:
in this case “people” includes being able to use dbs to prepare the relval injection. So that is a potential problem unless the timescale of backwards support is O(2-3 years)

vkuznet:
It can be a problem, since DBS relies on third-party packages, like pycurl.
If we'll only set proper pythonpath for dbs but not its dependencies dbs may
mis-behave.

geneguvo:
I'm not sure I'm following, but I don't think it would be just a matter of having things added PYTHONPATH. The underlying stack (version and build features) must match too (openssl, python, libcurl, c-ares, etc).

ericvaandering:
Here is CMSSW 5_3_22_patch1:

```
-bash-3.2$ which python
/cvmfs/cms.cern.ch/slc5_amd64_gcc462/cms/cmssw-patch/CMSSW_5_3_22_patch1/external/slc5_amd64_gcc462/bin/python
-bash-3.2$ python -V
Python 2.6.4
-bash-3.2$ python
Python 2.6.4 (r264:75706, Feb  7 2012, 16:03:40) 
[GCC 4.6.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import pycurl
```

So it should be OK. I presume old releases in a series are not a real problem and that new externals can be added to new patch releases if actually needed.

geneguvo:
No, it is NOT ok! Unless pycurl, libcurl, etc was built the same way.

geneguvo:
For instance, if curl was built against NSS (not openssl), it will not be possible to authenticate using a proxy certificate.

vkuznet:
> No, it is NOT ok! Unless pycurl, libcurl, etc was built the same way.

I agree with Diego.

Eric, we can get weird behavior too which may show up for some users and not
for others, since it will depend on their environment settings. All (shared) libs,
modules should be compiled/linked in the same way.

geneguvo:
If openssl version is too old, it may not be able to authenticate.
If you're not using c-ares, curl based clients are going to eventually face race conditions that are ultra hard debug
If python version is 2.7 or 3, who knows what's going be the behavior of the tools
If crab-client would need to run some grid tool...
There were other cases, always very painful to catch

geneguvo:
I think once the dbs2-client also had a problem with the httplib2 library version in use on the cmssw stack.

geneguvo:
I think we came into a decision about what to do here (DBS client support on CMSSW releases) during the O&C week. Were there any news?

IIRC: Offline would tell exactly which DBS client APIs they need to use, DMWM would write a command line client script to call them using the DBS client python library, and finally we'd wrap such script to run from the COMP environment like we do for Crab.

vkuznet:
We don't need this anymore since we use das_client wrapper now.

PR #1814, opened 2015-09-18
updated das wrapper to always unset the cmssw env and use cmsos from path

smuzaffar:


smuzaffar:
@vkuznet , I did these changes to das client wrapper in order to get it easily tests in the IBs.

vkuznet:
ok, looks fine to me.

PR #1840, opened 2015-09-25
upgrade to pythia8 212

mkirsano:


mkirsano:
This requires simultaneous changes in CMSSW: PR #11495

ghost:
@mkirsano Since we no longer take Pythia from the primary source and since we apply our own patches I would suggest to put create a cms-sw external repository for Pythia8 (as we do for root and others) and maintain it from there.

bendavid:
our own patches are expected to go away in the near future.

ghost:
@bendavid that's no problem: once the path won't be needed anymore we'll revert them from the source. The biggest reason for this moving to cms-externals is more for consistency. Note that I'd do the work myself, I just want your green light.

bendavid:
Ok, well if @mkirsano agrees then fine with me.

mkirsano:
@degano why you say that we take it not from the primary source?
@bendavid this is a question to you about the need of the matching patch - when it will not be needed.
For me it is less convenient to work with a package in git, but this is a question of practice.

bendavid:
Steve has a proposed fix for the matching patch.  We can test it once 8.212 is in 76x. (Then hopefully the patch will go away with the next pythia release)

bendavid:
(But I want to get 8.212 in with the existing patch first, since this is what we want to back port to 53x and 71x to start with)

davidlt:
@mkirsano it's slightly Debian-like source management for packages. The sources for packages are kept in repositories under _cms-externals_ account [1]. The repository contains the original code (imported from tarball or directly from package's source management system) and also clearly visible our changes. An extra branch is created as `cms/<ORIGINAL_VERSION_TAG_COMMIT>` and then on top our changes are applied.

E.g., ROOT 6.04.XX in DEVEL_X IBs. We are using ROOT at official `264f10d88003213be6110c9d4a5c669c2077a3e4`, thus we have a branch `cms/264f10d` and two additional commits (CMS specific) are applied on top of original code. Thus we use:
Branch: `cms/264f10d`
Commit: `2859404d22bce048706972c94e65ac9d39965485`

The difference in context and history, shows modification to the original code.

In your case, pythia source code needs to be important into pythia8 repository at cmd-externals, then a branch is created (`cms/212`, the second part is version) then on top of that branch CMS specific modifications are applied.

[1] https://github.com/cms-externals

mkirsano:
Yes, I worked a bit with ThePEG which is stored like this. I even tried to put some instructions here: https://twiki.cern.ch/twiki/bin/view/CMS/ExternalandGit

ghost:
@mkirsano @bendavid I created the external repository and a PR to cmsdist to include Pythia8.212 from that repository (where the only patch is already included in the CMS branch): #1845.
I'm closing this PR.

PR #1845, opened 2015-09-29
Update Pythia8 to 212, now taken from cms-externals.

ghost:
Depends on cms-sw/cmssw#11540.
@mkirsano @bendavid


davidlange6:
@degano -I think we have everything for this PR to go in?

ghost:
@davidlange6 yes, it should go in with the IB starting now (2015-09-30-1100)

PR #1873, opened 2015-10-06
make dxr depend on sqlite so maybe it will get rebuilt when sqlite is…

gartung:
… rebuilt

### DIFF
--- a/py2-dxr.spec
+++ b/py2-dxr.spec
@@ -1,8 +1,9 @@
 ### RPM external py2-dxr 1.0
 ## INITENV +PATH PYTHONPATH %i/$PYTHON_LIB_SITE_PACKAGES
 BuildRequires: llvm sqlite
-Requires: python zlib py2-setuptools py2-futures py2-jinja py2-markupsafe py2-ordereddict py2-parsimonious
-
+Requires: python llvm sqlite zlib py2-setuptools py2-futures py2-jinja py2-markupsafe py2-ordereddict py2-parsimonious
+#
+#
 %define dxrCommit 6ea764102a
 %define triliteCommit e64a2a1 
 %define re2Version 20140304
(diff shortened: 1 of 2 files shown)


davidlt:
Note, we still need to rewrite Makefile for it to look into our SQLite3 location.

gartung:
The trilite.patch addresses this.
https://raw.githubusercontent.com/cms-sw/cmsdist/IB/CMSSW_7_6_X/stable/trilite.patch

davidlt:
I completely forgot about it, I was looking at plain unpatched version last night.

davidlt:
This definitely LGTM.

gartung:
@degano Can this be tested?

ghost:
@gartung Yes, I'm on it.

ghost:
@gartung I've tested it and found no issue. However is it necessary for dxr to also depend on llvm?

davidlt:
Yes, because it uses Clang.

gartung:
Further testing showed that the statement must be enclosed in a conditional otherwise the trlilite library load segfaults.

gartung:
@davidlt @degano 
How can I compile this against sqlite 3.7. When I compile against sqlite 3.7 then the library does not segfault.

davidlt:
I thought you said it works fine after recompiling, isn't it? Is it the same segfault like before?

gartung:
Yes. It turns out I was compiling again sqlite 3.7 on one machine. When I tried to reproduce it on a VM running on my Mac the segfault happened again. There is some change to the sqlite_extenstion_init between sqlite 3.7 and sqlite 3.8 that I am trying to figure out.

On Oct 7, 2015, at 4:01 PM, davidlt <notifications@github.com<mailto:notifications@github.com>> wrote:

I thought you said it works fine after recompiling, isn't it? Is it the same segfault like before?

gartung:
I can't make DXR work with sqlite 3.8. DXR needs to be compiled and used with sqlite 3.7. I can do that by hand when I need to run DXR.

PR #1880, opened 2015-10-10
Fix py2 dxr linux

gartung:


gartung:
@davidlt @degano I finally have something that allows dxr to work with sqlite 3.8

py2-dxr.spec
```diff
@@ -1,8 +1,8 @@
 ### RPM external py2-dxr 1.0
 ## INITENV +PATH PYTHONPATH %i/$PYTHON_LIB_SITE_PACKAGES
 BuildRequires: llvm sqlite
```
davidlt, line 3: Move this to Requires, because their are needed for run-time.

py2-dxr.spec
```diff
@@ -33,7 +35,9 @@ mv trilite-%triliteCommit/* trilite
 # GCC + libstdc++. The ABIs are different, thus correct
 # it accordingly.
 # https://code.google.com/p/re2/issues/detail?id=99
+%if %isdarwin
```
davidlt, line 38: Do we really need this? Extra 3 lines for something that doesn't affect Linux.

py2-dxr-sqlite38.patch
```diff
@@ -0,0 +1,31 @@
...
++#import sqlite3
++from pysqlite2 import dbapi2 as sqlite3
+ import string
+ from sys import stdout
+ from urllib import quote, quote_plus
+@@ -112,6 +113,9 @@ def connect_db(dir):
+ 
+     """
+     conn = sqlite3.connect(join(dir, ".dxr-xref.sqlite"))
++    conn.enable_load_extension(True)
```
davidlt, line 26: So why not do this in C? sqlite3_enable_load_extension seems to existed even before 3.8. Extension loading is disabled by default since 2006 according to SQLite source code.

davidlt:
I looked at DXR repo and _trilite_ was officially removed in ~2014. They moved the project to ES and removed almost any reference to _trilite_. From commit: `Take into account the gone-ness of trilite and the here-ness of ES`.

Thus I would suggest moving away from something which is not supported anymore and completely removed.

py2-dxr.spec
gartung, line 3: This is handled by the toolfile.

py2-dxr-sqlite38.patch
gartung, line 26: This function requires a db connection first which is not available.

gartung:
Look, I worked all week on figuring this out. I don't want to use ES yet.

py2-dxr.spec
gartung, line 38: It doesn't hurt anything either.

ghost:
Testing now.

ghost:
@gartung The include of sqlite3.h in the setup.cfg of py2-sqlite was still necessary.
I created a new PR with that fix and squashing your commits here: #1881.
I'll merge that one and close this one.

PR #1903, opened 2015-10-15
WMAgent 1.0.11.pre3

ticoann:
For jenkins unittest, @amaltaro, @geneguvo, could you merge this?


amaltaro:
yep, I'm merging it

amaltaro:
Changes committed here https://github.com/cms-sw/cmsdist/commit/7ebc0f392baa321f8962c21a4e9bffb68b60ec19

geneguvo:
Why are you guys merging to comp_gcc493 instead of the current production architecture comp_gcc481? was it a mistake or there are breaking changes on this tag that are supposed to go only to comp_gcc493?

Otherwise, remember we're already synchonizing comp_gcc481 -> comp_gcc493 every once in a while, so eventually the new wmagent version would appear there on gcc493 too.

amaltaro:
Eric is setting up jenkins for the python2.7 version.
Diego, so the next time we can just PR and merge against gcc481 and it gets synced to gcc493 as well? Is it automatic or you guys from time to time push gcc481 updates to gcc493?

geneguvo:
we're doing by hand, since it is not all the changes we can/should push to gcc493. For top-level projects, however, we always synchronize.

ericvaandering:
@geneguvo can you do the sync by hand then? There was a huge compatibility-breaking change in WMCore a week or so ago so the current RPM does not work for me. 

And if I understand correctly I will find these in comp, not comp.pre?

geneguvo:
You meant to sync these wmagent changes from the py27 to the py26 branch? sure, I can do that. No problem.

And yes, to deploy the py27 stuff you have to use "comp", not "comp.pre".

ericvaandering:
Great

geneguvo:
done!

ericvaandering:
Awesome! I will give it a shot next week.

PR #1910, opened 2015-10-20
add patch for memory leak to 7_5_X

civanch:
Switch to use the same Geant4 as in 7_6_X. @degano, please, correct me if this is done in a wrong way.


ghost:
@civanch Should be okay, I'm testing this now.

ghost:
@civanch I notice now: you have made the PR to `IB/CMSSW_7_5_X/gcc481` which is not the production branch (and it's not even enabled). You probably want to target this PR to `IB/CMSSW_7_5_X/stable`.

civanch:
#1913 has been created to the correct branch

PR #1926, opened 2015-10-24
upgrade pythia8 to 212, add evtgen and photospp

mkirsano:


mkirsano:
Requires changes in GeneratorInterface/Pythia8Interface, PR cms-sw/cmssw#12083

covarell:
+1

davidlange6:
@mkirsano - you need to then remove the other evtgen version. Thanks

mkirsano:
It is not so simple to remove evtgenlhc, requires other changes in CMSSW. I can try to remove the libraries of evtgen.

bendavid:
I don't think we can remove the old evtgen from 53x sonce we still need to support the old pythia6 + evtgen interface.

bendavid:
If this is a problem we can even skip adding the new evtgen to 53x, since it is not needed there so far.

mkirsano:
No, this does not work. Is it possible to live for some short time with both evtgen and evtgenlhc? Then, if evtgenlhc is still needed, I will remove (comment out) evtgen from Pythia8Interface.

davidlange6:
and you can’t have both versions as they conflict….

covarell:
In 53x it is absolutely needed to have evtgenlhc. So I believe the best solution is to comment out the new evtgen interface from Pythia8 in this release.

covarell:
+1

ghost:
@mkirsano This is not the correct branch. Please reopen this PR with respect to: `IB/CMSSW_5_3_X/slc6_amd64_gcc472`. You can check the production branch for CMSDIST for any given release on cms-bot/config.map: https://github.com/cms-sw/cms-bot/blob/master/config.map#L63 looking for the keyword `PROD_ARCH=1`

PR #1951, opened 2015-11-11
[TEST] PR to pythia8 external

ghost:


bendavid:
why do we want to point this directly to steve's repository rather than just merging his pull request to the cms-external repository?

ghost:
@bendavid We don't, it's a just for testing. Once I know it actually compiles I'll merge it in the pythia repository and create the correct CMSDIST entry.

PR #1953, opened 2015-11-11
fastjet 1.020 specs from https://github.com/cms-sw/cmsdist/pull/1891

slava77:
remaking the last attempt of https://github.com/cms-sw/cmsdist/pull/1891
now to be tested with https://github.com/cms-sw/cmssw/pull/11791

PR #1993, opened 2015-11-25
fastjet 1.020 specs (same as 76X version in #1953)

slava77:
reused the same topic branch from 76X to make this PR for fastjet 1.020
The tag matches the head of https://github.com/cms-externals/fastjet-contrib/tree/cms/v1.020


smuzaffar:
merge

PR #2012, opened 2015-12-03
CRAB3 production release for December

mmascher:


mmascher:
That's only for the client

geneguvo:
Done. Pushed.

PR #2009, opened 2015-12-02
[INT-CHECK] Root6 patch for S3 key auth.

ghost:
https://github.com/holzman/root/commit/5ccdddc11a38ab5b3647f8f737f2d5115b956d21

PR #2017, opened 2015-12-06
updated root to include s3 session key fix

smuzaffar:
tests of this was done in #2009


smuzaffar:
this has been already tests in #2009

PR #2022, opened 2015-12-08
tbb: update to 44_20150728oss

davidlt:
The following includes support for AArch64 + Clang. TTB does not have
hand written assembly for atomic support. Clang provides `__sync_*`
builtins like GCC, thus we can fallback to builtin atomics on Clang.

This is needed, because TBB headers end up going through ROOT6 Cling
run-time (Clang-based).


ghost:
Please test

smuzaffar:
merge

PR #2028, opened 2015-12-09
Update fastjet-contrib.spec

smuzaffar:


davidlange6:
@smuzaffar, could you put this into 76x as well.

PR #2036, opened 2015-12-10
New version of das client with x509 support

vkuznet:


ghost:
please test

smuzaffar:
@davidlange6 , should we include it? Side effects (as we already know)
- if X509_USER_PROXY env variable is set or /tmp/x509up_u<uid> file exists  then das uses it 
  - In case of expired proxy it fails 
- if X509_USER_PROXY env variable is not set and missing /tmp/x509up_u<uid> file then das client prints out a warning and still return the results

We announce the change to hypernews first. I will do it once you agree. By the way which HN forum is appropriate for this?

ghost:
@davidlange6 @vkuznet @smuzaffar I suggest that we move this PR to the devel branch of 8_0_X (IB/CMSSW_8_0_X/next) to keep things going, meanwhile I'll close the other PRs.

PR #2045, opened 2016-01-05
starlight files uploaded

KiSooLee:
spec file and patch file for starlight are uploaded

### DIFF
--- a/starlight-r193-p01.patch
+++ b/starlight-r193-p01.patch
@@ -1,0 +1,256 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index fd0afe1..7d560f0 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -179,6 +179,17 @@ else()
+ 		)
+ endif()
+ 
++# setup CLHEP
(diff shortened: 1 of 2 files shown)


ghost:
Dear @KiSooLee, few comments:
1) New externals should always go to the development cycle, which right now is the 8_0_X
2) I though that we would be using Starlight from the external repository that we created here: cms-externals/starlight (which means changing the source and no need for the patch as it's already patched in the repository, see for example: https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_8_0_X/stable/pythia8.spec)
3) The spec file you provided isn't really in line with the spec files we have in this repository. (eg remove unwanted sections: license, doc...)
4) The spec file is still missing some piece (eg line 33), have you actually tested this?
5) The spec file `cmssw-tool-conf.spec` should also be updated to include `starlight.spec`
6) You should also provide a "toolconf" that will enable the use of this generator in CMSSW. See, for example, https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_8_0_X/stable/pythia8-toolfile.spec.

KiSooLee:
Ok.
Let me try again in the CMSSW_8_0_X.

PR #2048, opened 2016-01-06
tool file updated

KiSooLee:
tool file for starlight is updated and applied

### DIFF
--- a/cmssw-tool-conf.spec
+++ b/cmssw-tool-conf.spec
@@ -1,12 +1,11 @@
-### RPM cms cmssw-tool-conf 32.0
+### RPM cms cmssw-tool-conf 29.0
 ## NOCOMPILER
-# With cmsBuild, change the above version only when a new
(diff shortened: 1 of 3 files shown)


davidlt:
NAK, still wrong, look at the diff of `cmssw-tool-conf.spec`. It's wrong.

KiSooLee:
I found the reason. Let me make Pull Request again.

davidlange6:
@KiSooLee , @degano  -before we too much further, is there a presentation or other documentation as to how we expect to use starlight in CMSSW? The readme for starlight does not suggest an easy way to do that.

KiSooLee:
We expect to use starlight like pythia.
Unfortunately we don't have presentation or documentation yet

davidlange6:
is there at least a prototype generator interface ? (or else where in starlight would I find an entrance point into its results (e.g., hepmc particle list or equivalent) that you expect to use?

PR #2064, opened 2016-01-20
Fix missing CRAB3 dependency

mmascher:


geneguvo:
+1 to 90ea78ff7b53716c5cb247b0af78e40761b9d45d on slc6_amd64_gcc493.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/2064/test_90ea78f-20160120-014510.log" target="_blank">here</a>.

mmascher:
Let me mention that I had quite some troubles with this because I am taking WMCore from the head of a branch here https://github.com/dmwm/WMCore/tree/1.0.12_crab , and since the version number is always the same, then `cmsBuild` tried to take it from a cached file here http://cmsrep.cern.ch/cmssw/comp/SOURCES/cache/a05a28daa804f93dde7949b541e96128/WMCore-crabserver-1.0.12_crab.tar.gz . But this is the old version without https://github.com/dmwm/WMCore/commit/a566f8edfbd32295761b4296f881681d2d043762.

I had to workaround this modifying cmsBuild to have good and working RPMs. I hope that's not happening in comp as well, otherwise we'll have to think about something...

PR #2066, opened 2016-01-20
Updated root to tip of 87c769e

smuzaffar:


smuzaffar:
looks like I never merged it :-(

PR #2082, opened 2016-01-22
Herwig7: Updated config files

mharrend:
LHAPDF: Herwig7 requires two further PDFs
MG5aMC: Herwig7 allows to use it as a ME provider
gengetopt: Herwig7 needs it during compilation
HerwigPP: Removed obsolete patches and a small patch to compile it with our boost version is necessary


bendavid:
Why are the extra pdfs needed locally, given that they should already be available through cvmfs?

mharrend:
I had to add the following two PDFs to the LHAPDF config since
Herwig7 use them by default for its LO and NLO event generation and
already wants them during make check / make install.
MMHT2014lo68cl MMHT2014nlo68cl
For some reasons, they were not found during my local build using the
cmsdistbuild script.

bendavid:
Ok, indeed I don't think cvmfs access can be guaranteed during the build process, so adding them is necessary in that case.

gengetopt.spec
```diff
@@ -0,0 +1,39 @@
+### RPM external gengetopt 2.22.6
+Source: ftp://ftp.gnu.org/gnu/gengetopt/gengetopt-2.22.6.tar.gz
+Patch0: gengetopt-parallelbuild
+
+BuildRequires: autotools
+
+%define keep_archives true
```
davidlt: If you have `--disable-static` why do you `keep_archives true`?

gengetopt-toolfile.spec
```diff
@@ -0,0 +1,21 @@
...
+%prep
+
+%build
+
+%install
+
+mkdir -p %i/etc/scram.d
+cat << \EOF_TOOLFILE >%i/etc/scram.d/gengetopt.xml
+<tool name="gengetopt" version="@TOOL_VERSION@">
+  <client>
```
davidlt, line 12: Does this package provide any shared library which should be in `LD_LIBRARY_PATH`. If yes, you should add the lines `<lib name="[NAME]"/>` where you have, e.g., `lib[NAME].so`.

gengetopt.spec
```diff
@@ -0,0 +1,39 @@
...
+
+%if "%{?cms_cxx:set}" != "set"
+%define cms_cxx c++
+%endif
+
+%if "%{?cms_cxxflags:set}" != "set"
+%define cms_cxxflags -O2 -std=c++11
```
davidlt: You can set this to `c++14` as soonish it will be default for C++.

herwigpp.spec
```diff
@@ -1,29 +1,51 @@
...
+
+BuildRequires: autotools
 
 %if "%{?cms_cxx:set}" != "set"
 %define cms_cxx g++
 %endif
 
 %if "%{?cms_cxxflags:set}" != "set"
-%define cms_cxxflags -O2 -std=c++0x
+%define cms_cxxflags -O2 -std=c++11
```
davidlt: Same here, `c++14` as it will be default for C++11 soonish.

herwigpp.spec
```diff
@@ -1,29 +1,51 @@
...
 %endif
 
 %prep
-%setup -q -n herwig++/%{realversion}
+%setup -q -n Herwig-7.0.0
```
davidlt: `7.0.0` -> `%{realversion}`

herwigpp.spec
```diff
@@ -1,29 +1,51 @@
-### RPM external herwigpp 2.7.1
-Source: http://service-spi.web.cern.ch/service-spi/external/MCGenerators/distribution/herwig++/herwig++-%{realversion}-src.tgz
-Requires: boost thepeg gsl hepmc
+### RPM external herwigpp 7.0.0
+Source: https://www.hepforge.org/archive/herwig/Herwig-7.0.0.tar.bz2
```
davidlt: `7.0.0` -> `%{realversion}`

gengetopt.spec
mharrend: I was using some other package as a template, so I guess we can remove this line.

herwigpp.spec
```diff
@@ -1,29 +1,51 @@
...
+            --disable-silent-rules \
+            --with-thepeg=$THEPEG_ROOT \
+            --with-fastjet=$FASTJET_ROOT \
+            --with-gsl=$GSL_ROOT \
+            --with-boost=$BOOST_ROOT \
+            --with-madgraph=$MADGRAPH5AMCATNLO_ROOT \
+            --with-openloops=$OPENLOOPS_ROOT \
+            --prefix=%i \
+            CXX="$CXX" CC="$CC" CXXFLAGS="%{cms_cxxflags}" \
+	    BOOST_ROOT="$BOOST_ROOT" LDFLAGS="$LDFLAGS -L$BOOST_ROOT/lib"
```
davidlt, line 48: Why do we need `-L$BOOST_ROOT/lib` in `LDFLAGS`? Is `--with-boost=` not working properly?

gengetopt-toolfile.spec
mharrend, line 12: No, it does not provide libraries.

gengetopt.spec
mharrend: Yes, I am fine with this.

thepeg.spec
```diff
@@ -1,17 +1,20 @@
-### RPM external thepeg 1.9.2p1
+### RPM external thepeg 2.0.0
 ## INITENV +PATH LD_LIBRARY_PATH %{i}/lib/ThePEG
 ## INITENV +PATH DYLD_LIBRARY_PATH %{i}/lib/ThePEG
 
-%define tag 41e9a26f5ca9659e30e9c27e4dc86e65ecd4a4bd
-%define branch cms/v%realversion
+# Download from official webpage
+Source: http://www.hepforge.org/archive/thepeg/ThePEG-2.0.0.tar.gz
```
davidlt: `2.0.0` -> `%{realversion}`

thepeg.spec
```diff
@@ -23,7 +26,7 @@ BuildRequires: autotools
 %endif
 
 %prep
-%setup -q -n thepeg-%{realversion}-%{tag}
+%setup -q -n ThePEG-2.0.0
```
davidlt: Same here, `2.0.0` -> `%{realversion}`

thepeg.spec
```diff
@@ -58,4 +66,4 @@ find %{i}/lib -name '*.la' -exec rm -f {} \;
 %{relocateConfig}lib/ThePEG/Makefile.common
 %{relocateConfig}lib/ThePEG/Makefile
 %{relocateConfig}lib/ThePEG/ThePEGDefaults.rpo
-%{relocateConfig}lib/ThePEG/ThePEGDefaults-1.9.2.rpo
+%{relocateConfig}lib/ThePEG/ThePEGDefaults-2.0.0.rpo
```
davidlt: Same here, `2.0.0` -> `%{realversion}`

davidlt:
I would also like to have things divided into separate commits, e.g. gengetopt is fully a stand alone thing. Thus it can go into it's own commit. Things should not be monolithic (one big commit).

gengetopt also can have content in `/usr/share`, make sure you delete it.

herwigpp.spec
mharrend, line 48: Yes, we need it and we need also my boost patch. Somehow the m4 boost macro is not working properly.

davidlt, line 48: What exactly goes wrong? If it completely does not detect boost, then you also need `CPPFLAGS='${BOOST_ROOT}/include'` to find the correct headers.

mharrend:
@davidlt Yes, i can split the pull request. 
But can you please explain further what you mean with 
"gengetopt also can have content in /usr/share, make sure you delete it."

However, I would be fine with if you just implement the necessary code lines. I guess it is easier to just implement what you mean instead of explaining it to me first.

Thanks for the comments above I should have included them now.

davidlt:
I looked at gengetopt RPMs and they tend to install documentation in /usr/share directory. At CMS we don't ship man pages, documentations, examples, etc of such tools. We strip the package of not needed files.

govoni:
+1

mharrend:
Closing this request after splitting pull request in two requests:
#2099 Gengetopt request 
#2100 Herwig7 request

PR #2083, opened 2016-01-24
DQMGUI 8.2.8

rovere:
This version is needed since it will be used to patch the current index to insert StreamerInfo that is needed to properly transition to ROOT6


geneguvo:
+1 to 6f228d42838bd422f10688cd3151bb1577b2bbde on slc6_amd64_gcc481.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/2083/test_6f228d4-20160124-210505.log" target="_blank">here</a>.

PR #2090, opened 2016-01-25
build gmp standalone to get gmp.h needed for gcc plugin builds

gartung:


gartung:
@Dr15Jones

davidlt:
There is `gmp-static` package, which should be at 6.1.0 version in gcc530 branch and 5.0.2 in stable branch.

gartung:
The gmp version should match the one gcc was compiled against. The Atlas gcc plugin I am testing was written against gcc 4.9.3 which uses gmp 6.0a. The gcc internals have probably changed for gcc 5.3. This does not cause a rebuild of gcc since it depends on gcc.

davidlt:
That's why there is `gmp-static` which does not provide shared library, otherwise shared library would end up in `LD_LIBRARY_PATH` thus potentially overriding what was used in GCC.

You don't need `gmp.spec`, because our GCC packages provides `gmp` and `gmpxx` libraries and headers.

Simply put use `${GCC_ROOT}/include` and `${GCC_ROOT}/lib` and you will get exact version used for our GCC.

gartung:
OK. The Atlas gcc plugin was built again the LCG gcc distribution that does not include gmp.h. It required setting up a standalone gmp.

davidlt:
Yeah, it's wrong that our package is called GCC. Basically it's a toolchain package (GCC, binutils, elfutils, etc) + all the prerequisites (MPFR, ISL, GMP, etc).

PR #2107, opened 2016-01-28
Fix relative import problem caused by futurize

mmascher:


mmascher:
That's only for the crabclient

geneguvo:
+1 to 172b34c2542a2831d234f81d463a0323747d5ad1 on slc6_amd64_gcc493.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/2107/test_172b34c-20160128-114514.log" target="_blank">here</a>.

PR #2114, opened 2016-02-01
new external starlight added

smuzaffar:
same as #2049 with updated tag/hash of starlight

PR #2115, opened 2016-02-01
new commit to the updated cmsdist

KiSooLee:
starlight-toolfile.spec, starlight.spec are tested in previous pull request.
cmssw-tool-conf.spec conflict is resolved


smuzaffar:
@KiSooLee , I had made the #2114 which is identical to this. If you prefer, we can take yours PR (just remove the last line i.e. rm -rf %{i}/lib/libStarlib.a from the starlight.spec). otherwise we can merge #2114 too

smuzaffar:
done via #2114

PR #2116, opened 2016-02-01
change HepPDT ParticleTable internal from a map to a tbb::concurrent_unordered_map

dan131riley:
The HepPDT::ParticleDataTable class with HeavyIonUnknownID() as the handler for processing unknown particle IDs is not thread safe, as it can update an internal stl::map when the ParticleDataTable::particle() method is called with the ID of an ion not already present in the table.  This has been seen to cause infinite loops (and helgrind errors) with the multi-threaded MixingModule, as the SimTracker/SiStripDigitizer makes calls that can result in unsafe updates that corrupt the map.  This patch replaces the HepPDT::ParticleDataTable internal maps with tbb::concurrent_unordered_map, adding a dependency on tbb.  Since this changes the internal maps from an ordered type to an unordered type, and HepPDT::ParticleDataTable exposes iterators over its internal maps, this does result in an observable change in behavior (which, unfortunately breaks most of the HepPDT tests).  A search for PDTMap::const_iterator in CMSSW finds three places where this interface is called, none of which appear to depend on map ordering, but it is always possible that some ordering dependency has been missed.

This patch has been tested with the full matrix.  Detailed one-to-one and validation tests on one workflow found no differences (other than the same uninitialized values found in the MixingModule testing) but further IB testing is needed.


ghost:
please test

PR #2118, opened 2016-02-02
Python 2.7.11

davidlt:
Moving CMSSW to the latest and greatest version of Python 2.7.11. Note, that starting 2.7.9 version SSL host verification is enabled by default thus machines must have CERN CA bundles installed (i.e. CERN-CA-certs).

### DIFF
--- a/cmssw-tool-conf.spec
+++ b/cmssw-tool-conf.spec
@@ -1,4 +1,4 @@
-### RPM cms cmssw-tool-conf 32.0
+### RPM cms cmssw-tool-conf 33.0
 ## NOCOMPILER
 # With cmsBuild, change the above version only when a new
 # tool is added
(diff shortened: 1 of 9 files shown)


smuzaffar:
please test

davidlt:
So, what do we do with this PR?

ghost:
Probably we don't want it to be on 8_0_X, we might put it in the DEVEL branch for the next release.

davidlt:
It's in a DEVEL branch for quite some time already. We can discuss it's fate at Core Software meeting.

smuzaffar:
My worry is that once we have it in release then we have to deal with messages like my script XYZ does not work after setting  cmssw env ....

ghost:
This has been merged in 81X.

PR #2119, opened 2016-02-02
New TBB (fix for Clang) and Clang 3.7.1

davidlt:
This introduces LLVM/Clang 3.7.1. Currently we have pre-3.6, but ROOT 6.06 is based on Clang 3.7. With these changes we finally match what we have in ROOT. In addition to that OpenMP is added and Compiler Run-Time which will allow using Clang-based sanitisers.

Also TBB is updated to allow C++11 and above constructors if being compiled with Clang. Currently Clang based CMSSW IBs are failing without this change. The patch was submitted to Intel TBB team and is to be considered for the next TBB version.


davidlt:
CMSSW_8_0_DEVEL_X_2016-02-01-1100 was successfully compiled with Clang yesterday as a cross-check.

PR #2129, opened 2016-02-06
upgrade to t0wmadatasvc 0.0.6

hufnagel:
Adds a couple new methods and changes the return format of existing methods from string to json.
Needed schema changes are already deployed (they are safe to use with the old code).

Tested in my own personal VM deployment, everything seems to be working. Could wait for the HG1603 deployment (but if you see a chance to include it in HG1602 I also wouldn't mind).


geneguvo:
+1 to 9923350c93697ecc16b6eb15645b17a20310443b on slc6_amd64_gcc493.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/2129/test_9923350-20160206-020509.log" target="_blank">here</a>.

PR #2138, opened 2016-02-10
Update Jemalloc to 4.0.4.

ghost:


ghost:
please test

smuzaffar:
cmsbuild, please test
last time it has used a broken IB

ghost:
please test

ghost:
@davidlt @fwyzard The tests ran fine and didn't find any issue. Should we try to merge this upgrade in the DEVEL branch?

davidlange6:
would be great!
Thanks Alessandro

davidlt:
Someone should do memory consumption and performance comparison between older and newer jemalloc versions, not that I expect any significant differences, but it would be nice to know.

davidlange6:
Right - we can compare that in the devel IB and then push into the mainstream.

davidlange6:
Hi- looks like the memory gets quite a bit worse with this change (i’m comparing 02-09-1100 and 02-15-1100). Lets revert and see that indeed things get better again (it can be because of something else that I have not noticed)

davidlt:
You should be comparing a single IB with different jemallocs. It's just a single scram setup && cmsenv and you have new jemalloc.

davidlange6:
ok- then I’ll try to set up the old jemalloc in -15 and check

PR #2166, opened 2016-02-25
Advance data for RecoJets/JetProducers.

ghost:
https://github.com/cms-data/RecoJets-JetProducers/commit/90ae65aeac2c5b528578458717d300c83fa07b09

### DIFF
--- a/data-RecoJets-JetProducers.spec
+++ b/data-RecoJets-JetProducers.spec
@@ -1,4 +1,4 @@
-### RPM cms data-RecoJets-JetProducers V05-10-17
+### RPM cms data-RecoJets-JetProducers V05-10-18
 
 %prep
 


ghost:
@cmsbuild please test

slava77:
@davidlange6 
shouldn't we have equivalent CMSSW categories for externals?

In this particular case just new files are added, and cmssw code change is required to enable the feature.
However, sometimes file names stay but contents change and this case should be checked.

slava77:
Just making a note about https://github.com/cms-sw/cmssw/pull/13472 so that the merge status of this PR is visible there.

For new files added in this package (no indiscriminate includes are present in the code runtime) we can get the update of the external  in IBs and then proceed on cmssw side.

davidlange6:
is why we nominally want these requests raised at the ORP. But there are several ways to do it.

slava77:
@degano 
ping
This just adds new files.
Nothing will change or be noticed in the tests.

slava77:
We need 80X and 76X versions of this.

PR #2188, opened 2016-03-10
Fix readline issue with comp python

vkuznet:


davidlt:
So, the whole issue is not having `${READLINE_ROOT}` in dirs? `setup.py` should directly look into `${READLINE_ROOT}` for readline.

ericvaandering:
So it's just a configuration change, nothing to do with new packages? @geneguvo can you take care of this. I concur that this is a major annoyance for developers. (using pdb for instance)

vkuznet:
Yes

geneguvo:
+1 to 6365931a0c60fe9583a6dbe33295e10d21eaf514 on slc6_amd64_gcc493.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/2188/test_6365931-20160310-152510.log" target="_blank">here</a>.

geneguvo:
Hi,
I'm totally in favor of fixing it. It was actually supposed to be already there, see:
https://github.com/vkuznet/cmsdist/blob/fix-comp-python/python.spec#L68
https://github.com/vkuznet/cmsdist/blob/fix-comp-python/python.spec#L71

For that reason, I don't see how Valentin's change would fix it? Anyhow, if Valentin confirms it works, we could include as soon as HG1503 is over.

vkuznet:
Diego,
you missed the point. Previously dirs didn't contain READLINE_ROOT
and LDFLAGS haven't had readline libs, see
https://github.com/vkuznet/cmsdist/blob/fix-comp-python/python.spec#L66

Now I explicitly added READLINE_ROOT to dirs and LDFLAGS therefore get
properly updated with -L$READLINE_ROOT/lib.

This cause configure to find readline libraries which were not happened before.

Yes, I build python with my fix and did simple test
import readline
so it works now.

Best,
Valentin.

geneguvo:
LDFLAGS had it in line 71 as I pointed. Either that fullpath syntax is wrong or there are other libs under READLINE_ROOT/lib to be included too.

vkuznet:
Diego this line (line 71) will expand as

LDFLAGS=-L/path1/lib -L/path2/lib $NCURSES_ROOT/lib/libncurses.a $READLINE_ROOT/lib/libreadline.a

which means configure will only look-up in /path1/lib and /path2/lib
but it will not look-up libncurses.a and libreadline.a because this line
provides the path to libraries instead of path of lib location.

What I did will translate LDFLAGS into

LDFLAGS=-L/path1/lib -L/path2/lib -L$NCURSES_ROOT/lib -L$READLINE_ROOT/lib

as you can see now I tell LDFLAGS explicitly to look into library dirs

The configure only need paths since it tries to find libs in given lib dirs.

This slight difference cause that configure was not finding readline/ncurses
before because nobody provide lib dirs.

PR #2212, opened 2016-03-22
update t0 to 2.0.1

hufnagel:
Updates t0 spec to the latest version (which is actually used in prod). Want to finally switch production deployments to use the comp repository instead of my private one.

As the package isn't installed on cmsweb and currently no one uses the comp package this should be able to go in whenever (asap).

Question, when are the rpms actually available from the comp repository ? After the first HG1604 build ?


geneguvo:
+1 to a8a38c0f948f248b0df0276ebb835c2839e03dac on slc6_amd64_gcc493.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/2212/test_a8a38c0-20160322-113510.log" target="_blank">here</a>.

PR #2213, opened 2016-03-23
update t0 to 2.0.2

hufnagel:
Updates t0 spec to the latest version (which is actually used in prod). Want to finally switch production deployments to use the comp repository instead of my private one.

As the package isn't installed on cmsweb and currently no one uses the comp package this should be able to go in whenever (asap).

Question, when are the rpms actually available from the comp repository ? After the first HG1604 build ?

Superseeds #2212


geneguvo:
+1 to 9a1b606dfdba8eca26b11c926d02dbef080fa358 on slc6_amd64_gcc493.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/2213/test_9a1b606-20160323-134510.log" target="_blank">here</a>.

BrunoCoimbra:
Hi @hufnagel, I'm merging this PR now and the RPMs shall be available on comp repository this afternoon.

hufnagel:
Thanks @BrunoCoimbra

PR #2232, opened 2016-04-07
Updated Pede external to V04-03-04.

ghellwig:
Summary of changes: fixed uninitialized variable
- when used with option 'monitorresiduals' pede crashed when
   compiled with gcc530
- although no crash has been observed with older compilers the fix
   should be applied also there
- build was successful

### DIFF
--- a/millepede.spec
+++ b/millepede.spec
@@ -1,4 +1,4 @@
-### RPM external millepede V04-03-03 
+### RPM external millepede V04-03-04
 
 Source: svn://svnsrv.desy.de/public/MillepedeII/tags/%{realversion}/?scheme=http&module=%{realversion}&output=/%{n}-%{realversion}.tgz
 Requires: zlib


ghellwig:
My understanding is, that this will be also merged in the gcc530 branch. Is that correct?

davidlt:
Correct.

You need this change in all these release cycles?

ghellwig:
Well, it fixes a bug introduced in the last update (V04-03-03), which was applied to these release cycles.
It is certainly needed in 81X and 80X, but I created the PRs for the cycles where the update to V04-03-03 was applied to be consistent.

ghellwig:
@tlampen, @mschrode
What do you think about the above?

mschrode:
@ghellwig @davidlt We need this fix in 81X and 80X, which we will use in the upcoming alignment campaigns. The PRs for 7XX can be omitted.

mschrode:
For future reference, the validation of this new pede revision has been presented at the tracker-alignment meeting on April 6, 2016: https://indico.cern.ch/event/512679/session/0/contribution/8/attachments/1252066/1846582/cms-mp_160406.pdf

ghost:
please test

PR #2233, opened 2016-04-07
Updated Pede external to V04-03-04.

ghellwig:
Summary of changes: fixed uninitialized variable
- when used with option 'monitorresiduals' pede crashed when
   compiled with gcc530
- although no crash has been observed with older compilers the fix
   should be applied also there
- build was successful

Backport of #2232


ghost:
please test

PR #2255, opened 2016-04-15
updated utm tag to r44601-xsd310-patch

smuzaffar:

PR #2256, opened 2016-04-15
updated utm to r44601-xsd310-patch

smuzaffar:
same as #2255 which was made for production branch gcc530


smuzaffar:
Already tested here https://github.com/cms-sw/cmsdist/pull/2255

PR #2276, opened 2016-04-28
Upgrade frontier client to v 2.8.19.

### LINKED ISSUE #2245: Please upgrade latest CMSSW to frontier_client 2.8.19
These are the release notes:

v2_8_19 - 12 Apr 2016 (dwd)
- Fix bug in the function that gets the name & IP address of servers
  and proxies as a string.  This bug was introduced in v2_8_15.  The
  function is supposed to return the host name followed by the IP
  address in square brackets, however for proxies and for direct-
  connect servers it was returning only the IP address in square
  brackets.  This mostly affects warning & debug messages, but also
  causes the digital signature checking option (security=sig) to
  fail on direct-connect servers.
- Add support for $X509_CERT_DIR. If it is set and the capath option
  is not set, it will be used as the directory containing CA certs
  and CRLs instead of the default /etc/grid-security/certificates.

ghost:
Fix #2245.

### DIFF
--- a/frontier_client.spec
+++ b/frontier_client.spec
@@ -1,4 +1,4 @@
-### RPM external frontier_client 2.8.18
+### RPM external frontier_client 2.8.19
 ## INITENV +PATH PYTHONPATH %{i}/python/lib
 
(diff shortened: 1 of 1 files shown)


ghost:
please test

PR #2279, opened 2016-05-03
Update DBS Client to DBS_3_3_134 

yuyiguo:
This is to allow the DBS client work with early versions of pycurl (namely libcurl/7.28.0) so that it can run in the CMSSW environment.


geneguvo:
+1 to 9a77888c190cbaa772ddf74114766ec2284b6ad0 on slc6_amd64_gcc493.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/2279/test_9a77888-20160503-232509.log" target="_blank">here</a>.

PR #2288, opened 2016-05-11
Updated root to tip e4b3d70 of v6-06-00 patches branch

smuzaffar:


smuzaffar:
cmsbuild, please test

davidlt:
This should be done for "stable" branch.

smuzaffar:
yes I know. I did it so that tests can be run (including the comparison which we have for gcc530). Once it is tested for gcc530 I will make a PR for stable and merge that

PR #2294, opened 2016-05-13
move to pythia8 219 from tarball

mkirsano:


mkirsano:
I suggest this as an intermediate step, taking pythia8 from tarball, tater move to git mirror.

mkirsano:
This is to be tested after merging https://github.com/cms-sw/cmssw/pull/14496 (or together with it)

bendavid:
@cmsbuild please test with cmssw#14496

bendavid:
@degano did I get the syntax wrong or do I just not have permission to trigger tests here?

ghost:
@bendavid I believe the command is correct but has to be issued from cmssw repository. However I'm not entirely sure that's a good idea to have this "intermediate" version as I see no harm in creating a new branch of Pythia in cms-externals and work from there when needed.

bendavid:
Well there are a number of further changes which depend on this, and no cms patches are needed for now (and hopefully neither in the near future).  So it could even stay like this for a while/until a future pythia version.  Would be good to avoid additional delays anyways.  (We were waiting a while for this release to come from the pythia side)

bendavid:
This failure looks rather unrelated to the pythia update?

smuzaffar:
correct, looks like it was using the wrong arch slc6_amd64_gcc530 for IB/CMSSW_8_1_X/stable branch. @degano can you please look it to it.
for now I have re-started the test using the correct arch (slc6_amd64_gcc493)

bendavid:
Also seems unlikely this failure is related (and the linked summary report seems not to be there)

PR #2305, opened 2016-05-20
xcers3.1 changes for Ib/cmssw 8 1 x/gcc530next

iahmad-khan:


smuzaffar:
@iahmad-khan , looks like 3.1.3 sources are not available under current server. Please update the spec to get sources rom http://www-us.apache.org/dist//xerces/c/3/sources/xerces-c-3.1.3.tar.gz

PR #2316, opened 2016-05-30
updates heppdt.spec to use cms-externals/heppdt repo as source

iahmad-khan:


smuzaffar:
cmsbuild, please test

PR #2318, opened 2016-05-31
fix sherpa toofile to remove warning appearing in CMSSW_7_1_.. versions

pmillet:
message:
'****WARNING: Invalid tool root_cxxdefaults. Please fix sherpa file.'
due to modifications for root 6, which is not present in e.g. CMSSW_7_1_23

this does not modify the functionality of sherpa


smuzaffar:
looks good. Yes, root_cxxdefaults tool is only available in root 6 based releases.

PR #2336, opened 2016-06-12
upgrade photos++ to 3.61

mkirsano:


photospp-toolfile.spec
```diff
@@ -1,4 +1,4 @@
-### RPM external photospp-toolfile 3.56
+### RPM external photospp-toolfile 2.0
```
iahmad-khan: @mkirsano the version number is different in spec file than tool-file , shouldn't it be the same?

mkirsano: There is no such rule in cmsdist - see other toolfiles. The content of toolfiles is rarely changed.

ghost:
please test

davidlange6:
if we want to change the tool file version for some reason, it should at least increase in number, not go backwards...

photospp-toolfile.spec
smuzaffar: correct, version in toolfile has nothing to do with version of tool. toolfile version should only be changed when the tool definition is changed

smuzaffar:
agree with @davidlange6 
Although we do not deploy toolfile but it would still be nice to increase the version. I would suggest to set it to 4.0

mkirsano:
I agree with 4.0. Let's wait the test results.

mkirsano:
I think the errors are unrelated - I looked at some of them. I had doubts about compilation, but it is OK.

ghost:
please test

ghost:
@mkirsano Tests are fine. If you could please change the specfile version as discussed then we'll move on and merge this PR.

PR #2349, opened 2016-06-19
Oracle environment removal and Oracle related packages update

davidlt:
The change set removes Oracle environment package and users should move to use official packages (which are up to date). Users might need to set appropriate `SQLPATH` and `TNS_ADMIN` environment variables themselves. In addition to that update Oracle Instant Client version to `12.1.0.2.0`. One common version for macOS (x86_64) and Linux (x86_64, ppc64le, ppc64). Note, there is still no release for ARMv8 64-bit/ARM64/AArch64. At the same time update cx_Oracle Python package to the latest version.


smuzaffar:
please test

PR #2355, opened 2016-06-23
Update davix.spec

juztas:
Until 0.6.2 version davix is allocating multi-range buffers on the stack. If chunk size is bigger than the stack size, it segfaults.
Starting from 0.6.2 this was fixed to allocate on the heap. Upgrading to the newest version 0.6.3

### DIFF
--- a/davix.spec
+++ b/davix.spec
@@ -1,4 +1,4 @@
-### RPM external davix 0.6.0
+### RPM external davix 0.6.3
 
 %define tag %(echo R_%{realversion} | tr . _)
 %define branch master


smuzaffar:
please test

davidlt:
LGTM

Justas have discussed with me this issue, where Davix was allocating buffers on the stack and default stack size is different on various systems. If CMSSW requests a big enough chunk of ROOT file, it could not fit thus segfaulting. Good to see that the issue was already resolved.

smuzaffar:
@iahmad-khan , did you get this message?

PR #2367, opened 2016-07-04
updates geant4 spec file to use new geant4.10.2.p02

iahmad-khan:
For https://github.com/cms-externals/geant4/issues/10

PR #2368, opened 2016-07-04
updates spec files listed in geant4data for 81xgcc530next

iahmad-khan:
https://github.com/cms-externals/geant4/issues/10


iahmad-khan:
please test

iahmad-khan:
@davidlange6 @smuzaffar this error is in IB also:
 `1003.0 step1 DAS Error` 

while this one is not:

`140.53 step1 DAS Error`

smuzaffar:
I think geant4 PR #2367 and this one should both be tested together.
As it is for Devel, so I would suggest please merge these

PR #2370, opened 2016-07-05
adds spec files for md5 to be used as external

iahmad-khan:


smuzaffar:
addOn test error does not relate to this PR

PR #2373, opened 2016-07-05
updates geant4 spec file to use latest patch on top of 10.2.p02

iahmad-khan:
Fixes https://github.com/cms-sw/cmsdist/issues/2371


iahmad-khan:
please test

iahmad-khan:
please test

PR #2404, opened 2016-07-21
Added a new Geant4  patch to DEVEL

civanch:
Added a new Geant4 10.2 CMS-private patch to FTF-model configuration. This patch is now only for the DEVEL branch.

This include disabled Geant4 verbosity which should provide non-negligble speed-up of the production.


iahmad-khan:
@civanch first we need to merge the previous changes ( which also required new clhep version during build ) , i am compiling that locally now  because of some errors , after that is done , i will test this one.

smuzaffar:
correct PR is #2407

PR #2444, opened 2016-08-11
das_client.py symlink in CMS_PATH/share/overrides/bin

smuzaffar:
- New scram tag which adds CMS_PATH/share/overrides/bin (if exists) as first directory in PATH 
- das_client now creates CMS_PATH/share/overrides/bin/das_client.py symlink pointing to common/das_client wrapper

this is needed in order to get the old releases, which have old das_client in them and they call das_client.py instead on das_client wrapper, working

### DIFF
--- a/SCRAMV1.spec
+++ b/SCRAMV1.spec
@@ -1,4 +1,4 @@
-### RPM lcg SCRAMV1 V2_2_6_pre5
+### RPM lcg SCRAMV1 V2_2_6_pre6
 ## NOCOMPILER
 
 BuildRequires: gmake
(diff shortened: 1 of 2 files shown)


smuzaffar:
please test

smuzaffar:
thisis not going to test new SCRAM as jenkins test system uses scram from the CMS_PATH and not the one build locally. We will update the cmsdist testing to use locally build env instead of getting it from default CMS_PATH

davidlt:
Does this solve issue if we have `import das_client`?

smuzaffar:
No, it will not resolve that issue. That one can not be resolved easily as python in cmssw and needed by latest das_client could be different. We knew that whatever is using das python module will always get it from the cmssw env.

davidlange6:
right- import das_client is dead.. (I believe this was announced, or at least I asked it to be)

On Aug 11, 2016, at 10:03 AM, Malik Shahzad Muzaffar <notifications@github.com<mailto:notifications@github.com>> wrote:

No, it will not resolve that issue. That one can not be resolved easily as python in cmssw and needed by latest das_client could be different. We knew that whatever is using das python module will always get it from the cmssw env.

smuzaffar:
+externals

PR #2468, opened 2016-08-26
backports jemalloc numpy matplotlib and scikit-learn from 81devel

iahmad-khan:
For https://github.com/cms-sw/cmssw/issues/15602

### DIFF
--- a/cmake.spec
+++ b/cmake.spec
@@ -1,4 +1,4 @@
-### RPM external cmake 3.3.0
+### RPM external cmake 3.5.2
 %define downloaddir %(echo %realversion | cut -d. -f1,2)
 Source: http://www.cmake.org/files/v%{downloaddir}/%n-%realversion.tar.gz
 Requires: bz2lib curl expat zlib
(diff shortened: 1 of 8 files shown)


iahmad-khan:
please test

iahmad-khan:
@davidlange6 should we merge it now.

smuzaffar:
looks good to me. We still have some changes for heppdt.spec in devel (e.g. getting heppdt changes from out github repo instead of http://lcgapp.cern.ch).
I would suggest to make another PR to heppdt too.

davidlange6:
indeed I wasn’t sure what the heppdt changes were.. if those are safe, we can back port them too.

davidlt:
This resulted in failures on multiple architectures. matplotlib is failing to link to build png module.

davidlt:
Install step re-runs the configuration, finds out that there is no png installed on the system and fails. On the systems where png is installed, everything works.

davidlt:
I pushed a fix, but this needs a review if a more proper fix should be added.

PR #2532, opened 2016-09-26
adding histogrammar python package

davidlange6:
http://histogrammar.org 

I didn't manage to test within an actual CMSSW release - so likely there will be revisions to the PR after this one gets integrated.

### DIFF
--- a/cmssw-tool-conf.spec
+++ b/cmssw-tool-conf.spec
@@ -1,4 +1,4 @@
-### RPM cms cmssw-tool-conf 33.0
+### RPM cms cmssw-tool-conf 34.0
 ## NOCOMPILER
 # With cmsBuild, change the above version only when a new
 # tool is added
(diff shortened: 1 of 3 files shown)


davidlange6:
please test

cmssw-tool-conf.spec
```diff
@@ -175,6 +175,7 @@ Requires: freetype-toolfile
 Requires: tkonlinesw-toolfile
 Requires: py2-cx-oracle-toolfile
 Requires: oracle-toolfile
+Requires: histogrammar
```
smuzaffar, line 178: you need to add dependency on histogrammar-toolfile

rovere:
My I naively ask who's the customer of this new external?

smuzaffar:
@davidlange6 , IB/CMSSW_8_1_X/stable is also closed one :-) you need to use IB/CMSSW_8_1_X/gcc530.
I will update cms-bot to close PR for cmsdist branches which are not in use any more.

davidlange6:
@smuzaffar - I was actually more fooled by the forward porting matrix where stable appears to be at the bottom of the stack. Maybe we should also remove that (Or maybe it was commented out and I just didn't notice)..

anyway, will try again today...

smuzaffar:
you are right, we should have remove stable to gcc530 forward port map (done now).

smuzaffar:
closing it was it is not an active branch.

Issue #2527, opened 2016-09-22
External updates for 81x


davidlange6:
Hi @degano @iahmad-khan 

could you make a few small changes
LAPACK to version 3.6.1

and add

HDF5 1.8.17 https://support.hdfgroup.org/HDF5/release/obtain5.html (should not depend on anything except for zlib (and cmake if you use that build method) 

./configure --enable-shared --enable-cxx --with-zlib=/cvmfs/cms-ib.cern.ch/2016-39/slc6_amd64_gcc600/external/zlib/1.2.8-giojec2 --prefix $PWD/../install

(maybe we can add also --disable-static to reduce the size)

h5py 2.6.0 which is the corresponding python package  (https://pypi.python.org/pypi/h5py/2.6.0)

thx

ghost:
@davidlange6 Okay, I'll be looking into it now.

PR #2534, opened 2016-09-28
External updates Hdf5 and py2-h5py 

iahmad-khan:
https://github.com/cms-sw/cmsdist/issues/2527


iahmad-khan:
please test

smuzaffar:
@iahmad-khan , please also increment the cmssw-tool-conf version. When ever we update a toolfile or tool-conf then we should also update its version.

iahmad-khan:
please test

PR #2535, opened 2016-09-28
Ib/cmssw 8 1 x/stable: Rivet 2.5.2 and Yoda 1.6.4

xjanssen:
Moving to latest Rivet version 2.5.2 (and underlying Yoda 1.6.4)
I could compile the packages and test them in CMSSW_8_1_0_pre10
Also revisited the dependency as some were removed by the authors.
Xavier

### DIFF
--- a/rivet.spec
+++ b/rivet.spec
@@ -1,4 +1,4 @@
-### RPM external rivet 2.4.0
+### RPM external rivet 2.5.1
 Source: http://cern.ch/service-spi/external/MCGenerators/distribution/rivet/rivet-%{realversion}-src.tgz
 
 Requires: hepmc boost fastjet gsl yaml-cpp yoda
--- a/rivet.spec
+++ b/rivet.spec
@@ -1,7 +1,8 @@
-### RPM external rivet 2.5.1
+### RPM external rivet 2.5.2
 Source: http://cern.ch/service-spi/external/MCGenerators/distribution/rivet/rivet-%{realversion}-src.tgz
+#Source: http://www.hepforge.org/archive/rivet/Rivet-%{realversion}.tar.gz  
 
-Requires: hepmc boost fastjet gsl yaml-cpp yoda
+Requires: hepmc fastjet gsl yoda
 Requires: python cython
 
 Patch0: rivet-1.4.0
@@ -16,6 +17,7 @@ Patch0: rivet-1.4.0
 
 %prep
 %setup -n rivet/%{realversion}
+#%setup -n Rivet-%{realversion}
 %patch0 -p0
 
 ./configure --disable-silent-rules --prefix=%i --with-boost=${BOOST_ROOT} --with-hepmc=$HEPMC_ROOT \
--- a/yoda.spec
+++ b/yoda.spec
@@ -1,4 +1,4 @@
-### RPM external yoda 1.5.5
+### RPM external yoda 1.6.3
 
 Source: http://cern.ch/service-spi/external/MCGenerators/distribution/%{n}/%{n}-%{realversion}-src.tgz
 
--- a/yoda.spec
+++ b/yoda.spec
@@ -1,8 +1,10 @@
-### RPM external yoda 1.6.3
+### RPM external yoda 1.6.4
 
 Source: http://cern.ch/service-spi/external/MCGenerators/distribution/%{n}/%{n}-%{realversion}-src.tgz
+#Source: http://www.hepforge.org/archive/yoda/YODA-%{realversion}.tar.gz
 
 Requires: boost python cython
+#Requires: python cython
 
 %if "%{?cms_cxxflags:set}" != "set"
 %define cms_cxxflags -std=c++0x -O2
@@ -10,6 +12,7 @@ Requires: boost python cython
 
 %prep
 %setup -q -n %{n}/%{realversion}
+#%setup -q -n YODA-%{realversion}
 
 ./configure --prefix=%i --with-boost=${BOOST_ROOT} CXX="$(which %cms_cxx)" CXXFLAGS="%cms_cxxflags"
 


rivet.spec
```diff
@@ -1,7 +1,8 @@
-### RPM external rivet 2.4.0
+### RPM external rivet 2.5.2
 Source: http://cern.ch/service-spi/external/MCGenerators/distribution/rivet/rivet-%{realversion}-src.tgz
+#Source: http://www.hepforge.org/archive/rivet/Rivet-%{realversion}.tar.gz
```
davidlt, line 3: Why the comment (`#`)?

rivet.spec
```diff
@@ -1,7 +1,8 @@
-### RPM external rivet 2.4.0
+### RPM external rivet 2.5.2
 Source: http://cern.ch/service-spi/external/MCGenerators/distribution/rivet/rivet-%{realversion}-src.tgz
+#Source: http://www.hepforge.org/archive/rivet/Rivet-%{realversion}.tar.gz  
 
-Requires: hepmc boost fastjet gsl yaml-cpp yoda
+Requires: hepmc fastjet gsl yoda
```
davidlt, line 5: Make you that modify corresponding toolfile, i.e. `rivet-toolfile.spec` to change dependencies. That's what SCRAM uses.

rivet.spec
```diff
@@ -16,6 +17,7 @@ Patch0: rivet-1.4.0
 
 %prep
 %setup -n rivet/%{realversion}
+#%setup -n Rivet-%{realversion}
 %patch0 -p0
 
 ./configure --disable-silent-rules --prefix=%i --with-boost=${BOOST_ROOT} --with-hepmc=$HEPMC_ROOT \
```
davidlt, line 23: You just removed `boost` and `yaml-cpp`from run-time dependencies, but you ask to build with it at configure time. Clean it up.

yoda.spec
```diff
@@ -1,15 +1,18 @@
-### RPM external yoda 1.5.5
+### RPM external yoda 1.6.4
 
 Source: http://cern.ch/service-spi/external/MCGenerators/distribution/%{n}/%{n}-%{realversion}-src.tgz
+#Source: http://www.hepforge.org/archive/yoda/YODA-%{realversion}.tar.gz
 
 Requires: boost python cython
+#Requires: python cython
```
davidlt, line 7: Same here, make sure that dependencies mach-up in `yoda-toolfile.spec` for SCRAM.

yoda.spec
```diff
@@ -1,15 +1,18 @@
...
 
 %if "%{?cms_cxxflags:set}" != "set"
 %define cms_cxxflags -std=c++0x -O2
 %endif
 
 %prep
 %setup -q -n %{n}/%{realversion}
+#%setup -q -n YODA-%{realversion}
 
 ./configure --prefix=%i --with-boost=${BOOST_ROOT} CXX="$(which %cms_cxx)" CXXFLAGS="%cms_cxxflags"
```
davidlt, line 17: You have removed `boost` from run-time dependencies again, but still ask for it at configure time. In this case it will grab `boost` from system standard location if it's available. If package supports, explicitly disable boost at configuration time thus we don't pick it up from the standard system location.

davidlt (changes requested):

davidlt:
Btw, don't use `IB/CMSSW_8_1_X/stable`. You need `IB/CMSSW_8_1_X/gcc530`.

rivet.spec
xjanssen: Hi, there is nothing in the rivet and yoda toolfiles about these dependencies. Maybe it's missing since ever and I am not sure what has to go there. 

PS: I am cleaning the rest and will redo the PR in IB/CMSSW_8_1_X/gcc530.

davidlt: You can look around for examples (e.g. https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_8_1_X/gcc530/geant4-toolfile.spec) GEANT4 uses:

```
  <use name="clhep"/>
  <use name="xerces-c"/>
```

smuzaffar:
@xjanssen, please submit a PR for IB/CMSSW_8_1_X/gcc530 branch. stable branch is dead.

xjanssen:
Hi, I have one more question before I can proceed with a new PR in IB/CMSSW_8_1_X/gcc530
branch. In the mean time Yoda authors fixed the option to enable root support (that we were missing) and I have to compile enabling it. I am trying to understand what is the correct way to add root in the spec files. Is it simply 'root' or as I see in some case 'rootcore' or even 'root_cxxdefaults'. Seems we have quite some options and I am unsure what is involved in all these.
Thanks, Xavier.

davidlt:
@xjanssen if ROOT is used at run-time add `Requires: root`.  These are RPM dependencies. Things like `root_interface`, `root_cxxdefaults`, etc are SCRAM tool definitions. A single package by have multiple SCRAM tools defined. Make sure you add proper configuration time options to pick up ROOT and make sure that packages does not hardcode absolute paths to root. Anything that does hardcode path -- needs to be relocated.

xjanssen:
OK, I added 'Requires: root' to yoda.spec and checked the configure script which is getting root environment via 'root-config' calls. So I guessed that is ok and tried to build but it fails after root compilation:

Trying to install the rpm package lcg+root+6.06.08-cms just built.
Checking local path dependency for rpm package lcg+root+6.06.08-cms just build.
Requested to quit.
Requested to quit.
Requested to quit.
Requested to quit.
- The action "build-external+yoda+1.6.5" was not completed successfully because The following dependencies could not complete:
  install-lcg+root+6.06.08-cms
- The action "install-external+yoda+1.6.5" was not completed successfully because The following dependencies could not complete:
  build-external+yoda+1.6.5
- The action "final-job" was not completed successfully because The following dependencies could not complete:
  install-lcg+root+6.06.08-cms
- The action "install-lcg+root+6.06.08-cms" was not completed successfully because Traceback (most recent call last):
  File "/afs/cern.ch/work/x/xjanssen/cms/Rivet/81X_gcc530/PKGTOOLS/scheduler.py", line 199, in doSerial
  result = commandSpec[0](*commandSpec[1:])
  File "PKGTOOLS/cmsBuild", line 2917, in installPackage
  File "PKGTOOLS/cmsBuild", line 2757, in installRpm
  RpmInstallFailed: Failed to install package root. Reason:
  error: Failed dependencies:
  /bin/perl is needed by lcg+root+6.06.08-cms-1-1.x86_64

davidlt:
Why do you need to rebuild ROOT? The error means that your local build of root builds something that refers to `/bin/perl` thus dependency is added and Perl is not part of CMSSW bootstrap. This most likely means that your build host is different from CMS builder machines (e.g. contains extra packages installed). You should probably use `--repository cms.week1` for `cmsBuild`. We have alternating repositories for CMSSW IB: `cms.week0` and `cms.week1` (final stuff lands in `cms` once release is built). This way you will need to rebuild less.

For the build machine I would advice to look at https://github.com/cms-sw/cms-docker/ and use `slc7-builder` or `slc6-builder` Docker containers. They container the same setup used by CMSSW IB infrastructure.

xjanssen:
Hmm, I am trying to build at CERN on a lxplus host (slc6), so I can't really modify the installed package. I used the following environment to extract CMSDIST/PKGTOOL:

SCRAM_ARCH=slc6_amd64_gcc530;PKGTOOLS_TAG=V00-24-XX;CMSDIST_TAG=IB/CMSSW_8_1_X/gcc530;RELEASE_QUEUE=CMSSW_8_1_X

So I am unsure what went wrong such it decided to rebuild root in fact

davidlt:
It's very good way of figuring out which CMSDIST and PKGTOOLS branches to use, but I would avoid using lxplus machines, because they are configured differently.

Again, you can try with `--repository cms.week1` and as long as you haven't changed `root.spec` it should download the latest RPM and install instead of building local version of ROOT.

xjanssen:
OK, indeed even doing:
PKGTOOLS/cmsBuild --new-scheduler -i c -a $SCRAM_ARCH --builders 4 build yoda --repository cms.week1
on lxplus, it starts building root. So can you advice what machine to use because in the past I never ran into the need of using special ones to build these external packages and did build/test them on lxplus only.

davidlt:
Try using `cmsdev{01,02,03,04}` machines.

davidlange6:
though root should not be being rebuilt in this case… (As David already mentioned)

xjanssen:
Yes, using cmsdev01 machine and cms.week1 repository, it did not rebuild root (but it did on lxplus) and I am moving forward with compiling and testing yoda/rivet now, let see ...

PR #2548, opened 2016-10-03
added gcc-atomic toolfile

smuzaffar:


smuzaffar:
please test

smuzaffar:
@davidlt , should we explicitly add LD_LIBARY_PATH runtime pointing to install-path/gcc/530/lib64 for this tool or is it correct assumption that gcc env is enough?

PR #2581, opened 2016-10-13
fixes py2-h5py dependencies

iahmad-khan:


iahmad-khan:
please test

iahmad-khan:
@davidlt this should fix the py2-h5py dependency issues.

cython.spec
```diff
@@ -21,5 +21,3 @@ sed -ideleteme 's|#!.*/bin/python|#!/usr/bin/env python|' \
   %{i}/${PYTHON_LIB_SITE_PACKAGES}/Cython/Debugger/libpython.py
 
 find %{i} -name '*deleteme' -delete
-
-find %{i}/${PYTHON_LIB_SITE_PACKAGES} -name '*.egg-info' -print0 | xargs -0 rm -f
```
davidlt, line 25: Why this is removed? Long-term policy was that we do not provide _egg_ stuff.

py2-h5py.spec
```diff
@@ -2,8 +2,8 @@
 ## INITENV +PATH PYTHONPATH %{i}/$PYTHON_LIB_SITE_PACKAGES
 %define my_name %(echo %n | cut -f2 -d-)
 Source: https://github.com/%{my_name}/%{my_name}/archive/%{realversion}.tar.gz
-Requires: python py2-numpy hdf5
-BuildRequires: py2-setuptools hdf5
+Requires: python py2-numpy hdf5 py2-six
+BuildRequires: py2-setuptools cython py2-pkgconfig
```
davidlt, line 6: Have you looked into how pkgconfig is being used? We need to ensure it does not go to look into the system.

py2-six.spec
```diff
@@ -13,4 +13,3 @@ python setup.py build
 
 %install
 python setup.py install --single-version-externally-managed --record=/dev/null --skip-build --prefix=%{i}
-find %{i}/${PYTHON_LIB_SITE_PACKAGES} -name '*.egg-info' -print0 | xargs -0 rm -rf
```
davidlt, line 16: Again, way we keeping the eggs?

cython.spec
iahmad-khan: actually the h5py setup has no way to find our cython without this.

py2-six.spec
iahmad-khan: same reason as above

py2-h5py.spec
iahmad-khan: yes

smuzaffar: I think pkgconfig is only needed if you call setup.py test which we are not calling at build time (may be we should)

PR #2586, opened 2016-10-14
add professor2

mkirsano:


perrozzi:
please test

professor2-toolfile.spec
```diff
@@ -0,0 +1,22 @@
...
+%install
+
+mkdir -p %i/etc/scram.d
+cat << \EOF_TOOLFILE >%i/etc/scram.d/professor2.xml
+<tool name="professor2" version="@TOOL_VERSION@">
+<client>
+<environment name="PROFESSOR2_BASE" default="@TOOL_ROOT@"/>
+</client>
+<runtime name="PATH" value="$PROFESSOR2_BASE/bin" type="path"/>
+<runtime name="PYTHONPATH" value="$PROFESSOR2_BASE/lib/python@PYTHONV@/site-packages" type="path"/>
```
davidlt, line 23: You are missing these dependencies listed here: py2-numpy py2-sympy py2-matplotlib root cython yoda eigen

cmssw-tool-conf.spec
```diff
@@ -119,6 +119,7 @@ Requires: fftjet-toolfile
 Requires: lapack-toolfile
 Requires: pyminuit2-toolfile
 Requires: professor-toolfile
+Requires: professor2-toolfile
```
davidlt, line 122: Do we need to keep old version of professor?

professor2.spec
```diff
@@ -0,0 +1,16 @@
...
+
+Requires: py2-numpy py2-sympy py2-matplotlib root cython yoda eigen
+%prep
+%setup -n Professor-%{realversion}
+
+%build
+make CPPFLAGS=-I${EIGEN_ROOT}/include
+
+%install
+make install PREFIX=%{i}
```
davidlt, line 13: Does professor2 and professor provide the same binaries and shared libraries? If yes, then we might have an issue.

cmssw-tool-conf.spec
mkirsano: There was a request to keep old version

professor2.spec
mkirsano: The names of binaries (actually scripts) and shared libraries are different (actually the old professor has no library)

cmssw-tool-conf.spec
davidlange6: @mkirsano -please summarize the request here (and its expected longevity) - its certainly not normal to try to manage multiple versions of the "same" tool around.

mkirsano: professor2 is very different from professor. It does not yet have all functionality of professor. This is the reason to keep the old version. Note that CMSSW code is not compiled against this external.

smuzaffar:
testing

mkirsano:
still testing?

smuzaffar:
please test
@davidlange6 , this looks good to co-exist with professor.

smuzaffar:
@mkirsano, I see that there is include directory and lib/libProfessor2.so ? Do you want to make these part of professor2-toolfile?

mkirsano:
This is not needed right now (however it would not do any harm).

mkirsano:
Is this still waiting for some other upgrades (python for example)?
Is there automatically the corresponding PR in 90X or it is to be made?

PR #2611, opened 2016-10-28
enables tkinter support for matplotlib

iahmad-khan:
https://github.com/cms-sw/cmssw/issues/15660


iahmad-khan:
please test

smuzaffar:
@iahmad-khan , you need to modify pythion.spec too in order to get it rebuild and pick tk-devel from system.
Just add a comment in python.spec so that it triggers rebuild of it.

iahmad-khan:
please test

PR #2615, opened 2016-11-01
DQMGUI 8.4.5

rovere:
This will bring in DQMGUI 8.4.5 which contains an important bug fix that will prevent the crashes of the JSON API that were recently happening at a high rate on production servers (mainly offline)


geneguvo:
+1 to 1e1e1ab22111dfa303aad5c1ade9dd56ef693cde on slc6_amd64_gcc493.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/2615/test_1e1e1ab-20161101-175505.log" target="_blank">here</a>.

PR #2678, opened 2016-11-25
Python via pip in cmsdist

davidlange6:
Introduce the ability to build python packages from pip. More testing is certainly needed, however current implementation checks for any dependencies being installed by pip and fails in the case that there are (so still a one to one mapping of python package to spec file). 

Two examples are added (py2-configparser and py2-entrypoints) that confirm that dependencies are handled correctly (both are part of the stack of dependencies that jupiter depends on).

The main outstanding issue is to retain the source somehow..

Otherwise, I add rootpy (which I will eventually change to install via pip...)

### DIFF
--- a/build-with-pip.file
+++ b/build-with-pip.file
@@ -9,7 +9,7 @@ BuildRequires: py2-pip
 %build
 
 pip list
-pip install -t %{i}/$PYTHON_LIB_SITE_PACKAGES  %{my_name}==%{realversion}
+pip install --no-deps -t %{i}/$PYTHON_LIB_SITE_PACKAGES  %{my_name}==%{realversion}
 #see how many things got installed
 if [ `ls -d %{i}/$PYTHON_LIB_SITE_PACKAGES/*info | wc -l` == "1" ] ; then
    echo "Pip installed just what you asked for"
--- a/build-with-pip.file
+++ b/build-with-pip.file
@@ -1,20 +1,20 @@
 %define my_name %(echo %n | cut -f2 -d-)
 Source: none
 
+Requires: python
 BuildRequires: py2-pip
 
 %prep
 
 %build
 
-%define pythonVersion %(python -V |& cut -f2 -d' ' | cut -f1-2 -d.)
-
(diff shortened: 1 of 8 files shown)


build-with-pip.file
```diff
@@ -0,0 +1,25 @@
...
+
+Requires: python
+BuildRequires: py2-pip
+
+%prep
+
+%build
+
+pip list
+pip install --no-deps -t %{i}/$PYTHON_LIB_SITE_PACKAGES  %{my_name}==%{realversion}
```
smuzaffar: @davidlange6 , try the following. It allows pip to use our py2 tools
```
export PYTHONUSERBASE=%i
pip list
pip install --user %{my_name}==%{realversion}
```
and pip sysy
```
+ pip install --user entrypoints==0.2.2
Collecting entrypoints==0.2.2
  Using cached entrypoints-0.2.2-py2.py3-none-any.whl
Requirement already satisfied: configparser>=3.5; python_version == "2.7" in /var/build/muz/pip/w/slc6_amd64_gcc530/external/py2-configparser/3.5.0/lib/python2.7/site-packages (from entrypoints==0.2.2)
Installing collected packages: entrypoints
Successfully installed entrypoints-0.2.2
```

davidlange6:
Needs https://github.com/cms-sw/pkgtools/pull/134
@smuzaffar -can you offer more advice for cleaning this up?

smuzaffar:
@davidlange6 , looks good to me.
Looks like this is going to grow the PYTHONPATH a lot. Should we introduce a meta package for example py2-pip-packages.spec which basically depend on all the py2-* pip based packages and all it does is to rsync py2 packages in on single directory and we make cmssw only depend on this mate package. This way we will have only one directory (for these pip based packages) in PYTHONPATH?

smuzaffar:
please test
pkgtools changes are now in, so test should work now

davidlange6:
ah - must be me… will look tomorrow..

smuzaffar:
ok, was going to remind you

davidlange6:
please test

davidlange6:
Interface change that my unit tests need to adapt to..

PR #2680, opened 2016-11-25

fwyzard:


fwyzard:
please test

fwyzard:
This should be included in CMSSW only on top of https://github.com/cms-sw/cmssw-config/pull/56 .

smuzaffar:
no need to have it for gcc620. Once included in gcc530 then it should be forward ported to gcc620

PR #2681, opened 2016-11-25

fwyzard:


fwyzard:
please test

fwyzard:
This should be included in CMSSW only on top of https://github.com/cms-sw/cmssw-config/pull/56 .

intel-vtune.spec
```diff
@@ -0,0 +1,25 @@
...
+
+mkdir -p %i/etc/scram.d
+cat << \EOF_TOOLFILE >%i/etc/scram.d/intel-vtune.xml
+<tool name="intel-vtune" version="%{realversion}">
+  <info url="https://software.intel.com/en-us/intel-vtune-amplifier-xe"/>
+  <lib name="ittnotify"/>
+  <client>
+    <environment name="INTEL_VTUNE_BASE" default="/afs/cern.ch/sw/IntelSoftware/linux/x86_64/xe2017/vtune_amplifier_xe_%{realversion}"/>
+    <environment name="INCLUDE" default="$INTEL_VTUNE_BASE/include"/>
+    <environment name="LIBDIR" default="$INTEL_VTUNE_BASE/lib64"/>
```
smuzaffar, line 17: this directory contains libstdc++. this can cause problems. Looks like vtune executables linked with full path e.g.
```
        libstdc++.so.6 => /afs/cern.ch/sw/IntelSoftware/linux/x86_64/xe2017/vtune_amplifier_xe_2017.0.2.478468/bin64/./../lib64/libstdc++.so.6 (0x00007f4003603000)
```
So I guess we can drop it and instead add
```
<flags LDFLAGS="$(INTEL_VTUNE_BASE)/lib64"/>
```

fwyzard:
@smuzaffar I see your point, however that does not seem to happen: after setting up the intel-vtune tool in CMSSW, I do not see its library path added to LD_LIBRARY_PATH:
```
$ cmsenv
$ echo $LD_LIBRARY_PATH | tr : \\n
/data/user/fwyzard/IntelWorkshop/CMSSW_9_0_X_2016-11-24-2300/biglib/slc6_amd64_gcc530
/data/user/fwyzard/IntelWorkshop/CMSSW_9_0_X_2016-11-24-2300/lib/slc6_amd64_gcc530
/data/user/fwyzard/IntelWorkshop/CMSSW_9_0_X_2016-11-24-2300/external/slc6_amd64_gcc530/lib
/cvmfs/cms-ib.cern.ch/week0/slc6_amd64_gcc530/cms/cmssw-patch/CMSSW_9_0_X_2016-11-24-2300/biglib/slc6_amd64_gcc530
/cvmfs/cms-ib.cern.ch/week0/slc6_amd64_gcc530/cms/cmssw-patch/CMSSW_9_0_X_2016-11-24-2300/lib/slc6_amd64_gcc530
/cvmfs/cms-ib.cern.ch/week0/slc6_amd64_gcc530/cms/cmssw-patch/CMSSW_9_0_X_2016-11-24-2300/external/slc6_amd64_gcc530/lib
/cvmfs/cms-ib.cern.ch/2016-48/slc6_amd64_gcc530/cms/cmssw/CMSSW_9_0_X_2016-11-22-2300/biglib/slc6_amd64_gcc530
/cvmfs/cms-ib.cern.ch/2016-48/slc6_amd64_gcc530/cms/cmssw/CMSSW_9_0_X_2016-11-22-2300/lib/slc6_amd64_gcc530
/cvmfs/cms-ib.cern.ch/2016-48/slc6_amd64_gcc530/external/llvm/4.0.0-oenich/lib64
/cvmfs/cms-ib.cern.ch/2016-48/slc6_amd64_gcc530/external/gcc/5.3.0/lib64
/cvmfs/cms-ib.cern.ch/2016-48/slc6_amd64_gcc530/external/gcc/5.3.0/lib
```

smuzaffar:
@fwyzard , ah right. SCRAM only adds the path in LD_LIBRARY_PATH if tool has explicitly ask it to do so 
```
<runtime name="LD_LIBRARY_PATH" value="$TOOL_BASE/lib64" type="path"/>
```
so this toolfile should be fine.

intel-vtune.spec
```diff
@@ -0,0 +1,25 @@
...
+    <environment name="BINDIR" default="$INTEL_VTUNE_BASE/bin64"/>
+  </client>
+  <runtime name="PATH" value="$INTEL_VTUNE_BASE/bin64" type="path"/>
+  <runtime name="VTUNE_AMPLIFIER_XE_2017_DIR" value="$INTEL_VTUNE_BASE"/>
+  <runtime name="INTEL_LICENSE_FILE" value="28518@lxlicen01.cern.ch,28518@lxlicen02.cern.ch,28518@lxlicen03.cern.ch" type="path" handler="warn"/>
+</tool>
+EOF_TOOLFILE
+
+## IMPORT scram-tools-post
```
smuzaffar: I am not sure, if this is going to create the correct init.sh and init.csh files. As you do not have an explicit -toofile, so I would suggest to not import scram-tools-post but just add explicit %post section like https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_9_0_X/gcc530/scram-tools-post.file#L19 (use INTEL_VTUNE instead of %{BaseTool}_TOOLFILE).

intel-vtune.spec
```diff
@@ -0,0 +1,25 @@
...
+  <lib name="ittnotify"/>
+  <client>
+    <environment name="INTEL_VTUNE_BASE" default="/afs/cern.ch/sw/IntelSoftware/linux/x86_64/xe2017/vtune_amplifier_xe_%{realversion}"/>
+    <environment name="INCLUDE" default="$INTEL_VTUNE_BASE/include"/>
+    <environment name="LIBDIR" default="$INTEL_VTUNE_BASE/lib64"/>
+    <environment name="BINDIR" default="$INTEL_VTUNE_BASE/bin64"/>
+  </client>
+  <runtime name="PATH" value="$INTEL_VTUNE_BASE/bin64" type="path"/>
+  <runtime name="VTUNE_AMPLIFIER_XE_2017_DIR" value="$INTEL_VTUNE_BASE"/>
+  <runtime name="INTEL_LICENSE_FILE" value="28518@lxlicen01.cern.ch,28518@lxlicen02.cern.ch,28518@lxlicen03.cern.ch" type="path" handler="warn"/>
```
smuzaffar: drop this one too, I will set it via intel-license toolfile

fwyzard: OK - but there is not such toolfile for the time being. Do you want me to add one ?

smuzaffar: I have added a one here 
https://github.com/cms-sw/cmsdist/commit/7cfbfa592f8a8dfeb77d5058423ed82a44c86871 
it should be available in next IB

intel-vtune.spec
fwyzard: You mean like
```
%post
if [ "$CMS_INSTALL_PREFIX" = "" ] ; then CMS_INSTALL_PREFIX=$RPM_INSTALL_PREFIX; export CMS_INSTALL_PREFIX; fi
%{relocateConfig}etc/scram.d/*.xml
echo "INTEL_VTUNE_ROOT='$CMS_INSTALL_PREFIX/%{pkgrel}'" > $RPM_INSTALL_PREFIX/%{pkgrel}/etc/profile.d/init.sh
echo "set INTEL_VTUNE_ROOT='$CMS_INSTALL_PREFIX/%{pkgrel}'" > $RPM_INSTALL_PREFIX/%{pkgrel}/etc/profile.d/init.csh
```
?

But there is nothing to relocate, and there is nothing that makes use of $INTEL_VTUNE_ROOT .
Maybe I should just have an empty `%post` section ?

smuzaffar: scramv1-tool-conf.file (which collects all toolfiles and bundle them in tool-conf) needs to access the ```<tool>_ROOT``` env. So each tool should set it. 
If you provide an empty %post section then cmsBuild is anyway going to generate init.sh with default values for ROOT, VERSION, PATH, LD_LIBRARY_PATH etc (just a standard init.sh file). As for toofiles we do not need any of these that is why we re-create init.* with only _ROOT variable in them.

fwyzard:
Understood, thanks. I'll fix it.

fwyzard:
please test

fwyzard:
The error does not seem to come from this PR ?

smuzaffar:
@fwyzard , forgot to mention that you need to add ```intel-vtune``` dependency in cmssw-tool-conf.spec file. Without that this tool is not going to be part of cmssw.  You should remove INTEL_LICENSE_FILE too which is now available in cmssw 90C IBs via intel-license tool.

fwyzard:
please test

fwyzard:
please test

fwyzard:
> @fwyzard , forgot to mention that you need to add intel-vtune dependency in cmssw-tool-conf.spec file. Without that this tool is not going to be part of cmssw. You should remove INTEL_LICENSE_FILE too which is now available in cmssw 90C IBs via intel-license tool.

Both should be done. 

Should there be an explicit dependency on intel-license ?  How does one make sure it ends up in the "selected" tools instead of the "available" ones ?

smuzaffar:
All tools build via cmssw-tool-conf are by default selected unless explicitly ignored. intel-license is not ignored , so it should be selected and should already be visible in 90X IBs.

Issue #2686, opened 2016-11-28
Update fastjet-contrib to 1.025


mseidel42:
Hi all,

would it be possible to update fastjet-contrib from 1.020 to 1.025?
It includes some new energy correlations, see arXiv:1609.07483v1

Cheers,
Markus

iahmad-khan:
@intrepid42 the new version 1.25 have some problems with cmssw:

https://github.com/cms-sw/cmsdist/pull/2687#issuecomment-263540927


can you propose changes by making a pull request to either https://github.com/cms-externals/fastjet-contrib/tree/cms/v1.25 or cmssw

mseidel42:
Hm, it seems that the ConstituentSubtractor::use_common_bge_for_rho_and_rhom() disappeared with version > 1.0.0 of that contrib :(

smuzaffar:
@intrepid42 , should we close this issue or can you propose changes for cmssw?

davidlange6:
closing given no response here.

mseidel42:
Hi, I can just remove the call to use_common_bge_for_rho_and_rhom() so that everything compiles but I do not know if it is needed by someone/something.

@rappoccio, as you added the ConstituentSubtraction, could you comment, please?

rappoccio:
If the use_common_bge_for_rho_and_rhom method is gone it can probably just be commented out. Granted, basically no one is really using this anyway. 

We would still need to go through the exercise of ensuring thread safety for the contrib as we've done in the past.

PR #2687, opened 2016-11-29
updates fastjet-contrib to v1.25

iahmad-khan:
https://github.com/cms-sw/cmsdist/issues/2686


iahmad-khan:
please test

intrepid42:
Hi, I can just remove the call to use_common_bge_for_rho_and_rhom() so that everything compiles but I do not know the physics motivation behind that option -> it might be needed by someone.

@rappoccio, as you added the ConstituentSubtraction, could you comment, please?

PR #2692, opened 2016-11-29
added herwigpp bin in PATH; created herwig wrapper to use --repo option

smuzaffar:


smuzaffar:
please test

smuzaffar:
this is just a rebuild of herwigpp which properly relocate various herwig files and creates Herwig wrapper

PR #2704, opened 2016-12-05

fwyzard:
Add back a tool file for Intel VTune.
It does not export the ittnotify library dependency any more (that lives in the ittnotify tool) and is an optional tool.

The idea is for people to be able to just do
```
scram setup intel-vtune
cmsenv 
```
in order to use it.


fwyzard:
please test

PR #2720, opened 2016-12-08
adds py2-pippkgs to manage pip based python packages

iahmad-khan:


smuzaffar:
@davidlange6 , this is the meta package for pip based tools.

PR #2741, opened 2016-12-14
config tag updated; generate llvm-ccdb for cmssw projects

smuzaffar:
New cmssw-config tag V05-05-20 avoids picking up makefile fragments for deleted packages from release,


smuzaffar:
please test

PR #2748, opened 2016-12-20
Herwig7: Changed include directive in toolfile

mharrend:
I just adjusted the include directive in the toolfile of ThePEG and Herwig7, so that I can make a PR for the upcoming Herwig7 interface soon.

Since this PR does not break anything, testing is not required IMHO. You can just merge this PR.


davidlange6:
please test

PR #2753, opened 2017-01-03
move to pip for psutil python package

davidlange6:
for devel


davidlange6:
@davidlt - i'll move psutil to pip, test and then put into mainstream..

davidlange6:
please test

PR #2760, opened 2017-01-11
backport of Sherpa v2.2.1

pmillet:
As discussed in GEN technical meeting: backport of newest Sherpa version v2.2.1, which is already present in 81X and 90X.

### DIFF
--- a/sherpa.spec
+++ b/sherpa.spec
@@ -1,5 +1,5 @@
-### RPM external sherpa 2.2.0
-%define tag d905c22cc12b11b635a2b62766cfbe0b51133ba2
+### RPM external sherpa 2.2.1
+%define tag bc0ccbfdf07f4df4f0368106aa9c793bb61e1def
 %define branch cms/v%realversion
 %define github_user cms-externals
 Source: git+https://github.com/%github_user/%{n}.git?obj=%{branch}/%{tag}&export=%{n}-%{realversion}&output=/%{n}-%{realversion}-%{tag}.tgz


smuzaffar:
@cmsbuild, please test

pmillet:
I'm a bit puzzled by this. It seemed to work fine when I tried building Sherpa. Also I can't make sense of the error message. Is it related to Sherpa?

davidlt:
According to the page `CMSSW_7_1_X_2017-01-08-0800` had all add-on tests passed. If only Sherpa was updated one could assume that it could be a culprit. Someone needs to investigate.

pmillet:
I'll test building it again and report the result.

smuzaffar:
I have forced re-built 71X IB today to see it this addOn test is failing in IB too
https://cmssdt.cern.ch/SDT/html/showIB.html

smuzaffar:
@cmsbuild, please test
addOn tests works fine in IB, may be it was glitch during the previous run

smuzaffar:
I can run the failing addOn test in my local build.
@davidlange6, if every thing else look good for this PR then I would suggest that merge it and see how it behaves in next IB ?

pmillet:
I tried building it again locally. Did not see any problem.

smuzaffar:
@davidlange6 , this is failing due to some issue with cmsdist PR testing (still not understood why). I have built full cmssw 71X IB locally with openloop and sherpa updated and failing addOn works there.

pmillet:
@davidlange6 Should we continue with @smuzaffar suggestion or try something different?

PR #2791, opened 2017-01-26
add repoze.lru

davidlange6:
(I'm not sure if '.' will be allowed in the package name. if not, I'll need to change the pip build script to handle the case where the pip package name differs from the rpm spec file name


davidlange6:
please test

davidlange6:
please test

davidlange6:
please test

davidlange6:
please test

smuzaffar:
@davidlange6 , spec file name and name of the package should be same. 
why you call the spce  py2-repoze--lru.spec  (two - in the name?). the first line of the spec should also match the name, right now you have 
```### RPM external repoze.lru 0.6```

File: build-with-pip.file [hunk 1]
```diff
@@ -1,5 +1,7 @@
 %define my_name %(echo %n | cut -f2 -d-)
-Source: pip://%{my_name}/%{realversion}/source.tar.gz
+#assume that -- will never occur in a pip package name...
+%define pip_name %(echo %my_name | sed 's/--/./g')
+Source: pip://%{pip_name}/%{realversion}/source.tar.gz
```
smuzaffar, line 4: may be instead of putting this logic here, we can have
```
%if "%{?my_name:set}" != "set"
%endif
```
and then each package can set it to whatever they want in their spec file.

davidlange6: yes, this works better ... will make a new PR with this change

PR #2797, opened 2017-02-01
 pip package development

davidlange6:
allow pip package name to differ from rpm spec file name in case of special characters in the pip package name

add python stack for jupyter

I need to add a follow on commit that removes some now obsolete tool files and tools from cmssw tool conf before testing


davidlange6:
please test

davidlange6:
weird that I didn't see this locally..

davidlange6:
ok, now I can reproduce.. investigating.

davidlange6:
please test

davidlange6:
abort

davidlange6:
i fooled myself into thinking it worked locally - but not yet the case

davidlange6:
please test

davidlange6:
@smuzaffar - any issues you see?

build-with-pip.file
```diff
@@ -1,4 +1,10 @@
+
+%if 0%{?pip_name:1}
+%define my_name %{pip_name}
+%else
 %define my_name %(echo %n | cut -f2 -d-)
+%endif
+
```
smuzaffar: I would have done
```
%if "%{?pip_name:set}" != "set"
%define pip_name  %(echo %n | cut -d- -f2-5)
%endif
```

and use pip_name in the rest of the file (instead of my_name).

davidlange6:
yes, that is better.

py2-pippkgs.spec
```diff
@@ -18,9 +67,14 @@ for pkg in %builddirectpkgreqs ; do
...
     echo "Checking for duplicates ...."
     for f in $(ls $SOURCE) ; do
-      if [ -e %{i}/$PYTHON_LIB_SITE_PACKAGES/$f ] ; then
-        echo "  Duplicate file found: $f"
-        exit 1
+      if [ -e %{i}/$PYTHON_LIB_SITE_PACKAGES/$f ] ; then 
+        # https://github.com/jupyter/jupyter_core/issues/55
+        #backports is an example directory that can have multiple packages inside
+        if [ $f != "backports" && $f != "jupyter.py" ] ; then  
+           echo "  Duplicate file found: $f"
```
smuzaffar: so there are multiple package providing jupyter.py? May be better to update one package to delete its jupyter.py after the build.

davidlange6:
> On Feb 2, 2017, at 3:02 PM, Malik Shahzad Muzaffar <notifications@github.com> wrote:
> 
> @smuzaffar commented on this pull request.
> 
> In py2-pippkgs.spec:
> 
> > @@ -18,9 +67,14 @@ for pkg in %builddirectpkgreqs ; do
>    if [ -d $SOURCE ] ; then
>      echo "Checking for duplicates ...."
>      for f in $(ls $SOURCE) ; do
> -      if [ -e %{i}/$PYTHON_LIB_SITE_PACKAGES/$f ] ; then
> -        echo "  Duplicate file found: $f"
> -        exit 1
> +      if [ -e %{i}/$PYTHON_LIB_SITE_PACKAGES/$f ] ; then 
> +        # https://github.com/jupyter/jupyter_core/issues/55
> +        #backports is an example directory that can have multiple packages inside
> +        if [ $f != "backports" && $f != "jupyter.py" ] ; then  
> +           echo "  Duplicate file found: $f"
> 
> so there are multiple package providing jupyter.py? May be better to update one package to delete its jupyter.py after the build.
> 

ok, will think about that - I documented the github discussion on the topic

davidlange6:
please test

PR #2822, opened 2017-02-11
adding Keras/Theano

davidlange6:
splitting off pippkgs into a second file for things that depend on scipy (which will increase)


davidlange6:
please test

PR #2828, opened 2017-02-16
include boost chrono

fwyzard:
* build boost chrono
* add a boost_chrono tool to scram

### DIFF
--- a/boost-toolfile.spec
+++ b/boost-toolfile.spec
@@ -1,4 +1,4 @@
-### RPM external boost-toolfile 1.1
+### RPM external boost-toolfile 1.2
 Requires: boost
 %prep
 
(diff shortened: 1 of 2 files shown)


fwyzard:
I would like to use boost chrono for its `boost::chrono::thread_clock`

fwyzard:
please test

fwyzard:
Do you think we can merge this in by pre5 ?
It's necessary to have per-thread cpu time measurements, which in turn is needed to support multithreading in the FastTimerService .

smuzaffar:
@fwyzard , I do not see any issue with it.
@Dr15Jones , @davidlange6 any objections?

Dr15Jones:
I'm fine with it.

boost-toolfile.spec
```diff
@@ -28,6 +28,15 @@ cat << \EOF_TOOLFILE >%i/etc/scram.d/boost.xml
 </tool>
 EOF_TOOLFILE
 
+# boost_chrono toolfile
+cat << \EOF_TOOLFILE >%i/etc/scram.d/boost_chrono.xml
+<tool name="boost_chrono" version="@TOOL_VERSION@">
+  <info url="http://www.boost.org"/>
+  <lib name="@BOOST_CHRONO_LIB@"/>
+  <use name="boost"/>
```
smuzaffar: @fwyzard , looks like boost_chrono  needs boost_system
``` ldd libboost_chrono.so | grep boost
libboost_system.so.1.57.0 => not found
```
so please add dependency on ```<use name="boost_system"/>``` boost_system any way depends on boost.

can you also please update the version number to 1.2 of this spec?

fwyzard:
done !

fwyzard:
please test

PR #2860, opened 2017-02-27
Adding deepish python package

davidlange6:
which depends on tables. (looks like the rest would go away with a rebase - sorry)


davidlange6:
please test

PR #2864, opened 2017-02-27
use icc from /cvmfs/projects.cern.ch

smuzaffar:
To avoid AFS dependency we now have intel compiler available under /cvmfs/projects.cern.ch


smuzaffar:
please test

PR #2883, opened 2017-03-06
bump protobuf version

davidlange6:


smuzaffar:
please test

davidlange6:
I would think we'll have the same testing errors here as in the mainstream - but lets find out..

smuzaffar:
@davidlange6 , https://github.com/cms-sw/cmssw/pull/17798 should fix it.

rovere:
These files have to be re-generated every-time you change protoc.
A possibility is to include their "compilation" out of the proto-message in the buildfile itself.

davidlange6:
thanks! I'll add a note in the protobuf.spec file so we remember. since update rarely I don't think more work is needed to automate this

davidlt:
Looks like at the end you had to bump protobuf version ;)

PR #2886, opened 2017-03-06
upgrade t0wmadatasvc to 1.0.2

hufnagel:
Updates t0wmadatasvc to 1.0.2, the only change is to return an additional field in a REST API call which was already in the database but was mistakenly left out in the retrieval.

New version built in private repo and deployed on private cmsweb VM, work there.

Hope it can still make it into the Wednesday deployment.


BrunoCoimbra:
+1 to 7a50a9de29c6aad4278e20f2fe51f776836657a6 on slc6_amd64_gcc493.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/2886/test_7a50a9d-20170306-211506.log" target="_blank">here</a>.

PR #2708, opened 2016-12-06
Herwig7: Adding GoSam and update to newer Herwig7 version

mharrend:
In this PR besides updating of Herwig7 to Herwig7.0.4 from Herwig7.0.2,
also GoSam is added as a ME provider. This ME provider is necessary since it brings with itself the golem95 library which is also required by MG5aMC@NLO.

PR #2889, opened 2017-03-07
Adding GoSam and update to newer Herwig7

iahmad-khan:
https://github.com/cms-sw/cmsdist/pull/2708

### DIFF
--- a/form.spec
+++ b/form.spec
@@ -1,0 +1,33 @@
+### RPM external form 4.1.033e
+Source: https://gosam.hepforge.org/gosam-installer/form-%{realversion}.tar.gz
+
+
+%if "%{?cms_cxx:set}" != "set"
+%define cms_cxx g++
+%endif
+
+
+%prep
+%setup -q -n form-4.1
+
+
+%build
+
+
+
+CXX="$(which %{cms_cxx}) -fPIC"
+CC="$(which gcc) -fPIC"
+PLATF_CONF_OPTS="--enable-shared --disable-static"
+
+./configure $PLATF_CONF_OPTS \
+            --prefix=%i \
+            --bindir=%i/bin \
+            --without-gmp \
+            CXX="$CXX" CC="$CC" CXXFLAGS=-fpermissive
+
+make %makeprocesses
+
+%install
+make install 
+
+
--- a/form.spec
+++ b/form.spec
@@ -1,19 +1,14 @@
 ### RPM external form 4.1.033e
 Source: https://gosam.hepforge.org/gosam-installer/form-%{realversion}.tar.gz
 
-
 %if "%{?cms_cxx:set}" != "set"
 %define cms_cxx g++
 %endif
 
-
 %prep
 %setup -q -n form-4.1
 
-
 %build
-
-
 
 CXX="$(which %{cms_cxx}) -fPIC"
 CC="$(which gcc) -fPIC"
--- a/madgraph5amcatnlo.spec
+++ b/madgraph5amcatnlo.spec
@@ -35,6 +35,18 @@ rm bin/compile.py
 # Save patched config
 mv input/mg5_configuration_patched.txt input/mg5_configuration.txt
 
+# Start small NLO event generation to make sure that all additional packages are compiled
+cat <<EOF > basiceventgeneration.txt
+generate p p > t t~ [QCD]
+output basiceventgeneration
+launch
+set nevents 100
+EOF
+./bin/mg5_aMC ./basiceventgeneration.txt
+
+# Remove folder of basic event generation
+rm -rf basiceventgeneration
+
 # Remove all downloaded tgz files before building the package
 find . -type f -name '*.tgz' -delete
 
--- a/form-toolfile.spec
+++ b/form-toolfile.spec
@@ -1,0 +1,23 @@
+### RPM external form-toolfile 1.0
+Requires: form
+%prep
+
+%build
+
+%install
+
+mkdir -p %i/etc/scram.d
+cat << \EOF_TOOLFILE >%i/etc/scram.d/form.xml
+<tool name="form" version="@TOOL_VERSION@">
+  <client>
+    <environment name="FORM_BASE" default="@TOOL_ROOT@"/>
+    <environment name="LIBDIR" default="$FORM_BASE/lib"/>
+    <environment name="INCLUDE" default="$FORM_BASE/include"/>
+    <environment name="BINDIR" default="$FORM_BASE/bin"/>
+  </client>
+  <runtime name="PATH" default="$BINDIR" type="path"/>
+</tool>
+EOF_TOOLFILE
+
+## IMPORT scram-tools-post
+
--- a/form-toolfile.spec
+++ b/form-toolfile.spec
@@ -11,8 +11,6 @@ cat << \EOF_TOOLFILE >%i/etc/scram.d/form.xml
 <tool name="form" version="@TOOL_VERSION@">
   <client>
     <environment name="FORM_BASE" default="@TOOL_ROOT@"/>
-    <environment name="LIBDIR" default="$FORM_BASE/lib"/>
-    <environment name="INCLUDE" default="$FORM_BASE/include"/>
     <environment name="BINDIR" default="$FORM_BASE/bin"/>
   </client>
   <runtime name="PATH" default="$BINDIR" type="path"/>
--- a/gmp.spec
+++ b/gmp.spec
@@ -1,2 +1,2 @@
-### RPM external gmp 4.2.1
-Source: ftp://mirrors.kernel.org/gnu/%n/%n-%realversion.tar.gz
+### RPM external gmp 6.1.2
+Source: ftp://mirrors.kernel.org/gnu/%n/%n-%realversion.tar.xz
--- a/gosam-toolfile.spec
+++ b/gosam-toolfile.spec
@@ -11,14 +11,15 @@ cat << \EOF_TOOLFILE >%i/etc/scram.d/gosam.xml
 <tool name="gosam" version="@TOOL_VERSION@">
   <client>
     <environment name="GOSAM_BASE" default="@TOOL_ROOT@"/>
-    <environment name="LIBDIR" default="$GOSAM_BASE/lib"/>
-    <environment name="INCLUDE" default="$GOSAM_BASE/include"/>
     <environment name="BINDIR" default="$GOSAM_BASE/bin"/>
   </client>
-<runtime name="PATH" default="$BINDIR" type="path"/>
-<use name="python"/>
-<use name="cython"/>
+  <runtime name="PYTHONPATH" value="$GOSAM_BASE/lib/python@PYTHONV@/site-packages" type="path"/>
+  <runtime name="PATH" default="$BINDIR" type="path"/>
+  <use name="python"/>
+  <use name="cython"/>
 </tool>
 EOF_TOOLFILE
+
+export PYTHONV=$(echo $PYTHON_VERSION | cut -f1,2 -d.)
 
 ## IMPORT scram-tools-post
--- a/gosam-toolfile.spec
+++ b/gosam-toolfile.spec
@@ -1,0 +1,24 @@
+### RPM external gosam-toolfile 2.0
+Requires: gosam
+%prep
+
+%build
+
+%install
+
+mkdir -p %i/etc/scram.d
+cat << \EOF_TOOLFILE >%i/etc/scram.d/gosam.xml
+<tool name="gosam" version="@TOOL_VERSION@">
+  <client>
+    <environment name="GOSAM_BASE" default="@TOOL_ROOT@"/>
+    <environment name="LIBDIR" default="$GOSAM_BASE/lib"/>
+    <environment name="INCLUDE" default="$GOSAM_BASE/include"/>
+    <environment name="BINDIR" default="$GOSAM_BASE/bin"/>
+  </client>
+<runtime name="PATH" default="$BINDIR" type="path"/>
+<use name="python"/>
+<use name="cython"/>
+</tool>
+EOF_TOOLFILE
+
+## IMPORT scram-tools-post
--- a/gosam.spec
+++ b/gosam.spec
@@ -7,12 +7,8 @@ Requires: form
 Requires: gosamcontrib
 Requires: python cython
 
-
-
 %prep
 %setup -q -n gosam
-
-
 
 %build
 CXX="$(which c++) -fPIC"
@@ -20,13 +16,13 @@ CC="$(which gcc) -fPIC"
 FC="$(which gfortran)"
 PLATF_CONF_OPTS="--enable-shared --disable-static"
 export PYTHONPATH=${CYTHON_ROOT}/${PYTHON_LIB_SITE_PACKAGES}
-
-
 python setup.py install --prefix=%{i}
 
 %install
 perl -p -i -e "s|^#!.*python|#!/usr/bin/env python$1|" $(grep -r -e "^#\!.*python.*" %i | cut -d: -f1)
 find %{i}/lib -name '*.la' -exec rm -f {} \;
 
-
-
+%post
+%{relocateConfig}bin/gosam.py
+%{relocateConfig}bin/gosam-config.py
+%{relocateConfig}lib/python2.7/site-packages/golem/installation.py
--- a/gosam.spec
+++ b/gosam.spec
@@ -1,0 +1,32 @@
+
+### RPM external gosam 2.0.4-12f4de9
+Source: http://www.hepforge.org/archive/gosam/gosam-%{realversion}.tar.gz
+
+Requires: qgraf
+Requires: form
+Requires: gosamcontrib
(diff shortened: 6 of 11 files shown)


iahmad-khan:
please test

smuzaffar:
please test

davidlange6:
hold

form.spec
```diff
@@ -0,0 +1,28 @@
...
+%if "%{?cms_cxx:set}" != "set"
+%define cms_cxx g++
+%endif
+
+%prep
+%setup -q -n form-4.1
+
+%build
+
+CXX="$(which %{cms_cxx}) -fPIC"
```
davidlange6, line 13: is this fPIC needed here?

smuzaffar: looks like it is needed. Its makefile is not adding it resulting in non-position independent code.

mharrend:
Dear all,

are there any reasons why this PR is not merged?

I would be glad if you could give @andrej-cms and I would get some feedback.

Thanks, 

Marco

Andrej-CMS:
Dear all, @smuzaffar, @iahmad-khan, @davidlange6 

Did you have a chance to look into this PR? As far as I understand there was an issue with the GNU GPL V3 licence of GoSam. In this case the GoSam external does not only provide the calculation of one loop amplitudes for Herwig7, but it also provides the Golem95 [1] library which is required for Herwig7 in order to be able to use external matrix element providers. This is very important for the Herwig7 Interface in CMSSW.

If it is impossible to take in GoSam into CMSSW because of the licence issue, is it then possible to make a similar PR with the Golem95 library instead of GoSam?

Kind regards,
Andrej

[1] https://golem.hepforge.org/

davidlt:
Based on a quick look, it's true that we cannot incl. GPLv3 code into CMSSW which is covered by Apache 2.0.

mharrend:
@davidlt Okay, maybe GPLv3 code is an issue.
But may I ask why OpenLoops which is also GPLv3 licensed is included in CMSDIST / CMSSW since a long time?

Can you please explain why including OpenLoops was not a problem and GoSam or Golem95 both also GPLv3 is?

davidlange6:
gplv3 is not a major problem for externals (but it is a problem).  the issues with gosam are not gosam itself but its dependencies, which golem appears not to share

davidlt:
> This licensing incompatibility applies only when some Apache project software becomes a derivative work of some GPLv3 software, because then the Apache software would have to be distributed under GPLv3. This would be incompatible with ASF's requirement that all Apache software must be distributed under the Apache License 2.0.
> 
> We avoid GPLv3 software because merely linking to it is considered by the GPLv3 authors to create a derivative work. We want to honor their license. Unless GPLv3 licensors relax this interpretation of their own license regarding linking, our licensing philosophies are fundamentally incompatible. This is an identical issue for both GPLv2 and GPLv3.

It sounds that linking to GPLv3 is already considered a derivate work. You are allowed to incl. Apache 2.0 license into GPLv3, but not reverse.

mharrend:
Yes, I have seen this also. 
But according to this, we should have never included the old Herwigpp 2.7.1 package in CMSDIST since Herwig++ 2.7.1 was linked to ThePEG which are also both GPLV3.

If we dig a little bit deeper with the same argumentation we have to remove the CLHEP package, too, since it links against the NINJA library which is also GPLv3.  I did not look for further packages so far.
So, if we are really going in this way I would believe that we would have to remove almost all physics packages from CMSDIST.

Let me just add, I really do not want to annoy anyone and I know that we all are not lawyers and at least I always struggle with license issues. But I really would like to understand why in some cases we decided to include GPLv3 software while it seemst that strictly speaking we should have never done this.

davidlt:
CLHEP uses ninja (which is like make) to build it, but that is under Apache 2.0 and we don't link to it.

We are now a bit more cautious probably since we just added a license file into CMSSW: https://github.com/cms-sw/cmssw/blob/master/NOTICE

We are more and more officially claiming Apache 2.0 for CMSSW.

mharrend:
Okay, but what is now the conclusion?
So, we do not include any longer Herwig7 at all?
Because GPLv3 Herwig7 is linked 
* against the independent GPLv3 ThePEG framework
* against either the GPLv3 GoSam library which is linked against GPLv3 form, qgraf and golem95 or only against golem95
* against GPLv3 OpenLoops

I am pretty sure some analysts will have trouble to understand this decision.

BTW: Why is CMSDIST Apache2.0 licensed and not GPLv3?
The source code is anyway public available via Github.

perrozzi:
Hi, I think we need a solution for this.  Can computing suggest ASAP way to go? This issue is blocking a long awaited development in GEN land.
We will likely have a discussion during the CMS week but I would like to get there with an executive decision.
Can you please comment?

davidlange6:
if you avoid non-HEP developed GPLed software (eg form) and software whose distributions are behind authentication layers (eg qgraf), it will be ok for me. Requests for exceptions to that go to the management..

perrozzi:
Hi, I am discussing with some HW7 authors and they are telling me that the redistribution "issue" is there also for GLPv2. In fact, if one goes to the apache website
https://www.apache.org/licenses/GPL-compatibility.html
one can CLEARLY read:
"We avoid GPLv3 software because merely linking to it is considered by the GPLv3 authors to create a derivative work. We want to honor their license. Unless GPLv3 licensors relax this interpretation of their own license regarding linking, our licensing philosophies are fundamentally incompatible. **This is an identical issue for both GPLv2 and GPLv3**."
I think that declaring CMSSW as Apache **prevents distribution** and the problem is not (at all) only related to Herwig7. I remind you that Pythia is GLPv2.... and stuf like GNU scientific package, GNU readline are GLPv3........................................

perrozzi:
and by the way, as far as I understand, already having the cmssw code publicly available on github, is effectively a way of *distributing* it.

davidlt:
Please, same issue applies to GPLv2 (unless you have a special section modifying rules regarding the patents). Source code and binary distribution are usually different things.

Currently Apache 2.0 does not hold for CMSSW (personal opinion). The idea is not to introduce more stuff which we would deal with later on.

Andrej-CMS:
Hello,

As far as I understand, we are not really progressing here. So let me ask a question: If I manage to implement Herwig7 and its interface to CMSSW without using GoSam and the other packages like qgraf and form, but still using Golem95 library (which is necessary and currently provided by GoSam!), will this PR then be accepted?

If this is a possible way to go, I will direct my efforts in this direction.

Kind regards,
Andrej

perrozzi:
Hi Andrej, it looks to me that CMSSW claims a license without complying to its standards. I believe the only way forward is to continue accepting this kind of PRs and sort of the license issues in parallel. It is not acceptable that something enforced upstream blocks the CMS software developments. I will bring this issue during the next ORP.
Luca

fwyzard:
Standard disclaimer: I Am Not A Lawyer.

Said that, my understanding is the following:
  - the fact that CMSSW is on GitHub does _not_ imply distribution
  - the fact that CMSSW is on GitHub and has a NOTICE (why not LICENSE ?) file does imply intent to distribute the source code under that license (Apache v2)
  - we actively distribute older versions of CMSSW together with our "open data" 
  - the source code itself is not derivative work of the GPL externals (assuming we only use the public interface in the header files, or nothing at all)
  - distributing CMSSW in binary form is a different matter: we definitely link with ROOT (LGPLv2) and we may have externals with other copyleft licenses (GPLv3 it seems), so the final product can only be distributed as GPLv3 (or not at all, if we run into some incompatible licenses)

Now, running the externals and using their output is explicitly not covered by GPLv3:
> This License explicitly affirms your unlimited permission to run the unmodified Program. The output from running a covered work is covered by this License only if the output, given its content, constitutes a covered work.

unless we directly link with them, or communicate other than by input/output:
> For example, Corresponding Source includes interface definition files associated with source files for the work, and the source code for shared libraries and dynamically linked subprograms that the work is specifically designed to require, such as by intimate data communication or control flow between those subprograms and other parts of the work.

If it turns out that CMSSW should not be considered a derivative work of the GPLv3 externals, there should be no problems in including them.

If it turns out that CMSSW *should* be considered a derivative work of the GPLv3 externals, and that we are distributing ("conveying") it, then we have to distribute the whole of CMSSW + externals + build tools under GPLv3:

Section 6:
> You may convey a covered work in object code form under the terms of sections 4 and 5, provided that you also convey the machine-readable Corresponding Source under the terms of this License

Section 5.c:
> You must license the entire work, as a whole, under this License to anyone who comes into possession of a copy. This License will therefore apply, along with any applicable section 7 additional terms, to the whole of the work, and all its parts, regardless of how they are packaged.

Of course this applies only to CMSSW as a whole: anybody can take the source code, and if it can be built leaving the GPLv2/GPLv3 externals out, the result can be distributed under the Apache 2.0 license.

Again, I Am Not A Lawyer.
CERN has lawyers, we should talk to them if deemed necessary.

perrozzi:
Ciao Andrea,
thanks a lot for the comments.

Standard disclaimer: I Am Not A Lawyer.
>
indeed this is a delicate legal issue that I would like to delegate to
experts
without waiting for the resolution before to proceed with software
integration...

> Said that, my understanding is the following:
>
>    - the fact that CMSSW is on GitHub does *not* imply distribution
>
> if so, this is even better from the point of view of continuing to
integrate GPL code for now...

>
>    - the fact that CMSSW is on GitHub and has a NOTICE (why not LICENSE
>    ?) file does imply intent to distribute the source code under that license
>    (Apache v2)
>    - we actively distribute older versions of CMSSW together with our
>    "open data"
>
> it would be interesting to know whether the NOTICE is distributed together
with these CMSSW versions

>
>    - the source code itself is not derivative work of the GPL externals
>    (assuming we only use the public interface in the header files, or nothing
>    at all)
>
> I assume so too. we have cases like the Pythia and Sherpa interfaces which
are part of CMSSW and should only call public interface of these programs

>
>    - distributing CMSSW in binary form is a different matter: we
>    definitely link with ROOT (LGPLv2) and we may have externals with other
>    copyleft licenses (GPLv3 it seems), so the final product can only be
>    distributed as GPLv3 (or not at all, if we run into some incompatible
>    licenses)
>
> I agree with this. Another solution would be to remove all the possible
GLP linking in the binaries to be distributed publicly.
provided that the receivers are not interested, but this has nothing to do
with the integration of GPL code into CMSSW
for pure scientific purpose...

>
>
> Now, running the externals and using their output is explicitly not
> covered by GPLv3:
>
> This License explicitly affirms your unlimited permission to run the
> unmodified Program. The output from running a covered work is covered by
> this License only if the output, given its content, constitutes a covered
> work.
>
> unless we directly link with them, or communicate other than by
> input/output:
>
> For example, Corresponding Source includes interface definition files
> associated with source files for the work, and the source code for shared
> libraries and dynamically linked subprograms that the work is specifically
> designed to require, such as by intimate data communication or control flow
> between those subprograms and other parts of the work.
>
> If it turns out that CMSSW should not be considered a derivative work of
> the GPLv3 externals, there should be no problems in including them.
>
> If it turns out that CMSSW *should* be considered a derivative work of
> the GPLv3 externals, and that we are distributing ("conveying") it, then we
> have to distribute the whole of CMSSW + externals + build tools under GPLv3:
>
> Section 6:
>
> You may convey a covered work in object code form under the terms of
> sections 4 and 5, provided that you also convey the machine-readable
> Corresponding Source under the terms of this License
>
> Section 5.c:
>
> You must license the entire work, as a whole, under this License to anyone
> who comes into possession of a copy. This License will therefore apply,
> along with any applicable section 7 additional terms, to the whole of the
> work, and all its parts, regardless of how they are packaged.
>
> Of course this applies only to CMSSW as a whole: anybody can take the
> source code, and if it can be built leaving the GPLv2/GPLv3 externals out,
> the result can be distributed under the Apache 2.0 license.
>
I agree

> Again, I Am Not A Lawyer.
> CERN has lawyers, we should talk to them if deemed necessary.
>
I think this is already scheduled by CMS management.
What matters to me is that we hold software development and/or integration
(which is effectively happening)
waiting for a final answer...

Ciao
Luca

davidlange6:
> On Jul 5, 2017, at 11:33 AM, perrozzi <notifications@github.com> wrote:
> 
> 
> > Of course this applies only to CMSSW as a whole: anybody can take the
> > source code, and if it can be built leaving the GPLv2/GPLv3 externals out,
> > the result can be distributed under the Apache 2.0 license.
> >
> I agree

and its easy to see this is not possible..

> 
> > Again, I Am Not A Lawyer.
> > CERN has lawyers, we should talk to them if deemed necessary.
> >
> I think this is already scheduled by CMS management.
> What matters to me is that we hold software development and/or integration
> (which is effectively happening)
> waiting for a final answer...

because there is a license.. maybe that should be revisited..

fwyzard:
>>> Of course this applies only to CMSSW as a whole: anybody can take the source code, and if it can be built leaving the GPLv2/GPLv3 externals out, the result can be distributed under the Apache 2.0 license.

>> I agree

> and its easy to see this is not possible..

Let me rephrase: a large part of CMSSW can be built without the generator externals; everybody is free to rip that part out of the CMSSW repository, and do whatever they want with it.

>>> Again, I Am Not A Lawyer.
>>> CERN has lawyers, we should talk to them if deemed necessary.

>> I think this is already scheduled by CMS management.
>> What matters to me is that we hold software development and/or integration (which is effectively happening) waiting for a final answer...

> because there is a license.. maybe that should be revisited..

... or we just move CMSSW to GPLv3.

Anyway, was this discussed in some ORP that I missed (which is most of them recently) ?

perrozzi:
There was some discussion during the last ORP but things didn't move very much forward, there will be a reprise soon

perrozzi:
We received green light from the L1 for this

Andrej-CMS:
Hello,

This is great news. So now I can make the PR, which has all the changes to make Herwig7 and its interface work in 90X?
There were some changes required to this PR and to the H7Interface.

Kind regards,
Andrej

davidlange6:
Hi- this work should move to the CMSSW_9_3_X branch for integration.

perrozzi:
would it be worth to give it a try directly changing the base branch of the pull request?

smuzaffar:
@perrozzi , that might work. let me try

smuzaffar:
please test

alberto-sanchez:
@davidlange6, can you please unhold

madgraph5amcatnlo.spec
```diff
@@ -35,6 +35,18 @@ rm bin/compile.py
 # Save patched config
 mv input/mg5_configuration_patched.txt input/mg5_configuration.txt
 
+# Start small NLO event generation to make sure that all additional packages are compiled
+cat <<EOF > basiceventgeneration.txt
```
davidlange6, line 39: hi- @alberto-sanchez - i find this bit of proposed code to be crazy. What is the problem its trying to solve? I would think we could get a proper library build from the authors..

Andrej-CMS: Hello @davidlange6 ,
from https://herwig.hepforge.org/tutorials/installation/prerequisite.html

"Madgraph writes some files to the source location the first time it is run, which can cause problems if you install it on a read-only filesystem. We would therefore recommend you run a simple process during install to avoid this problem."

If we can get a proper library for this, it's also fine. We just need to have Mg5amc@NLO  runnin properly when we call it with Herwig7 matchbox

Just to give an update: I'm revising some changes to this PR right now to make Herwig 7.1.1 (newest version) and its interface run in CMSSW. This should be finished in the next few days.

alberto-sanchez: Dear @Andrej-CMS, Any news about H7?

Andrej-CMS: @alberto-sanchez 
I'm implementing Herwig 7.1.1. and found a bug in combination of Herwig 7.1.1. with Madgraph. I'm in contact with the authors in regards to this. The bug also exists in the standalone Herwig 7.1.1 installation, so it's not a result of the CMSSW implementation. As soon as Herwig 7.1.1 can be interfaced correctly with MG5aMC@NLO, I will update the PR.

Kind regards,
Andrej

perrozzi:
@iahmad-khan can you please solve the conflict?

perrozzi:
@Andrej-CMS any news from the Herwig authors?

Andrej-CMS:
@perrozzi 
Unfortunately, I haven't received the patch from the authors yet, but they are working on it. In the mean time, I'm preparing and cleaning the rest of the code.

perrozzi:
Hi, is there any news on this?

Andrej-CMS:
Hello @perrozzi ,
I've got the patch last week and currently I'm building the latest version locally. If the build works properly, I will update the PR for a working version of Herwig7.1.1. with NLO matrix element providers and couple of additional bug fixes for the H7Interface

perrozzi:
great, looking forward!

perrozzi:
I think this PR can be closed as is superseeded by https://github.com/cms-sw/cmsdist/pull/3519, @Andrej-CMS please confirm

Andrej-CMS:
Yes, this PR can be closed in favor of https://github.com/cms-sw/cmsdist/pull/3519

PR #2891, opened 2017-03-07
Add professor2

mkirsano:


smuzaffar:
please test

professor2-toolfile.spec
```diff
@@ -0,0 +1,29 @@
...
+%install
+
+mkdir -p %i/etc/scram.d
+cat << \EOF_TOOLFILE >%i/etc/scram.d/professor2.xml
+<tool name="professor2" version="@TOOL_VERSION@">
+<client>
+<environment name="PROFESSOR2_BASE" default="@TOOL_ROOT@"/>
+</client>
+<use name="py2-numpy"/>
+<use name="py2-sympy"/>
```
davidlange6, line 16: here and elsewhere, the py2-sympy tool won't be directly available - you should depend on py2-pippkgs instead.

smuzaffar: @davidlange6 , all of the py2-pippkgs tools are also separately available. So it is better to explicitly mention the dependency.
```
scram tool info py2-sympy
Tool info as configured in location /tmp/muzaffar/CMSSW_9_1_X_2017-03-07-2300
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++

Name : py2-sympy
Version : 1.0
++++++++++++++++++++
SCRAM_PROJECT=no
USE=py2-pippkgs
```

davidlange6:
ah - ok - so its only the Requires:... in specs that should be updated..(so in professor2.spec)

smuzaffar:
ah right ..... having Requires directly on py2-sympy (and others) means these packages will be distributed.
Agree, to avoid the double distribution, Requires: in professor2.spec should be changed to
```
Requries: py2-pippkgs
```

smuzaffar:
please test

davidlange6:
assuming tests are ok, this looks good to me.

PR #2894, opened 2017-03-08
Update py2-tensorflow.spec

davidlange6:
Seems that I had picked up an old version - getting up to the modern age.


davidlange6:
please test

PR #2898, opened 2017-03-14
To avoid latency disable uniq sorting by default and made it optional

vkuznet:
I found that recent change to use unique results may significantly hurt performance for large queries, such as `file dataset=/Neutrino_E-10_gun/RunIISpring15PrePremix-PUMoriond17_80X_mcRun2_asymptotic_2016_TrancheIV_v2-v2/GEN-SIM-DIGI-RAW` which returns 125K+ LFNs. Sorting of such list will take a minute and not really necessary. While for other DAS queries, such as site for given dataset, it is required since users want to see final set of sites, while results shows all possible replicas. As compromise I made unique flag and made it optional. By default sorting/uniq is disabled to provide ability to look-up results as fast as possible.

### DIFF
--- a/dasgoclient-binary.spec
+++ b/dasgoclient-binary.spec
@@ -1,4 +1,4 @@
-### RPM cms dasgoclient-binary v00.00.05
+### RPM cms dasgoclient-binary v00.00.06
 Source0: git+https://github.com/vkuznet/dasgoclient?obj=master/%{realversion}&export=dasgoclient&output=/dasgoclient.tar.gz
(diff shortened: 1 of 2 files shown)


vkuznet:
please test

vkuznet:
@smuzaffar, Shahzad could you please trigger tests build, I submit PR early last week.

smuzaffar:
please also update dasgoclient.spec

smuzaffar:
@cmsbuild, please test

PR #2903, opened 2017-03-20
disable atlas usage in py2-numpy

smuzaffar:
if atlas package is installed on system then py2-numpy picks it up from system but as we do not seed atlas in our bootstrap so it is an issue.
If we need atlas package then we should add it in our externals.


smuzaffar:
please test

davidlange6:
Hi @smuzaffar - I found my notes - indeed I had decided atlas package was not critical for us. So this is a good solution.

smuzaffar:
closing it in favor of #2907

PR #2906, opened 2017-03-21
remove /bin/perl from memprobe in root

davidlange6:
my build of root on gcc630 happens to notice a /bin/perl in memprobe. Removing it following the usual solution.

### DIFF
--- a/root.spec
+++ b/root.spec
@@ -167,6 +167,7 @@ make %{makeprocesses} -l $(getconf _NPROCESSORS_ONLN) install
 
 find %{i} -type f -name '*.py' | xargs chmod -x
 grep -R -l '#!.*python' %{i} | xargs chmod +x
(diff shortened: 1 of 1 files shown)


davidlange6:
please test

root.spec
```diff
@@ -167,6 +167,7 @@ make %{makeprocesses} -l $(getconf _NPROCESSORS_ONLN) install
 
 find %{i} -type f -name '*.py' | xargs chmod -x
 grep -R -l '#!.*python' %{i} | xargs chmod +x
+perl -p -i -e "s|#!/bin/perl|#!/usr/bin/env perl|" %{i}/bin/memprobe
```
smuzaffar, line 170: I would suggest to change it to
```perl -p -i -e "s|#!.*/perl|#!/usr/bin/env perl|" %{i}/bin/memprobe```

PR #2907, opened 2017-03-21
Fix scipy numpy

smuzaffar:
- Fix lapack package to properly add its lib64 in to library path
- Fix numpy and scipy to disable ATLAS explicitly (otherwise it can pick it from system if available)

### DIFF
--- a/lapack.spec
+++ b/lapack.spec
@@ -1,4 +1,5 @@
 ### RPM external lapack 3.6.1
+## INITENV +PATH LD_LIBRARY_PATH %{i}/lib64
 Source0: http://www.netlib.org/lapack/lapack-%realversion.tgz
 
 Requires: cmake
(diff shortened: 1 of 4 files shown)


smuzaffar:
please test

davidlange6:
@smuzaffar 

fwiw, I had the same problem today, after adding your fixes, I got to a different one

Raise RpmInstallFailed (pkg, output)
RpmInstallFailed: Failed to install package py2-scikit-learn. Reason:
error: Failed dependencies:
        libcblas.so.3()(64bit) is needed by external+py2-scikit-learn+0.18.1-cms-1-1.x86_64

which I though I would get around by adding lapack to py2-scikit-learn requires list. but this seems not to be the solution

smuzaffar:
@davidlange6 , yes scikit-learn also need to disable ATLAS. I am goin to update the PR soon (once locally build)

smuzaffar:
please test

davidlange6:
interesting... lets see how far this takes use - but ok, we should depend on ATLAS it seems...

smuzaffar:
I tired building ATLAS but it fail as it found that cpu  throttling is enabled ( http://math-atlas.sourceforge.net/atlas_install/node5.html ) and disabling it is not trivial. So I decided to not build it

davidlange6:
interesting....ok, will look tomorrow.

PR #2909, opened 2017-03-22
add forgotten funcsigs python package

davidlange6:
the spec is already there and in use, but only as build requires.


davidlange6:
please test

smuzaffar:
please test

davidlange6:
hum, that is a strange error..will try to reproduce.

PR #2910, opened 2017-03-22
Change python to unicode ucs4

davidlange6:
avoids the need for building tensor flow from source.  From what I read, this is the "typical" way of building python, but it is different from the default on SLC6/centos7. Basic tests (building cmssw, running relvals, python interactively) all look ok to me.

Would like to try in devel for now (even if that means I will ask for a slc7 version of devel....)

### DIFF
--- a/build-with-pip.file
+++ b/build-with-pip.file
@@ -29,7 +29,7 @@ if [ `cat files.list | wc -l` == "1" ] ; then
    echo ${PIPFILE}
    export PYTHONUSERBASE=%i
    pip list
-   pip install --user %{PipBuildOptions} $PIPFILE
+   pip install --user -v %{PipBuildOptions} $PIPFILE
 else
    echo "Sorry I don't know how to handle no/multiple install files yet"
    cat %{_sourcedir}/files.txt
--- a/py2-pippkgs.spec
+++ b/py2-pippkgs.spec
@@ -76,6 +76,7 @@ BuildRequires: py2-pbr
 BuildRequires: py2-mpmath
 BuildRequires: py2-sympy
 BuildRequires: py2-tqdm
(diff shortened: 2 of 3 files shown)


davidlange6:
please test

smuzaffar:
@davidlange6 , as it rebuilds all pip based packages, can you please add -v option to 
https://github.com/davidlange6/cmsdist/blob/1b85f4a437509b155347a2a711ca100256845c44/build-with-pip.file#L32
this will print full log of pip build command.

davidlange6:
done - and hopefully even correctly

davidlt:
This change will double unicode string memory usage as "UCS-4" should be 4-byte Unicode encoding for each character. This flag changes how internally Python represents Unicode strings (the most popular encoding [and basically powers everything] is UTF-8 which represents chars (points) in 1 to 6 bytes).

This change is ABI change. Thus could create breakage if our Python is mixed with system stuff, or our Python is mixed with COMP Python. There are recorded references where packages compiled with UCS-2 and UCS-4 don't play well.

Looks like system default Python is compiled with UCS-4 and majority distros do that.

I would suggest double checking RHEL5/6/7 defaults, COMP defaults. I guess, you will want the same change in COMP to avoid any issues.

davidlange6:
ok - I should check for memory changes. Since we don't keep the python config around in the job, it should be no issue. Good point about the COMP python..

davidlt:
In general I am for UCS-4 (+1 from me), because that seems to bring us closer to system defaults. It does provide full Unicode code point coverage (but not sure how much that is helpful). Well, probably will be able to use recently added emojis in your strings :)

davidlange6:
will debug in the IB tomorrow -I don't see the error here.

PR #2919, opened 2017-03-28
Update cmssw-tool-conf.spec

davidlange6:
require the OpenBLAS tool not OpenBLAS


davidlange6:
please test

davidlange6:
cmssw-tool-conf built ok - since numpy already has problems, this is ok to go in

PR #2922, opened 2017-03-29
Fix issue with json output and custom separator

vkuznet:


smuzaffar:
@vkuznet, we should think an otherwise of distributing the dasgoclient-binary. Right now, for each of these updates, we need to build/upload dasgoclient-binary to cms repositry and after that we can ask for the automatic tests.
One easy thing I can think of is to get a eos based website at CERN https://webservices.web.cern.ch/webservices/Services/CreateNewSite/Default.aspx 
and then you can build the dasgoclient-binary using cmsBuild and copy the generate dasgoclient-binary RPM to the new site. We can then just update the Source to point to new localtion to get the rpm.

vkuznet:
Shahzad,
I don't mind to change distribution of dagoclient-binary to eos.

I followed instructions and created a personal web site.
Then I put over there newly created dasgoclient-binary RPM,
here is shared link:
https://cernbox.cern.ch/index.php/s/sJNUcbZrvEXXFHv

or full URL

https://cernbox.cern.ch/index.php/apps/files/ajax/download.php?dir=%2Fwww%2Fdasgoclient&files=cms%2Bdasgoclient-binary%2Bv00.00.07-1-1.slc6_amd64_gcc530.rpm

so, which one should I put into dasgoclient.spec?

Feel free to give me more instruction or change dasgoclient.spec itself.
Best,
Valentin.

smuzaffar:
what is the name of web site. One should be able to directly access it via cern.ch/site-name ...right?

smuzaffar:
ok I found it . it is https://cern.ch/valya/dasgoclient/cms+dasgoclient-binary+v00.00.07-1-1.slc6_amd64_gcc530.rpm

vkuznet:
I think this is what you're asking
https://valya.web.cern.ch/valya/dasgoclient/cms+dasgoclient-binary+v00.00.07-1-1.slc6_amd64_gcc530.rpm
right?

smuzaffar:
@cmsbuild, please test
@vkuznet , yes that was what I was looking for. I have updated the PR to use the correct url.

smuzaffar:
+externals

PR #2938, opened 2017-04-05
DQMGUI 9.0.1

rovere:
This will bring in DQMGUI 9.0.1 which has quite a number of important features:
   * The handling of the streamerInfo from ROOT has been completely rewritten with the goal to avoid in future the schema evolution problems we had in the past, that forces us to patch the index once per year.
   * Introduce ROOTJS so that the user can dynamically interact with the plots directly in the browser.


BrunoCoimbra:
+1 to 86a7eddc513d8b8c5d48be82dd55ee9e1e0e9c00 on slc6_amd64_gcc493.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/2938/test_86a7edd-20170405-141506.log" target="_blank">here</a>.

BrunoCoimbra:
Hey @rovere should we merge this PR?

rovere:
Ciao @BrunoCoimbra 
I usually submit code for integration only when I'd like it to be integrated. So the answer is yes.

PR #2971, opened 2017-04-21
bump xrootd to rc2 and add python support

davidlange6:


davidlange6:
please test

xrootd-toolfile.spec
```diff
@@ -19,6 +19,7 @@ cat << \EOF_TOOLFILE >%i/etc/scram.d/xrootd.xml
   </client>
   <runtime name="PATH" value="$XROOTD_BASE/bin" type="path"/>
   <runtime name="ROOT_INCLUDE_PATH" value="$INCLUDE" type="path"/>
+  <runtime name="PYTHONPATH" value="$XROOTD_BASE/lib/python@PYTHONV@/site-packages" type="path"/>
```
davidlt, line 22: Python support should have been as a separate commit.

xrootd.spec
```diff
@@ -1,16 +1,21 @@
...
-%define tag 6a5db61a4ef13d1e79da71af47fbb8a660cf8aac
-%define branch cms/v%{realversion}
-%define github_user cms-externals
+%define tag 10ba4d7adfc5c66f33bcc77cf38ca45f426cb041
+%define branch stable-4.6.x
+
+%define github_user xrootd
 Source: git+https://github.com/%github_user/xrootd.git?obj=%{branch}/%{tag}&export=%{n}-%{realversion}&output=/%{n}-%{realversion}.tgz
 
+Patch1: xrootd-add-version
```
davidlt, line 9: Note, that this patch will go away once we have final 4.6.1 (if was done as temporary thing in GCC 7.0.1 IBs).

xrootd.spec
```diff
@@ -36,7 +41,8 @@ cmake ../ \
   -DENABLE_KRB5=TRUE \
   -DENABLE_READLINE=FALSE \
   -DENABLE_CRYPTO=TRUE \
-  -DCMAKE_SKIP_RPATH=TRUE
+  -DCMAKE_SKIP_RPATH=TRUE \
+  -DENABLE_PYTHON=TRUE -DPYTHON_LIBRARY=$PYTHON_ROOT/lib/libpython2.7.so
```
davidlt, line 45: Styling. `-DENABLE_PYTHON=TRUE` should be on it's own line.
Btw, why we enable Python support here?
Instead of `-DPYTHON_LIBRARY` you can use `CMAKE_PREFIX_PATH`. See: https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_9_1_X/gcc700/root.spec#L139

davidlt (changes requested):

xrootd.spec
davidlange6: i'm up to 3 requests for it.  - i'll make the other changes

PR #2986, opened 2017-04-25
Add py2-jsonpickle packages

davidlt:
This packages allows to dump Python objects in JSON format which is much
more human readable than standard pickle format.

Signed-off-by: David Abdurachmanov <David.Abdurachmanov@cern.ch>


smuzaffar:
please test

PR #2990, opened 2017-04-28
Set DYNAMIC_ARCH and NUM_THREADS

davidlt:
Make sure that we don't build multi-libs, instead we have on generic
library. Set NUM_THREADS otherwise it will be set to number of cores on
build machine. Fedora is using 128 as default, but we already have
machines with higher number of threads. Thus let's pick 256.

Signed-off-by: David Abdurachmanov <David.Abdurachmanov@cern.ch>


smuzaffar:
please test

PR #2995, opened 2017-05-02
set env CMSSW_FULL_RELEASE_BASE for full release used in a patch release

smuzaffar:
backported from devel IBs


smuzaffar:
please test

PR #2997, opened 2017-05-03
Remove %{makeprocesses} to avoid random build failures

davidlt:
I looked into Fedora spec for OpenBLAS and they do not use parallel
builds. Looks like our builds are failing to build randomly -- a
command does not get executed and OpenBLAS library is missing some
symbols.

Signed-off-by: David Abdurachmanov <david.abdurachmanov@gmail.com>


smuzaffar:
please test

PR #2999, opened 2017-05-03
Added coral patch to protect frontier access

Dr15Jones:


Dr15Jones:
I didn't know how to modify the coral.spec file, so I did not try.

smuzaffar:
@Dr15Jones , we will take care of it

smuzaffar:
please test

coral-CORAL_2_3_21-global-mutex-frontier.patch
```diff
@@ -0,0 +1,47 @@
+--- a/src/FrontierAccess/src/Connection.h	2017-04-28 09:43:29.945972505 -0500
++++ b/src/FrontierAccess/src/Connection.h	2011-03-22 05:36:50.000000000 -0500
+@@ -91,7 +91,7 @@
+       /// The type converter
+       TypeConverter*                                   m_typeConverter;
+       /// The connection lock
+-      static boost::mutex s_lock;
++      mutable boost::mutex m_lock;
```
smuzaffar, line 8: @Dr15Jones,  this is already fixed in coral 2.3.21 https://svnweb.cern.ch/trac/lcgcoral/browser/coral/tags/CORAL_2_3_21/FrontierAccess/src/Connection.h#L94
https://svnweb.cern.ch/trac/lcgcoral/browser/coral/tags/CORAL_2_3_21/FrontierAccess/src/Connection.cpp?rev=21698#L31
In fact all changes in this patch are already part of coral 2.3.21.

Dr15Jones:
The link you gave, has

	mutable boost::mutex m_lock;

which is definitely not the same as
	static boost:mutex

Dr15Jones:
@smuzaffar Is the patch I created backwards?

davidlange6:
you want static s_lock? if so, yes, your patch is backwards

smuzaffar:
@Dr15Jones , I only updated your patch to have correct paths so that it can be applied. Your original path https://github.com/cms-sw/cmsdist/pull/2999/commits/99f884210f971d419ec7e364d16289471fdb657e shows that you want to remove s_lock and replace it with m_lock. Did you actually want to revert the change?

Dr15Jones:
I meant to replace the mutable with a static. Clearly my call to diff was wrong. sorry

smuzaffar:
closing this PR in favour of #3001 which contains the updated patch

PR #3020, opened 2017-05-09

smuzaffar:
use new build rule tag V05-05-24 which contains a bug fix for python/*__init__.py files. Previously there was a chance that two build process can write in the same __init__.py file at same time.


smuzaffar:
please test

PR #3033, opened 2017-05-13
update coral: Extended the lock to encompass all the frontier client calls

smuzaffar:
https://github.com/cms-externals/coral/pull/1


smuzaffar:
please test

PR #3083, opened 2017-06-02
fix utm build for gcc630/700 where looks like boost is messing up with std namespace

smuzaffar:
This is just a workaround to avoid the utm build error for gcc630 and gcc700. A proper fix should go in utm.


smuzaffar:
please test

PR #3114, opened 2017-06-21
added OpenMPI support for sherpa

pmillet:
Adds openmpi support to Sherpa as requested in cms-sw/cmssw#18174


smuzaffar:
please test

sherpa.spec
```diff
@@ -15,6 +15,18 @@ Requires: openloops
 %endif # isamd64
 %endif # islinux
 
+%if "%{?cms_cxx:set}" != "set"
```
davidlt, line 18: These are all redundant and not needed. We don't need this macro. Just directly put what's needed on the configure/make lines.

Btw, `-O2 -std=c++0x` is it required, or was it just copy & paste from somewhere?

PR #3127, opened 2017-06-28
Update data-L1Trigger-L1TMuon.spec

smuzaffar:
new tag V01-01-00 as requested cms-data/L1Trigger-L1TMuon#12 ( #3126)
this is needed by cms-sw/cmssw#19451


davidlange6:
this looks like it won't be needed in 92x. 930 should come first

PR #3145, opened 2017-07-04
updated root 6.10.01 for normal 93X IB

smuzaffar:
also needs cmssw change

### DIFF
--- a/root.spec
+++ b/root.spec
@@ -47,6 +47,7 @@ cmake ../%{n}-%{realversion} \
   -DCMAKE_CXX_COMPILER=g++ \
   -DCMAKE_Fortran_COMPILER=gfortran \
   -DCMAKE_LINKER=ld \
+  -Droot7=ON \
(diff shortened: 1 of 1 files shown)


davidlange6:
Hi @smuzaffar -seems we can go ahead with this? The PythonAnalysis unit test failure looks real - but presumably fixable.

smuzaffar:
yes this looks ok now. I shall merge it just before 23h00 IB to avoid normal cmssw PR tests picking up https://github.com/cms-sw/cmssw/pull/19542/commits/dadb48b62e6b80f0cb00f244a4dfe4e50751a0b7 without having new root.

davidlange6:
sounds good- -could you also add the one line change I proposed yesterday (which looks to be fine)

smuzaffar:
I am going to test that change (now) first in root6 IBs just to make sure that it does not break anything for root 6.10.01.

PR #3147, opened 2017-07-05
OpenCL updated to 2.2

davidlange6:


davidlange6:
please test

davidlt:
Just as reminder Nvidia officially supports OpenCL 1.1 / 1.2 IIRC (not following it). They did OpenCL 2.0 for evaluation only for Windows according to releases notes, but that's not full implementation. Surprisingly that also works on Linux, but not mentioned in release notes IIUC. Even if you can use some bits of 2.0 the cards still will report 1.2 (or similar) until it's ready.

davidlange6:
ah - so there is a _good_ reason that we have such an old version.. so I'll close this. Google suggests support is coming (or even in place for some architectures)

davidlt:
Nvidia didn't show interest in OpenCL 1.2+ because it's not what majority of customers use and it generally slows them dow introducing features. It was a big surprise to see OpenCL 2.0 evaluation being mentioned in Windows Driver Beta few months ago.

It's inevitable in the future as Vulkan will merge with OpenCL to provide a single API: http://hexus.net/tech/news/software/105895-vulkan-opencl-will-merge-single-api/

I don't think there are any customers for these specs within CMSSW right now. They are also not loaded into CMSSW environment by default. It's also binned to what CUDA [Nvidia] supports.

PR #3154, opened 2017-07-10
upgrade pythia8 to 226 in 53X

mkirsano:


mkirsano:
To be tested with https://github.com/cms-sw/cmssw/pull/19636

PR #3156, opened 2017-07-11
Update py2-rootpy.spec

smuzaffar:
new version contains the fix https://github.com/rootpy/rootpy/commit/ef99f63d13a665afad3384ba67888f4b5fad4668 which is needed with root 6.10 releases


smuzaffar:
+externals

PR #3324, opened 2017-08-01
update lhapdf pdfsets.index

mkirsano:


mkirsano:
pdfsets.index on cvmfs was a bit corrected (4 sets added) AFTER PR to update sets in 9XX. In 71X there is no such problem.

smuzaffar:
please test

PR #3345, opened 2017-08-09
To build pull request https://github.com/cms-sw/root/pull/114

gartung:
https://github.com/cms-sw/root/pull/114


smuzaffar:
#3348 uses tip of root master plus cms-sw/root#114

PR #3347, opened 2017-08-10
Add Pypi hooks for pip based python stack

davidlange6:


smuzaffar:
@davidlange6 , not able to digest this :-) so what's the idea here?
Should we create new pyCMSSW github repo for this instead of making it part of cmsdist?

davidlange6:
we could - but there won't be much more to this work (and things are derived from CMSDIST)

python/setup.py
```diff
@@ -0,0 +1,169 @@
...
+        'uncertainties==3.0.1',
+        'urllib3==1.21.1',
+        'virtualenv==15.1.0',
+        'virtualenvwrapper==4.7.2',
+        'wcwidth==0.1.7',
+        'webencodings==0.5.1',
+        'werkzeug==0.12.2',
+        'wheel==0.30.0a0',
+        'widgetsnbextension==1.2.6',
+        'xgboost==0.6a2'
```
smuzaffar, line 140: does this mean we have to update this list whenever we add new pip based package or change its version?

davidlange6:
no - that would be terrible - i have most of a script to update this file when its important

PR #3351, opened 2017-08-11
Update SCRAMV1.spec

smuzaffar:
new bug fix scram tag V2_2_7_pre3: 'scram build help' command is fixed to show build-rules help when it is run from within a scram-base project otherwise show scram help.


smuzaffar:
please test

PR #3361, opened 2017-08-24
Update tensorflow-sources.spec

davidlange6:
use sse3 as suggested in cms-sw/cmssw#19893

### DIFF
--- a/tensorflow-sources.spec
+++ b/tensorflow-sources.spec
@@ -12,7 +12,7 @@ Requires: py2-numpy python py2-wheel
 export PYTHON_BIN_PATH=`which python`
 export TF_NEED_JEMALLOC=0
 export TF_NEED_HDFS=0
-export CC_OPT_FLAGS="-march=prescott"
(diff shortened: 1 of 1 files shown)


davidlange6:
please test

davidlt:
`sse3` is not a valid option for `-march`.  The lowest you can go is `prescott` (the first one to enable SSE3). Another safe alternative could be `core2`. Later on, e.g. `nehalem`, already brings SSE4.2 which is probably still not available on every worker node (should be minority).

davidlange6:
hi @davidlt - thanks - i couldn't find definitive documentation on this (undoubtedly because I didn't search for the right thing)

davidlange6:
please test

(I'm trying to match what we did with openblas)

davidlange6:
please test

davidlange6:
merge

davidlt:
The list is here: https://gcc.gnu.org/onlinedocs/gcc-6.4.0/gcc/x86-Options.html#x86-Options

PR #3362, opened 2017-08-24
Build tensorflow-cc library with bazel and copy out the headers needed.

gartung:


davidlange6:
you should instead incorporate most of  this into tensorflow-sources.spec

Dr15Jones:
please test

Dr15Jones:
please abort test
Patrick is working on the code change recommended by David Lange.

gartung:
This is an attempt to implement what was done with cmake here
https://github.com/FloopCZ/tensorflow_cc

gartung:
@davidlange6 The cmake config files want to link in libprotobuf.a when libtensorflow_cc.so is used. 
https://github.com/FloopCZ/tensorflow_cc/blob/master/tensorflow_cc/CMakeLists.txt#L69
What is the environment variable I need to set in the toolfile.xml to have that included?

tensorflow-cc-toolfile.spec
```diff
@@ -0,0 +1,22 @@
...
+%install
+
+mkdir -p %i/etc/scram.d
+cat << \EOF_TOOLFILE >%i/etc/scram.d/tensorflow-cc.xml
+<tool name="tensorflow-cc" version="@TOOL_VERSION@">
+  <lib name="tensorflow_cc"/>
+  <client>
+    <environment name="TENSORFLOW_CC_BASE" default="@TOOL_ROOT@"/>
+    <environment name="LIBDIR" default="$TENSORFLOW_CC_BASE/lib:$TENSORFLOW_C_BASE/lib"/>
+    <environment name="INCLUDE" default="$TENSORFLOW_CC_BASE/include:$TENSORFLOW_C_BASE/include"/>
```
smuzaffar: tensorflow-c-toolfile is already setting 
 - $TENSORFLOW_C_BASE/lib
 - $TENSORFLOW_C_BASE/include
so you do not need to set them here. all you need to do is to add dependency on tensorflow-c i.e. cleanup the above two lines and add ```<use name="tensorflow-c"/>```

tensorflow-cc-toolfile.spec
```diff
@@ -0,0 +1,22 @@
+### RPM external tensorflow-cc-toolfile 1.0
+Requires: tensorflow-cc tensorflow-c-toolfile
```
smuzaffar: you do not need to depend on tensorflow-c-toolfile

Dr15Jones:
@gartung I also found a simple shell script described here
https://stackoverflow.com/questions/33620794/how-to-build-and-use-google-tensorflow-c-api

gartung:
@davidlange6 @smuzaffar 
Both libtensorflow.so and libtensorflow_cc.so have 2 undefined references to Eigen:: symbols and many weak references to Eigen:: symbols.

>  nm -C libtensorflow*.so | grep "u Eigen::" 
>0000000000000000 u Eigen::NonBlockingThreadPoolTempl<tensorflow::thread::EigenEnvironment>::GetPerThread()::per_thread_
>0000000003d82630 u Eigen::internal::manage_caching_sizes(Eigen::Action, long*, long*, long*)::m_cacheSizes

>nm -C libtensorflow_cc.so | grep "W Eigen::"
>0000000002d37b40 W Eigen::QInt32 tensorflow::FloatToQuantized<Eigen::QInt32>(float, float, float)

The tensorflow build downloads many third party dependencies including Eigen and protobuf. I compiles the object files and links them into the library. This is probably OK when using libtensorflow.so with python. When using libtensorflow_cc.so in CMSSW it could potentially be loaded at the same time as the Eigen shared library causing symbol collisions.

Did you check for this when you built the tensorflow python package?

davidlange6:
> On Aug 25, 2017, at 9:37 AM, Patrick Gartung <notifications@github.com> wrote:
> 
> @davidlange6 @smuzaffar
> Both libtensorflow.so and libtensorflow_cc.so have 2 undefined references to Eigen:: symbols and many weak references to Eigen:: symbols.
> 
> nm -C libtensorflow*.so | grep "u Eigen::"
> 0000000000000000 u Eigen::NonBlockingThreadPoolTempltensorflow::thread::EigenEnvironment::GetPerThread()::per_thread_
> 0000000003d82630 u Eigen::internal::manage_caching_sizes(Eigen::Action, long*, long*, long*)::m_cacheSizes
> 
> nm -C libtensorflow_cc.so | grep "W Eigen::"
> 0000000002d37b40 W Eigen::QInt32 tensorflow::FloatToQuantizedEigen::QInt32(float, float, float)
> 
> The tensorflow build downloads many third party dependencies including Eigen and protobuf. I compiles the object files and links them into the library. This is probably OK when using libtensorflow.so with python. When using libtensorflow_cc.so in CMSSW it could potentially be loaded at the same time as the Eigen shared library causing symbol collisions.
> 
> Did you check for this when you built the tensorflow python package?

Its a known issue with using tensorflow, yes.

davidlt:
There is an upstream ticket and they are slowing working on this issue. TL;DR Linking to tensorflow C/C++ is unsafe with many reports from various distributions. This usually affects bigger projects with large dependencies tree.

gartung:
@Dr15Jones Can you start the test.

Dr15Jones:
Please test

gartung:
@Dr15Jones As far as I can tell Eigen is a header only package. See for instance
/cvmfs/cms.cern.ch/slc7_amd64_gcc630/external/eigen/3.3.0/
There is no library, just header files and config files for how to include them.

gartung:
@davidlange6 @Dr15Jones  this is the file to modify use our sources for eigen and protpbuf
https://github.com/tensorflow/tensorflow/blob/v1.1.0/tensorflow/workspace.bzl#L77
https://github.com/tensorflow/tensorflow/blob/v1.1.0/tensorflow/workspace.bzl#L227
There is no guarantee that our sources are compatible.

davidlange6:
Hi Patrick
do you want to try to make a patch to see if our sources work?

gartung:
I am working on a kludge right now which is to copy our header files into the eigen_archive directory that bazel created.

Dr15Jones:
@davidlange6 instead of a patch, why not just fork our own tensorflow repository? That's how we now handle modifying 3rd party sources.

gartung:
@davidlange6 is there a way to have rpmbuild pull the sources used for the eigen and protobuf builds?

davidlange6:
implementation detail. but lets see if something works as a first step

davidlange6:
likely - but I don't know of an example that we have for it. presumably the tensorflow patch is to make the source derived from an envvar - a version for seeing what else goes wrong could hardwire the sources given the dependency and its version (from cmsrep)

gartung:
Just to note, when I changed workspace.bzl to use the CMS eigen source the build works. When I changed the protobuf source the build fails. I think this is because the CMS source is c++ only and the bazel build is expecting protobuf source for java as well.

gartung:
The version of eigen CMS uses is too new for tensorflow 1.1.0. I need to move to v1.2.1 or the just released 1.3.0:

tensorflow/core/lib/core/threadpool.cc:91:56: error: no matching function for call to 'Eigen::NonBlockingThreadPoolTempl<tensorflow::thread::EigenEnvironment>::NonBlockingThreadPoolTempl(int&, bool&, tensorflow::thread::EigenEnvironment)'
             EigenEnvironment(env, thread_options, name)) {}
                                                        ^
In file included from external/eigen_archive/unsupported/Eigen/CXX11/ThreadPool:58:0,
                 from external/eigen_archive/unsupported/Eigen/CXX11/Tensor:72,
                 from ./third_party/eigen3/unsupported/Eigen/CXX11/Tensor:4,
                 from tensorflow/core/lib/core/threadpool.cc:19:
external/eigen_archive/unsupported/Eigen/CXX11/src/ThreadPool/NonBlockingThreadPool.h:22:3: note: candidate: Eigen::NonBlockingThreadPoolTempl<Environment>::NonBlockingThreadPoolTempl(int, Environment) [with Environment = tensorflow::thread::EigenEnvironment]
   NonBlockingThreadPoolTempl(int num_threads, Environment env = Environment())
   ^~~~~~~~~~~~~~~~~~~~~~~~~~
external/eigen_archive/unsupported/Eigen/CXX11/src/ThreadPool/NonBlockingThreadPool.h:22:3: note:   candidate expects 2 arguments, 3 provided

gartung:
https://bitbucket.org/eigen/eigen/branches/compare/3.3.4%0Df3a22f35b044#pull-requests

gartung:
https://bitbucket.org/eigen/eigen/branches/compare/f3a22f35b044%0D3.3.4#diff

davidlange6:
since I understood the opposite in the meeting - this commit is in the eigen master branch...I would guess tensorflow just picked up the tip of the recipe at some moment in time (from late April). I don't see any fundamental reason against trying to move to it in CMSSW (other than if we have a second external package that requires a specific version of eigen). Lets start that in the devel branch..

davidlange6:
that said, I don't immediately see how fixes in the 3.3 branch get upstreamed. maybe you do?

gartung:
Maybe the primary author merge changes from the 3.3 branch to default?

gartung:
@davidlange6 I tested this on my desktop and it builds. It needs to be tested with CMSSW.

riga:
Hi.
In the context of cms-sw/cmssw#19893, are there any updates on this matter? Compared to the C API, the C++ version would do a great job providing (some of) the convenience to work with TF in CMSSW. However, if you find that right now the current setup process is not stable / compatible with cmsdist, we will of course continue to use the C version.

gartung:
@davidlange6 @smuzaffar This builds properly once the eigen update is merged to 94x/gcc630.

slava77:
please test

eigen #3435 was merged on Sept 20.
Does this PR need anything else?

gartung:
no

gartung:
It needs the version of eigen on the 630next branch. This is a non tagged version from the default eigen branch.

gartung:
I see that was merged in the pull request you mentioned.

slava77:
OK, all good then. 

it looks like jenkins ran out of workers for PR testing: this PR tests haven't started yet (and there is only one PR test in progress)

slava77:
2 hours later, tests haven't started.
@smuzaffar @mrodozov @davidlange6 
is there some service interruption that prevents using more nodes for tests?

davidlange6:
not answering the question - but it is the day that all the IBs get redone from scratch....

slava77:
On 9/24/17 10:09 AM, David Lange wrote:
> not answering the question - but it is the day that all the IBs get
> redone from scratch....

OK, this should explain.
I should just do something else on Sundays

gartung:
I never tested this on SL6. That might be the source of this error:

error: undefined reference to 'clock_gettime'

davidlt:
Please fix your last two commits. The commit messages do not match the content.

We don't enable anything above SSE3 in CMSSW.  We still have some nodes which do not support SSE4.X (which is surprising as it's 2017 already). Unless you know this package will only be used on newer HW. You even enabled AVX, which landed in CPUs in ~2011. Double check the requirements and flags.

gartung:
SSE4.1 and SSE4.2 have been available on AMD and Intel cpus since bdver1 and nehalem.

gartung:
The Fermilab people have run into this issue when using -march=core2

https://github.com/tensorflow/tensorflow/issues/12011

davidlt:
Correct, SSE4 stuff is old, but so are some SKUs in WLCG. It depends on particular world region. E.g., for optimized zlib we select functions based on hardware capability (HWCAP) to avoid this issue.

If you plan to use Ternsorflow on CPUs then you would probably highly benefit from SSE, AVX, AVX2, etc. IIRC, all CERN OpenStack [new] machines guarantee that you get AVX. If Tensorflow is good chunk of workflow then you might consider architecture specific builds and then just pick one based on HWCAP.

Solaris + SPARC has something like that built-in ( https://blogs.oracle.com/rie/shared-object-filters ). Used ISA extensions are recorded in ELF section and later dynamic loader could pick optimized one based on HWCAP.

gartung:
@Dr15Jones @davidlange6 @smuzaffar I have fixed the SLC6 build and verified that tensorflow-sources build reliably. Can someone please issue the "please build" command.

slava77:
please test

gartung:
OK, why did rpm decide that a no-source rpm has a dependence on /usr/bin/python2.4????????

* The action "install-external+tensorflow-cc+1.3.0" was not completed successfully because Traceback (most recent call last):
File "/build/cmsbld/jenkins-workarea/workspace/ib-any-integration/PKGTOOLS/scheduler.py", line 199, in doSerial
result = commandSpec[0](*commandSpec[1:])
File "PKGTOOLS/cmsBuild", line 2816, in installPackage
File "PKGTOOLS/cmsBuild", line 2629, in installRpm
RpmInstallFailed: Failed to install package tensorflow-cc. Reason:
error: Failed dependencies:
	/usr/bin/python2.4 is needed by external+tensorflow-cc+1.3.0-1-1.x86_64

gartung:
@slava77 can you start another test please

slava77:
please test

gartung:
@davidlange6 @smuzaffar This needs to be added to a nightly build to resolve this problem
https://github.com/cms-sw/cmssw/pull/19893#issuecomment-332535700

slava77:
@davidlange6
please clarify when can this be merged.
We are waiting for this to be able to proceed with  https://github.com/cms-sw/cmssw/pull/19893

slava77:
ping

smuzaffar:
+externals

PR #3371, opened 2017-08-30
DQMGUI 9.1.2

rovere:
Solve a problem while rendering histograms with many nans that was causing the server to crash.
The user will receive a completely black plot, instead.


h4d4:
+1 to 7a398e6b9ff0ac8f8f56ace837765b1b4fdc12c5 on slc6_amd64_gcc493.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/3371/test_7a398e6-20170830-122505.log" target="_blank">here</a>.

PR #3382, opened 2017-09-05
Use version of protobuf used internally by Tensorflow.

gartung:


smuzaffar:
please test

smuzaffar:
@gartung , looks like we also need to update/regenerate https://github.com/cms-sw/cmssw/blob/master/DQMServices/Core/src/ROOTFilePB.pb.h with new protocol buffer.

davidlange6:
Right - protobuf updates need to be handled with some care..

rovere:
I again ask if it's possible/useful to trigger pb compilation from within the Buildfile, so that upgrades/downgrades to ProtocolBuffer will be somehow easier. Right now we have the compiled version of the PC Message in the release that must be refreshed with the upgraded/downgraded pc compiler.

smuzaffar:
@rovere , I have no objection on add a build rule for converting .proto to .pb.cc and .pb.h
would you like to have controlled it via a BuildFile flag e.g
`<flag PROTOBUF="file.proto"/>`
or it should be automatic i.e if scram sees a .proto file then convert it to  .pb.cc and .pb.h? Currently we only have one such file DQMServices/Core/src/ROOTFilePB.proto

rovere:
Ciao @smuzaffar I'd vote for the automatic rule to handle .proto files with pc compiler.
The only problem I could see is the output location of the compiled files since I honestly do not recall if I had to manually relocate them in different folders.

smuzaffar:
@rovere , I see that the generated pb.h file is used by other packages too e.g.
https://github.com/cms-sw/cmssw/search?utf8=%E2%9C%93&q=ROOTFilePB.pb.h&type=
in this case I would suggest to generate these files by hand whenever we update protobuf otherwise compilation of every source file in packages which depend (directly or in-directly) on package which provides pb.h will be delayed causing the whole build process to slow down.

If generated pb.h files is only used with in its package then we can automatically generate it otherwise I am afraid it will slow down the build.

smuzaffar:
+externals
this should be merged along with cms-sw/cmssw#20406

davidlange6:
Hi- it seems this pr is out of sync now? Could it get rebased?

smuzaffar:
please test

PR #3384, opened 2017-09-06
[94X] Add GeneralBrokenLines as external library.

ghellwig:
GeneralBrokenLines (GBL) is already used in CMSSW but was integrated some time ago by copying the files directly into CMSSW.
Using it as external would ensure that code updates by the maintainers can be included much easier.

### DIFF
--- a/gbl-toolfile.spec
+++ b/gbl-toolfile.spec
@@ -1,0 +1,24 @@
+### RPM external gbl-toolfile 1.0
+Requires: gbl
+%prep
+
+%build
+
+%install
+
+mkdir -p %i/etc/scram.d
+cat << \EOF_TOOLFILE >%i/etc/scram.d/gbl.xml
+<tool name="gbl" version="@TOOL_VERSION@">
+  <info url="https://www.wiki.terascale.de/index.php/GeneralBrokenLines"/>
+  <lib name="GBL"/>
+  <client>
+    <environment name="GBL_BASE" default="@TOOL_ROOT@"/>
+    <environment name="LIBDIR" default="$GBL_BASE/lib"/>
+    <environment name="INCLUDE" default="$GBL_BASE/include"/>
+  </client>
+  <use name="eigen"/>
+</tool>
+EOF_TOOLFILE
+
+## IMPORT scram-tools-post
+
--- a/cmssw-tool-conf.spec
+++ b/cmssw-tool-conf.spec
@@ -1,4 +1,4 @@
-### RPM cms cmssw-tool-conf 38.0
+### RPM cms cmssw-tool-conf 39.0
 ## NOCOMPILER
 # With cmsBuild, change the above version only when a new
 # tool is added
@@ -32,6 +32,7 @@ Requires: evtgen-toolfile
 Requires: expat-toolfile
 Requires: fakesystem
 Requires: fastjet-toolfile
+Requires: gbl-toolfile
 Requires: gcc-toolfile
 Requires: gdbm-toolfile
 Requires: geant4-toolfile
--- a/gbl.spec
+++ b/gbl.spec
@@ -1,0 +1,29 @@
+### RPM external gbl V02-01-03
+
+Source: svn://svnsrv.desy.de/public/GeneralBrokenLines/tags/%{realversion}/cpp/?scheme=http&module=%{realversion}&output=/%{n}-%{realversion}.tgz
+
+BuildRequires: cmake
+Requires: eigen
+
+%prep
+%setup -q -n %{realversion}
+
+%build
+rm -rf build
+mkdir build
+cd build
+
+cmake .. \
+  -DCMAKE_INSTALL_PREFIX=%{i} \
+  -DCMAKE_BUILD_TYPE=Release \
+  -DEIGEN3_INCLUDE_DIR=${EIGEN_ROOT}/include/eigen3 \
+  -DSUPPORT_ROOT=False
+
+make %{makeprocesses}
+
+%install
+cd build
+make install
+
+%post
+%{relocateConfig}GBLConfig.cmake


ghellwig:
please test

gbl-toolfile.spec
```diff
@@ -0,0 +1,25 @@
...
+<tool name="gbl" version="@TOOL_VERSION@">
+  <info url="https://www.wiki.terascale.de/index.php/GeneralBrokenLines"/>
+  <lib name="GBL"/>
+  <client>
+    <environment name="GBL_BASE" default="@TOOL_ROOT@"/>
+    <environment name="LIBDIR" default="$GBL_BASE/lib"/>
+    <environment name="INCLUDE" default="$GBL_BASE/include"/>
+  </client>
+  <runtime name="ROOT_INCLUDE_PATH" value="$INCLUDE" type="path"/>
+  <use name="root_cxxdefaults"/>
```
ghellwig: @smuzaffar I have added these two lines because I have seen them in all other spec files, but the meaning of them is not really clear to me.
Are they actually required, if one has no ROOT dependency?

smuzaffar: @ghellwig , for now its fine to have them. They are only needed if headers files of this package are used (directly or indirectly) in root dictionaries.

ghellwig: Thanks for the explanation. I think they won't be needed in root dictionaries, but let's review that after I integrated this external in the CMSSW code base.
Anyway, it looks like it does not really hurt to have those lines just in case, right?

smuzaffar: root uses ROOT_INCLUDE_PATH to find the headers files, so more items in it can slow down the search. So if you are sure that headers of this package is not going to be part of root dict then feel free to drop these.

By the way, as you have eigen dependency so please add eigen depdency in tool file too. 
```
<use name="eigen"/>
```
This will instruct scram to also add eigen include path when someone uses gbl to build/link.

ghellwig: @smuzaffar regarding the dictionaries I am not 100% sure as I am sometimes surprised by how ROOT works. Maybe you can guide me a bit?
The headers that are in GBL are currently all in `Alignment/ReferenceTrajectories/interface`.
Since there is nothing in the package related to dictionaries, I assume we can indeed drop it, but as I said, I am not 100% sure.

Anyway, I'll add the eigen dependency also to the tool file.

smuzaffar: ROOT_INCLUDE_PATH is needed by rootcling (for relocatability)

```
Do not store the headers' directories in the dictionary
Instead, rely on the environment variable $ROOT_INCLUDE_PATH at runtime
```

https://github.com/root-project/root/blob/94a28b7c2552f3428a9605649aa6d9cde22da056/core/dictgen/src/rootcling_impl.cxx#L86
https://github.com/root-project/root/blob/cf9ce8347d28a4eeaca2ffa01670d83e5907207d/core/clingutils/src/TClingUtils.cxx#L4837

ghellwig:
please test

ghellwig:
@smuzaffar done
can you remind me what the syntax is to test a PR in CMSSW requiring a certain PR in cmsdist to be included?
Or do I have to wait for an IB where this is integrated?

smuzaffar:
@ghellwig , you can comment 
```
please test with cms-sw/cmsdist#<PR>
``` 
on a cmssw PR to test it along with a cmsdist PR. See the full cms-bot commands here http://cms-sw.github.io/cms-bot-cmssw-cmds.html (which are referenced in each cmssw PR)

ghellwig:
@smuzaffar thanks (also for reminding me that reading is a highly underrated skill...)

smuzaffar:
@ghellwig , in the gbl build area I see it had GBLConfig.cmake file with build directory paths hardcoded in it. 
```
FIND_PATH( GBL_INCLUDE_DIR
        NAMES GblTrajectory.h
        PATHS /build/cmsbld/jenkins-workarea/workspace/ib-any-integration/testBuildDir/tmp/BUILDROOT/bbc69e4bbfe56ea9a66b17eff268e6c3/opt/cmssw/slc6_amd64_gcc630/external/gbl/V02-01-03/include
        NO_DEFAULT_PATH
)
```

although no tool is using it for now but still it is better to relocate it. You can add something like 
```
%post
%{relocateConfig}GBLConfig.cmake
```
at the end gbl.spec

ghellwig:
@smuzaffar thanks for the hint
Is there anything else I should look out for?
I assume the `%{relocateConfig}` macro modifies the paths it finds in the file put after it such that it uses variable names?

smuzaffar:
@ghellwig , every thing else looks good. 
%{relocateConfig} macro (defined here  https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_9_4_X/gcc630/rpm-preamble.file#L230 ) is going to replace the buildpath with RPM_INSTALL_PREFIX (only available at install time). Note that there is no space between %{relocateConfig} and file name

ghellwig:
@smuzaffar are you sure about the command for the relocation?
I tried the following two options, but I still have hard-coded names in my build area:
```
%post
%{relocateConfig}GBLConfig.cmake
```
and
```
%post
cd build
%{relocateConfig}GBLConfig.cmake
```

I am still learning the details of the build process and tools. Therefore, I fail to detect the problem.
Do I need to rebuild from scratch or can I just reuse the existing area?
My impression was that I can reuse it, just that the newer output get's some suffix (`cms?`) in the directory name.

smuzaffar:
@ghellwig , after `%{relocateConfig}GBLConfig.cmake` you will still have hard coded paths which will be relocated at the install time. But if you see hard coded path with tmp/BUILDROOT/somehash/.... in then this means relocation failed otherwise everything went fine.

ghellwig:
that's the point, I see exactly this pattern in the example you posted, modulo a different hash:
```
FIND_PATH( GBL_INCLUDE_DIR
        NAMES GblTrajectory.h
        PATHS /build/ghellwig/ext/CMSSW_9_3_X/20170905_1055/a/tmp/BUILDROOT/da34f0a2a0fdd7732613bc3f46b6dcfd/opt/cmssw/slc6_amd64_gcc630/external/gbl/V02-01-03-cms4/include
        NO_DEFAULT_PATH
)
```

ghellwig:
Is there a way to debug those rpm files? E.g., I would like to print out the directory content before running the `%{relocateConfig}` macro.

smuzaffar:
Relocation works for me (so as for many packages we use https://github.com/cms-sw/cmsdist/search?utf8=%E2%9C%93&q=relocateConfig&type= )

```
>cd cmsdist
>git diff
diff --git a/gbl.spec b/gbl.spec
index fbb45a9..a56f52a 100644
--- a/gbl.spec
+++ b/gbl.spec
@@ -24,3 +24,6 @@ make %{makeprocesses}
 %install
 cd build
 make install
+
+%post
+%{relocateConfig}GBLConfig.cmake
> grep PATHS ../w/slc6_amd64_gcc630/external/gbl/V02-01-03/GBLConfig.cmake
        PATHS /build/muz/u/w/slc6_amd64_gcc630/external/gbl/V02-01-03/include
        PATHS /build/muz/u/w/slc6_amd64_gcc630/external/gbl/V02-01-03
```

ghellwig:
ok, this works, too...
I was looking into `../w/BUILD/slc6_amd64_gcc630/external/gbl/V02-01-03/GBLConfig.cmake` and there nothing changed.
Thanks for your help, I'll push the change then.

ghellwig:
please test

davidlange6:
merge

PR #3386, opened 2017-09-06
New dasgoclient version

vkuznet:
This PR fixes dasgoclient error with specific data-service queries.

PR #3390, opened 2017-09-07
New dasgoclient version

vkuznet:
Replacement of #3386 

It is new build on slc6_amd64_gcc630 (instead of slc6_amd64_gcc530).

dasgoclient-binary is uploaded to eos area:
/eos/user/v/valya/www/dasgoclient/cms+dasgoclient-binary+v01.00.08-1-1.slc6_amd64_gcc630.rpm


ghellwig:
please test

PR #3412, opened 2017-09-13
Update llvm.spec to build 5.0.0

smuzaffar:
updated LLVM to 5.0.0 for DEVEL IBs.
- disabled py2-numba and py2-llvmlite as there is no these are not yet ready for llvm 5.0 . Once we have llvmlite available based on llvm 5.0 then we can include it in normal IBs.


smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
lets get this in DEVEL IBs. Once llvmlite is available based on llvm 5.0 then we can look to include it in normal IBs.

PR #3414, opened 2017-09-13
update python.spec: drop register keywork to avoid warning in gcc7 cmssw build

mrodozov:
Updating python tag for gcc7
- cms-sw/cmsdist/issues/3413
- cms-externals/cpython#3


mrodozov:
please test

PR #3415, opened 2017-09-13
upgrade to lhapdf 6.2.1

mkirsano:


perrozzi:
@cmsbuild please test

perrozzi:
dear all, these (#3415, #3416, #3417, #3419) are the urgent updates needed by GEN for MC v2 production we were discussing at the last ORP. 
Please consider approving them at your earliest convenience.
Thanks a lot in advance

smuzaffar:
+external

perrozzi:
Just to clarify: the backport of this PR in particular to 92x (#3419) and 71x (#3417) and corresponding new releases are **very urgent** because we need the new PDF to start producing gridpacks [*]. The gridpacks production is a pre-condition for the GEN-SIM campaign, it would be good not to wait for 930 to be out.

[*] We use 92x (gcc53x) for POWHEG and 71x (gcc48x) for Madgraph gridpacks, they can be then used on all the CMSSW versions

davidlange6:
hi Luca
It doesn't make sense to do anything in 92x for gen going forward. we'll already have to support 93x for the same reason, so we should converge to that cycle (and a back port will show up in 930 before 92x regardless...) 71x was noted at the release meeting.

perrozzi:
Yes I agree, but for the moment we have validated POWHEG using gcc530. We were not aware of the switch to gcc630 for MC v2 production.
We can try to validate it asap in 93x/gcc630, hoping that there are no hiccups. I assume that it would be ok.
Anticipating possible problems, what shall we do in case gcc630 doesn't work?

davidlange6:
It sounds like POWHEG workflows need to become part of the routine CMS validation... then we'd all be aware of its status on gcc630 or other compilers that CMS may wish to move towards. 

Lets find out if POWHEG+gcc630 works or not and worry next week if it doesn't.

perrozzi:
I am collecting the issues here
https://github.com/cms-sw/genproductions/issues/1307#issuecomment-329958537
they are not necesarily related (only) to gcc630...
For the moment we can probably live with 93x, I will let you know in case of "unsolvable" problems

PR #3437, opened 2017-09-20
downgrade herwigpp and thepeg until herwig7 works in CMSSW

mkirsano:


mkirsano:
To be tested together with cmssw #20597

davidlange6:
merge

PR #3457, opened 2017-09-27
Make python link to python3

vkuznet:
This will create python link to python3 and avoid headache in all python based packages which should pick up proper version of python via various constructs, e.g.
```
#!/usr/bin/env python
```
or
```
python /path/file.py
```
The new link will ensure that code will call python3 rather then python. And, this will only be valid for packages which will explicitly depend on python3.

### DIFF
--- a/python3.spec
+++ b/python3.spec
@@ -195,17 +195,13 @@ find %{i}/lib -type f -name "_tkinter.so" -exec rm {} \;
 # Remove .pyo files
 find %i -name '*.pyo' -exec rm {} \;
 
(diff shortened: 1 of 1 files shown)


h4d4:
+1 to 7acddcbceeffa4380d32f19c4bbe3620ff4ec69c on slc7_amd64_gcc630.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/3457/test_7acddcb-20170927-163504.log" target="_blank">here</a>.

PR #3461, opened 2017-09-28
CUDA 9.0.176 for 9.2.x

fwyzard:
Update CUDA to version 9.0.176.

Include `libcudart_static.a` in the external.

No longer include `libcuda.so.*` and `libnvidia-fatbinaryloader.so.*` in the external; instead, rely on the libraries installed on the system, since they would anyway need to match the version of the installed (kernel) driver.

To make them available at link time, we add `libcuda.so` to the libraries, and add `lib64/stubs` to the linker library path.


fwyzard:
please test

fwyzard:
backport #3460

smuzaffar:
+externals

fwyzard:
@cmsbuild, please test

fwyzard:
@cmsbuild, please test

fwyzard:
@cmsbuild, please test

smuzaffar:
please test

fwyzard:
Moved to 9.4.x and later.

PR #3484, opened 2017-10-04
do not include CUDA drivers in the external

fwyzard:
No longer include `libcuda.so.*` and `libnvidia-fatbinaryloader.so.*` in the external; instead, rely on the libraries installed on the system, since they would anyway need to match the version of the installed (kernel) driver.

To make them available at link time, we add `libcuda.so` to the libraries, and add `lib64/stubs` to the linker library path.

### DIFF
--- a/cuda-toolfile.spec
+++ b/cuda-toolfile.spec
@@ -10,6 +10,7 @@ mkdir -p %{i}/etc/scram.d
 cat << \EOF_TOOLFILE >%{i}/etc/scram.d/cuda.xml
 <tool name="cuda" version="@TOOL_VERSION@">
   <info url="https://developer.nvidia.com/cuda-toolkit"/>
+  <lib name="cuda"/>
   <lib name="cudart"/>
   <lib name="nppc"/>
   <lib name="nvToolsExt"/>
@@ -18,6 +19,7 @@ cat << \EOF_TOOLFILE >%{i}/etc/scram.d/cuda.xml
     <environment name="NVCC"      default="$CUDA_BASE/bin/nvcc"/>
     <environment name="BINDIR"    default="$CUDA_BASE/bin"/>
     <environment name="LIBDIR"    default="$CUDA_BASE/lib64"/>
+    <environment name="LIBDIR"    default="$CUDA_BASE/lib64/stubs"/>
     <environment name="INCLUDE"   default="$CUDA_BASE/include"/>
   </client>
   <flags CUDA_CFLAGS="-fPIC"/>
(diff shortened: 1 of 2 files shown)


fwyzard:
@cmsbuild, please test

fwyzard:
hold

fwyzard:
unhold

fwyzard:
@cmsbuild, please test

smuzaffar:
@fwyzard , as I mentioned here https://hypernews.cern.ch/HyperNews/CMS/get/sw-develtools/2518/1/1/1/1/1.html . the stub/libcuda.so will still be in available via LD_LIBRARY_PATH as scram is going to create symlink for it under cmssw/externals/arch/lib directory.

For libcuda, create a separate tool and add
```
```
so that scram does not create the symlink.

fwyzard:
Hi Shahzad,
thanks - I'm away for few days, I'll address this by the end of the week.

.A

fwyzard:
@cmsbuild, please test

fwyzard:
@smuzaffar I've followed your suggestion, and indeed now I do not get the libraries under `stubs/` synlinked under `external/.../lib`.
However, how do I tell SCRAM that it should look for them at build (link) time ?

I've tried both
```
  <use name="cuda"/>
  <use name="cuda-stubs"/>
```
but the `cuda-stubs` library path is not added to the command line of the link step.

smuzaffar:
@fwyzard , sorry there was bug in build rules due to which it  was not adding the -LLIBDIR for tools which set SKIP_TOOL_SYMLINKS flag. This is fixed for today's 11h00 IBs. 
You cuda-stubs tool looks good just replace CUDA_BASE with CUDA_STUBS_BASE in it

fwyzard:
>  You cuda-stubs tool looks good just replace CUDA_BASE with CUDA_STUBS_BASE in it

Sure, I can do that.
Is that to keep things tidy,  or is there a technical reason ?
I kept the same variable name since it points to the same directory.

fwyzard:
By the way, is there a way to pull the latest SCRAM in a local area and test it ?

fwyzard:
OK, I've tested it with CMSSW_9_4_X_2017-10-12-1100, and now the CUDA samples compile and link correctly; as expected, they do fail to run on a machine without a local NVIDIA installation.

fwyzard:
@cmsbuild, please test

fwyzard:
@cmsbuild, please test

smuzaffar:
@fwyzard, yes for now it is just to keep things tidy plus this allows us to have tool pointing to different local.
It is on my to-do-list to allow to use BASE (instead of TOOL_BASE).

PR #3501, opened 2017-10-16
Fixed decay channels or K0 and antiK0

civanch:
Back-ported fix to Geant4 10.2p02 for the treatment of K0 and antiK0, which is a part of Geant4 10.3.
The problem was faced in attempt to start simulation production for XeXe run but may affect any workflow.

In G4VDecayChannel the fixes concerns so-called 1 body decay - transformation of K0 into KS or KL.

In G4Decay extended printout for G4Exception is added.

### DIFF
--- a/geant4.spec
+++ b/geant4.spec
@@ -1,5 +1,5 @@
 ### RPM external geant4 10.02.p02
-%define tag f3271a58f63c955d68aab76c31aeb10e196ff544
+%define tag a7c691a6fb5f3a6669a2929f5d24f8b8671821c4
 %define branch cms/4.%{realversion}
 %define github_user cms-externals
 Source: git+https://github.com/%github_user/%{n}.git?obj=%{branch}/%{tag}&export=%{n}.%{realversion}&output=/%{n}.%{realversion}-%{tag}.tgz


civanch:
please test
@mrodozov , this is an official test of #25 for cms-externals/geant4

civanch:
@davidlange6 , please, consider for 9_4_X

mrodozov:
+externals

davidlange6:
merge

davidlange6:
hi @civanch - thinking more- what in this patch should have changed the simulation results in these tests?

civanch:
@davidlange6 , it should fix rare crashes only. Normally, antiK0 should not appear in Geant4 tracking.

davidlange6:
so maybe these tests are a sign of a bug?

civanch:
Yes, it is Geant4 bug which was fixed in the next release. I say rare problem because it was missed in Geant4 10.2 validation and was identified later. The same for us - Geant4 10.2 was part of CMSSW for long but only in XeXe run the bug become visible.

davidlange6:
no - you miss my point. if this is fixing a rare bug then

	• DQMHistoTests: Total files compared: 26
	• DQMHistoTests: Total histograms compared: 2767139
	• DQMHistoTests: Total failures: 85046

is naively a problem

civanch:
@davidlange6 , sorry now I see the problem. For me it is strange, because what we see is a very different tracking history. This may normal but may be not.

davidlange6:
its not normal... of course once the random sequences get out of sync, things diverge. The question is if they should get out of sync?

civanch:
there are clean-ups taken additionally to this PR:

if      (rbranch <0.  ) rbranch = 0.0;  
else if (rbranch >1.0 ) rbranch = 1.0;  

and

if (width<=0.0) return massPDG

which are correct itself but may be responsible for history change, the second skip usage of random numbers in following lines below the fix....

davidlange6:
ahhhh - can't resist adding random things not part of the actual bug fix..nice.  Certainly the check on width<=0.0 changes the way random numbers are used... For the purposes of 93x, please first make a PR without that change so that we can avoid a long validation step and instead do an identity check.

PR #3505, opened 2017-10-17

civanch:
This patch for K0 and antiK0 decay should fix XeXE production but not change simulation histories.


civanch:
please test

mrodozov:
+1

civanch:
@davidlange6 , the fix help to keep sim histories.

davidlange6:
hi @civanch - do we need both this and #3510?

civanch:
Hi @davidlange6 , we have production gcc/architecture and another one, I would keep cmsdist the same for both - it is more clear for the long term maintenance. So, to me an optimal solution is to merge both PRs.

davidlange6:
merge

PR #3504, opened 2017-10-17
New tag for geant4

mrodozov:
Testing changes in the latest PR: https://github.com/cms-externals/geant4/pull/26

PR #3511, opened 2017-10-18

civanch:
This is a patch for K0 and antiK0 decay cloned from #3505

### DIFF
--- a/geant4.spec
+++ b/geant4.spec
@@ -1,5 +1,5 @@
 ### RPM external geant4 10.02.p02
-%define tag a7c691a6fb5f3a6669a2929f5d24f8b8671821c4
+%define tag f9e758d95096c7956792a541549515b3c767d1fb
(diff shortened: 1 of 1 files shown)


civanch:
please test

mrodozov:
I think this one is the same https://github.com/cms-sw/cmsdist/pull/3504

civanch:
Yes, I am trying propagate the same patch to Geant4 to 9_3_X and 9_4_X. The new commits (of the same geant4.spec) are needed because I see that geant4.spec are not identical in these branches.

civanch:
@davidlange6 , @mrodozov , large number of failures is expected here, because the IB for comparisons includes #3501 which changes sim histories. Now the histories should be as before #3501.

davidlange6:
right - thats why you were asked to do it for 93x...

davidlange6:
merge

PR #3514, opened 2017-10-19
Remove google-perftools which just produces a stub library.

gartung:


gartung:
Test with https://github.com/cms-sw/cmssw/pull/20972

PR #3519, opened 2017-10-23
Herwig 7.1.1 with External NLO Matrix Element Providers

Andrej-CMS:
Dear all,

this PR introduces Herwig 7.1.1, ThePEG 2.1.1. as well as a MG5aMCatNLO patch to make the interface between ME generators work in CMSSW . This is an further development of PR #2889 . Please see the discussion on externals there.

Herwig7.1.1 brings matching and merging to the Herwig7 framework. With the version in this PR the following combinations of generators work:

- Herwig 7.1.1. +Openloops
- Powheg ME implementation in Herwig 7.1.1
- Showering of LHE files (Tested with PowhegV2 LHE files ) 
- Herwig 7.1.1+MG5aMC@NLO 2.5.5
- Herwig 7.1.1+Madgraph+GoSam
- Herwig 7.1.1+Madgraph+Openloops

Kind regards,

Andrej


perrozzi:
please test

Andrej-CMS:
Hello,

Should the code of the error above be updated to the newer version of thePEG or do we want to have two different versions of thePEG, one for older herwig++ and one for herwig7?

Kind regards,

Andrej

perrozzi:
good question. herwig++ was requested by JME to produce alternative samples to estimate JEC systematics.
I believe that HW7 would be a more appropriate comparison at this point, and we should deprecate hewrig++ (in case people need it, we still have CMSSW_9_3_1 working).
@mkirsano @alberto-sanchez @efeyazgan what do you think?

efeyazgan:
I will check with JME guys.

perrozzi:
good idea. In any case I would say that we should kindly suggest what to do because they were also using herwig++ with MLM matching option in pythia8 (@kdlong, @qliphy), so not sure they know exactly what they are doing, they just need an alternatie sample for systematics... (this is my gut feeling)

alberto-sanchez:
@perrozzi @efeyazgan , having both, from the integration point of view, could be difficult. both uses the same namespace, so the possibility to have clashes is high. We already have problems of this kind in the past when we keep both evtgen versions available. So, I think we should move to h7, do whatever validation needed and for specific request on herwig++, use the release where is present.

efeyazgan:
@alberto-sanchez, I agree.

perrozzi:
@Andrej-CMS do you need any further input?

Andrej-CMS:
@perrozzi 

I'm not quite sure how to proceed.

If we don't want to keep the herwig++ interface, then it should be removed from the jenkins test here and instead be tested with https://github.com/cms-sw/cmssw/pull/21050 . The error generated in the request here comes from the ThePEG 2.1.1

perrozzi:
@alberto-sanchez could you please take care of removing it from the jenkins tests?

alberto-sanchez:
it just matter of remove ThePEGInterface and test with #21050 cms-sw/cmsdist#3519.
I have create a PR #21052, but this could be added to #21050 to make this clear.

Andrej-CMS:
I fixed the buildfile in https://github.com/cms-sw/cmssw/pull/21050 . can you please test again?

perrozzi:
Hi, I lost track of all the PR concerning HW7 and related topic among cmssw and cmsdist...
Can you please specify which PR need to be tested in combination with which other PRs?
Thanks :-)

Andrej-CMS:
I fear that I have committed an older version of the code. Let me recheck and come back at you in the morning.

Andrej-CMS:
Dear experts,

Would you please comment on the issue with release validation? I don't fully understand why it fails. I doubt that there are validation subroutines for Herwig 7 events, since this has never been in CMSSW before. 
Is this an issue with GEANT4? Does it maybe fail because Herwig6 cannot be tested with the new version of ThePEG interface?

Kind regards,

Andrej

madgraph5amcatnlo.spec
```diff
@@ -35,6 +38,19 @@ rm bin/compile.py
...
 mv input/mg5_configuration_patched.txt input/mg5_configuration.txt
 
+# Start small NLO event generation to make sure that all additional packages are compiled
+cat <<EOF > basiceventgeneration.txt
+generate p p > t t~ [QCD]
+output basiceventgeneration
+launch
+set nevents 5
+EOF
+./bin/mg5_aMC ./basiceventgeneration.txt
```
davidlange6, line 49: as before - why do we need this - the package must build itself.

davidlange6:
its likely an unrelated failure (transient problem in a unit test)

madgraph5amcatnlo.spec
Andrej-CMS: Hello @davidlange6  ,

This is required for the installation of subpackages that MG5aMC@NLO uses.
I summarized the issue and asked in regards of the best implementation in the hypernews thread https://hypernews.cern.ch/HyperNews/CMS/get/softwareDistrib/757.html

Kind regards,
Andrej

perrozzi:
please test

mseidel42:
Herwig 7.1.2 is out. It may be needed for showering LHE files with the dipole shower (tbc by authors)

Andrej-CMS:
Dear @all,

I just pushed the update to Herwig 7.1.2 and MG5 2.6.0. The software was tested with the current version of the Herwig 7 Interface.

Could you please run the tests and merge if the compilation is OK? If there is anything missing for this PR to be merged, please don't hesitate to let me know!

Thank you very much and kind regards,

Andrej

perrozzi:
@Andrej-CMS thanks a lot.
Can you remind us how should this be tested, provided?
This PR directly, standalone? 
The cms-sw/cmssw#21050 together with this PR?

Andrej-CMS:
Hello @perrozzi ,
https://github.com/cms-sw/cmssw/pull/21052 should be tested with this PR . The important part is the compilation of the PRs. There is currently no workflow that tests that Herwig7 also runs properly. However, I already tested it locally with my own build, making sure that I have read only access to the CMSDIST build. 

If you have any further questions, please let me know!

Kind regards,
Andrej

perrozzi:
this seems to have finally passed the tests!

herwigpp.spec
```diff
@@ -1,35 +1,89 @@
-### RPM external herwigpp 2.7.1
-Source: http://service-spi.web.cern.ch/service-spi/external/MCGenerators/distribution/herwig++/herwig++-%{realversion}-src.tgz
-Requires: boost thepeg gsl hepmc
+### RPM external herwigpp 7.1.2
```
davidlange6, line 1: I understand this to be herwig rather than herwig++ - why not name the spec file accordingly?

thepeg.spec
```diff
@@ -33,29 +34,44 @@ CXX="$(which %{cms_cxx}) -fPIC"
 CC="$(which gcc) -fPIC"
 PLATF_CONF_OPTS="--enable-shared --disable-static"
 
-case %{cmsplatf} in
-  osx*)
-    LIBQUADMATH="-lquadmath"
-    ;;
-esac
+
+# Update to detect aarch64 and ppc64le
```
davidlange6, line 38: these should be picked up as sources rather than here so that its more easily understood when the build configuration changes.

madgraph5amcatnlo.spec
davidlange6: it seems we need a discussion of this topic to understand what is really happening.. (from your HN I only add to by confusion)

perrozzi:
for the record, this can be closed as it has been superseeded by https://github.com/cms-sw/cmsdist/pull/3642

PR #3522, opened 2017-10-24

smuzaffar:
new build rules V05-05-50 which contains update cuba rules (use of device link step)


smuzaffar:
please test

PR #3510, opened 2017-10-18
Added geant4 patch for decay of K0 and antiK0

civanch:
This is a clone of #3505, which was done for gcc630.

PR #3525, opened 2017-10-24

civanch:
This is a clone of #3505 and #3510

### DIFF
--- a/geant4.spec
+++ b/geant4.spec
@@ -1,5 +1,5 @@
 ### RPM external geant4 10.02.p02
-%define tag f3271a58f63c955d68aab76c31aeb10e196ff544
+%define tag f9e758d95096c7956792a541549515b3c767d1fb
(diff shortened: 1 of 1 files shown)


mrodozov:
are you sure about the compiler ? we cannot test this as we don't have an IB for it, we could've if it was for IB/CMSSW_9_3_X/gcc630

civanch:
@mrodozov , I am not sure if this PR should be tested more than for 1 compiler but for me seems to be correct if PRs are merged for all branches - 9_3X/gcc530, 9_3_0/gcc630 and 9_3_0/gcc700 - probably only 1 test for the production release is enough.

davidlange6:
prs are merged for all branches via an automatic process.

civanch:
What I have observed , at least, in 9_3_X - geant4.spec file is not identical in different gcc branches. Was this needed for purpose or is a result of historical process - do not know.

PR #3531, opened 2017-10-26

civanch:
This is a clone of #3505 and #3510


smuzaffar:
please test

civanch:
differences are expected

davidlange6:
the releases we care about are 93x and 10x.

PR #3543, opened 2017-10-30
Update CUDA 9.0.176 external and toolfiles

fwyzard:
Update the list of static and dynamic CUDA libraries included in the external
Update build rules for CUDA code:
  - drop CUDA_CFLAGS; instead pass all CXXFLAGS to the host compiler via `nvcc --compiler-flags '...'`
  - use a separate tool for each CUDA library
  - drop the `-lcudadevrt` flag for linking CUDA device code, and hardcode it in the SCRAM rules (see cms-sw/cmssw-config#60)

In GCC, reword '-Wa,...' as '-Xassembler ...' to support CUDA `nvcc --compiler-flags '...'`

### DIFF
--- a/cuda-toolfile.spec
+++ b/cuda-toolfile.spec
@@ -1,4 +1,4 @@
-### RPM external cuda-toolfile 2.0
+### RPM external cuda-toolfile 2.1
 Requires: cuda
 %prep
 
@@ -9,11 +9,12 @@ Requires: cuda
 mkdir -p %{i}/etc/scram.d
 cat << \EOF_TOOLFILE >%{i}/etc/scram.d/cuda-stubs.xml
 <tool name="cuda-stubs" version="@TOOL_VERSION@">
-  <info url="https://developer.nvidia.com/cuda-toolkit"/>
+  <info url="https://docs.nvidia.com/cuda/index.html"/>
   <lib name="cuda"/>
   <client>
     <environment name="CUDA_STUBS_BASE" default="@TOOL_ROOT@"/>
     <environment name="LIBDIR"          default="$CUDA_STUBS_BASE/lib64/stubs"/>
+    <environment name="INCLUDE"         default="$CUDA_STUBS_BASE/include"/>
   </client>
   <flags SKIP_TOOL_SYMLINKS="1"/>
 </tool>
@@ -21,10 +22,10 @@ EOF_TOOLFILE
 
 cat << \EOF_TOOLFILE >%{i}/etc/scram.d/cuda.xml
 <tool name="cuda" version="@TOOL_VERSION@">
-  <info url="https://developer.nvidia.com/cuda-toolkit"/>
+  <info url="https://docs.nvidia.com/cuda/index.html"/>
   <use name="cuda-stubs"/>
   <lib name="cudart"/>
-  <lib name="nppc"/>
+  <lib name="cudadevrt"/>
   <lib name="nvToolsExt"/>
   <client>
     <environment name="CUDA_BASE" default="@TOOL_ROOT@"/>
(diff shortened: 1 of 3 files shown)


fwyzard:
This goes together with cms-sw/cmssw-config#59 .

fwyzard:
@cmsbuild, please test with #3542

cuda-toolfile.spec
```diff
@@ -33,10 +34,142 @@ cat << \EOF_TOOLFILE >%{i}/etc/scram.d/cuda.xml
...
+cat << \EOF_TOOLFILE >%{i}/etc/scram.d/cuda-cublas.xml
+<tool name="cuda-cublas" version="@TOOL_VERSION@">
+  <info url="https://docs.nvidia.com/cuda/cublas/index.html"/>
+  <use name="cuda"/>
+  <lib name="cublas"/>
+  <lib name="cublas_device"/>
+  <client>
+    <environment name="CUDA_CUBLAS_BASE" default="@TOOL_ROOT@"/>
+    <environment name="LIBDIR"           default="$CUDA_CUBLAS_BASE/lib64/stubs"/>
+    <environment name="INCLUDE"          default="$CUDA_CUBLAS_BASE/include"/>
```
smuzaffar: @fwyzard , this is going to make libs under stub directory available at runtime. I would suggest to create a new tool 
```
<tool name="cuda-stubs-interface" version="@TOOL_VERSION@">
  <info url="https://docs.nvidia.com/cuda/index.html"/>
  <client>
    <environment name="CUDA_STUBS_INTERFACE_BASE" default="@TOOL_ROOT@"/>
    <environment name="LIBDIR"          default="$CUDA_STUBS_INTERFACE_BASE/lib64/stubs"/>
    <environment name="INCLUDE"         default="$CUDA_STUBS_INTERFACE_BASE/include"/>
  </client>
</tool>
```
and make all other tools to use it e.g. 
```
<tool name="cuda-stubs" version="@TOOL_VERSION@">
  <use name="cuda-stubs-interface"/>
  <lib name="cuda"/>
</tool>
<tool name="cuda-cublas" version="@TOOL_VERSION@">
  <use name="cuda"/>
  <lib name="cublas"/>
  <lib name="cublas_device"/>
  <flags CUDA_LDFLAGS="-lcublas_device"/>
</tool>
```

fwyzard: Actually, that is just a mistake... the individual libraries should not need to define LIBDIR or INCLUDE, they should pick them up from "cuda-stubs" and "cuda".

fwyzard:
@cmsbuild, please test

fwyzard:
OK, I have only one last issue remaining: the order of the `CUDA_LDFLAGS`.
In cuda.xml we have
```xml
  <flags CUDA_LDFLAGS="-L$(CUDA_BASE)/lib64"/>
  <flags CUDA_LDFLAGS="-lcudadevrt"/>
 ```
and in cuda-cublas.xml we have
```xml
  <flags CUDA_LDFLAGS="-lcublas_device"/>
  <use name="cuda"/>
```

Using cuda-cublas results in 
> nvcc -dlink ... -L$(CUDA_BASE)/lib64 -lcudadevrt -lcublas_device ...

However if `-lcudadevrt` is not the last library on the command line, nvlink fails with 

> @E@nvlink fatal   : unexpected object after cudadevrt


How can we make scram pass the flags in the right order ?

smuzaffar:
@fwyzard , order depends on the dependency. For flags and libs order is opposite. For example if we have tool T2 depends on tool T1 then order for
- flags will be `flags of T1` `flags of T2` (last flag wins)
- libs will be `libs of T2` `libs of T1`  (first lib with symbol wins)
As we are using CUDA_LDFLAGS that is why cuda-cublas flags are shown at the last. 

If cudadevrt is special library then we can set special flag CUDA_DEVRT and jst add it at the end here https://github.com/cms-sw/cmssw-config/blob/master/SCRAM/GMake/Makefile.rules#L627

fwyzard:
Indeed, I was thinking about it later this morning, and I realised that for FLAGS this is the expected behaviour.

`cudadevrt` is indeed special, in that (it looks like) it needs to be the last one - so we could just hard-code at the end of the in the `nvcc -dlink` command.
Or, we could add `CUDA_LIBS` and make it work like `LIBS`.

I have a mild preference for the latter - because we may end up with something like a CUDA package depending on a second CUDA package, and that take care of it - but for the moment hardcoding would work as well.

smuzaffar:
Yes for now we should hard-code it. 
To support `CUDA_LIBS` , I can extend the lib tag to accept `type` attribute e.g.
```
<lib name="name" type="cuda"/>
```
but this need some non-trivial changes in scram.

fwyzard:
OK, I'll submit PRs with the changes (here and for cmssw-config)

smuzaffar:
cmssw-config changes might not be need if you add
```
<flags CUDA_LDFLAGS="-lcuda-devrt"/>
```
in your BuildFile then it will endup as a last item.  You might need to clean-up  cuda toolfiles though.

fwyzard:
True... that's what I was doing to begin with, but I'd rather not add it by hand to all (or just some) CUDA binaries...

Rather: cms-sw/cmssw-config#60

fwyzard:
@cmsbuild, please test with cms-sw/cmsdist#3544

smuzaffar:
@fwyzard , can you make this PR for IB/CMSSW_10_0_X/gcc630next branch which is used for DEVEL IBs (where we have latest cmssw-config)

fwyzard:
sure: cms-sw/cmsdist#3545

smuzaffar:
closing in favor of #3566

PR #3545, opened 2017-10-31
Update CUDA 9.0.176 external and toolfiles

fwyzard:
Update the list of static and dynamic CUDA libraries included in the external
Update build rules for CUDA code:
  - drop CUDA_CFLAGS; instead pass all CXXFLAGS to the host compiler via `nvcc --compiler-flags '...'`
  - use a separate tool for each CUDA library
  - drop the `-lcudadevrt` flag for linking CUDA device code, and hardcode it in the SCRAM rules (see cms-sw/cmssw-config#60)

In GCC, reword '-Wa,...' as '-Xassembler ...' to support CUDA `nvcc --compiler-flags '...'`

### DIFF
--- a/cuda-toolfile.spec
+++ b/cuda-toolfile.spec
@@ -1,4 +1,4 @@
-### RPM external cuda-toolfile 2.0
+### RPM external cuda-toolfile 2.1
 Requires: cuda
 %prep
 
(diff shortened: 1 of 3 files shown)


fwyzard:
@cmsbuild, please test with cms-sw/cmsdist#3544

smuzaffar:
merging it for devel IBs

PR #3595, opened 2017-11-22
New dasgoclient version

vkuznet:
Add user-agent string to distinguish dasgoclient from other go-clients in cmsweb frontend logs


smuzaffar:
please test

PR #3607, opened 2017-11-27
updated latest scram and build rules

smuzaffar:
get latest scrama nd build rules in 92X. This should fix the wrong python directory path https://github.com/cms-sw/cmsdist/pull/3599


smuzaffar:
please test

PR #3616, opened 2017-12-06
Switch to CLHEP 2.4.0.0

civanch:
Main modification - improved MixMax random number generator and make it default. The test should show no difference with the baseline, because in CMSSW explicit random number generators are defined per producer, so the old ones will be used.

### DIFF
--- a/clhep.spec
+++ b/clhep.spec
@@ -1,6 +1,6 @@
-### RPM external clhep 2.3.4.2
+### RPM external clhep 2.4.0.0
 
-%define tag 03bdb22fe139303b43ca834a641f451f8e79bbcf
+%define tag e56c9f2bea9ba3014c6ab799ebc321b8f75a0808
 %define branch cms/v%{realversion}
 %define github_user cms-externals
 Source: git+https://github.com/%github_user/%{n}.git?obj=%{branch}/%{tag}&export=%{n}-%{realversion}&output=/%{n}-%{realversion}.tgz


civanch:
please test

civanch:
It is strange, that the difference is two WFs only: 5.1 and 135.4

civanch:
For WF 5.1 one of the plots:
jetmet__jetvalidation_ak4pfjetschs_photonmultiplicity
It is clear that random numbers are different

smuzaffar:
@davidlange6 , any objection getting this one in?

davidlange6:
hi @civanch - but we don't expect different random number sequences unless there is a change. Lets check that its reproducible - otherwise is there something in the release notes that suggests a bug fix that might affect fast sim?

davidlange6:
please test

civanch:
@smuzaffar , @davidlange6 the difference in two FWs means that there are places were where CMS Service for random numbers do not used, instead CLHEP random engine is called directly. CLHEP random default was changed but this does not affect CMS RandomService which is explicit per producer. The suspect is also generators, in these WFs they likely from run-1.

davidlange6:
hi @civanch - it can not be gen - as the generators are the same here and in full sim. (both run1/run2)

i notice that the HepRandom defaults have changed to mixmax - do we just discover that someone is using the default (that probably shouldn't be)

civanch:
Hi, @davidlange6 , I agree with you, that it is the FastSim feature - we discover that somewhere in FastSim the default CLHEP generator is used. Also FastSim is tightly coupled with TRandom3 - I see inside FastSim even checks this... We need discuss this with FastSim experts but this should not stop from merge of this CLHEP PR - when FatsSim will be fully understood likely a CMSSW PR will be needed but should we wait for that?

civanch:
@ssekmen - we see a problem in FastSim WFs.

davidlange6:
the question is why are we now ready to use minmax in production? I don't think so. so should we not just control its usage?

i don't actually see something in fast sim creating a HepRandom object - do you know where to look?

civanch:
The sequence is long. I grep all places in FastSim without full understanding of the situation.

davidlange6:
what should i search for? I don't see much from looking at HepRandom.

civanch:
there may search for G4UniformRand() it is only one idea.

davidlange6:
G4UniformRand seems to be used in lots of places (but almost no place in fast sim) - I conclude we must have initialized it correctly.

ssekmen:
@civanch , checking.  This could be happening in the calorimetry code.

ssekmen:
G4UniformRand seems to be used in fastsim HF showers: FastSimulation/ShowerDevelopment/src/FastHFShowerLibrary.cc .

civanch:
Also FastSimulation/CTPPSRecHitProducer in several places use directly CLHEP for Gauss.

davidlange6:
I hate guessing games - can you find the actual constructor that uses the "default"?

Eg, the syntax in FastSim.CTPPS is no different from that in the rest of CMS (and its TRandom3 it would seem, strangely)

https://cmssdt.cern.ch/lxr/source/FastSimulation/CTPPSRecHitProducer/plugins/CTPPSRecHitProducer.cc#0151

PR #3639, opened 2017-12-18
Add NVIDIA libraries needed by the CUDA runtime.

fwyzard:
On a machine with CUDA installed we should use the local NVIDIA drivers and libraries.  
However on a machine without CUDA, or even without an NVIDIA GPU, these libraries are needed to run some of the CUDA tests.


fwyzard:
@smuzaffar , I don't think there is any way to build this as part of the "cuda" package, but with the libraries in a different directory and in a different tool ?

fwyzard:
@cmsbuild, please test

fwyzard:
@cmsbuild, please test

fwyzard:
@smuzaffar @davidlange6 can we have this in 10.0.0-pre3 ?

fwyzard:
Thanks,
.A

PR #3641, opened 2017-12-18
New dasgoclient version

vkuznet:
This version includes various fixes:
- https://github.com/dmwm/dasgoclient/issues/15
- https://github.com/dmwm/DAS/issues/4277
- fix for -json output
- fix for site block=/a/b/c#123 queries

### DIFF
--- a/dasgoclient-binary.spec
+++ b/dasgoclient-binary.spec
@@ -1,11 +1,11 @@
-### RPM cms dasgoclient-binary v01.01.00
+### RPM cms dasgoclient-binary v01.01.02
 Source0: git+https://github.com/dmwm/dasgoclient?obj=master/%{realversion}&export=dasgoclient&output=/dasgoclient.tar.gz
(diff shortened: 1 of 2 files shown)


smuzaffar:
please test

vkuznet:
The failure comes from invalid query in runTheMatrix:
```
file dataset=/MinimumBias/Run2012A-v1/RAW run=191226 site=T2_CH_CERN
```
This query produces 0 results which is considered as a failure in your test, but
it is the correct result since given dataset is not located at T2_CH_CERN:
```
dasgoclient -query 'site dataset=/MinimumBias/Run2012A-v1/RAW'
T0_CH_CERN_Export
T1_US_FNAL_Buffer
T1_US_FNAL_MSS
T0_CH_CERN_MSS
T1_US_FNAL_Disk
T3_US_Colorado
T2_BE_UCL
```
Please re-test with a dataset which properly placed on a site you're looking for.

smuzaffar:
yes, we have the same das error in the IBs.

PR #3651, opened 2017-12-25
included cms patches for geant4 10.04

smuzaffar:
These patches were part of 10.02.p02 branch but were not picked up for 10.04. This is the cause of failures on clang IBs.
- Do not use __has_feature(c_atomic) in C++ to detect std::atomic support : https://github.com/cms-externals/geant4/commit/7788cfbecfee824344bd3761cb32efe2bdc53046
- Do not use __has_feature(c_atomic) in C++11/C++14/C++1z: https://github.com/cms-externals/geant4/commit/f5a27d9b0a47f2a40fc27db06a3851685005bf72


smuzaffar:
please test

PR #3657, opened 2018-01-08

smuzaffar:
Add rule on macOS to use library ID @rpath/libname and to add lib productstore as an rpath: https://github.com/cms-sw/cmssw-config/pull/63


smuzaffar:
please test

PR #3668, opened 2018-01-16
Update the tag

mrodozov:
Testing the changes added here: https://github.com/cms-data/RecoEgamma-PhotonIdentification/pull/6 , new xml files

### DIFF
--- a/data-RecoEgamma-PhotonIdentification.spec
+++ b/data-RecoEgamma-PhotonIdentification.spec
@@ -1,4 +1,4 @@
-### RPM cms data-RecoEgamma-PhotonIdentification V01-00-04
+### RPM cms data-RecoEgamma-PhotonIdentification V01-00-05
 
 %prep
 


mrodozov:
please test

mrodozov:
please test

mrodozov:
please test

slava77:
new files were added, which are not used in the code yet.
So, the integration should be straightforward at this point: I propose to merge this.

slava77:
@mrodozov @smuzaffar 
please make the same change for 94_MAOD_X as well

Thank you.

This is needed for integration of https://github.com/cms-sw/cmssw/pull/22302

smuzaffar:
@slava77 , for now both 9.4.X and 9.4.MAOD.X share the same externals. Is it safe to put this in 9.4.X  (maod will also pick it up) otherwise we have to add separate branch for maod.

smuzaffar:
@slava77 , looks like this PR just add few new files, so I think this is same to put in 94X.

PR #3673, opened 2018-01-18
Add fortran flag for aarch64

mrodozov:
AArch64 builds were failing because of LoopTools failing to compile some fortran code, where the error log was suggesting to use the flag so to avoid the build error.

### DIFF
--- a/herwigpp.spec
+++ b/herwigpp.spec
@@ -1,10 +1,10 @@
 ### RPM external herwigpp 7.1.2
 Source: https://www.hepforge.org/archive/herwig/Herwig-%{realversion}.tar.bz2
 
-
 # Tried to comment out the parts which build HerwigDefaults.rpo during make install
 
 %define isamd64 %(case %{cmsplatf} in (*amd64*) echo 1 ;; (*) echo 0 ;; esac)
+%define isaarch64 %(case %{cmsplatf} in (*_aarch64_*) echo 1 ;; (*) echo 0 ;; esac)
 
 Requires: lhapdf
 Requires: boost 
@@ -58,6 +58,9 @@ PLATF_CONF_OPTS="--enable-shared --disable-static"
 %endif
             --prefix=%i \
             CXX="$CXX" CC="$CC" \
+%if %isaarch64
+            FCFLAGS="-fno-range-check" \
+%endif
 	    BOOST_ROOT="$BOOST_ROOT" LDFLAGS="$LDFLAGS -L$BOOST_ROOT/lib" \
             LD_LIBRARY_PATH=$LHAPDF_ROOT/lib:$GSL_ROOT/lib:$GOSAMCONTRIB_ROOT/lib:$MADGRAPH5AMCATNLO_ROOT/HEPTools/lib:$LD_LIBRARY_PATH
 


mrodozov:
please test

davidlt:
Does herwigpp have a testsuite? If yes, could you run it before and after the change, e.g. on x86_64? Just to figure out what could be potential impact.

Also could you share the actual error for the record?

mrodozov:
The actual error was a fortran lib fails to build in a subpackage of Herwig called LoopTools
```
PPF77    D/libHwLooptoolsXFC_la-Dget.lo
D/D0func.F:486:34:

  parameter (nz2 = -2147483648) ! O'20000000000'
                                  1
Error: Integer too big for its kind at (1). This check can be disabled with the option -fno-range-check
D/D0func.F:491:27:

  parameter (nz2p1234 = nz2 + p1234)
                           1
Error: Parameter 'nz2' at (1) has not been declared or is a variable, which does not reduce to a constant expression
D/D0func.F:495:27:
```
and after the error the program 'suggests' 

> This check can be disabled with the option -fno-range-chec

so that's why we've put the flag. I'm reading the manual of Herwig now to find if there are tests https://herwig.hepforge.org/index.html
The code trying to compile can be seen here: 
https://na62-sw.web.cern.ch/sites/na62-sw.web.cern.ch/files/doxygen/d5/dde/LoopTools214_2src_2LoopTools214_2build_2D0funcC_8F_source.html
although seems this is an older version.

mrodozov:
please test

mrodozov:
+externals

mrodozov:
@Andrej-CMS We integrated Herwig in 10_0_X but we have now this error that the PR fixes, could you ask the developers why do we have it in aarch64 or maybe for any other way to fix it ? 
Cheers,
M.

Andrej-CMS:
@mrodozov 

I just wrote an email to the Herwig authors. I will let you know when I get an answer.

Andrej-CMS:
@all 

I just put the answer of the Herwig authors here. Is there a way of giving them access to the test account, so they can debug it?

Hi Andrej,

First of all, aarch64 is an architecture we do not currently support
with testing in any way, mainly for lack of test machines. If you have a
way of giving us access to a test account, I can investigate a bit more.

The rest is therefore just an initial guess...

We use a copy of Looptools http://www.feynarts.de/looptools/ internally,
and the linking of that fortran code to the C++ was often tricky to get
right, especially with differing default sizes of numeric types. Your
error is in the pure fortran part, though, so not directly related to
the linking that Herwig does. Can you try compiling Looptools from the
original source above, and see if that works or not. It may be that the
problem is there upstream already.

I looks like on aarch64 the default fortran 'integer' (that your
compiler chooses) is somehow not big enough to hold -2147483648 (usually
this is the minimum that a signed 32-bit integer can be)

Specifying the -fno-range-check will silence the error, but will
potentially write a nonsense value into the variable. This is not a good
way to solve this problem until you know what's going on that causes the
error.

See you,

David

Andrej-CMS:
Dear @all,

Is there any statement on the possibility of giving a test account to the Herwig authors, so they can debug this issue?

Kind regards,
Andrej

fabiocos:
@Andrej-CMS I have punt this point into the ORP agenda for January 30.

perrozzi:
@Andrej-CMS is there any Herwig author able to access lxplus? We discuss in the meeting that this should be a sufficient condition

Andrej-CMS:
Dear @perrozzi , @all,

I talked to the authors . They can access lxplus. It depends on the number of users of the Aarch64 architecture whether they want to put manpower into this issue. 

How important is the support for this architecture for CMS? If this is an important issue, I think we should setup an email thread with the authors on this issue. 

Kind regards, 
  Andrej

smuzaffar:
@Andrej-CMS , same warning are also there for AMD64 with GCC7 compiler
https://cmssdt.cern.ch/SDT/cgi-bin/buildlogs/slc6_amd64_gcc700/CMSSW_10_1_X_2018-03-07-1100/GeneratorInterface/AMPTInterface

smuzaffar:
ah sorry , the warnings https://cmssdt.cern.ch/SDT/cgi-bin/buildlogs/slc6_amd64_gcc700/CMSSW_10_1_X_2018-03-07-1100/GeneratorInterface/AMPTInterface are new one for uninitialized value

Andrej-CMS:
Hello @smuzaffar ,
I made the authors aware of this PR and the new warning. 
lets see what they have to say.

PR #3677, opened 2018-01-22
Update cms-git-tools.spec

smuzaffar:
Included latest updates: Add an option to setup only the upstream repository https://github.com/cms-sw/cms-git-tools/pull/93
For now added only for Devel IBs. Once tested there then we will include it in normal IBs/Releases


smuzaffar:
please test

PR #3680, opened 2018-01-22
Root update v6-10-00 patches

smuzaffar:
This also includes "Fix ROOT-9173: add per object flag for the statsoverflow" https://github.com/root-project/root/pull/1503


smuzaffar:
please test

PR #3691, opened 2018-01-25
Add C++ wrappers for the CUDA API

fwyzard:
Add C++ wrappers for the CUDA API from https://github.com/eyalroz/cuda-api-wrappers

### DIFF
--- a/cmssw-tool-conf.spec
+++ b/cmssw-tool-conf.spec
@@ -1,7 +1,6 @@
-### RPM cms cmssw-tool-conf 40.0
+### RPM cms cmssw-tool-conf 41.0
 ## NOCOMPILER
-# With cmsBuild, change the above version only when a new
-# tool is added
+# With cmsBuild, change the above version only when a new tool is added
 
 %define islinux %(case $(uname -s) in (Linux) echo 1 ;; (*) echo 0 ;; esac)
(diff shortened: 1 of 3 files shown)


fwyzard:
@cmsbuild, please test

smuzaffar:
@fwyzard , I see following error message in the build log https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-3691/25635/external/cuda-api-wrappers/20180125/log . Is this expected?
```
-- Found CUDA: /build/cmsbld/jenkins/workspace/ib-any-integration/testBuildDir/slc6_amd64_gcc630/external/cuda/9.1.85-cms (found suitable version "9.1", minimum required is "7.0") 
/tmp/cuda-compute-version-helper-23001-1516918207.316623085: error while loading shared libraries: libcudart.so.9.1: cannot open shared object file: No such file or directory
-- CUDA device-side code will assume compute capability 

```

fwyzard:
Hi @smuzaffar, that would only affect the tests, which we don't actually build.
Anyway, I've updated the spec file to pass some parameters explicitly, which should avoid the error.

fwyzard:
@cmsbuild, please test

fwyzard:
@cmsbuild, please test

PR #3702, opened 2018-01-31
Fix vecgeom build fail on arm64 gcc700

mrodozov:


mrodozov:
please test

mrodozov:
please test

mrodozov:
please test

vecgeom.spec
```diff
@@ -19,7 +26,13 @@ cmake ../%{n}-%{realversion} \
   -DBACKEND=Scalar \
   -DUSOLIDS=ON \
   -DUSOLIDS_VECGEOM=ON \
+%if %isamd64
   -DVECGEOM_VECTOR=sse3 \
+%else
+  -DVECGEOM_VECTOR=native \
```
davidlt: Do we really want to have this? Native could be problematic if it picks specific CPU/SOC instructions.

vecgeom-fix-for-arm64.patch
```diff
@@ -0,0 +1,36 @@
...
+@@ -31,7 +31,18 @@ using uint = unsigned int;
+ #endif
+ #else
+ // Functionality of <mm_malloc.h> is automatically included in icc
++#if defined(__i386__) || defined(__x86_64__)
+ #include <mm_malloc.h>
++#else
++#if __cplusplus >= 201703L
++#include <cstdlib>
++#define _mm_malloc(a, b) ::aligned_alloc(b, a)
```
davidlt, line 27: Just a comment. I couldn't use here std:: namespace due to a bug in GCC. It's easy to fix and there is a ticket for GCC 8, but apparently it could break target in Windows thus not being fixed at this moment.

vecgeom.spec
mrodozov: I didn't want to cut any of the flags as I don't know how it would change the installation. The sse3 was what broke the arm64 IB so that's how we ended up with this PR. It transforms to -msse3 compiler option which the compiler doesn't recognize.

mrodozov:
please test

smuzaffar:
please test

PR #3717, opened 2018-02-06
updated bazel 0.10.0; tensorflow 1.5.0

smuzaffar:
Bazel/Tensorflow are failing for aarch64 IBs and we can not rebuilt it because of java version update which breaks bazel. This PR updates
- bazel 0.10.0 which also contains fixes for new java 1.8.0
- tensorflow 1.5.0
- eigen commit c7dc0a897676

We need to update bazel as existing version 0.4.5 does not work with the java 1.8.0_162 (https://github.com/bazelbuild/bazel/issues/4483). Bazel version 0.10.0 works with latest java but tensorflow 1.3.0 does not build with it (because of some deprecated functionality in bazel). So we propose to move to tensorflow 1.5.0.


smuzaffar:
please test

smuzaffar:
please test

PR #3723, opened 2018-02-08
Cherry-pick upstream changeset 10711:034b6c3e1017

fwyzard:
For cuda 9.1 replace math_functions.hpp with cuda_runtime.h


fwyzard:
@cmsbuild, please test

PR #3732, opened 2018-02-09
updated fastjet to 3.3.3 and fastjet-contrib to 1.032

smuzaffar:
This integrates the following requests in 10.1.X DEVEL IBs. Once it is validated in there then we will backport it to normal 10.1.X IBs.
- https://github.com/cms-sw/cmsdist/issues/3725
- https://github.com/cms-sw/cmsdist/issues/3731

### DIFF
--- a/fastjet-contrib.spec
+++ b/fastjet-contrib.spec
@@ -1,6 +1,6 @@
-### RPM external fastjet-contrib 1.026
-%define tag 63154ae21b8f40acfb4ee163ac720c39e260aabe
-%define branch cms/v%realversion
+### RPM external fastjet-contrib 1.032
+%define tag 6000f18568bd2f03fb6bbb89cedf0baeed5344ad
+%define branch cms/v%{realversion}
 %define github_user cms-externals
 Source: git+https://github.com/%github_user/%{n}.git?obj=%{branch}/%{tag}&export=%{n}-%{realversion}&foo=1&output=/%{n}-%{realversion}.tgz
 Requires: fastjet
--- a/fastjet.spec
+++ b/fastjet.spec
@@ -1,5 +1,5 @@
 ### RPM external fastjet 3.3.0
-%define tag 2db2b530b753af3ab39c91f2551e6a90597e1c13
+%define tag 26e1f4828f4622742062f1465ae8de9986f96f86
 %define branch cms/v%{realversion}
 %define github_user cms-externals
 Source: git+https://github.com/%{github_user}/fastjet.git?obj=%{branch}/%{tag}&export=%{n}-%{realversion}&output=/%{n}-%{realversion}.tgz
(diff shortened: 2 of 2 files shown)


smuzaffar:
please test

smuzaffar:
please test

slava77:
> Reco comparison results: 22 differences found in the comparisons

however, the link to comparisons is broken.

@smuzaffar  do you know where I can find the comparisons?

smuzaffar:
link was wrong it is pointing to DEVEL IB where as the actual contents are under the base release name. Please try the foolowing link (I will fix the cms-bot, to use the correct link).
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_10_1_X_2018-02-11-2300+3732/25025/

slava77:
@rappoccio

I'm trying to understand the differences in the tests.
ak4Cs4PFJets plots in HI workflow show up with differences 

there are 20% more jets (most increase is with tiny energies)
all_oldvsnew_runhi2011wf140p53c_log10recopfjets_akcs4pfjets__rereco_obj_energy
Based on the eta distribution, I guess the extra 20% are all directed along the beam somewhere
all_oldvsnew_runhi2011wf140p53c_recopfjets_akcs4pfjets__rereco_obj_eta


The constituents now come out with a huge spike at pt = 0
all_oldvsnew_runhi2011wf140p53c_log10recopfcandidates_akcs4pfjets_pfparticlescs_rereco_obj_pt

The CH frac looks funny with a bunch of values above 1
all_oldvsnew_runhi2011wf140p53c_recopfjets_akcs4pfjets__rereco_obj_chargedhadronenergyfraction

@mverwe

rappoccio:
Hmmm. These were only supposed to contain technical changes to my understanding (3.3.0 added a python interface, and contrib 1.032 made the N3 variable faster). 

The changes to constituent subtraction are:

http://fastjet.hepforge.org/svn/contrib/contribs/ConstituentSubtractor/tags/1.2.4/NEWS

@mverwe any ideas there?

rappoccio:
I think it's due to this:

> 
 - Important change for the output corrected particles in case of correction of massive particles. The corrected zero pt particles with non-zero corrected delta_m (massive particles in rest) are now included in the output by default. This is the recommended usage. The user can change this behaviour by using function "set_remove_all_zero_pt_particles". By using argument true, all the zero pt corrected particles are removed.

slava77:
@mverwe
would it be OK for you to use set_remove_all_zero_pt_particles?

@rappoccio 
is the interface new in this version or can we already ask for set_remove_all_zero_pt_particles in the older version (even though it will do nothing).
I'm trying to understand the ordering of PRs in different repos to converge to the right result

mverwe:
@slava77 : I'm ok with activating set_remove_all_zero_pt_particles.

rappoccio:
@slava77 This was added in their version 1.2.0, unfortunately, so no.

slava77:
perhaps the most transparent for the physics performance is to make a PR in cmssw that uses  set_remove_all_zero_pt_particles
and then merge that and the external at the same time

rappoccio:
Great. @mverwe Do you want to handle that or should I?

mverwe:
@rappoccio : I will handle that

PR #3740, opened 2018-02-15
Fix nvcc.profile and update CUDA compiler flags (10.1.x)

fwyzard:
- let `nvcc` find its components when invoked from the command line
  - update default CUDA compilation flags to include `-O2 -std=c++14`

### DIFF
--- a/cuda.spec
+++ b/cuda.spec
@@ -10,9 +10,8 @@ AutoReq: no
 %build
 
 %install
-cp %{SOURCE0} %_builddir
 mkdir -p %_builddir/tmp
-/bin/sh %_builddir/%{n}_%{realversion}_%{driversversion}_linux --silent --tmpdir %_builddir/tmp --extract %_builddir
+/bin/sh %{SOURCE0} --silent --tmpdir %_builddir/tmp --extract %_builddir
 # extracts:
 # %_builddir/NVIDIA-Linux-x86_64-387.26.run
 # %_builddir/cuda-linux.9.1.85-23083092.run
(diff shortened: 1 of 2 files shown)


fwyzard:
@cmsbuild, please test

fwyzard:
type bugfix

cuda.spec
```diff
@@ -59,3 +58,22 @@ ln -sf libcuda.so.%{driversversion}                                     %{i}/dri
...
 cp -ar %_builddir/nvidia/libnvidia-ptxjitcompiler.so.%{driversversion}  %{i}/drivers/
 ln -sf libnvidia-ptxjitcompiler.so.%{driversversion}                    %{i}/drivers/libnvidia-ptxjitcompiler.so.1
+
+%post
+# let nvcc find its components when invoked from the command line
+cat > $RPM_INSTALL_PREFIX/%{pkgrel}/bin/nvcc.profile << @EOF
+
+TOP              = $CMS_INSTALL_PREFIX/%{pkgrel}
+
+NVVMIR_LIBRARY_DIR = \$(TOP)/nvvm/libdevice
```
smuzaffar, line 68: why not use sed or perl to replace `_HERE_` with `$CMS_INSTALL_PREFIX/%{pkgrel}/bin`?

fwyzard: because we also make other adjustments to the directory structure to skip the `targets/x86_64-linux/` part

smuzaffar: so can we also replace `_TARGET_DIR_`  with empty string then?
Overriding the full file means we might drop/add something in it which cuda might have updated in new version.

fwyzard: Sorry for the delay - can you check #3745 ?

PR #3741, opened 2018-02-15
Fix nvcc.profile and update CUDA compiler flags (10.0.x)

fwyzard:
- let `nvcc` find its components when invoked from the command line
  - update default CUDA compilation flags to include `-O2 -std=c++14`


fwyzard:
backport #3740

fwyzard:
@cmsbuild please test

fwyzard:
type bugfix

PR #3753, opened 2018-02-20
updated scram and cmssw-config tag

smuzaffar:
- new scram: which make sure any special env variables (e.g. LD_PRELOAD) are not set if scram runtime env is not set.
- cmssw-config: Allow to override default library name via `<export><lib name="myname"/></export>` section. This helps if you have packages with depth more than 2 for example one can have src/SubSystem/Package/Package/src and would like to give special name to this public library.


smuzaffar:
please test

PR #3788, opened 2018-03-02
Python3 spec file

davidlange6:
Thanks to @smuzaffar for a working python3 spec.

in addition, we've added a PYTHON27PATH variable to eventually remove our usage of PYTHONPATH (so possible to support python2 and 3 for some period)

### DIFF
--- a/python3.spec
+++ b/python3.spec
@@ -1,50 +1,21 @@
-### RPM external python3 3.6.2
+### RPM external python3 3.6.4
 ## INITENV +PATH PATH %{i}/bin
 ## INITENV +PATH LD_LIBRARY_PATH %{i}/lib
-## INITENV SETV PYTHON_LIB_SITE_PACKAGES lib/python%{python_major_version}/site-packages
+## INITENV SETV PYTHON_LIB_SITE_PACKAGES lib/python%{pythonv}/site-packages
 ## INITENV SETV PYTHONHASHSEED random
 # OS X patches and build fudging stolen from fink
-%{expand:%%define python_major_version %(echo %realversion | cut -d. -f1,2)}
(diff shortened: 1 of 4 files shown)


davidlange6:
please test

davidlange6:
(most of cmssw-tool-conf is rebuild in my area by now-so the test should be ok) - i would propose to merge both sets of python changes before the 2300 tonight

python3.spec
```diff
@@ -0,0 +1,85 @@
...
+find %i -name '*.pyo' -exec rm {} \;
+
+# Remove documentation, examples and test files.
+rm -rf %{i}/share %{i}/lib/python*/test %{i}/lib/python%{pythonv}/distutils/tests %{i}/lib/python%{pythonv}/lib2to3/tests
+
+echo "from os import environ" > %i/lib/python%{pythonv}/sitecustomize.py
+echo "if 'PYTHON3PATH' in environ:" >> %i/lib/python%{pythonv}/sitecustomize.py
+echo "   import sys,site" >> %i/lib/python%{pythonv}/sitecustomize.py
+echo "   for p in environ['PYTHON3PATH'].split(':'):">> %i/lib/python%{pythonv}/sitecustomize.py
+echo "      site.addsitedir(p)" >> %i/lib/python%{pythonv}/sitecustomize.py
```
smuzaffar: should we change this logic to something like
```
from os import environ
if 'PYTHON3PATH' in environ:
  import sys
  sys.path =environ['PYTHON3PATH'].split(':') + sys.path
```
this way without any changes in SCRAM or buildrules (to dump .pth files) we can just update toolfiles to replace PYTHONPATH with PYTHON3PATH

davidlange6:
please test

PR #3791, opened 2018-03-04
Update Eigen to the HEAD as of 10726:de7544f256bd

fwyzard:
Update Eigen to the tip of the development branch, as of 10726:de7544f256bd, Fri Feb 16 22:26:01 2018 +0400.

### DIFF
--- a/eigen.spec
+++ b/eigen.spec
@@ -1,4 +1,4 @@
-### RPM external eigen 034b6c3e1017
+### RPM external eigen de7544f256bd
 ## INITENV +PATH PKG_CONFIG_PATH %{i}/share/pkgconfig
 ## NOCOMPILER
 %define tag %{realversion}


fwyzard:
@cmsbuild please test

fwyzard:
@cmsbuild please test

fwyzard:
A complete rebuild of CMSSW_10_1_X_2018-03-05-1100 with this updated version of Eigen seems to have been successful.
Let's see if rerunning the tests is now OK.

smuzaffar:
@fwyzard , we also need to update tensorflow (which internally has a copy of eigen: https://github.com/cms-externals/tensorflow/blob/654f6427ef945ea5a69c217cec4de1102e78f2b3/tensorflow/workspace.bzl#L96) to match the version you are proposing here.
As tensorflow update just went in and only be available for today;s 23h00 IB. So lets wait for couple of IBs and then we will update eigen and tensorflow.

fwyzard:
So maybe tensorflow should *not* have a copy of eigen, and instead use the one we package ?

fwyzard:
Actually, I don't see any failed tests ?

fwyzard:
Actually, I don't see any failed tests ?

davidlange6:
Indeed. Unfortunately that’s not the development path they choose

Cheers,
David


On 5 Mar 2018, at 23:22, Andrea Bocci <notifications@github.com<mailto:notifications@github.com>> wrote:


So maybe tensorflow should not have a copy of eigen, and instead use the one we package ?

fwyzard:
OK, so, what about doing
```
rm -rf tensorflow/third_party/eigen3/
ln -s $EIGEN_ROOT/include/eigen3/ tensorflow/third_party/eigen3
```
when building tensorflow ?

fwyzard:
@cmsbuild please test

davidlange6:
and then?

davidlange6:
whoops didn't mean to press send - and then? - one still has to watch for no conflicts. But otherwise this can work if its supported by rpm (which I presume it is)

fwyzard:
I mean:
  - if Tensorflow needs (or just wants) to use its own version of Eigen, and does not "leak" it or export it in any way, then we should let it use it, *but* we should not need to synchronise the versions of Eigen used by Tensorflow and CMSSW
  - if the version of Eigen used by Tensorflow needs to be in sync with the one used by CMSSW and the rest of our stack, then we should make Tensorflow pick Eigen from the external package

In the latter scenario (which  I guess is the one we are in) we still have to make sure that whatever version of Eigen we want to use, Tensorflow is fine with it; but at least we don't need to push the same updates in two different places.

Is there some other benefit from having two separate versions of Eigen (one for Tensorflow, one for CMSSW), but having to keep them in sync ?

fwyzard:
@cmsbuild please test

smuzaffar:
please test

smuzaffar:
closing in favor of https://github.com/cms-sw/cmsdist/pull/3809 which now uses eigen for cms-external/eigen-git-mirror

smuzaffar:
@fwyzard , https://github.com/cms-sw/cmsdist/pull/3801 should allow to update eigen and protobuf sources in one place and tensorflow should be pick those sources.

fwyzard:
I see... not what I had in mind, but I guess it does get the libraries to be consistent.

PR #3709, opened 2018-02-02
updates of TrackerDAQ package (version 4.2)

rgerosa:
New version of trackerDAQ which contains updates needed to commit in the cmssw SiStrip commissioning workflow (part of DQM) a new analysis identifying bad channels in the detector. The new spec file/ tarball package has been tested for slc6_amd64_gcc630 and slc7_amd64_gcc630.

PR #3810, opened 2018-03-07
Tracker daq 4.2 gcc700

rgerosa:
With this PR, we have updated the tracker online source code to be compliant with gcc 7.0. This PR follows [0], which was merged in the gcc630 branch. With respect to the past, where a huge patch file with O(20k) lines was used to patch the tkonline code to compile under gcc7,  we have now produced a new tarball for the source code, as indicated in the spec file.

[0]https://github.com/cms-sw/cmsdist/pull/3709


smuzaffar:
please test

smuzaffar:
@rgerosa , looks like there are still some places left with throw(). Can you please fix these.

rgerosa:
@smuzaffar, ah yes there were some "throw" calls left in the code which were not failing the compilation of tkonline itself. Should be solved now. The tarball has been updated so it does not require an update of the spec files.

smuzaffar:
please test

PR #3824, opened 2018-03-08
Bump bazel to 0.11.1

mrodozov:
Update bazel to 11.1 after testing if it builds tf (on amd64 with gcc63 and on arm with gcc7)


smuzaffar:
please test

PR #3831, opened 2018-03-09
updated to use patched pythia8 from cms-external/pythia8

smuzaffar:
Ref: https://github.com/cms-sw/cmsdist/issues/3829


smuzaffar:
please test

PR #3846, opened 2018-03-15
New dasgoclient version

vkuznet:
Fixed summary query issue when user send run list with between clause.


smuzaffar:
please test

PR #3858, opened 2018-03-21
remove use of pythonpath in cmssw 

davidlange6:
most of this migration is already done - this should remove the last few uses 
[bonus, I removed beautifulsoup.spec as its not used and not recently updated.


davidlange6:
please test

PR #3869, opened 2018-03-22
Make a switch from das_client to dasgoclient (DO NOT MERGE)

vkuznet:


vkuznet:
Please do not merge yet this PR. It is proposal to review before making a switch. @smuzaffar could you please review and let me know if this is what we need. If so, per our DM agreement we need to perform a switch on last day of March.

dasgoclient.spec
```diff
@@ -44,3 +44,13 @@ if [ "`ls ${RPM_INSTALL_PREFIX}/*/%{pkgcategory}/%{pkgname}/v*/etc/profile.d/ini
   /bin/cp -f ${RPM_INSTALL_PREFIX}/%{pkgrel}/etc/dasgoclient $RPM_INSTALL_PREFIX/common/dasgoclient.tmp
   mv $RPM_INSTALL_PREFIX/common/dasgoclient.tmp $RPM_INSTALL_PREFIX/common/dasgoclient
 fi
+
+# make soft link between dasgoclient and das_client
+if [ -f $RPM_INSTALL_PREFIX/common/das_client ]; then
+   rm $RPM_INSTALL_PREFIX/common/das_client
+fi
+ln -s $RPM_INSTALL_PREFIX/common/dasgoclient $RPM_INSTALL_PREFIX/common/das_client
```
cmsbot: @vkuznet , one problem which we will have is that if someone install an old cmssw release after installing this? old release will bring an old das_client which will override the das_client symlink. I would suggest to drop the above change and make use of override/bin/das_client. No package yet creates share/overrides/bin/das_client so I would suggest to create override/bin/das_client symlink pointing to common/dasgocleint.

Only issue left is for those release where we call das_client.py (which is already in available in share/override/bin) and I do not see any simple solution for this.

smuzaffar: @vkuznet , one problem which we will have is that if someone install an old cmssw release after installing this? old release will bring an old das_client which will override the das_client symlink. I would suggest to drop the above change and make use of override/bin/das_client. No package yet creates share/overrides/bin/das_client so I would suggest to create override/bin/das_client symlink pointing to common/dasgocleint.

Only issue left is for those release where we call das_client.py (which is already in available in share/override/bin) and I do not see any simple solution for this.

smuzaffar:
@vkuznet , please see my comment https://github.com/cms-sw/cmsdist/pull/3869/files#r176945764
plus I would suggest to drop the das_client-toolfile dependency here https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_10_1_X/gcc630/cmssw-tool-conf.spec#L25

vkuznet:
Shahzad, I dropped requested block and I agree that we can remove das_client-toolfile but probably after the switch. Please review and decide when to merge.

smuzaffar:
please test

smuzaffar:
@vkuznet , I am fine with the change. Should we merge it today so that it become part of 10.1.0 release and get deployed? Note that 10.1.0 will be out tomorrow or day after tomorrow.

vkuznet:
Yes, let's merged today.

PR #3883, opened 2018-03-26
Add Google common python package

davidlange6:
This is nothing more than a __Init__.py file that makes protobuf (and eventually other google python products) work without their internal pth file.


davidlange6:
please test

PR #3885, opened 2018-03-27
Use copy rather than rsync in pippkgs packaging step

davidlange6:
deal with duplicate directories in different python tools (more) correctly - rsync was overwriting the first by the second - cp -r instead keeps both (at least it works in the one case we have at the moment)

### DIFF
--- a/py2-pippkgs_depscipy.spec
+++ b/py2-pippkgs_depscipy.spec
@@ -14,6 +14,7 @@ BuildRequires: py2-rootpy
 %if %isamd64
 BuildRequires: py2-tensorflow
 %endif
+BuildRequires: py2-googlecommon
 BuildRequires: py2-protobuf
 
 BuildRequires: py2-tables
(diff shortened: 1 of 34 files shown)


davidlange6:
please test

py2-pippkgs_depscipy.spec
```diff
@@ -58,14 +59,18 @@ for pkg in %builddirectpkgreqs ; do
...
         fi
+        fi
 
       fi
     done
     echo "Copying $SOURCE in %{pkgrel}"
-    rsync -av $SOURCE/ %{i}/${PYTHON_LIB_SITE_PACKAGES}/
+#try cp instead of rsync as we don't want to overwrite duplicate directories
+#    rsync -av $SOURCE/ %{i}/${PYTHON_LIB_SITE_PACKAGES}/
+    cp -r ${SOURCE}/* %{i}/${PYTHON_LIB_SITE_PACKAGES}/
```
smuzaffar, line 73: @davidlange6 , do we need such a package any more? We can just explicitly add these py2-* packages dependency in cmssw-tool-conf ?

davidlange6:
> On Mar 27, 2018, at 11:33 AM, Malik Shahzad Muzaffar <notifications@github.com> wrote:
> 
> @smuzaffar commented on this pull request.
> 
> In py2-pippkgs_depscipy.spec:
> 
> >  
>        fi
>      done
>      echo "Copying $SOURCE in %{pkgrel}"
> -    rsync -av $SOURCE/ %{i}/${PYTHON_LIB_SITE_PACKAGES}/
> +#try cp instead of rsync as we don't want to overwrite duplicate directories
> +#    rsync -av $SOURCE/ %{i}/${PYTHON_LIB_SITE_PACKAGES}/
> +    cp -r ${SOURCE}/* %{i}/${PYTHON_LIB_SITE_PACKAGES}/ 
> 
> @davidlange6 , do we need such a package any more? We can just explicitly add these py2-* packages dependency in cmssw-tool-conf ?

We can indeed move in that directlon (modulo the performance cost of very long lists inside a pth).  I wasn't quite ready to do that yet.

However it doesn't actually solve the protobuf issue that is behind this thread of commits. To work without pths, protobuf needs the google-common piece installed on its same path.

[ok, that said, the DEVEL build of last night doesn't show problems as I think it should (and did in the testing PR)]

davidlange6:
merge

PR #3896, opened 2018-04-04
Python3 builds for packages not depending on root

davidlange6:
ROOT appears to support only one python version at a time, so exclude that for now. I'll need a better solution for tensorflow later. 

Also removed the no longer needed large python bundles. (performance consequence to be tested)

[and this doesn't merge for some reason]


davidlange6:
interesting, its too complex to resolve the conflicts here - the non-devel branch does merge, so I guess there is some large difference between the two now. I'll resolve this tomorrow.

PR #3902, opened 2018-04-06
Get llvmlite from source & bring numba back

mrodozov:
This gets numba 23 from source as it's not yet available as pip package and brings back llvmlite and numba for DEVEL IBs


mrodozov:
please test

PR #3926, opened 2018-04-17
Geant4: Update the commit number after fix on that branch

mrodozov:
Fix description: https://github.com/cms-externals/geant4/pull/28

### DIFF
--- a/geant4.spec
+++ b/geant4.spec
@@ -1,5 +1,5 @@
 ### RPM external geant4 10.04
-%define tag 7788cfbecfee824344bd3761cb32efe2bdc53046
+%define tag d3f235ecd35702375b6c6f60b90e062d1051d7da
 %define branch cms/4.%{realversion}
 %define github_user cms-externals
 Source: git+https://github.com/%github_user/%{n}.git?obj=%{branch}/%{tag}&export=%{n}.%{realversion}&output=/%{n}.%{realversion}-%{tag}.tgz


mrodozov:
please test

civanch:
@mrodozov , thanks

civanch:
No difference versus baseline is expected, because fixed code is called only in special circumstances, like tau particle gun and should not affect baseline productions.

mrodozov:
+externals

fabiocos:
@civanch @smuzaffar @mrodozov I think we should include this fix in 10_2_0_pre1. But as far as I see it is not added as a patch, it is a change of the baseline distribution, am I correct?

civanch:
@fabiocos , Geant4 distribution includes base release (like 10.4) and patches (like 10.4p01). CMS may add CMS patches to Geant4 on top of a Geant4 release (as we in this PR) and on top of a Geant4 patch. This particular fix may be included in 10.4p02 which not yet exists. In the current branch we already add 3 another fixes in December, this one will be the 4th.

Possible migration of CMSSW from 10.4 to one of Geant4 patches will require PPD validation. For Geant4 patches, which are taken by CMS, we create separate branches. In previous cases, we migrated to a new Geant4 patches with care and practically in all production versions we have extra CMS private patches, which may come very late. Usually, if a Geant4 crash happens in the middle of big CMS production we fixed it always as a CMS private patch.

fabiocos:
@civanch thank you Vladimir, I see that now things are managed differently compared with what we were doing some 5 years ago. Fine for the integration.

fabiocos:
@smuzaffar we can integrate this immediately as far as I am concerned, unless you prefer to avoid a full rebuild for the time being

fabiocos:
+1

PR #3928, opened 2018-04-18
renamed tensorflow specs; add tfcompile

smuzaffar:
Replaced Remove tensorflow-cc and tensorflow-c specs with tensorflow.spec. Added tfcompile executable along with tensorflow-runtime and xla_compiled_cpu_function


smuzaffar:
please test

PR #3931, opened 2018-04-18
Update pythia8.spec

smuzaffar:
A crash was found in Phase2 production using 9_3_6: see detail here https://github.com/cms-externals/pythia8/pull/10


smuzaffar:
please test

PR #3937, opened 2018-04-20
Updates for DEVEL branch

davidlange6:
1) remove developments only in DEVEL that are not needed any longer
2) add future package to help with python3 migration work


davidlange6:
lets use the 1100 DEVEL IB to confirm these changes

smuzaffar:
@davidlange6 , we do not run DEVEL IBs at 11h. I will start a special one later in the evening when jenkins is bit free.

davidlange6:
ah - no rush! 2300 is ok too:)

PR #3962, opened 2018-04-26
Hack CUDA to support GCC 7

fwyzard:
First, patch CUDA headers not to reject GCC 7.

Then, force the host compiler to use C++14 for CUDA code, because with GCC 7 we cannot safely mix C++17 preprocessing with C++14 compilations.

### DIFF
--- a/cuda-toolfile.spec
+++ b/cuda-toolfile.spec
@@ -41,6 +41,8 @@ cat << \EOF_TOOLFILE >%{i}/etc/scram.d/cuda.xml
   <flags CUDA_LDFLAGS="-dlink"/>
   <flags CUDA_LDFLAGS="-shared"/>
   <flags CUDA_LDFLAGS="-L$(CUDA_BASE)/lib64"/>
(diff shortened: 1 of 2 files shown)


fwyzard:
@cmsbuild, please test

fwyzard:
@cmsbuild, please test

fwyzard:
Do you know if this build error is already there in the IBs ?

smuzaffar:
please test 
@fwyzard , the error is due to the fact that we updated rot to 6.12 (which is not yet part of the IB but is merged in cmsdist). Last night IB has the new root. I have restarted the test which should use the new IB

fwyzard:
Ah OK.

I did test these changes locally, but only on the CUDA code.

.A

fwyzard:
Hi @smuzaffar ,
could you let me know when the SCRAM rules are updated as well ?

Thank you,
.A

smuzaffar:
@fwyzard , sorry not yet. Will do it later in the evening.

PR #3967, opened 2018-04-30
Update jemalloc to version 5.0.1

fwyzard:


fwyzard:
For a simple benchmark (running a recent HLT menu), running with jemalloc 5.0.1 is almost 6% faster than running with jemalloc 4.5.0 .

fwyzard:
@cmsbuild please test

PR #4004, opened 2018-05-16
Update boost to 1.67

mrodozov:
Try to update boost to 1.67.0 applying all patches existing (and not applied already) from 1.63


mrodozov:
please test

mrodozov:
please test

PR #4012, opened 2018-05-18
backported xrootd and git from DEVEL IBs

smuzaffar:
Back ported updated from DEVEL IBs
- xrootd to 4.8.3
- git 2.17.0 with fix for git-gui


smuzaffar:
please test

PR #4024, opened 2018-05-22
New dasgoclient version

vkuznet:
This is new dasgoclient version based on up-coming das2go server which is scheduled to be in production on cmsweb in early June. It fixed several user-based queries and aligned with new das2go version.


smuzaffar:
please test

PR #4029, opened 2018-05-24
RFC: mirror and build from the (unofficial) LLVM git "monorepo"

fwyzard:
The "monorepo" is hosted at https://github.com/llvm-project/llvm-project-20170507/ .
See http://llvm.org/docs/GettingStarted.html#for-developers-to-work-with-a-git-monorepo for details about the "monorepo" and build commands.


fwyzard:
If we like this approach (to me the benefit is that it seems to be easier to build LLVM and CLANG) I can make a PR for the 630 arch.

fwyzard:
@cmsbuild please test

smuzaffar:
thanks @fwyzard. Yes it should be cleaner/easier to update from one repo.

PR #4030, opened 2018-05-24
Mirror and build from the (unofficial) LLVM git "monorepo"

fwyzard:
The "monorepo" is hosted at https://github.com/llvm-project/llvm-project-20170507/ .
See http://llvm.org/docs/GettingStarted.html#for-developers-to-work-with-a-git-monorepo for details about the "monorepo" and build commands.


fwyzard:
@cmsbuild, please test

PR #4063, opened 2018-05-31
[10.2.X] RecoBTag-Combined: Update to retrained DeepDoubleB models

smuzaffar:
To include https://github.com/cms-data/RecoBTag-Combined/pull/13

### DIFF
--- a/data-RecoBTag-Combined.spec
+++ b/data-RecoBTag-Combined.spec
@@ -1,4 +1,4 @@
-### RPM cms data-RecoBTag-Combined V01-00-10
+### RPM cms data-RecoBTag-Combined V01-00-11
 
 %prep
 


fabiocos:
@slava77 from your comment https://github.com/cms-sw/cmssw/pull/23033#issuecomment-393649652 I understand that https://github.com/cms-sw/cmssw/pull/23033 and this external PR are technically independent, but I assume they are not from the physics content point of view. So it would be useful to have an explicit "+1" to merge when the tests are done

slava77:
> So it would be useful to have an explicit "+1" to merge when the tests are done

it sounds like the tests need to be redone because they never completed.
The full build approach we are using for external data files is excessive.

slava77:
> it sounds like the tests need to be redone because they never completed.

nevermind, the "Comparison is ready" just never made it to this thread, while it shows up in https://github.com/cms-sw/cmssw/pull/23033#issuecomment-393739110

slava77:
> So it would be useful to have an explicit "+1" to merge when the tests are done

+1 for reco
based primarily on the two plots in https://github.com/cms-sw/cmssw/pull/23033#issuecomment-393649652 . Jenkins comparisons show that changes are in the expected places.

fabiocos:
+1

fabiocos:
merge

smuzaffar:
+externals

PR #4064, opened 2018-05-31

smuzaffar:
Build rules for 10.2.X were changed to generate filename.cc.o instead of filename.o . This change broke the classes.h and headers.h (serialization) dependency. This PR should fix the issue.


smuzaffar:
please test

PR #4067, opened 2018-06-01
added python-paths tools to provide python27 and python3 files

smuzaffar:
Avoid setting cmsw-tool-conf python paths via cmssw-config/Self.xml file. This breaks the cmsdist PR testing as in this case we always points to pythons paths from release. Thsi change should allow us to pick things from local area.


smuzaffar:
please test

smuzaffar:
please test

PR #4099, opened 2018-06-08
Improve the procedure to install the CUDA tools in CMSSW

fwyzard:
### Leave out nvidia `nsight` (IDE) and `nvvp` (visual profiler) from the CUDA distribution.

These tools are only useful on the developer's machine, not on the worker nodes, and we expect the developers to run them on their local desktops/notebooks rather than on lxplus.
Leaving them out saves almost 500 MB from the archive and installation.

###  Handle the case where `$CUDA_BASE/include` is a symlink

This is needed for the AMR64 packaging of CUDA.

###  Add a wrapper for `cuda-gdb` and package its support files
Set `PYTHONHOME` to scram's `$PYTHON_BASE` to fix the error:

> Could not find platform independent libraries <prefix>
> Could not find platform dependent libraries <exec_prefix>
> Consider setting $PYTHONHOME to <prefix>[:<exec_prefix>]
> ImportError: No module named site

Package `$CUDA_BASE/share/gdb` to fix the error:

  > Python Exception <type 'exceptions.NameError'> Installation error: gdb.execute_unwinders function is missing:

### DIFF
--- a/cuda.spec
+++ b/cuda.spec
@@ -28,8 +28,6 @@ mkdir -p %_builddir/tmp
 rm -rf %_builddir/lib64/libcublas.so.9.2.88
 rm -rf %_builddir/lib64/libnvblas.so.9.2.88
 
-ln -sf ../libnvvp/nvvp %_builddir/bin/nvvp
-ln -sf ../libnsight/nsight %_builddir/bin/nsight
 mkdir -p %{i}/lib64
 # package only runtime and device static libraries
 cp -ar %_builddir/lib64/libcudart_static.a %{i}/lib64/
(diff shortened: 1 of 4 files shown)


fwyzard:
@cmsbuild, please test

fwyzard:
@cmsbuild, please test

fwyzard:
hold

fwyzard:
@cmsbuild, please test

fwyzard:
unhold

cuda.spec
```diff
@@ -60,9 +63,21 @@ rm -rf %_builddir/bin/nsight
...
 rm -rf %_builddir/bin/nvvp
 rm -rf %_builddir/bin/computeprof
+
+# add a wrapper for cuda-gdb and package the support files
+mkdir %{i}/share
+cp -ar %_builddir/share/gdb/ %{i}/share/
+mv %_builddir/bin/cuda-gdb %_builddir/bin/cuda-gdb.real
+cat > %_builddir/bin/cuda-gdb << @EOF
+#! /bin/bash
+export PYTHONHOME=\$(scram tool tag python PYTHON_BASE)
```
smuzaffar: @fwyzard , you can add python dependency and use $PYTHON_ROOT in the wrapper. Do not forget to relocate the wrapper in %post

fwyzard: Mhm, I'd rather not add to cuda the dependency on python, to avoid rebuilding cuda too often.
I'll stick with finding it at runtime, and maybe change it later.

fwyzard:
@cmsbuild, please test

cuda.spec
smuzaffar: @fwyzard , ok valid point. How about creating a new cuda-gdb.spec  which depends on both cuda and python? You will also need cuda-gdb-toolfile.spec .

fwyzard: I thought about it, but the hour was getting late...

What happens if both `cuda` and `cuda-gdb-wrapper` have a binary called `cuda-gdb` ?

fwyzard:
@cmsbuild, please test

cuda.spec
smuzaffar: @fwyzard , if there is no dependency in between cuda-gdb-wrapper and cuda toolfiles then itis undefined behaviour. But if you make cuda-gdb-wrapper toolfile depend on cuda then cuda-gdb from cuda-gdb-wrapper will be in PATH via exteranls/arch/bin directory and cuda-gdb from cuda will end up in externals/arch/bin00 direcotry (which is not in PATH).

fwyzard: Ah, so that's what `.../bin00` is about...

Anyway, I've kept the renaming in `cuda.spec` and moved the rest to a separate spec and tool file, so it should be OK now.

smuzaffar:
+externals

PR #4106, opened 2018-06-11
Add xrootd lib/python27 in PYTHON27PATH

smuzaffar:
This should fix the issue reported here https://github.com/cms-sw/cmssw/issues/23560


smuzaffar:
please test

PR #4129, opened 2018-06-20
use seperate tool for qd_f_main

smuzaffar:
libqd_f_main.so contains main() so it should not be linked directly to cmssw libs. This PR breaks the qd toolfile in to two
- qd.xml:  which contains qd and qdmod libs
- qd_f_main.xml: which depends on qd.xml and adds qd_f_main.


smuzaffar:
please test

PR #4165, opened 2018-07-02
Updated root to tip of branch v6-14-00-patches

smuzaffar:


smuzaffar:
please test

slava77:
in 136.8311
```
----- Begin Fatal Exception 02-Jul-2018 23:01:59 CEST-----------------------
An exception of category 'FatalRootError' occurred while
   [0] Calling function StreamSchedule::processOneEvent
   Additional Info:
      [a] Fatal Root Error: @SUB=TStreamerInfo::Build
atomic<vector<const void*>*>: __atomic_base<vector<const void*>*> has no streamer or dictionary, data member "_M_b" will not be saved
```
this is the same as we've seen in tests of #22594 

it looks like the head of the root repo is broken for us.

smuzaffar:
I have opened a JIRA ticket https://sft.its.cern.ch/jira/browse/ROOT-9513 for this. Both ROOT 6.14 and master branches are broken

PR #4167, opened 2018-07-03
Add MXNet prediction tool

hqucms:
This is to add [MXNet](https://mxnet.incubator.apache.org/) to CMSSW externals following the discussions in https://github.com/cms-sw/cmssw/issues/21314. It is needed for the integration of the [DeepAK8](https://twiki.cern.ch/twiki/bin/view/CMS/DeepAKXTagging) tagger into CMSSW. 

This includes both the C and C++ APIs. MXNet is built in prediction-only mode (`MXNET_PREDICT_ONLY=1`) here, as so far we only intend to use it for the evaluation of the neural networks, not for training (which is typically done on a GPU w/ the python API). By compiling in the prediction-only mode, MXNet is also set to run in [Native Engine](https://mxnet.incubator.apache.org/faq/env_var.html#engine-type) mode such that it runs in the master thread instead of creating its own thread pool.

### DIFF
--- a/cmssw-tool-conf.spec
+++ b/cmssw-tool-conf.spec
@@ -152,6 +152,7 @@ Requires: madgraph5amcatnlo-toolfile
 Requires: python_tools
 Requires: dasgoclient
 Requires: OpenBLAS-toolfile
+Requires: mxnet-predict-toolfile
 
 # Only for Linux platform.
 %if %islinux
--- a/mxnet-predict-toolfile.spec
+++ b/mxnet-predict-toolfile.spec
@@ -1,0 +1,22 @@
+### RPM external mxnet-predict-toolfile 1.2.0
+Requires: mxnet-predict
+%prep
+
+%build
+
+%install
+
+mkdir -p %i/etc/scram.d
+cat << \EOF_TOOLFILE >%i/etc/scram.d/mxnet-predict.xml
+<tool name="mxnet-predict" version="@TOOL_VERSION@">
+  <lib name="mxnetpredict"/>
+  <client>
+    <environment name="MXNET_PREDICT_BASE" default="@TOOL_ROOT@"/>
+    <environment name="INCLUDE" default="$MXNET_PREDICT_BASE/include"/>
+    <environment name="LIBDIR" default="$MXNET_PREDICT_BASE/lib"/>
+  </client>
+  <use name="openblas"/>
+</tool>
+EOF_TOOLFILE
+
+## IMPORT scram-tools-post
--- a/mxnet-predict.spec
+++ b/mxnet-predict.spec
@@ -1,0 +1,22 @@
+### RPM external mxnet-predict 1.2.0
+%define projectname mxnet-predict
+%define releasename apache-mxnet-src-%{realversion}-incubating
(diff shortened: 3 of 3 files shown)


hqucms:
@gouskos who would also like to follow this thread.

davidlange6:
please  test

smuzaffar:
@hqucms , please also update cmssw-tool-conf.spec to depend on mxnet-predict-toolfile. This is needed to get it within cmssw env.

smuzaffar:
@mrodozov , can you please check if this builds correctly on slc7/aarch64 and , gcc7/gcc8?

hqucms:
@smuzaffar Done.

mrodozov:
tested on slc6_amd64_gcc630,700 slc7_amd64_gcc700,810 slc7_aarch64_gcc700
- builds fine.

mrodozov:
please test

smuzaffar:
+externals

slava77:
@smuzaffar @fabiocos 
Please let me know what is the plan for merging this.
Thank you.

fabiocos:
@slava77 as the build looks functional I think that we have no technical obstacle to integrate it, @smuzaffar please comment in case.

Anyway, looking at the discussions at the root of this request, I would like to better understand the overall strategy, as this is the 3rd deep learning-oriented external tool that we are integrating to my knowledge, besides tensorflow and lwtnn.  From the slides of @Dr15Jones and @makortel at the november O&C meeting I understand that the idea is to use the DeepAK8 tagger based on MXNet as a testbed for that approach. As I did not take part to that discussion, I would like to understand whether this request is part of an overall strategy, as it seems. Do we want to have the 3 of them within CMSSW for comparison, with the idea of possibly moving in the longer term towards a single approach?

smuzaffar:
@fabiocos , yes there are not technical obstacle integrating it. As this is new tool, so we can integrate it now OR we can wait for https://github.com/cms-sw/cmsdist/pull/4185 (which also include this PR changes) to be fully tested.

makortel:
My feeling (could be wrong) is that at the moment we don't have enough experience on these tools on the inference side to make a clear choice (are there any other downsides than supporting yet another external?). Also, if we start restricting the inference tools, it should be clearly communicated to the developer community.

davidlange6:
Hi
I agree. If we want to move out of the “let 1000 flowers bloom” mode in this problem space, a pr is not the place to decide (nor is first come first served the approach to take). I haven’t seen any research to suggest that one solution is now covering the problem space, so am biased toward continuing to integrate data science supported tools.

Cheers,
David


On 11 Jul 2018, at 22:14, Matti Kortelainen <notifications@github.com<mailto:notifications@github.com>> wrote:


My feeling (could be wrong) is that at the moment we don't have enough experience on these tools on the inference side to make a clear choice (are there any other downsides than supporting yet another external?). Also, if we start restricting the inference tools, it should be clearly communicated to the developer community.

fabiocos:
@davidlange6 I agree that general long term strategies goes beyond the single PR discussion. But the problem is practically posed through PRs. This tool was already discussed and mentioned, so ok, but in case more arrives, I think we need to understand where we want to go

fabiocos:
+1

PR #4168, opened 2018-07-03
Improve CUDA build rules to support linking device code across different libraries 

fwyzard:
Update to the CUDA tool:
  - drop `-dlink`, which should be moved to the SCRAM build rules
  - drop `-shared`, which should be replaced by `--compiler-options '$(CXXSHAREDOBJECTFLAGS)'` in the SCRAM build rules, which includes `-fPIC`
  - drop `-L$(CUDA_BASE)/lib64`, which is included automatically by `nvcc`

Update to the SCRAM build rules for CUDA code (by Shahzad):
  - See https://github.com/cms-sw/cmssw-config/issues/65
  -  Update cmssw-config to V05-07-30: Update cuda compilation rules for device code


fwyzard:
@smuzaffar , these changes should go together with the upcoming changes to the build rules for CUDA code.

fwyzard:
These upcoming changes: https://github.com/cms-sw/cmssw-config/issues/65 .

fwyzard:
@cmsbuild, please test

PR #4169, opened 2018-07-05
bug fix: make sure that share/data has correct content after local deprecate local and re-install

smuzaffar:
This is a bug fix to avoid the following situation when building with `--reference /cvmfs/cms-ib.cern.ch/weekN` as reference repository
- A data package is build locally means arch/cms/data-SubSys-Pack/version/SubSys is symlink to share/cms/data-SubSys-Pack/version/SubSys
- newly built RPMs are uploaded
- run deprecate local to force get newly uploaded RPMs. This means rpm erase will remove the contents of arch/cms/data-SubSys-Pack/version/SubSys (which is symlink to share/cms/data-SubSys-Pack/version/SubSys).
- We when re-install this data package from the uploaded repo it does not re-populate the contents of share directory.

This PR should fix this situation.


smuzaffar:
please test

smuzaffar:
+externals

PR #4173, opened 2018-07-06
Update Eigen to leverage the latest updates to CUDA and AVX support

fwyzard:


fwyzard:
@cmsbuild, please test

smuzaffar:
+externals

smuzaffar:
@fwyzard , do you understand why the latest commit here https://github.com/cms-externals/eigen-git-mirror/tree/cms/64060da  is 64060da while if you click on commits of this branch then there are many other commits on top of it?

fwyzard:
Is it possible that in one case they are sorted by date, and in the other
by graph ?

PR #4176, opened 2018-07-09
Update data-RecoBTag-Combined.spec

mrodozov:
Description: https://github.com/cms-data/RecoBTag-Combined/pull/15

PR #4185, opened 2018-07-11
add mxnet and update data-RecoBTag-Combined.spec (merge of #4167 and #4176) needed for cmssw tests

slava77:
this is a combined PR to be able to test https://github.com/cms-sw/cmssw/pull/23768


smuzaffar:
both of these externals are in 10.2.X, so no need to have this PR open.

PR #4187, opened 2018-07-12
Add gcc toolfile with location of gcc plugin headers and library

gartung:


gartung:
@Dr15Jones @smuzaffar

smuzaffar:
@gartung , does this only work for gcc7 and above? If yes then we should update the cmssw PR to only build it for gcc7 archs

gartung:
It should work with any gcc. I just happened to make the change on the gcc7 branch.

gartung:
I will add the gcc-plugin in its own spec so I do not need this toolfile.

PR #4190, opened 2018-07-16
The Atlas GccCheckerPlugins package as an external.

gartung:
Although the PR is based on the gcc7 branch it can be back ported to any branch with a compiler that supports plugins and c++11.

### DIFF
--- a/gcc-checker-plugin-toolfile.spec
+++ b/gcc-checker-plugin-toolfile.spec
@@ -34,6 +34,7 @@ cat << \EOF_TOOLFILE >%i/etc/scram.d/gcc-checker-plugin.xml
       <environment name="LIBRARY"   default="$GCC_CHECKER_PLUGIN_ROOT/lib"/>
     </client>
   </tool>
+  <runtime name="GCC_CHECKER_PLUGIN" default="$GCC_CHECKER_PLUGIN_ROOT/lib/libchecker_gccplugins.so"/>
 EOF_TOOLFILE
 export GCC_CHECKER_PLUGIN_ROOT
 
--- a/gcc-checker-plugin-toolfile.spec
+++ b/gcc-checker-plugin-toolfile.spec
@@ -1,0 +1,46 @@
+### RPM cms gcc-checker-plugin-toolfile 1.0
+
+# gcc has a separate spec file for the generating a 
+# toolfile because gcc.spec could be not build because of the 
+# "--use-system-compiler" option.
+
+Source: none
+Requires: gcc-checker-plugin
+
+%prep
+%build
+%install
+mkdir -p %{i}/etc/scram.d
+
+# Determine the GCC_ROOT if "use system compiler" is used.
+if [ "X$GCC_ROOT" = X ]
+then
+    export GCC_PATH=$(which gcc) || exit 1
+    export GCC_ROOT=$(echo $GCC_PATH | sed -e 's|/bin/gcc||')
+    export GCC_VERSION=$(gcc -dumpversion) || exit 1
+    export G77_ROOT=$GCC_ROOT
+else
+    export GCC_PATH
+    export GCC_ROOT
+    export GCC_VERSION
+    export G77_ROOT=$GCC_ROOT
+fi
+# GCC tool file for explicity linking gcc plugin library
+cat << \EOF_TOOLFILE >%i/etc/scram.d/gcc-checker-plugin.xml
+  <tool name="gcc-checker-plugin" version="@GCC_VERSION@">
+    <lib name="checker_gccplugins"/>
+    <client>
+      <environment name="GCC_CHECKER_PLUGIN_ROOT" default="@GCC_CHECKER_PLUGIN_ROOT@"/>
+      <environment name="LIBRARY"   default="$GCC_CHECKER_PLUGIN_ROOT/lib"/>
+    </client>
+  </tool>
+EOF_TOOLFILE
+export GCC_CHECKER_PLUGIN_ROOT
+
+# General substitutions
+perl -p -i -e 's|\@([^@]*)\@|$ENV{$1}|g' %{i}/etc/scram.d/*.xml
+
+%post
+%{relocateConfig}etc/scram.d/*.xml
+echo "GCC_CHECKER_PLUGIN_TOOLFILE_ROOT='$CMS_INSTALL_PREFIX/%{pkgrel}'; export GCC_CHECKER_PLUGIN_TOOLFILE_ROOT" > $RPM_INSTALL_PREFIX/%{pkgrel}/etc/profile.d/init.sh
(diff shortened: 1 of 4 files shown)


gartung:
@smuzaffar @Dr15Jones

gartung:
@smuzaffar 

When I try to run the plugin as compiled is fails with

cc1plus: error: cannot load plugin /home/gartung/cms-build/slc7_amd64_gcc700/external/gcc-checker-plugin/1.0-cms2/lib/libchecker_gccplugins.so
   /home/gartung/cms-build/slc7_amd64_gcc700/external/gcc-checker-plugin/1.0-cms2/lib/libchecker_gccplugins.so: undefined symbol: _Z19get_inner_referenceP9tree_nodePlS1_PS0_P12machine_modePiS5_S5_

If I link the plugin against libcp1plugin.so, it fails with

cc1plus: error: cannot load plugin /home/gartung/cms-build/slc7_amd64_gcc700/external/gcc-checker-plugin/1.0-cms/lib/libchecker_gccplugins.so
   /home/gartung/CMSSW_10_2_0/external/slc7_amd64_gcc630/lib/libcp1plugin.so.0: undefined symbol: cp_binding_oracle

If I link the plugin against libcc1plugin.so, it fails with

cc1plus: error: cannot load plugin /home/gartung/cms-build/slc7_amd64_gcc700/external/gcc-checker-plugin/1.0-cms4/lib/libchecker_gccplugins.so
   /home/gartung/CMSSW_10_2_0/external/slc7_amd64_gcc630/lib/libcc1plugin.so.0: undefined symbol: c_binding_oracle

gartung:
Nevermind. I was able to determine that set up a project area for gcc630 and I compiled the plugin with gcc700.

gartung:
Once the gcc-checker-plugin toolfile is configure in a project area the gcc checker can be run with 

g++ -c -o /dev/null --std=c++14 -fplugin=$GCC_CHECKER_PLUGIN -fplugin-arg-libchecker_gccplugins-checkers=all $file

gartung:
@Dr15Jones I took the twiki text from your email and reformatted it.
https://twiki.cern.ch/twiki/bin/view/Sandbox/PatrickGartungSandbox

gartung:
@davidlange6 Are you opposed to this one too?

davidlange6:
not specifically - presumably if we want this for more than a quick test, we should get it put into a proper distribution (add bonus, with a makefile or similar) rather than downloading all the root/fastjet etc etc etc sources that atlas uses..

gcc-checker-plugin-toolfile.spec
```diff
@@ -0,0 +1,47 @@
...
+EOF_TOOLFILE
+export GCC_CHECKER_PLUGIN_ROOT
+
+# General substitutions
+perl -p -i -e 's|\@([^@]*)\@|$ENV{$1}|g' %{i}/etc/scram.d/*.xml
+
+%post
+%{relocateConfig}etc/scram.d/*.xml
+echo "GCC_CHECKER_PLUGIN_TOOLFILE_ROOT='$CMS_INSTALL_PREFIX/%{pkgrel}'; export GCC_CHECKER_PLUGIN_TOOLFILE_ROOT" > $RPM_INSTALL_PREFIX/%{pkgrel}/etc/profile.d/init.sh
+echo "setenv GCC_CHECKER_PLUGIN_TOOLFILE_ROOT '$CMS_INSTALL_PREFIX/%{pkgrel}'" > $RPM_INSTALL_PREFIX/%{pkgrel}/etc/profile.d/init.csh
```
smuzaffar: @gartung , I don't think you need all this logic in this spec file. Just copy the one e.g. curl-toolfile.spec and update variables.

gcc-checker-plugin.spec
```diff
@@ -0,0 +1,23 @@
+### RPM external gcc-checker-plugin 1.0
```
smuzaffar: @gartung , use version 2.0.9 here and replace 2.0.9 with %{realversion} in rest of the file.

davidlange6:
the trouble is that 2.0.9 has nothing to do with the version number of this specific package being built..

smuzaffar:
but at least it points to a tag in gitlab ... right?

davidlange6:
no. It seems to be a version of a tarball full of sources from many projects

cd atlasexternals-2.0.9-671e8f2b1413ff87772ea5dc024abeb5ef6ee441
ls
APE			CheckerGccPlugins	FastJetContrib		HEPUtils		MCUtils			SoQt
Acts			Coin3D			GPerfTools		HepMCAnalysis		MKL			XRootD
BAT			Davix			Gdb			KLFitter		PyAnalysis		dSFMT
Blas			Eigen			Geant4			Lapack			Python			dcap
Boost			FFTW			GoogleTest		Lhapdf			ROOT			lwtnn
CLHEP			FastJet			HDF5			LibXml2			Simage			yampl

gartung:
I created this PR for review. It should not be merged yet. I plan to create a repo under cms-externals with just CheckerGccPlugin. I also want to change the Altas references to CMS. Is there a way to mark this PR WIP or For Review. I would mark the codecompass PR the same way.

gcc-checker-plugin-toolfile.spec
gartung: Eventually you will need the path to the plugin library.

gartung:
I put the CheckerGccPlugins source in its own repository here
https://github.com/gartung/CheckerGccPlugins
I added a Makefile and changed all of the ATLAS references to CMS. This includes the checker_macros.h header that needs to be included to use the pragma macros needed to activate the thread checkers. The header could be placed in CMSSW if that is easier.

gartung:
Mostly unmodified source now in cms-external repo
https://github.com/cms-externals/CheckerGccPlugins

gartung:
Create gcc-analyzer-c(xx)compiler.xml so gcc checker can be run with
scram b COMPILER=gcc-analyzer

smuzaffar:
please test

gcc-checker-plugin-toolfile.spec
```diff
@@ -0,0 +1,43 @@
...
+mkdir -p %{i}/etc/scram.d
+
+cat << \EOF_TOOLFILE >%i/etc/scram.d/gcc-checker-plugin.xml
+  <tool name="gcc-checker-plugin" version="@TOOL_VERSION@">
+    <client>
+      <environment name="GCC_CHECKER_PLUGIN_ROOT" default="@TOOL_ROOT@"/>
+      <environment name="INCLUDE"   default="$GCC_CHECKER_PLUGIN_ROOT/include"/>
+    </client>
+  </tool>
+  <runtime name="GCC_CHECKER_PLUGIN" default="$GCC_CHECKER_PLUGIN_ROOT/lib/libchecker_gccplugins.so"/>
```
smuzaffar: do we need GCC_CHECKER_PLUGIN env variable?

smuzaffar:
please test

gcc-checker-plugin-toolfile.spec
gartung: It is no longer needed.

PR #4194, opened 2018-07-18
suppress usage of AVX instructions in mxnet

slava77:
I was testing https://github.com/cms-sw/cmssw/pull/23768 on non-AVX machine and saw crashes due to invalid instructions.
In the specific case, it was a call to ```vpxor``` in ```mxnet::op::AdaptiveAvgPoolParam::__MANAGER__() ()```.
After recompiling with ```USE_F16C=0```, the same place in the code is using a simple ```pxor```.


slava77:
@cmsbuild please test

slava77:
i was testing compilation on cmsdev14 (AVX node), which is supposedly representative of the build nodes used for cmsdist.

smuzaffar:
+externals
build log shows that `-mf16c` flags was not used.

fabiocos:
@slava77 I understand that this is a technical protection that is not supposed to change the output of the code, am I correct? Fine for me integrating it

fabiocos:
+1

slava77:
On 7/19/18 9:56 AM, Fabio Cossutti wrote:
> @slava77 I understand that this is a technical protection that is not supposed to change the output of the code, am I correct? Fine for me integrating it

yes.
Also, this external is not used yet by any existing modules in CMSSW yet
(only in code from outside analysis repositories).

PR #4198, opened 2018-07-19
suppress usage of AVX instructions in mxnet (102X version of #4194)

slava77:
this is a 10_2_X version of #4194 

a copy of that PR description follows

I was testing https://github.com/cms-sw/cmssw/pull/23768 on non-AVX machine and saw crashes due to invalid instructions.
In the specific case, it was a call to ```vpxor``` in ```mxnet::op::AdaptiveAvgPoolParam::__MANAGER__() ()```.
After recompiling with ```USE_F16C=0```, the same place in the code is using a simple ```pxor```.


slava77:
@cmsbuild please test

smuzaffar:
+extenals

PR #4201, opened 2018-07-20
Fix ROOT-9173: add per object flag for the statsoverflow

smuzaffar:
Cherry-pick https://github.com/root-project/root/pull/1503 in to root v6-12-00-patches branch to fix the issue found in online DQMGUI (due to change in version of the TH1 class from 8 to 7).


smuzaffar:
please test

PR #4203, opened 2018-07-20
root: sync in changes from gcc700 branch; drop castor and use ninja

smuzaffar:


smuzaffar:
@gudrutis , can you please update IB/CMSSW_10_3_X/rootmaster too?

PR #4264, opened 2018-08-15
Update xgboost patch file name

mrodozov:
Match the file name, fix https://cmssdt.cern.ch/jenkins/job/build-any-ib/63594/console


mrodozov:
please test

mrodozov:
I've just checked manually how this will go and first, the patch is rejected, second - this PR won't pick the difference as it is aimed for arm only while it's tested on amd. I'll have to fix it manually

mrodozov:
abort

mrodozov:
the changes from the patch are present in the makefile itself now (xgboost/rabit/Makefile) and the arm if-endif section is not required anymore. py2-xgboost just built on arm64, so I'll just update this PR and merge it before the test is finished.

mrodozov:
please test

PR #4282, opened 2018-08-22
Bump fastHadd version again

schneiml:
The 2.3 version was most likely build with a ROOT6.12 version that was missing the ROOT-9173 fix and therefore has the wrong `TH1` `ClassDef`. It must not be used in P5.

By now we should have that worked out, and we might need to switch to 2.4 in online once HLT switches to CMSSW10_2.

### DIFF
--- a/fasthadd.spec
+++ b/fasthadd.spec
@@ -1,7 +1,7 @@
-### RPM external fasthadd 2.3
+### RPM external fasthadd 2.4
 
 #Change the commit hash every time a new version is needed.
-%define commit 905af02d3369428df677c232537da6fca8982ff3
+%define commit 0fec4de6785e45f6535eff44a6dc23bd0039d154
 Source0: https://raw.githubusercontent.com/cms-sw/cmssw/%commit/DQMServices/Components/bin/fastHadd.cc
 Source1: https://raw.githubusercontent.com/cms-sw/cmssw/%commit/DQMServices/Core/src/ROOTFilePB.proto
 Requires: protobuf root


mrodozov:
please test

davidlange6:
there is no need to bump versions due to dependency changes. Everything gets rebuilt when one of its dependencies changes. what is the goal of this pr?

mrodozov:
This fail has nothing to do with the build. the dataset required for this wf 4.22 is listed by the dasgoclient but not available on eos. 
```
dasgoclient --query 'file dataset=/Cosmics/Run2011A-v1/RAW ' |  grep 160/960 | wc -l
29
eos ls /store/data/Run2011A/Cosmics/RAW/v1/000/160/960/4C2B0A43-8353-E011-A104-0030487CAF5E.root
Unable to stat /eos/cms/store/data/Run2011A/Cosmics/RAW/v1/000/160/960/4C2B0A43-8353-E011-A104-0030487CAF5E.root (errc=2) (No such file or directory)
```

davidlange6:
there are cmseos problems at cern today...

mrodozov:
please test

schneiml:
@davidlange6 If we have packages for a fastHadd built with the right ROOT (the one used in 10_2_5, say), we are good and can use that at P5 when HLT switches to 10_2. 

This is all a bit scary since we are talking about changes to ROOT that are not reflected in its version number, so it is hard to tell if a given package will work or not... Let's wait for the HI test next week and see if we even need that.

davidlange6:
I guess the issue is potentially a cms desired patch to root... (for dqm if I remember).

Anyway, that’s all fine and has no bearing on my comment. Just sync the root.spec and it’s dependencies to the version you want and rebuild fasthadd


On 14 Sep 2018, at 17:59, Marcel Schneider <notifications@github.com<mailto:notifications@github.com>> wrote:


@davidlange6<https://github.com/davidlange6> If we have packages for a fastHadd built with the right ROOT (the one used in 10_2_5, say), we are good and can use that at P5 when HLT switches to 10_2.

This is all a bit scary since we are talking about changes to ROOT that are not reflected in its version number, so it is hard to tell if a given package will work or not... Let's wait for the HI test next week and see if we even need that.

smuzaffar:
@schneiml is this still needed or shall we close it?

fabiocos:
@schneiml we should clarify whether this is still needed, as there have been several updates of ROOT, and this code effectively sits inside CMSSW...

schneiml:
@fabiocos this is currently a time bomb: We are running `fastHadd` with a version mismatch at P5 since last year, and it works fine. However, at some point (HLT version change), it might blow up. We decided to not do the update (that is not strictly needed for now) since it involves a bit of manual work (requires building a wrapper rpm pointing into the right package provided by cmsdist for the DAQ machines).

So, I think we can close this PR and revive it once we get a new, incompatible version of ROOT.

PR #4301, opened 2018-08-29
SCRAM:Avoid invalid timestamp on EOS

smuzaffar:
This should fix the issue reported here: https://hypernews.cern.ch/HyperNews/CMS/get/softwareDistrib/758/1/1/1/1/1.html


smuzaffar:
please test

PR #4332, opened 2018-09-12
Update data-RecoParticleFlow-PFProducer.spec

mrodozov:
Test new data files

### DIFF
--- a/data-RecoParticleFlow-PFProducer.spec
+++ b/data-RecoParticleFlow-PFProducer.spec
@@ -1,4 +1,4 @@
-### RPM cms data-RecoParticleFlow-PFProducer V15-00-00
+### RPM cms data-RecoParticleFlow-PFProducer V15-00-01
 
 %prep
 
--- a/data-RecoParticleFlow-PFProducer.spec
+++ b/data-RecoParticleFlow-PFProducer.spec
@@ -1,4 +1,4 @@
-### RPM cms data-RecoParticleFlow-PFProducer V15-00-01
+### RPM cms data-RecoParticleFlow-PFProducer V16-00-00
 
 %prep
 


mrodozov:
please test

slava77:
if we are going to try to make sense out of the package tags, then V15-00-01 is not the most appropriate in this case.
The file format was changed, even though they intend to yield the same results after the update in the parser.
Perhaps make it V16-00-00 ? 

IIUC the tests with cmssw baseline will fail.
We need to test with cms-sw/cmssw#24432  (but even that test will start to make sense only after the copy of the new files is removed from cms-sw/cmssw#24432)

slava77:
ah, the tests actually passed.
@guitargeek are these PF files even used now?

guitargeek:
I am surprised! I certainly know from my own tests that __PfElectrons23Jan_BDT.weights.xml.gz__ and __pfConversionAug0411_BDT.weights.xml.gz__ are used in workflow __4.53__, which cmsbot tested here.

However, I don't know of a workflow that uses __finalID_hzz-pions_BDT.weights.xml.gz__.

Maybe the tests are organized in such a way that the old files are still there?

guitargeek:
Here is the list of GBRForests constructed from weights files in __4.53__:

https://rembserj.web.cern.ch/rembserj/scratch/4.53_gbr.out

Just a curiosity: notice I run my local tests all in single thread, so we would expect every file to only appear once in the log. It's interesting to see that some electron BDTs used in the reconstruction are loaded 2 or 3 times!

guitargeek:
In fact I found out some time ago that they are also used multiple times in each event, i.e. the MVA value is calculated redundantly. It's not much overhead since it's fast, but just to keep in mind.

slava77:
Just to repeat: the tests with results in https://github.com/cms-sw/cmsdist/pull/4332#issuecomment-420876216 were done on top of the baseline with this PR which corresponds to removal of the .txt files in one format and replacement of them with .xml.gz files with different names. The tests passed. Given a separate indications that the files are used, the explanation can be that there is a separate location of the same files.

@smuzaffar @mrodozov 
please clarify if in the tests we were still using a full rebuild strategy and the old source of data-RecoParticleFlow-PFProducer is not visible anymore in the test jobs search paths.

mrodozov:
> Maybe the tests are organized in such a way that the old files are still there?

They shouldn't be, consult with the jenkins console output to clarify this:

> 23:12:21 Starting to process package cms+data-RecoParticleFlow-PFProducer+V15-00-01
> 23:12:21 Checking if data-RecoParticleFlow-PFProducer is cached.
> 23:12:21 Package cms+data-RecoParticleFlow-PFProducer+V15-00-01 not found in repository. Queuing for build.
> 23:12:21 Dependencies for cms+data-RecoParticleFlow-PFProducer+V15-00-01: []

meaning - it tries to get data-RecoParticleFlow-PFProducer+V15-00-01, finds out it's not available
and gets('rebuild') it. this excludes the possibility that the old package files existed, you see each PR is tested on a 'clean' area

smuzaffar:
@slava77, cmsdist PR tests do not use full release build strategy. The CMSSW_SEARCH_PATH for cmsdist PR tests contains the following directories. So if a file is not found in CMSSW_BASE/data then it will be looked in to CMSSW_RELEASE_BASE/data. We can update cmsdist PRs test to drop CMSSW_RELEASE_BASE/external/slc6_amd64_gcc700/data  but this will not work for data files for src.

```
CMSSW_BASE/src
CMSSW_BASE/external/slc6_amd64_gcc700/data
CMSSW_RELEASE_BASE/src
CMSSW_RELEASE_BASE/external/slc6_amd64_gcc700/data
```

slava77:
On 9/13/18 6:17 AM, Malik Shahzad Muzaffar wrote:
> @slava77
> cmsdist PR tests do not use full release build strategy. The
> CMSSW_SEARCH_PATH for cmsdist PR tests contains the following
> directories. So if a file is not found in CMSSW_BASE/data then it will
> be looked in to CMSSW_RELEASE_BASE/data. We can update cmsdist PRs test
> to drop CMSSW_RELEASE_BASE/external/slc6_amd64_gcc700/data but this will
> not work for data files for src.
> 
> |CMSSW_BASE/src CMSSW_BASE/external/slc6_amd64_gcc700/data
> CMSSW_RELEASE_BASE/src CMSSW_RELEASE_BASE/external/slc6_amd64_gcc700/data |
> 
> —


Thank you for the clarification. This explains the test results.
This also then reminds that we should be really careful with removal of
the files fro the externals because we do not test it until the
deployment in the release.

About /src dependence of data files:
when we test a CMSSW PR, is there a "poison" or other method that
prevents the build from using the RELEASE_BASE files?
IIUC, we have it for py and headers.

It would be best to have the tests look as much as possible as what the
final release will be.

smuzaffar:
@slava77 , no, we do not have any poisoned files for data packages .... I guess we can create either empty files or broken symlinked files so that program fail when try to open such files.

slava77:
On 9/14/18 12:17 AM, Malik Shahzad Muzaffar wrote:
> @slava77
> , no, we do not have any poisoned files for data packages .... I guess
> we can create either empty files or broken symlinked files so that
> program fail when try to open such files.
> 
> —

it sounds like this will be a good improvement to the testing procedure.
Thank you.

smuzaffar:
please test
testing it without cmssw PR to make sure it fails as now PR tests will drop release/external/data from CMSSW_SEARCH_PATH

smuzaffar:
+externals

fabiocos:
+1

PR #4336, opened 2018-09-14
changes needed for poisoning the cmssw data files

smuzaffar:
These changes are need to make sure that deleted data files in cmssw area properly poisoned and any try to open those deleted files cause file open error.


smuzaffar:
please test

PR #4353, opened 2018-09-19
 Update llvm/clang to version 7.0.0 (release) 

fwyzard:
Also
  -  enable support for compiling CUDA code to NVPTX on amd64;
  -  disable automatic package dependencies.

For the moment, skip building dxr, numba and llvmlite, until they support LLVM 7.

### DIFF
--- a/llvm.spec
+++ b/llvm.spec
@@ -8,6 +8,7 @@ Requires: gcc zlib
 %if %{isamd64}
 Requires: cuda
 %endif
+AutoReq: no
 
 %define llvmCommit ff0a5e8a591ed8bfc14320740863b357b1774f49
 %define llvmBranch cms/release_70/342187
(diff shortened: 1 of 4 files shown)


fwyzard:
@cmsbuild please test

fwyzard:
numba / llvmlite for clang 7 is tracked here: https://github.com/numba/llvmlite/issues/396

fwyzard:
As for DXR, we are using a revision that is 4 years old; the current version of DXR does not use trilite and sqlite,  and instead relies on elasticsearch etc.

Would it make sense to try and update to the current version, rather than patching up the old one we use ?

smuzaffar:
@fwyzard , for now, we can merge this for DEVEL IB to get it fully tested.
About DXR, problem is the DXR server side setup. Newer DXR requires Elesticsearch. It was not a trivial update when last time we checked also was requiring too many resources (3GB/release in ES).

Anyway, I have made a PR ( https://github.com/fwyzard/cmsdist/pull/1)  to fix the dxr compilation with new llvm.

fwyzard:
Thanks, I will merge it and re-submit the tests.

About ElasticSearch, could we make use of the centralised CERN instance ?

fwyzard:
@cmsbuild, please test

smuzaffar:
+externals
merging it for DEVEL IBs.

PR #4356, opened 2018-09-21
set the COMPILER_PATH env variable instead of COMPILER_RUNTIME_OBJECTS

smuzaffar:
clang use COMPILER_PATH env variable to find the GCC toolchain. We were patching clang ( https://github.com/cms-externals/llvm-project-20170507/commit/85a9c36d893831f7c7ea73eaa6a74997739a0a0b#diff-c338d6d0b1a666a15fa597e194768aa3R1571 ) to get the gcc from the cms external area. With this change, the clang patch is not needed any more.


smuzaffar:
please test

PR #4359, opened 2018-09-24
Update llvm/clang to version 7.0.0 (release) 

smuzaffar:
backport of #4353 
 - enable support for compiling CUDA code to NVPTX on amd64;
 - disable automatic package dependencies.
For the moment, skip building numba and llvmlite, until they support LLVM 7.

### DIFF
--- a/python_tools.spec
+++ b/python_tools.spec
@@ -25,8 +25,8 @@ Requires: py2-bottleneck
 Requires: py2-downhill 
 Requires: py2-theanets
 Requires: py2-xgboost
-Requires: py2-llvmlite
-Requires: py2-numba
+#Requires: py2-llvmlite
+#Requires: py2-numba
 Requires: py2-hep_ml
 Requires: py2-rep
 Requires: py2-uncertainties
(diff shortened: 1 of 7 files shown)


smuzaffar:
please test

smuzaffar:
+externals

smuzaffar:
hold
We will integrate it in 10.4.X

smuzaffar:
please test

smuzaffar:
unhold

smuzaffar:
+externals

python_tools.spec
```diff
@@ -25,8 +25,8 @@ Requires: py2-bottleneck
 Requires: py2-downhill 
 Requires: py2-theanets
 Requires: py2-xgboost
-Requires: py2-llvmlite
-Requires: py2-numba
+#Requires: py2-llvmlite
```
fabiocos, line 28: @smuzaffar this seems to trigger a couple of failures in the unit tests of PythonAnalysis since CMSSW_10_4_X_2018-10-11-2300 because these tools are not found in the import. Is this needed?

davidlange6:
the unit test should be updated to remove the two imports corresponding to these packages until llvmlite supports llvm7 (presumably soon)

fabiocos:
@davidlange6 provided we do not forget about them. I am not sure whether it is better to temporarily remove the two tests and keep a github issue open (we have plenty of them), or declare them as known to fail in the IB, but at least we keep track of them

davidlange6:
right - either way - the downside to unit tests that fail is that any PR that checks out this package will fail tests..

fabiocos:
yes, this is also an issue to be taken into account, people might get confused by failures unrelated to their changes. Ok, I will go for the former solution

PR #4362, opened 2018-09-24
Update CUDA to version 10.0.130 (10.2.x backport)

fwyzard:
See https://docs.nvidia.com/cuda/pdf/CUDA_Toolkit_Release_Notes.pdf .

Update the SCRAM tool files for CUDA 10:
  - drop the device-side cublas libraries;
  - add a tool for the new nvjpeg library.

### DIFF
--- a/cuda.spec
+++ b/cuda.spec
@@ -1,9 +1,8 @@
-### RPM external cuda 9.2.148
-%define driversversion 396.37
+### RPM external cuda 10.0.130
+%define driversversion 410.48
 %define cudaversion %(echo %realversion | cut -d. -f 1,2)
 
-Source0: https://developer.nvidia.com/compute/cuda/%{cudaversion}/Prod2/local_installers/%{n}_%{realversion}_%{driversversion}_linux
-Source1: https://developer.nvidia.com/compute/cuda/%{cudaversion}/Prod2/patches/1/%{n}_%{realversion}.1_linux
+Source0: https://developer.nvidia.com/compute/cuda/%{cudaversion}/Prod/local_installers/%{n}_%{realversion}_%{driversversion}_linux
 AutoReq: no
 
 %prep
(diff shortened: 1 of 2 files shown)


fwyzard:
@cmsbuild please test

cuda.spec
```diff
@@ -14,78 +13,76 @@ AutoReq: no
...
-rm -rf %_builddir/lib64/libcusolver.so*
-rm -rf %_builddir/lib64/libcusparse.so*
-rm -rf %_builddir/lib64/libnpp*.so*
-rm -rf %_builddir/lib64/libnvgraph.so*
-rm -rf %_builddir/lib64/libnvrtc.so*
-rm -rf %_builddir/lib64/libnvrtc-builtins.so*
+mkdir -p %{i}/bin
+
+# package only the runtime static libraries
+mv %_builddir/lib64/libcudart_static.a %{i}/lib64/
```
fwyzard: @smuzaffar moving the files instead of copying them should make the build faster and use less disk space.
Can you think of any reasons why we would not want to do it ?

smuzaffar: +1
No, there should not be any reason to not do that.
While you are at it, could you remove, create and use %_builddir/build to extract the files. Removing and creating it will make sure that in case of re-build (due to previous failures) we hae a clean build directory.

smuzaffar (approved):

cuda.spec
fwyzard: Do you mean `%_builddir/build` instead of `%_builddir` ?

fwyzard: Also, should I move the repackaging from `%install` to `%build` ?

smuzaffar: yes `%_builddir/build` instead of `%_builddir`. Extracting directly in %_builddir might override some files which cmsBuild generates e.g log file.

No, repacking in `%install` is fine.

fwyzard:
@cmsbuild, please test

fwyzard:
@cmsbuild, please test

fwyzard:
@cmsbuild, please test

fwyzard:
We are moving the CUDA-based development to the 10.4.x release series.

PR #4364, opened 2018-09-25
Updated scram build rules to fix the missing cuda dependency

smuzaffar:
cuda automatic dependency, due to .cu files, was dropped when -fsyntax-only flag is passed via USER_CXXFLAGS. New build rules tag should fix this issue.


smuzaffar:
please test

PR #4366, opened 2018-09-25
Add missing python tool dependencies add onnx ML library and nbdime library for notebook diffing

davidlange6:
Changes fix a number of tools that I found were not working when enabling the python unit tests..

All of the back port packages are now installed in a common place (like the google namespace packages) to avoid directory name conflicts in the python path.

Some old stuff is removed.


davidlange6:
please test

smuzaffar:
+externals

PR #4367, opened 2018-09-25
Added basic support for compiling with CUDA 10.0

fwyzard:
Backported from llvm 8 master branch.


fwyzard:
@cmsbuild, please test

PR #4368, opened 2018-09-25
Update CUDA to version 10.0.130

fwyzard:
See https://docs.nvidia.com/cuda/pdf/CUDA_Toolkit_Release_Notes.pdf .

Update the SCRAM tool files for CUDA 10:
  - drop the device-side cublas libraries;
  - add a tool for the new nvjpeg library.

### DIFF
--- a/cuda-toolfile.spec
+++ b/cuda-toolfile.spec
@@ -51,8 +51,6 @@ cat << \EOF_TOOLFILE >%{i}/etc/scram.d/cuda-cublas.xml
   <info url="https://docs.nvidia.com/cuda/cublas/index.html"/>
   <use name="cuda"/>
   <lib name="cublas"/>
-  <lib name="cublas_device"/>
-  <lib name="cublas_device" type="cuda"/>
 </tool>
 EOF_TOOLFILE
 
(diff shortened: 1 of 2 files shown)


fwyzard:
@cmsbuild, please test

smuzaffar:
merging for DEVEL IBs.

fwyzard:
@smuzaffar thanks.
By the way, do we actually build IBs from the gcc700next branch ?

smuzaffar:
@fwyzard yes. CMSSW_10_3_DEVEL_X IBs are based on gcc700next branch.

smuzaffar:
we only build those for 23h00 though.

fwyzard:
ah, I see, we build them only for SLC6 .

smuzaffar:
Yes only for slc6. If needed we can build for slc7 too but may be not every night.

davidlange6:
+1 for dropping slc6 in favor of cc7... (didn't we agree to do that already some time ago?)

smuzaffar:
generators workflows (500's) are still not ready for slc7

davidlange6:
whats the eta for that? (maybe it shouldn't hope up the cmssw development branch, as after all the generators developers don't really care about it)

smuzaffar:
generators do not care for devel branch but mostly we use it to test new externals which we want to move back to production branch. So I would like to run these for same arch/compiler. Of course we can add CC7 IBs for DEVEL too.

davidlange6:
> On Sep 27, 2018, at 2:46 PM, Malik Shahzad Muzaffar <notifications@github.com> wrote:
> 
> generators do not care for devel branch but mostly we use it to test new externals which we want to move back to production branch. So I would like to run these for same arch/compiler. Of course we can add CC7 IBs for DEVEL too.

Thats reasonable indeed - but no ETA for gen supporting cc7? Seems we need to again discuss how to move forward (1.5 years after agreeing we could in 6 months)

PR #4393, opened 2018-09-28
Update llvm/clang to 7.0.0-svn342187 (release) (10.3.x backport)

fwyzard:
Update IWYU to the current master.

Disable automatic package dependencies: llvm 7 includes a new tool, llvm-exegesis, that depepnds on libpfm.so. Rather than including this library in the CMSSW stack, we assume that it will be available form the host system; as it's not something strictly necessary to have, we avoid a hard dependency on it in the RPM package.

Enable support for compiling CUDA code on amd64.

Backport basic support for compiling with CUDA 10.0 from the llvm 8 master branch.

### DIFF
--- a/llvm.spec
+++ b/llvm.spec
@@ -8,6 +8,7 @@ Requires: gcc zlib
 %if %{isamd64}
 Requires: cuda
 %endif
+AutoReq: no
 
 %define llvmCommit ff0a5e8a591ed8bfc14320740863b357b1774f49
 %define llvmBranch cms/release_70/342187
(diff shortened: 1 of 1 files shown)


fwyzard:
backport #4353 and #4367 from 10.3.x devel IBs

fwyzard:
@cmsbuild please test

fwyzard:
@cmsbuild, please test

fwyzard:
OK, so it's not a transient error... do we know where it comes from ?
Probably I will be able to look into it only after next week.

smuzaffar:
closing due to integration of https://github.com/cms-sw/cmsdist/pull/4359

PR #4406, opened 2018-10-03
New dasgoclient version

vkuznet:
It includes the following changes:
- filter original placement record for site dataset queries (these records occurs from DBS look-up but are not required for production datasets which are moved by PhEDEx)
- add **ecode** into DAS json output, if it is not empty it means that DAS encounter errors while processing data-service records, e.g. outage of DBS. It complements **status** value in DAS json output which mostly state of DAS itself had error.
- CRIC records are now supported for siteDB related queries.


smuzaffar:
please test

PR #4412, opened 2018-10-09
crabTaskWorker 3.3.1810.patch4 and WMCore 1.1.17.pre3

lecriste:
Use only blocks with locations in checkBlocksSize function to avoid Exception from DBS API


lecriste:
merge

PR #4413, opened 2018-10-09
CRABTaskWorker 3.3.1810.patch4

lecriste:
In the TaskWorker checkBlocksSize function, use only blocks with locations to avoid Exception from DBS API


lecriste:
merge

PR #4419, opened 2018-10-11
backport cuda from 104X devel

smuzaffar:
backport of #4368

### DIFF
--- a/cuda-toolfile.spec
+++ b/cuda-toolfile.spec
@@ -52,8 +52,6 @@ cat << \EOF_TOOLFILE >%{i}/etc/scram.d/cuda-cublas.xml
   <info url="https://docs.nvidia.com/cuda/cublas/index.html"/>
   <use name="cuda"/>
   <lib name="cublas"/>
-  <lib name="cublas_device"/>
-  <lib name="cublas_device" type="cuda"/>
 </tool>
 EOF_TOOLFILE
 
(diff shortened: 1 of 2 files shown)


smuzaffar:
please test

smuzaffar:
@fwyzard , any idea what to do with these failing unit tests? They fail now as there was no cuda device.

fwyzard:
I didn't know we had pushed them upstream... I'll have a look and see if I can add a simple check

fwyzard:
Hi @smuzaffar , 
I've had a look at the tests, and while they could print a more meaningful error message (i.e. "No CUDA devices found, skipping the test"), any error is already being caught, and the tests return success (return 0).

How should they be changed to keep the bot happy ?

smuzaffar:
@fwyzard , thanks for looking in to this. 
Can you please propose a change where these tests do not fail (i.e. return 0) when no cuda device is found?

fwyzard:
As far as I can see, `main()` already always does `return 0`:

### cudaAtan2Test.cu
```
  ...
} catch(cuda::runtime_error ex) {
   std::cout << "cuda error " << ex.what() << std::endl;
} catch(...) {
   std::cout << "a non cuda error" << std::endl;
}

  return 0;
}
```
### cudaMathTest.cu
```
   ...
} catch(cuda::runtime_error ex) {
   std::cout << "cuda error " << ex.what() << std::endl;
} catch(...) {
   std::cout << "a non cuda error" << std::endl;
}

  return 0;
}
```

What else would I need to change ?

smuzaffar:
@fwyzard , I think problem is here https://github.com/cms-sw/cmssw/blob/master/DataFormats/Math/test/cudaMathTest.cu#L100 where we exit if device count is 0.

fwyzard:
Should be fixed by
  - https://github.com/cms-sw/cmssw/pull/24938 (master)
  - https://github.com/cms-sw/cmssw/pull/24939 (10.3.x)
  - https://github.com/cms-sw/cmssw/pull/24940 (10.2.x)

smuzaffar:
please test

smuzaffar:
+externals

fwyzard:
by the way, when changes to CMSDIST are forward- or back-ported across different branches, would it be possibly to keep the original split in the commits, or at least the original author ?

PR #4438, opened 2018-10-20
add ofast-flag tool and updated SCRAM

smuzaffar:
- new ofast-flag tools added to allow to include -Ofast for compilations. It also takes care of properly adding the no-builtin flags for arch where needed.
- updated scram version to V2_2_8 which now supports 
```
<if[condition] name|value|match="value">
  <use name="foo"/>
<else/>
  <use name="bar"/>
<if[condition]>
```


smuzaffar:
please test

PR #4448, opened 2018-10-24
use default to openloops docs optimization, pick up factorial calc improvement

davidlange6:
IB tests notices very long startup times for herwig NLO workflows (eg 3 hours for "simple" physics process. While probably this is a sign that a different approach is needed to cache these long to compute results, this PR speeds this calculation up.

Two changes are proposed
1) to use the optimization flags suggested by openloops documentation for openloops. Note that adding the last possible -02 flag makes compilation go forever.
2) to improve one of the hotspots in openloops for the benchmark workflow.

There are other clear optimizations to make. I need to work a bit on my integer mathematics skills before pushing them forward.

### DIFF
--- a/openloops.spec
+++ b/openloops.spec
@@ -1,5 +1,5 @@
 ### RPM external openloops 2.0.b
-%define tag 0f0826bd718dc28dcc8a457acb59de678b011b96
+%define tag 5141c0eab4286a4a685f03c4731ff0358e635e54
 %define branch cms/v%{realversion}
 %define github_user cms-externals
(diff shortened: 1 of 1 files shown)


davidlange6:
please test workflow 511.1

davidlange6:
looks like 511.1 finished in 1.5 hours. Great.

davidlange6:
seems the add-on test failure is in a pythia only workflow - not related to this PR. It would be great if this could get merged today

fabiocos:
+1

smuzaffar:
+externals

PR #4474, opened 2018-11-03
Improve Eigen compatibility with CUDA (10.4.x)

fwyzard:
Update Eigen to the master branch as of Tue Sep 25 20:26:16 2018 +0200
  - hg hash  66ba78bf7efa93f69a075830a87a010ed1b1fe30
  - git hash 01ae86b9aad30b1e65cf1b749fd6cd9a645ac00d

Patch Tensorflow to follow Eigen internal changes
  - cherry-pick changes from upstream repository
  - add local changes for the latest updates

Improve Eigen compatibility with CUDA
  - add support for cache-size queries on CUDA devices.
  - extend support for matrix inversion on CUDA devices above 4x4 matrices;
the size of the matrices that can be inverted is limited at runtime by the per-thread stack size.
  - extend support for diagonal matrices on CUDA devices.
  - fix deprecation warning in CUDA 10.0.

### DIFF
--- a/eigen.spec
+++ b/eigen.spec
@@ -1,12 +1,12 @@
-### RPM external eigen 64060da8461a627eb25b5a7bc0616776068db58b
+### RPM external eigen 01ae86b9aad30b1e65cf1b749fd6cd9a645ac00d
 ## INITENV +PATH PKG_CONFIG_PATH %{i}/share/pkgconfig
 ## INITENV SETV EIGEN_SOURCE %{source0}
 ## INITENV SETV EIGEN_STRIP_PREFIX %{source_prefix}
 ## NOCOMPILER
 %define tag %{realversion}
 
-#These are needed by Tensorflow sources
-#NOTE: Never apply any patch in the spec file, this way tensorflow gets the exact same sources
+# These are needed by Tensorflow sources
+# NOTE: Never apply any patch in the spec file, this way tensorflow gets the exact same sources
 %define source0 https://github.com/cms-externals/eigen-git-mirror/archive/%{tag}.tar.gz
 %define source_prefix eigen-git-mirror-%{realversion}
 Source: %{source0}
--- a/eigen.spec
+++ b/eigen.spec
@@ -1,4 +1,4 @@
-### RPM external eigen 01ae86b9aad30b1e65cf1b749fd6cd9a645ac00d
+### RPM external eigen 1f44b667dd9aeeb153284b15fc7fe159d2c09329
 ## INITENV +PATH PKG_CONFIG_PATH %{i}/share/pkgconfig
 ## INITENV SETV EIGEN_SOURCE %{source0}
 ## INITENV SETV EIGEN_STRIP_PREFIX %{source_prefix}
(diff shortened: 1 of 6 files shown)


fwyzard:
@cmsbuild, please test

fwyzard:
When I build all the externals including these changes, I run into a problem with the `tensorflow-python3-sources` package, where the build fails with a cryptic message:

> Server terminated abruptly (error code: 14, error message: '', log file: '/data/user/fwyzard/patatrack/build/slc7_amd64_gcc700.patatrack/BUILD/slc7_amd64_gcc700/external/tensorflow-python3-sources/1.6.0-patatrack/build/72fcdeb2f560249cbc23c63d6d0200b0/server/jvm.out')

but jvm.out is empty.
If I re-run the same build command in the same build area, the second time it succeeds.

@smuzaffar , @davidlange6 have you seen this before ? do you have any suggestions ?

davidlange6:
I haven't see this, no. I see this possibly related issue

https://github.com/bazelbuild/bazel/issues/3020

you may have less memory than on other machines that we've used to build tensorflow? if it happens more than once in a while we may want to add --jobs=<something reasonable> to try to constrain the resources the build uses.

fwyzard:
Mhm, the strange thing is that it always goes like this:
  - it builds successfully the python 2 version
  - it fails the python 3 version
  - at the second attempt,  it builds successfully the python 3 version as well

I saw that bazel leaves some build files under $HOME/.cache/bazel . I did not check how much space it uses during a build, is it possible that it runs out of space when doing multiple builds?

smuzaffar:
Could it be that first time we started two bazel servers (one for python2 and other for python3) and these two servers stepped out each other? Try adding
```
BuildRequires: tensorflow-python2-sources
```
in tensorflow-python3-sources to make sure that only one of these run at one time

davidlange6:
> On Nov 15, 2018, at 9:51 AM, Andrea Bocci <notifications@github.com> wrote:
> 
> Mhm, the strange thing is that it always goes like this:
> 
> 	• it builds successfully the python 2 version
> 	• it fails the python 3 version
> 	• at the second attempt, it builds successfully the python 3 version as well
> I saw that bazel leaves some build files under $HOME/.cache/bazel . I did not check how much space it uses during a build, is it possible that it runs out of space when doing multiple builds?

Right - thats annoying. I have run into bazel filling my $HOME - but then you get an error message telling you that you are out of space...

fwyzard:
> Try adding
> 
> ```
> BuildRequires: tensorflow-python2-sources
> ```
> 
> in tensorflow-python3-sources to make sure that only one of these run at one time

Thanks, I will give it a try.


> I have run into bazel filling my $HOME - but then you get an error message telling you that you are out of space...

Is there any way to tell bazel to use a different temp directory ?

davidlange6:
> On Nov 15, 2018, at 10:17 AM, Andrea Bocci <notifications@github.com> wrote:
> 
> Try adding
> 
> BuildRequires: tensorflow-python2-sources
> 
> in tensorflow-python3-sources to make sure that only one of these run at one time
> 
> Thanks, I will give it a try.
> 
> I have run into bazel filling my $HOME - but then you get an error message telling you that you are out of space...
> 
> Is there any way to tell bazel to use a different temp directory ?
> 

Indeed - unfortunately we already tell bazel to use a different temp area (which gets very big).. i think someone spent a bit of time on this bug, but we didn't solve it yet.

PR #4482, opened 2018-11-06
cleanup of py2 packages; make sure to install their dependencies

smuzaffar:
Cleanup for `py2-*` packages. These changes work with `pkgtools V00-32-XX` branch.
- match cms package name to pip package name. `.` in package name should be replaced with `-` e.g. pip package `backports.lzma` becomes `py2-backports-lzma.spec`
- dropped most of `py2-*-toolfile.spec` and automatically generate these via `cmssw-tool-conf.spec`
- Make sure that dependency of a `py2-*` package satisfied and fail at build time if missing
- Simplified `py2-*` packages to be of the form
```
### RPM external py2-PipPackageName Version
## IMPORT build-with-pip

<extra build/requires flags>
```
one day we can auto generate these from a single file. For packages with `<extra build/requires flags>` can have py2-PipPackageName.file which can be included in the auto-generated package specs (using new `## INCLUDE file-name` feature of `cmsBuild`).
- make `build-with-pip.file` generic enough so that we can customize the build from the package spec.


smuzaffar:
FYI @davidlange6  and @amaltaro

smuzaffar:
please test

smuzaffar:
please test

davidlange6:
hi @smuzaffar - cool this will really help. 

We talked some time ago about dependencies. You've removed the few declared dependencies that the py2*spec files had. So this means if package X really does depend at build time on package py2-Y, then it needs to declare that dependency and any others in the full dependency tree of py2-Y. Is that right? 

Hopefully we will continue to have o(1) such package (tensorflow) where this is important.

smuzaffar:
@davidlange6 , yes package X needs to declare its dependencies and it will also recursively get all the dependencies from packages its depend on.

I only added the dependencies which are needed. e.g. I removed 47 deps of `py2-jupyter.spec` but all of them should come from `py2-ipywidgets py2-jupyter_console py2-qtconsole` (56 in total).

amaltaro:
Thanks Shahzad. Once this gets merged, I'll pick your build-with-pip changes and adapt according to the comp_gcc630 needs (aka PYTHONPATH only, since we don't have the same py2/py3 distinction as in cmssw).

The question I had is about the pkgtools tag. Do you see any reason why we shouldn't start using the 00-32 tag ASAP in comp_gcc630?

smuzaffar:
@amaltaro , there is no reason why you should not be able to move to V00-32. It is back word compatible.
In order to test it I would suggest that just take an existing tag of cmsdist/comp_gcc630  and try to build `comp` or any other package (which is already available in comp repo) using V00-32/cmsBuild. If it re-builds any package then we have a issue and we should fix it.

amaltaro:
Ok, I built comp with the pkgtools `V00-32-XX` and it actually didn't build anything, since all the sources were already available in the repo.
I'll circulate an email tomorrow to Lina and the service responsibles asking to upgrade pkgtools to this version. Is this pkgtools branch required to work with the new build-with-pip changes? Or there are other important features in 00-32 that aren't available in 00-30?

smuzaffar:
@amaltaro, V00-32 works with both (old and new) build-with-pip.

smuzaffar:
hold
I have a better option for checking the dependencies. looks like with this current change ever package download its dependencies sources which is not good as base level packages sources are downloaded multiple times. As we have a relatively newer pip in cmsdist, so I can make use of `pip show` and `pip check` to make sure if package dependencies are resolved. 

I will update this PR later with the changes for build-with-pip

smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
+externals

py2-pyOpenSSL.spec
```diff
@@ -0,0 +1,4 @@
+### RPM external py2-pyOpenSSL 18.0.0
+## IMPORT build-with-pip
+
+Requires: py2-cryptography
```
amaltaro, line 4: you missed a dependency on py2-six here

smuzaffar: py2-cryptography brings in that dependency
https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_10_4_X/gcc700/py2-cryptography.spec#L4

PR #4489, opened 2018-11-09
New dasgoclient version

vkuznet:
Fix bug in fetch number of files used in site dataset query for file-replicas presence metric.


smuzaffar:
please test

PR #4492, opened 2018-11-12
New dasgoclient version

vkuznet:
Fix bug associated with improper look-up of data-services for site dataset=X queries. Once I introduced Rucio data-serivce now it was look-up by default. For CERN users it is not a problem since rucio service is visible on CERN network but for off-site users this query hang for long time trying to resolve rucio auth server. This bug fix resolve the issue and remove rucio from a list of default services to look at. I strongly advise to deploy this dasgoclient version asap to cvmfs to avoid users complaining about these types of queries which will hang for long time.


mrodozov:
please test

PR #4502, opened 2018-11-14
[10.4.X] updated boost version 1.67

smuzaffar:
backported from IB/CMSSW_10_4_X/gcc700next bracnh.
This version of boost has been in DEVEL IBs. Note that with this update unit test `condTestRegression` will fail while reading condition written using new boost (1.67) and reading with old boost (1.63 in 9.0.X).
```

SELF: CMSSW_10_4_DEVEL_X_2018-11-13-2300 [slc6_amd64_gcc700]

                            Write | Read: CMSSW_9_0_1 [slc6_amd64_gcc530] | Read: SELF |
  CMSSW_7_6_6 [slc6_amd64_gcc493] |                                  True |       True |
 CMSSW_8_0_26 [slc6_amd64_gcc530] |                                  True |       True |
  CMSSW_8_1_0 [slc6_amd64_gcc530] |                                  True |       True |
  CMSSW_9_0_1 [slc6_amd64_gcc530] |                                  True |       True |
                             SELF |                                 False |       True |
```


smuzaffar:
please test

fabiocos:
@ggovi FYI, you test should be updated, right?

smuzaffar:
please test

smuzaffar:
@slava77 , @fabiocos,  we are seeing comparison failures for workflow 27434.0 (HGCal). We see these errors for even trivial PRs which should not change any physics (e.g. cms-git-tools).
Can you please check if these failures are real?

fabiocos:
@smuzaffar I have already seen these differences in recent integration, and I was attributing to a recent update to HF GEN-SIM. But may be they are not related to that. @kpedro88 do you see any other obvious reason for this?

slava77:
see cms-sw/cmssw#25234
it's probably solved in cms-sw/cmssw#25257

fabiocos:
@slava77 ok, good, it looks anyway independent on the boost addition

smuzaffar:
thanks @slava77 . I will go ahead with the boost update then.

PR #4503, opened 2018-11-14
Disable CUDA 10.0 deprecation warnings

fwyzard:
CUDA 10.0 deprecates some `cudaPointerAttributes` members.
The development branch of the cuda-api-wrappers works around this, but
those changes have not yet been merged into the stable branch.

For the time being, simply disable the CUDA deprecation warning. We can
re-enable them once the stable branch will include support for CUDA 10.


fwyzard:
@cmsbuild, please test

PR #4511, opened 2018-11-19
use source tag for config so that we can test it directly with new source option

smuzaffar:


smuzaffar:
please test

smuzaffar:
lets merge it for devel IB

PR #4517, opened 2018-11-21
use source tag for config so that we can test it directly with new source option

smuzaffar:
backport of #4511


smuzaffar:
please test

PR #4479, opened 2018-11-05
Update pyopenssl to 18.0.0; created spec for all its dependencies

amaltaro:
Main goal was to update the pyopenssl version in order to get two security vulnerabilities fixed, as reported in the WMCore repo:
https://github.com/dmwm/WMCore/network/alert/requirements.txt/pyopenssl/closed

However, several other changes were needed and in short they are:
* updated several other specs, reusing those from the CMSSW 10_4_X branch
* updated build-with-pip to be in sync with the CMSSW branch (rules for py2-* packages installed via pip)
* openssl spec had to be updated to include those 3 cryptograpic algorithms
* python library versions updated as much as possible (including numpy and scipy). Still a bunch to be updated though.

@h4d4 please review and merge when convenient

PR #4521, opened 2018-11-22
Added macro to auto-generate dependencies-setup*

amaltaro:
As discussed in this other PR:
https://github.com/cms-sw/cmsdist/pull/4479#issuecomment-440573696

we need to have these two macros in order to automatically generate the dependencies-setup.* scripts  when building RPMs. If a spec already generates then, then that macro doesn't run.

@smuzaffar did I miss anything? Apparently the only difference with the macro @vkuznet already has in place is that this one checks whether dependencies has already been created (if so, do nothing) and it also uses that `_CMSBUILD_BUILD_ENV_` var, which I wonder whether we need it or not?

### DIFF
--- a/rpm-preamble.file
+++ b/rpm-preamble.file
@@ -252,3 +252,29 @@ for f in `find %{pkginstroot} -type f`; do \
         perl -p -i -e 's{.*}{#!/usr/bin/env python3} if $. == 1 && m{#!.*/bin/python.*}' $f \
     fi \
 done
+
+# Generate dependencies-setup.{sh,csh} so init.{sh,csh} picks full environment.
+%define AutoSetupDependencies \
+if [ ! -e %i/etc/profile.d/dependencies-setup.sh ] ; then \
+mkdir -p %i/etc/profile.d \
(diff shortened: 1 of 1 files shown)


h4d4:
+1 to 4d0f5532033cbe6cc32f484ef029bc63263a1290 on slc7_amd64_gcc630.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/4521/test_4d0f553-20181122-092504.log" target="_blank">here</a>.

amaltaro:
I just ran the cmsBuild command and it didn't create any RPMs. I believe this is all we need as initial test, isn't it?

smuzaffar:
@amaltaro , `_CMSBUILD_BUILD_ENV_` is an internal env variable set by `cmsBuild` ( at build time) to avoid setting the env twice.

Yes adding these macros should not rebuild any package, they will be used by those packages which are re-built due to any other reason e.g. spec file of a package was updated or dependency of the package was changed.

vkuznet:
+1

PR #4559, opened 2018-12-06
revert t0_reqmon to 1.1.18.pre6

ticoann:


h4d4:
+1 to 88ae585eda01321781adf3aba10d34b48f099da9 on slc7_amd64_gcc630.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/4559/test_88ae585-20181206-173004.log" target="_blank">here</a>.

amaltaro:
@ticoann Seangchan, sorry for bugging you, but was it reverted because the T0 agent has to be updated with the same latest changes we pushed to the production agents (posting workflow summary instead of replicating)?

ticoann:
HI Alan, yes that is correct we have to use this version until Tier0 is updated to the same version as production agent.

PR #4565, opened 2018-12-07
Enable error reporting for CUDA library calls

fwyzard:
Define `CUB_STDERR` when compiling code that uses `cub` to enable (non fatal) error reporting for calls to CUDA library functions.


fwyzard:
@cmsbuild, please test

smuzaffar:
+externals

PR #4564, opened 2018-12-07
Extend support for self-adjoint matrices on CUDA devices

fwyzard:

PR #4560, opened 2018-12-06
Update tensorflow to 1.12.0

mrodozov:
Move tensorflow to 1.12.0 and adjust py2-tensorflow. libjpeg-turbo to be moved to variable later

PR #4566, opened 2018-12-07
Add build dependency on tensorflow-sources to avoid build conflicts

fwyzard:
By default, cmsBuild tries to build `tensorflow-sources` and `tensorflow-python3-sources` in parallel, leading to conflicts with the Bazel builds.
Declaring a build dependency on `tensorflow-sources` should force cmsBuild to build these packages one after the other.

### DIFF
--- a/tensorflow-python3-sources.spec
+++ b/tensorflow-python3-sources.spec
@@ -12,6 +12,7 @@ Patch3: tensorflow-1.6.0-eigen-rename-sigmoid
 
 BuildRequires: bazel eigen protobuf gcc
 BuildRequires: py2-setuptools java-env
+BuildRequires: tensorflow-sources
 Requires: py2-numpy python3 py2-wheel
 
 %prep


fwyzard:
@cmsbuild please test

fwyzard:
it would make sense to merge this and #4564 at the same time, since updating eigen will anyway trigger a rebuild of tensorflow.

mrodozov:
I understand you'll probably want this because it's a fast(er) solution, but consider we are nearly finished with tf 1.12 (see https://github.com/cms-sw/cmsdist/pull/4560). Fails only on arm (which I'm dealing with), thats why we are not adjusting cmssw already. I'm just saying this might be redundant. 
However it's good to know building tf_src and tf_py3_src at the same time might be problematic, I'll try to trick it the same way

fwyzard:
@mrodozov I'm curious, am I the only one that builds packages with `--builders 4` or anyway more than one tool at a time ?

mrodozov:
No, https://github.com/cms-sw/cms-bot/blob/master/run-cmsdist-tests.sh#L120 but I'm not using it (out of habit) hence haven't had that trouble building tensorflow.

fwyzard:
OK, no problem.

PR #4588, opened 2018-12-14
[10.4.X] drop sm_35 from cuda toolfile

smuzaffar:
@fwyzard , we had added sm_61 for P100 and in order to keep the build time in control we are dropping sm_35.


smuzaffar:
please test

smuzaffar:
+externals

PR #4616, opened 2019-01-07
[GCC 8.3] Bump gcc 8.3 to latest 2019-02-25

smuzaffar:


smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
@Dr15Jones , bumping gcc to latest gcc 8.2.0 generates class checksum error
```
error: class 'reco::JetCrystalsAssociation' has a different checksum for ClassVersion 11. Increment ClassVersion to 12 and assign it to checksum 348938133
```
this class is declared here https://github.com/cms-sw/cmssw/blob/master/DataFormats/BTauReco/interface/JetCrystalsAssociation.h 

Any idea why updating gcc could have generated such an error.

should I go ahead and merge this so that we can get an IB with latest GCC and can debug more?

Dr15Jones:
@pcanal changing gcc version appears to have made ROOT think a class was changed. Thoughts?

Dr15Jones:
@smuzaffar that class inherits from `std::pair` which is a terrible thing to do.

pcanal:
Unfortunately change in the base class layout requires an increment of the class version :(

smuzaffar:
@Dr15Jones , looks like it is related to this change https://gcc.gnu.org/bugzilla/show_bug.cgi?id=87822
So what is the solution here? All latest GCC versions of 6, 7 and 8 series have this issue. Should we add the following in https://github.com/cms-sw/cmssw/blob/master/DataFormats/BTauReco/src/classes_def.xml#L280 ?
```
<version ClassVersion="12" checksum="348938133"/>
```

Dr15Jones:
@smuzaffar we need the hash to version number to be the same across all compilers we are using to build. I see two choices
1. The correct one: make JetCrystalsAssociation not inherit from std::pair. The easiest way would just be to do `using JetCrystalsAssociation = std::pair<....>`
1. The poor way: Remove the ClassVersion information for that class from the classes_def.xml. I believe under that case ROOT will just make its own numbers at run time.

@pcanal what mechanism would allow us to read old data?

pcanal:
For I/O backward compatibility the 2nd option is the easiest (by a tiny bit).  In the First option you also need to add a rule that allows reading a ```JetCrystalsAssociation``` in to the equivalent ```std::pair```

Dr15Jones:
assign reconstruction

Dr15Jones:
At this point, we need to consult with reconstruction as to how they want to have the code changed.

smuzaffar:
@fabiocos , for your info.

Dr15Jones:
ping

smuzaffar:
See https://github.com/cms-sw/cmssw/issues/25965 , GCC 7.4 is also complaining about the checksum for this JetCrystalsAssociation class. We need to resolve this issue by either fixing/removing this class or updating the checksum and move to GCC 82 and GCC 74

davidlange6:
whoops, I commented on the new thread. There is reason to think that this is obsolete stuff

mrodozov:
please test

smuzaffar:
please test

PR #4624, opened 2019-01-08
upgrade pythia8 to 240, vincia to 2.2.04, dire to 2.003

mkirsano:


mkirsano:
To be tested together with CMSSW PR to change GeneratorInterface/Pythia8Interface: CMSSW #25602

fabiocos:
+1

fabiocos:
I would like to merge this for this morning IB, together with cms-sw/cmssw#25602

fabiocos:
hold

fabiocos:
before moving forward, @mkirsano we had a number of fixes on top of pythia8 230 listed in https://github.com/cms-externals/pythia8/commits/cms/230 . Have they been integrated by the authors in this new version? Otherwise I think we should forward port them (or better ask the authors to consider them), in order to avoid rediscovering old problems over and over again

mkirsano:
The main fix, as I understand, is the fix of the memory leak with 
LHAPDF. I checked with the version 235 that the authors fixed it in a 
different way.

  Other fixes probably simply anticipated future changes in pythia8.


04/02/19 11:00, Fabio Cossutti пишет:

fabiocos:
@mkirsano yes, LHAPDF6.h was modified to a large extent. I checked other updates, which were a bunch of fixes for missing initializations. Most of them seem to have been included in the new main version, not all of them anyway, I saw for instance one constructor in Event.h (and it is not the only one). Before moving forward I would like to have a technical check showing that we are not having obvious memory leaks, possibly lack of initializations. Did you run a long (thousands events) run with LHAPDF and monitored the memory for instance?

mkirsano:
@fabiocos I made this check with version 235. It is essential to have 
many lumi sections, not simply long run. It is improbable that it is 
broken again in 240, but I can try to find those tests and run again.

04/02/19 13:32, Fabio Cossutti пишет:

fabiocos:
@mkirsano yes, it would be important. Moreover concerning switch between lumi sections we have a pending issue cms-sw/cmssw#25708 and it would be good to check whether this is still there

mkirsano:
@fabiocos i made a pair of tests of GeneratorFilter with big statistics 
and many lumi sections, no signs of memory leak. I particular I tested 
the configuration with LHAPDF that had a significant memory leak with 
pythia8 230 before patching it.

04/02/19 13:46, Fabio Cossutti пишет:

kpedro88:
I ran the code-check again to find areas where there were potentially missing member variable initializations. I made a branch with the fixes applied (just a handful), after adjusting false positives manually: https://github.com/kpedro88/pythia8/tree/cms/240

There were a few places where I noticed none of my previous fixes were applied in the upstream: FJCore.h, FJCore.cc. Upon closer examination, these are copied directly from fastjet, so I think it is reasonable to avoid modifying them. I have excluded any changes to these files in the above branch.

fabiocos:
@mkirsano @kpedro88 thank you for the checks. At this point I would suggest to merge this PR, and then patch it with Kevin's code. In case you agree, I may do the first merge tonight

mkirsano:
I agree of course. I would like to see after what Kevin has corrected.

fabiocos:
unhold

fabiocos:
merge

Issue #4626, opened 2019-01-09
update vecgeom or patch to resolve uninit value use


slava77:
I was running valgrind on wf 150.0
and it looks like I rediscovered the issue fixed in https://gitlab.cern.ch/VecGeom/VecGeom/merge_requests/545
This is now present in present in v01.00.00

In valgrind the UninitCondition in an instantiation of a G4Polycone called from ```DDG4SolidConverter::polycone_rrz(DDSolid const&)``` appears at
https://gitlab.cern.ch/VecGeom/VecGeom/blob/v00.05.00/source/UnplacedPolycone.cpp#L754

smuzaffar:
@slava77 , https://github.com/cms-sw/cmsdist/pull/4628 patches vecgeom. The latest version does not go well with geant4 version we have.

PR #4628, opened 2019-01-09
Vecgeom patch for 00.05.00

mrodozov:
Resolve: https://github.com/cms-sw/cmsdist/issues/4626


mrodozov:
please test

mrodozov:
please test

slava77:
> Reco comparison results: 8 differences found in the comparisons

there is apparently nothing relevant related to the update.
Our MC tests with jenkins are only on about 20 sim events.
It looks like the fix for the uninit variable has fairly small impact (at least not affecting every event).

smuzaffar:
+externals

PR #4630, opened 2019-01-10
Add Nsight Compute CLI python libraries

fwyzard:
Properly package the Nsight Compute command line client and python libraries.

### DIFF
--- a/cuda.spec
+++ b/cuda.spec
@@ -63,7 +63,14 @@ rm -f %_builddir/build/bin/nvvp
 rm -f %_builddir/build/bin/computeprof
 
 # package the Nsight Compute command line tool
-mv %_builddir/build/NsightCompute-1.0/target/linux-desktop-glibc_2_11_3-glx-x64/* %{i}/bin/
+mkdir %{i}/NsightCompute-1.0
+mv %_builddir/build/NsightCompute-1.0/target                  %{i}/NsightCompute-1.0/
+mv %_builddir/build/NsightCompute-1.0/ProfileSectionTemplates %{i}/NsightCompute-1.0/
+cat > %{i}/bin/nv-nsight-cu-cli <<@EOF
+#! /bin/bash
+exec %{i}/NsightCompute-1.0/target/linux-desktop-glibc_2_11_3-glx-x64/nv-nsight-cu-cli "\$@"
+@EOF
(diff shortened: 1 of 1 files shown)


fwyzard:
@smuzaffar do you think this should work ?

The idea is to have in the PATH the file `nv-nsight-cu-cli`, which internally will launch `$CUDA_BASE/NsightCompute-1.0/target/linux-desktop-glibc_2_11_3-glx-x64/nv-nsight-cu-cli` .

If it works, do you think it would make sense to merge the `cuda-gdb` wrapper package here as well ?

smuzaffar:
@fwyzard , this should work. By the way, `nv-nsight-cu-cli` is already in PATH, so why you want to add wrapper?

Looks like `nv-nsight-cu-cli` is broken for slc6 (due to GLIBC version mismatch)
```
nv-nsight-cu-cli -v
nv-nsight-cu-cli: /lib64/libc.so.6: version `GLIBC_2.14' not found (required by nv-nsight-cu-cli)
```

fwyzard:
> By the way, `nv-nsight-cu-cli` is already in PATH, so why you want to add wrapper?

I put it in the PATH moving it and its support libraries from `$CUDA_BASE/NsightCompute-1.0/target/linux-desktop-glibc_2_11_3-glx-x64/nv-nsight-cu-cli` to `$CUDA_BASE/bin` .
However I just discovered that it looks for some support files in `../../ProfileSectionTemplates` by default, which then gets broken.

fwyzard:
> Looks like `nv-nsight-cu-cli` is broken for slc6 (due to GLIBC version mismatch)

Not much we can do - RHEL 6 and RHEL 7 use the same installer - apart from a bug report to NVIDIA.

fwyzard:
@cmsbuild, please test

fwyzard:
@smuzaffar, is it possible to test the external built by Jenkins interactively ?

fwyzard:
OK, from a local build it seems to work as intended.

smuzaffar:
@fwyzard , no it is not possible to test it interactively from jenkins. Once the jenkins job is done then the build area is cleaned up. 
We can upload the RPMS to e.g. `cms-PR<NUM>`  repository for any extra tests if needed.

PR #4634, opened 2019-01-14
[Do not merge] Testing genreflex to generate pcms for external libraries

yamaguchi1024:


smuzaffar:
please test

yamaguchi1024:
It looks that pcms are actually generated (They were there under tmp/slc7_amd64_gcc700/src/DataFormats/TCDS/src/DataFormatsTCDS/a for example). The error looks like it's from MakeFile. Would it be possible that CMSSW is checking how many files generated by genreflex, or to emit error if there was a warning from genreflex?

smuzaffar:
@yamaguchi1024 , do you see any DataFormatsMath_xr.cc file generated ? The build error is [a] which you should be able to reproduce by using CXXMODULE IB and passing correct cxxmodule flag to gereflex.

[a]
```
>> Compiling  /build/cmsbld/jenkins/workspace/ib-any-integration/CMSSW_10_5_CXXMODULE_X_2019-01-10-2300/src/DataFormats/Math/src/SSEVec.cc 
  Warning: Unused class rule: ROOT::Math::MatRepSym<*>
   gmake: *** [tmp/slc7_amd64_gcc700/src/DataFormats/Math/src/DataFormatsMath/a/DataFormatsMath_xr.cc] Error 1
 >> Compiling LCG dictionary: tmp/slc7_amd64_gcc700/src/DataFormats/Math/src/DataFormatsMath/a/DataFormatsMath_xr.cc
c++: error: DataFormatsMath/a/DataFormatsMath_xr.cc: No such file or directory
c++: fatal error: no input files
```

yamaguchi1024:
Hi @smuzaffar ,
Btw I can see cms logs since this morning)

The only difference between the genreflex with and without -cxxmodule is
```
Warning: Unused class rule: ROOT::Math::MatRepSym<*>
```
this, and a pcm generated under tmp directory. So I suspected that no warning is allowed to be emitted by genreflex.

Yes I see DataFormatsMath_xr.cc, not sure why it's complaining that this file doesn't exist..
```
[yuka@cmsdev15 a]$ ls
DataFormatsMath_xr.cc  DataFormatsMath_xr.pcm  DataFormatsMath_xr_rdict.pcm  DataFormatsMath_xr.rootmap
```

smuzaffar:
yes we pass `--fail_on_warnings` flag so on warning it should fail and delete generate xr.cc files.
While testing locally (using CMSSW_10_5_CXXMODULE_X_2019-01-10-2300 IB and passing `--cxxmodule` I can not reproduce the error. I would suggest that either we wait for tonight's CXXMODULE IB and re-do the test or merge this PR for tonight IB. I would prefer to merge it and see how it behaves in the full IB.

yamaguchi1024:
Actually, when the genreflex is used via scram it doesn't generates (or scram deletes?) DataFormatsMath_xr.cc
```
[yuka@cmsdev15 a]$ ls
DataFormatsMath_xr.pcm  DataFormatsMath_xr_rdict.pcm  DataFormatsMath_xr.rootmap
```

However when I execute genreflex directly, it generates DataFormatsMath_xr.cc
```
[yuka@cmsdev15 a]$ ls
DataFormatsMath_xr.cc  DataFormatsMath_xr.pcm  DataFormatsMath_xr_rdict.pcm  DataFormatsMath_xr.rootmap
```

yamaguchi1024:
@smuzaffar Ah I didn't see your comment. Indeed, `--fail_on_warnings` does looks like our thing. Ok, could you merge and let's see tonight's IB! :)

yamaguchi1024:
@smuzaffar About the reproducing, I think you were using IB build ROOT, which is not cmstest-masterv6. I have few patches on top of IB's ROOT in this tag, which allows genreflex to generate pcms.

yamaguchi1024:
I think merging this now into IB will just cause the same error on IB, can we test this PR without `--fail_on_warnings` ?

smuzaffar:
merging is fine, at least we will have an IB based on your latest changes and new flag. Local testing will be much faster after that :-)

yamaguchi1024:
https://github.com/cms-sw/cmssw/pull/25670
^ This fixes the error at least locally

PR #4672, opened 2019-02-04
Cxxmodule modulemap src

smuzaffar:
- Clean up CMSSW Queues speification logic. Now both cmssw-patch and cmssw spec uses the same information from cmssw-queue-override.file
- Added exra module.modulemap src for CXXMODULE IBs


smuzaffar:
please test

PR #4717, opened 2019-02-20
added py2-pybind11 dependency for fwlite

smuzaffar:
this is needed to get FWLite builds work. New dependency is needed due to FWCore/PyDevParameterSet dependency of cmsRun


smuzaffar:
merging it as it is only needed by fwlite builds.

PR #4729, opened 2019-02-25
Add cmsweb-exporters

vkuznet:


h4d4:
+1 to 031b8df5b6508b0c22351cc92b224eea5c67569d on slc7_amd64_gcc630.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/4729/test_031b8df-20190225-195505.log" target="_blank">here</a>.

amaltaro:
@vkuznet how about adding some very basic description to this PR? Since we're talking about a new package, such information might be useful in the future when we look at specs and assess what is what and what it is used for. Thanks

vkuznet:
cmsweb-exporters is a set of daemons to be run on worker node to monitor cmsweb data-services. They are designed with Prometheus guidelines and provide basic monitoring metrics for underlying services such as CPU, Memory usage, number of open file descriptors, etc. Currently we package the following exporters:
- `process_exporter` to monitor arbitrary UNIX process
- `das2go_exporter` to monitor DAS server
- `wmcore_exporter` to monitor WMCore based services if they implement necessary metrics and provide status endpoint
- `reqmgr_exporter` is an example of WMCore exporter for ReqMgr data-service based on [ProcessStats.py](https://github.com/dmwm/WMCore/blob/master/src/python/Utils/ProcessStats.py) class

The package also build 3rd party `apache_exporter` which can monitor apache frontends. In a future it can be extended to cover any other 3rd party exporter, e.g. see full list of them [here](https://prometheus.io/docs/instrumenting/exporters/)

amaltaro:
So these are daemons that will run in every single CMSWEB backend and they will be constantly polling some REST API made available by one of those services (like that one that Seangchan implemented some time ago for the Prometheus monitoring)? If so, then possibly they don't even need to run on the same CMSWEB backends(?)

vkuznet:
The information will be only requested by Prometheus server which will have
configuration to do such actions, e.g. we can configure Prometheus server
to fetch metrics every 5sec or every 5hours.

And, in `das2go|wmcore|reqmgr_exporters` can run on separate servers since I
made code aware of cmsweb x509 authentication, i.e. if someone setup valid proxy
file we can use it to access remote service. But `process_exporter` is designed
to work locally since it trace particular UNIX process (via /proc fs), therefore
it must be run on a node where you will monitor the UNIX process.

And, yes Seangchan implemented Prometheus monitoring based on these exporters
which by that time I explained to him. Now I decided to package them as
cms RPMs that we can install them as any other CMS DMWM service.

amaltaro:
I see.
And where would be the prometheus server running on? Who would maintain that service? How easy is it to change the configuration and add/remove a service from the monitoring?

vkuznet:
Alan,
I gave the link to full proposal. Here is for completeness:
https://docs.google.com/document/d/1A3IbiT60_ZEvihUHl-p1UQPHmpZPInohh9lr1dbHOxk/edit?usp=sharing
Please read it first before asking all these questions. It addresses everything.

amaltaro:
This proposal has never reached my inbox then. Thanks for the pointers though.

vkuznet:
Alan, it certainly did :), I sent you email to ping Caio and it contained the
link to the document.

PR #4731, opened 2019-02-26
[BuildRule] Disable rootmap dependency rule added for CXXMODULE IBs

smuzaffar:
this was causing shared libraries to be copied at `scrab build runtests` time causing unit tests to fail during PR tests


smuzaffar:
PR unit tests passed, so merging it for 23h00 IB

PR #4738, opened 2019-02-28
[SCRAM] updates build rules and SCRAM

smuzaffar:
- source_only dependency attribute added for dependency e.g.
```
<use name="foo" source_only="1"/>
```
to add a source level dependency without linking.

- Rootmap build rule fixed to avoid rebuilds


smuzaffar:
+externals

PR #4757, opened 2019-03-06
[10.6.X] cherry pick root-project/root#3485

smuzaffar:
Set offset of the used-for-write element in case of I/O rule on 'current' StreamerInfo.

This fixes ROOT-10016


smuzaffar:
please test

smuzaffar:
+externals
unit test is known to fail (hang)

PR #4772, opened 2019-03-11
update tnsnames.ora file

amaltaro:
Given the last time this file was updated was about 4 years ago, let's update it once again.
Just a FYI @belforte


smuzaffar:
Few years ago we removed it from cmsdist as CERN IT was not happy about it.
Can't you use it from system?

h4d4:
+1 to 21014136771b7fbad457cd182bfc051e7fd210b3 on slc7_amd64_gcc630.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/4772/test_2101413-20190311-172504.log" target="_blank">here</a>.

belforte:
indeed if CERN makes it available, what's the point in having out own ?  ANyhow I have no idea where CRAB uses this and hope I never need to.

amaltaro:
I think we can use it from the system. But the system here means either a) AFS, b) EOS, c) a web-service. Given that we don't want to have any dependency on AFS, we might then update the deployment procedure of services connecting to oracle such that we download/copy that file from the web-service during deployment time.

smuzaffar:
on CERN machines one can install oracle-instantclient-tnsnames.ora and set TNS_ADMIN to point to /etc/tnsnames.ora (this is what cmssw do).

Reading the thread here https://hypernews.cern.ch/HyperNews/CMS/get/swDevelopment/3283.html
CERN IT do not want us to make this file available in a way that search bots can index it. For outside CERN you can deploy it on CVMFS (I think contents of it are not searchable by search bots).

amaltaro:
/etc/tnsnames.ora is only available on lxplus/aiadm nodes, which are centrally maintained.
Do you maintain it yourself for the CMSSW IB machines?

In our service VOBoxes, we build and deploy our own version of oracle tools and instantclient:
https://github.com/cms-sw/cmsdist/blob/comp_gcc630/oracle.spec#L34

Maybe the way forward here would be to set TNS_ADMIN to a generic location (as you suggested) and sync the official tnsnames.ora file back to each VOBox through puppet. We just need to make sure the file replacement happens as an atomic operation, otherwise problems will happen eventually.

smuzaffar:
yes, on all of our build and dev nodes we deploy oracle-instantclient-tnsnames.ora package via puppet and in docker we mount -v /etc/tnsnames.ora:/etc/tnsnames.ora
CMSSW sets TNS_ADMIN environment pointing to /etc/tnsnames.ora

PR #4774, opened 2019-03-13
Updated root to tip of branch v6-14-00-patches

smuzaffar:


smuzaffar:
please test

fabiocos:
@smuzaffar @Dr15Jones starting from CMSSW_10_6_X_2019-03-17-0000 3 unit tests fail in the ROOT614 IB, see https://cmssdt.cern.ch/SDT/cgi-bin/logreader/slc7_amd64_gcc700/CMSSW_10_6_ROOT614_X_2019-03-17-0000/unitTestLogs/PhysicsTools/PythonAnalysis#/ . Anyway this looks to me posterior to the merge of this PR, is it correct? I was trying to correlate with the most recent changes...

Dr15Jones:
@davidlange6 please comment

fabiocos:
@Dr15Jones @davidlange6 I discussed with @smuzaffar we agree that the issue seems unrelated to this update

smuzaffar:
Looks like when packages were re-build over the weekend then some of python packages were not properly build. We are looking in to the logs files to see what went wrong with these packages.

smuzaffar:
Issue is with permissions. When py2-root_numpy was build then it set the lib directory permissions to 0700 which means no one else can read it. We are looks why it happened.
```
>ls -lart /cvmfs/cms-ib.cern.ch/nweek-02568/slc7_amd64_gcc700/external/py2-root_numpy/4.8.0-pohgbh
total 2
drwx------. 3 cvmfs cvmfs 3 Mar 17 12:04 lib
drwxr-xr-x. 3 cvmfs cvmfs 3 Mar 17 12:04 etc
```

smuzaffar:
@fabiocos , looks like building RPMs build /home do not go well. For few VMs, we started using /home/cmsbld as jenkins work directory (/home is SSD disk, so should be fast as compare to /build which is Ceph storage). We can reproduce the issue if we build py-2root_numpy under /home but building under /build sets the correct permission. 

For now, we have move back to /build and we will force build this package for tonight so that we get clean 23h IB.

PR #4745, opened 2019-03-01
Branch IB/CMSSW_10_6_X/gcc700 - Cherry-picked  6c1e592 from root pr.

gudrutis:

PR #4555, opened 2018-12-05
Update libjpeg-turbo to 2.0.1

mrodozov:
Update to 2.0.1 as tf 1.12 now works with 2.0.x

PR #4523, opened 2018-11-23
[10.4.X] Updated root: Support new Hexagon type in TEveBoxSet + 2 fixes

smuzaffar:
Included patch https://github.com/cms-sw/root/pull/117 on top on root v6.12 patches brach

PR #4430, opened 2018-10-17
Revert "Updated root to tip of branch v6-12-00-patches"

smuzaffar:
Reverts cms-sw/cmsdist#4417

PR #4775, opened 2019-03-13
Updated root to tip of branch v6-12-00-patches

smuzaffar:


smuzaffar:
please test

fabiocos:
```
13-Mar-2019 15:01:25 CET  Initiating request to open file root://eoscms.cern.ch//eos/cms/store/user/cmsbuild/store/relval/CMSSW_9_2_2/RelValTTbar_13/MINIAODSIM/PU25ns_92X_upgrade2017_realistic_v1-v1/10000/8E7EE25F-294E-E711-A5CC-0025905B8610.root
13-Mar-2019 15:01:29 CET  Successfully opened file root://eoscms.cern.ch//eos/cms/store/user/cmsbuild/store/relval/CMSSW_9_2_2/RelValTTbar_13/MINIAODSIM/PU25ns_92X_upgrade2017_realistic_v1-v1/10000/8E7EE25F-294E-E711-A5CC-0025905B8610.root
Begin processing the 1st record. Run 1, Event 54, LumiSection 2 on stream 0 at 13-Mar-2019 15:01:41.093 CET


A fatal system signal has occurred: segmentation violation
The following is the call stack containing the origin of the signal.

Wed Mar 13 15:01:42 CET 2019

Current Modules:

Module: PoolOutputModule:out (crashed)

A fatal system signal has occurred: segmentation violation
/build/cmsbld/jenkins/workspace/ib-any-integration/CMSSW_10_6_X_2019-03-12-1100/src/PhysicsTools/PatAlgos/test/runtests.sh: line 42: 12935 Segmentation fault      (core dumped) cmsRun ${LOCAL_TEST_DIR}/patTuple_updateMet_fromMiniAOD_cfg.py
Failure using patTuple_updateMet_fromMiniAOD_cfg.py: status 139
status = 35584

---> test runtestPhysicsToolsPatAlgos had ERRORS
```

@smuzaffar @Dr15Jones apparently this update does not work out of the box. I wonder whether it is reasonable to move forward and build 10_6_0_pre2 with what we have now, so far we have seen a failure only in one unit test. Otherwise we need to delay it further together with the validation. In this case perhaps it is better to temporarily revert cms-sw/#26053 , that apparently triggers this issue, build 10_6_0_pre2 to allow the validation of all the rest, and then go back to merge the PR and test further ROOT

Dr15Jones:
I vote to just go forward with the one unit test failure and not do the revert.

fabiocos:
@Dr15Jones we have so far no evidence of other issues after several IBs. I am not so much concerned about the single unit test, that in any case needs to be fixed, but it is probably not crucial in itself for the pre-release test, but whether this could be the hint of a deeper problem.

fabiocos:
BTW, I see that the Validation/Geometry test is now running ok, so that problem looks fixed, but apparently there is another one, so it isn't clear that this latest version is a priori better than what we have so far in absolute terms, more checks are needed I would say

slava77:
> Reco comparison results: 197 differences found in the comparisons

It looks like a similar  issue reported in https://github.com/cms-sw/cmssw/issues/25919#issuecomment-463786503

why was this merged ???
It looks like the second attempt to ignore problems and still merge, similar to what was done on Oct 12.


Looking at the past merges:
- #4757 on Mar 6: no reco diffs 
- #4745 on Mar 2: no reco diffs
- #4605 on Dec 21: no reco diffs (build process changes only)
- #4555 on Dec 13: no reco diffs (build deps changes only)
- #4523 on Nov 22: no reco diffs
- #4430 on Oct 17 reverts #4417 
- #4417 on Oct 12: reco diffs and a failure in the same unit test with apparently the same inputs

I'd like to repeat my message from Oct 17: 
https://github.com/cms-sw/cmsdist/pull/4417#issuecomment-430538548

smuzaffar:
@slava77 , my bad. I though these 15 failure are standard failures we get while testing root.
I will revert it for next IB.

slava77:
> @slava77 , my bad. I though these 15 failure are standard failures we get while testing root.
> I will revert it for next IB.

Thanks.
BTW, the last time the crash in runtestPhysicsToolsPatAlgos also coincided with all validate.C used in the reco comparisons failing/crashing and prevented any practical non-trivial signoff by reco.

fabiocos:
I preferred not to rush for pre2 to the newest version, and keep the previous one (we have an issue with unit tests there as well, but we saw stability in RelVals across several IBs)

slava77:
> I preferred not to rush for pre2 to the newest version, and keep the previous one (we have an issue with unit tests there as well, but we saw stability in RelVals across several IBs)

It looks like with this measured approach a bigger disaster of this entering also in pre2 was averted.
Nice.

smuzaffar:
This is reverted by #4782. Today's 11h00 IB has been restarted to pick up the change.

PR #4776, opened 2019-03-13
Set RUCIO_HOME env var; define a standard config file; update version

amaltaro:
Define a default rucio.cfg template in case someone wants to only use the client. It will then find RUCIO_HOME and one can properly `from rucio.client import Client`.


h4d4:
+1 to e033e878c6c88e4d2fd2930108dbc2bb5ecec4d3 on slc7_amd64_gcc630.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/4776/test_e033e87-20190313-104006.log" target="_blank">here</a>.

PR #4783, opened 2019-03-14
Update root 6-12

mrodozov:
Get one commit on top of 6-12

### DIFF
--- a/root.spec
+++ b/root.spec
@@ -2,7 +2,7 @@
 ## INITENV +PATH PYTHON27PATH %{i}/lib
 ## INITENV +PATH PYTHON3PATH %{i}/lib
 ## INITENV SET ROOTSYS %{i}
-%define tag 78c5af9a8bf237b4e4a07145b1b92579903b6e57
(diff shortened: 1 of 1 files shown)


mrodozov:
please test

smuzaffar:
please test

smuzaffar:
+externals
@fabiocos , @slava77 , @ Dr15Jones, this PR adds on commit [a]. This was merged in root 6.14 and above a long time ago and now we have requested to include in 6.12 too. This is needed to move forward with GCC 7.4 and GCC 8.3. PR tests looks good. We have one unit test failing which is already failing in IBs. Comparison also looks good.

Any objections on integrating it in 10.6.X?

[a] https://github.com/root-project/root/pull/3518
[IO] Do not loop over base classes of pair since it is an stl template 
https://github.com/cms-sw/root/commit/32b81d13cf2bae1fb9f41a9a5ce46cb28657d55f

fabiocos:
+1

this does not look to solve the existing issue, but the tests do not show evidence for new problems as well

PR #4787, opened 2019-03-18
force build py2-root_numpy

smuzaffar:
Previous build was done on a machine whcih has /home as jenkins work directory and looks like root_numpy does not set correct permissions if build in /home
https://github.com/cms-sw/cmsdist/pull/4774#issuecomment-473982222


smuzaffar:
please test

PR #4794, opened 2019-03-21
[106X] New build rules to avoid rebuild after a successful build

smuzaffar:


smuzaffar:
PR tests shows nothing was rebuilt, so I am merging this for 23h IB.

```
grep ' newer ' scram-rebuild.log  | grep -v 'cache/xlibs.backup' | wc -l
0
```

PR #4845, opened 2019-04-05
[GCC9] Update ThePeg 2.1.5 and HerwigPP 7.1.5

smuzaffar:
Updated ThePeg and HerwigPP which now builds with GCC 9.0


smuzaffar:
merging and triggering and IB

PR #4846, opened 2019-04-05
Update gcc version to 7.4.1

fabiocos:
Update gcc version to 7.4.1 already running in the DEVEL_X IB without apparent problems

### DIFF
--- a/gcc.spec
+++ b/gcc.spec
@@ -1,10 +1,10 @@
-### RPM external gcc 7.3.1
+### RPM external gcc 7.4.1
 ## INITENV +PATH LD_LIBRARY_PATH %{i}/lib64
 #Source0: ftp://gcc.gnu.org/pub/gcc/snapshots/4.7.0-RC-20120302/gcc-4.7.0-RC-20120302.tar.bz2
 # Use the svn repository for fetching the sources. This gives us more control while developing
 # a new platform so that we can compile yet to be released versions of the compiler.
 # See: https://gcc.gnu.org/viewcvs/gcc/branches/gcc-6-branch/?view=log
-%define gccRevision 257125
+%define gccRevision 268351
 %define gccBranch branches/gcc-7-branch
 
 %define moduleName gcc-%(echo %{gccBranch} | tr / _)-%{gccRevision}


fabiocos:
please test

fabiocos:
@smuzaffar is this test going to rebuild everything? Basically this is already done in the DEVEL_X IB, not sure we really need it...

smuzaffar:
@fabiocos , for externals it is going to reuse all the pre-build packages but yes for cmssw it is going to rebuild every thing. I agree we do not need the test but does not hurt much

fabiocos:
@smuzaffar ok let it run, this will provide anyway the comparison which is useful to see

fabiocos:
BTW, pre-build with the new gcc? I guess not, while this is done in DEVEL_X...

smuzaffar:
yes, as architecture is same (slc7_amd64_gcc700) so what ever has been build by DEVEL IBs will be reused by normal IBs. Ideally no external package should be rebuild.

fabiocos:
please test

fabiocos:
please abort

not sure that the bot is in the correct state

smuzaffar:
bot is in correct state and I have restarted tests (for those which were using bad bot). I did not realized that you already have revert the change.

fabiocos:
@slava77 @perrotta @civanch all the DQM and comparison test failures are concentrated in a single workflow, 9.0, which is a simulation one. And indeed there are differences already a generator level.
I guess that the subsequent differences in the comparison are due to that

fabiocos:
I have performed the following test: rerun wf 9 starting from either step1 or step2 produced with the default CMSSW (gcc730). Rerunning the DQM comparison with the baseline in the former case I see

```
 - summary of 121564 tests:
 o Failiures: 1.43% (1736/121564)
 o Nulls: 0.00% (0/121564) 
 o Successes: 98.57% (119828/121564) 
 o Skipped: 0.00% (0/121564) 
 o Missing objects: 0
```

and in the latter

```
 - summary of 121564 tests:
 o Failiures: 0.00% (1/121564)
 o Nulls: 0.00% (0/121564) 
 o Successes: 100.00% (121563/121564) 
 o Skipped: 0.00% (0/121564) 
 o Missing objects: 0
```

So this confirms that the non reproducibility is located in the simulation part, and not only at GEN-SIM level.

smuzaffar:
+externals

fabiocos:
@smuzaffar I have put the discussion about this PR in agenda for next ORP, I would like to merge it after that

fabiocos:
+1

fabiocos:
@smuzaffar we are not tracking minor version update in the architecture name (it was gcc700 already with gcc 731...), am I correct?

smuzaffar:
that is correct @fabiocos

PR #4857, opened 2019-04-08
fwlite:drop FWCore/PyDevParameterSet and FWCore/Framework/bin

smuzaffar:
FWLite buildset updated
- Drop FWCore/PyDevParameterSet
- Remobe FWCore/Framework/bin i.e. cmsRun* executables which should not be needed by fwlite release.


smuzaffar:
this can only be test in fwlite ib. So merging it for next FWLite IB build

PR #4869, opened 2019-04-10
Condition for ppc64 for tfcompile

mrodozov:
cms-sw doesn't compile as it can't find tfcompile binary, since it's not compiled on ppc (llvm missing references)


mrodozov:
please test

PR #4876, opened 2019-04-15
Move xgboost back to 0.81

mrodozov:
py2-xgboost is failing on arm64 and it breaks the IB build, I want to put it back on 0.81 until I fix the pip build on arm
@davidlange6

### DIFF
--- a/pip/requirements.txt
+++ b/pip/requirements.txt
@@ -192,5 +192,5 @@ webencodings==0.5.1
 Werkzeug==0.15.2
 wheel==0.33.1
 widgetsnbextension==3.4.2
-xgboost==0.82
+xgboost==0.81
 xrootdpyfs==0.1.5


mrodozov:
please test

mrodozov:
note: same pkg breaks the powerpc build.

davidlange6:
hi @mrodozov - where can I see the error you are getting? (though it sounds like you've understood it already..)

mrodozov:
@davidlange6 I copied the last log here: /afs/cern.ch/user/m/mrodozov/public/xgboost.log 
as it fails only on arm and ppc when I use pkgtools to build it, complains about missing header:
```
g++ -c -DDMLC_LOG_CUSTOMIZE=1 -std=c++11 -Wall -Wno-unknown-pragmas -Iinclude   -Idmlc-core/include -Irabit/include -I/include -O3 -funroll-loops -fPIC -fopenmp src/common/hist_util.cc -o build/common/hist_util.o
    src/common/hist_util.cc:16:12: fatal error: xmmintrin.h: No such file or directory
       #include <xmmintrin.h>
                ^~~~~~~~~~~~~
    compilation terminated.
    make: *** [build/common/hist_util.o] Error 1

```
 I'm reading on this related issue
https://github.com/yquake2/yquake2/issues/204

mrodozov:
Here's the PR where the devs expressed some concerns about non Intel systems
and I think there is something wrong with the conditions here:
https://github.com/dmlc/xgboost/pull/3957/files#diff-c1e1ba4a6b32de4e6256b149b4c3547bR11
which gets into:
https://github.com/dmlc/xgboost/pull/3957/files#diff-daec1d0ffe267847ee82f8b885a06dbbR17
and then complains about
'_MM_HINT_T0' was not declared in this scope
when I added
#include <arm_neon.h> instead of  <xmmintrin.h> (arm_neon was installed).
This is fixable but not in the current situation while building with pip (I just don't see fast enough way to patch the pip recipe on the fly)

davidlange6:
I think you can follow the example in CMSDIST/pip/py2-tables.file (not that its pretty..)

%define PipPreBuild export HDF5_DIR=${HDF5_ROOT} && tar -xzf tables-%{realversion}.tar.gz && pushd tables-%{realversion} && for pch in %{patches} ; do 
patch -p1 < ${pch} ; done && popd && rm -f tables-%{realversion}.tar.gz && tar czf tables-%{realversion}.tar.gz  tables-%{realversion}

mrodozov:
Thanks for the suggestion, solution is here: 
https://github.com/cms-sw/cmsdist/pull/4879
not very pretty but builds

PR #4905, opened 2019-04-24
Reverting cuda.spec to 3e9f7f9

gudrutis:
Reverting cuda.spec to 3e9f7f9f61bfa61a5a03e5a34aa7812d2780db9a Due to problems building on gcc900


gudrutis:
please test

PR #4906, opened 2019-04-25
Add override option and python dependency

fwyzard:
Pass the `--override` option to the CUDA installer, to "ignore compiler, third-party library, and toolkit detection checks which would prevent the CUDA Toolkit and CUDA Samples from installing."
This is needed in the GCC 9 builds.

Add an explicit dependency on `python` in order to fix the `cuda-gdb` wrapper.


fwyzard:
@cmsbuild, please test

PR #4916, opened 2019-04-30
[BuildRules] By default run code-format only for changed files

smuzaffar:
This make it consistent with code-checks. Use `scram b code-format-all` if you want to run code-format for all files under src directory.


smuzaffar:
please test

PR #4928, opened 2019-05-05
[BuildRules] Improved check-headers rules

smuzaffar:
New rule should 
- avoid internal compiler errors
- only run for public shared libraries interface headers (h, hh, hpp)
- Use local package dependencies

### DIFF
--- a/scram-project-build.file
+++ b/scram-project-build.file
@@ -64,7 +64,7 @@ Requires: glibc
 %endif
 
 %if "%{?configtag:set}" != "set"
-%define configtag       V05-08-30
+%define configtag       V05-08-31
 %endif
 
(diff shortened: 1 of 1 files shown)


smuzaffar:
please test

smuzaffar:
+externals

fabiocos:
@smuzaffar this should avoid all the fake internal compiler errors that are polluting the reports, do I understand correctly?

smuzaffar:
yes this should avoid those fake errors. For now I have disabled header checks for PR as we first need to fix these already existing errors
https://cmssdt.cern.ch/SDT/jenkins-artifacts/check_headers/CMSSW_10_6_X_2019-05-05-2300/slc7_amd64_gcc700/headers_with_error.log

PR #4944, opened 2019-05-08
[BuildRukes] Bugfix: update cmssw-config tag to V05-08-35

smuzaffar:
code-format rule in V05-08-33 was no broken and it was not picking up the src/.clang-format.
New tag should fix this issue.


smuzaffar:
please test

PR #4965, opened 2019-05-13
[Python] Added hepdata lib python package

### LINKED ISSUE #4954: Include hepdata_lib python package
[hepdata_lib](https://pypi.org/project/hepdata-lib/) is a tool for users to convert their ROOT files into the standard HepData format for analysis archival and distribution. It would be useful to distribute this to users in CMSSW.

smuzaffar:
This resolves https://github.com/cms-sw/cmsdist/issues/4954
New python package hepdata-lib added. Note that it is only added for Python2 as it depends on root and we only have root build with python2.


smuzaffar:
please test for slc7_aarch64_gcc700

fabiocos:
@smuzaffar @kpedro88 I see on the package page a dependency on ImageMagick that anyway is not reported in this update. Is that not really needed, and the ROOT interface is enough?

smuzaffar:
@fabiocos , ImageMagick (convert command) is needed at runtime. We can built and distribute it but as it is available for most distributions we can just use it from system (just like many other commands we use from system).

fabiocos:
@smuzaffar sure, I was just concerned about adding another dependency...

smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
+externals

PR #4981, opened 2019-05-17
New dasgoclient version

vkuznet:
Due to recent outages of DBS on cmsweb we found that users use dasgoclient in their scripts without input validation. This new version adds query validation to avoid passing input to backend server and enhance error codes.


smuzaffar:
please test

vkuznet:
I found a flaw in one regexp. I'm closing this PR and will submit a new one once I'll prepare it.

smuzaffar:
abort test

PR #4985, opened 2019-05-17
New dasgoclient version

vkuznet:
Due to recent outages of DBS on cmsweb we found that users use dasgoclient in their scripts without input validation. This new version adds query validation to avoid passing input to backend server and enhance error codes.

Replace PR #4981  fixed regular expression with dataset/block patterns


smuzaffar:
please test

smuzaffar:
+externals

PR #4480, opened 2018-11-06
Update cherrypy to 17.4.0 in a new spec file

amaltaro:
Created a new CherryPy spec with version 17.4.0 (the latest one supporting python2) and all its python dependencies. Also updated a bunch of other packages.
This PR also contains the matplotlib changes as in #4998 

And I take advantage to also update WMCore tag to 1.2.3.pre3 (and remove an unneeded cherrypy dependency from the wmagent.spec).

PR #4998, opened 2019-05-22
update matplotlib to 2.2.4 and freetype to 2.5.3

amaltaro:
Synchronizing matplotlib and freetype with the `IB/CMSSW_10_6_X` branch. Also this will be necessary if I want to upgrade CherryPy, as in: https://github.com/cms-sw/cmsdist/pull/4480

@smuzaffar Shahzad, I noticed it triggers the build of almost all packages in comp (I assume it's the freetype spec doing so). Do you see any possible issues with this PR?

### DIFF
--- a/py2-matplotlib.spec
+++ b/py2-matplotlib.spec
@@ -1,29 +1,42 @@
-### RPM external py2-matplotlib 1.4.3
+### RPM external py2-matplotlib 2.2.4
 ## INITENV +PATH PYTHONPATH %i/${PYTHON_LIB_SITE_PACKAGES}
-Source: http://sourceforge.net/projects/matplotlib/files/matplotlib/matplotlib-%{realversion}/matplotlib-%{realversion}.tar.gz
-Requires: py2-pytz py2-numpy py2-python-dateutil zlib libpng freetype py2-pyparsing py2-six
-BuildRequires: py2-setuptools
+
+Source: https://github.com/matplotlib/matplotlib/archive/v%{realversion}.tar.gz
+Requires: py2-pytz py2-numpy py2-python-dateutil zlib libpng freetype py2-pyparsing py2-six py2-cycler py2-subprocess32 py2-kiwisolver
+#Requires: py2-backports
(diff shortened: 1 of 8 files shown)


smuzaffar:
@amaltaro , in comp, freetype is only used by 
```
py2-matplotlib.spec
py2-pil.spec
root.spec
```
so it should only build these packages plus any package which depend on these.

h4d4:
+1 to 38b1f41c8fb9b37ef63cd040267d5b8705bf1bca on slc7_amd64_gcc630.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/4998/test_38b1f41-20190522-170507.log" target="_blank">here</a>.

amaltaro:
So, services that should be affected by this PR are:
```
(PY2) alannewmacpro:cmsdist amaltaro$ grep freetype * | grep Requires
py2-matplotlib.spec
root.spec

(PY2) alannewmacpro:cmsdist amaltaro$ grep -r py2-matplotlib * | grep Requires
PHEDEX-web.spec:Requires
crabhb.spec
happyface.spec
sreadiness.spec
```

My concern goes only to `root` and `phedex-web`, but still I think we should move forward...

py2-matplotlib.spec
```diff
@@ -1,29 +1,42 @@
...
 macosx = False
 EOF
 
+mkdir no-pkg-config
+(echo '#!/bin/sh'; echo 'exit 1') > no-pkg-config/pkg-config
+chmod +x no-pkg-config/pkg-config
+
 %build
-python setup.py build
+#python setup.py build
```
amaltaro, line 29: @smuzaffar is this commented out on purpose? I just x-checked against IB/11_0 and it's the same there.

smuzaffar: @amaltaro this line was needed when we were building it without pip. Now the package is build with pip , so pip internally take care of it.

amaltaro: thanks for the confirmation, Shahzad.

PR #4999, opened 2019-05-22
added CXXModules.mk for CXXModules IBs to build new pcms for selected packages

smuzaffar:
Added CXXModules.mk for CXX modules IBs. By default new modules are disabled.


smuzaffar:
there is no point running tests are these files are not used during PR tests

PR #5006, opened 2019-05-24
[BuildRules] Updated tag V05-08-51

smuzaffar:
- Added support for REM_GENREFLEX_ARGS via Package's BuildFile
- Bug fix: Make sure scram fails when genrelfex exit with non-zero exit code.


smuzaffar:
please test

PR #5010, opened 2019-05-28
[cms-git-tools] Get default protocol from git config

smuzaffar:
Includes https://github.com/cms-sw/cms-git-tools/pull/104


smuzaffar:
No need to test as it has been tested via cms-sw/cms-git-tools#104

PR #5044, opened 2019-06-17
[BuildRules] Pass full path to tmp direcotry to avoid errors

smuzaffar:


smuzaffar:
this is a trivial update which is only used during `scram build code-format`.

PR #5057, opened 2019-06-25
Update lz4, ROOT on cmsdist master

mrodozov:
Since lz4 will rebuild root, add also xxHash and its paths.

PR #5060, opened 2019-06-27
Test commit to check cms-bot

gudrutis:
It should check that if the test fails, the correct label is set.

https://github.com/cms-sw/cmsdist/pull/5057


gudrutis:
please test

gudrutis:
Bot works as intended, closing.

PR #5063, opened 2019-06-28
use openblas instead of gslcblas

smuzaffar:
- Build OpenBlas with `DYNAMIC_ARCH=1`
- Drop `gslcblas` library from gsl toolfile and add `openblas` dependency to make sure that any cmssw package using gsl links `openblas` instead of  `gslcblas`
- set GSL_CBLAS_LIB so that gsl-config properly return `-lopenblas` instead of `-lgslcblas`
- Fixed `geneva`, `thepeg` and `herwigpp` specs to link against `openblas`

### DIFF
--- a/OpenBLAS.spec
+++ b/OpenBLAS.spec
@@ -10,12 +10,12 @@ Source: https://github.com/xianyi/OpenBLAS/archive/v%{realversion}.tar.gz
 
 # PRESCOTT is a generic x86-64 target https://github.com/xianyi/OpenBLAS/issues/685 
 %ifarch x86_64
-make FC=gfortran BINARY=64 TARGET=PENRYN NUM_THREADS=256 DYNAMIC_ARCH=0
+make FC=gfortran BINARY=64 TARGET=PENRYN NUM_THREADS=256 DYNAMIC_ARCH=1
 %else
 %ifarch aarch64
-make FC=gfortran BINARY=64 TARGET=ARMV8 NUM_THREADS=256 DYNAMIC_ARCH=0
+make FC=gfortran BINARY=64 TARGET=ARMV8 NUM_THREADS=256 DYNAMIC_ARCH=1
 %else
-make FC=gfortran BINARY=64 NUM_THREADS=256 DYNAMIC_ARCH=0
+make FC=gfortran BINARY=64 NUM_THREADS=256 DYNAMIC_ARCH=1
 %endif # aarch64
 %endif # x86_64
 
--- a/geneva.spec
+++ b/geneva.spec
@@ -4,7 +4,7 @@ Source: git+https://stash.desy.de/scm/geneva/geneva-public.git?obj=master/%{real
 
 BuildRequires: cmake gmake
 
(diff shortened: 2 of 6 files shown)


smuzaffar:
please test

slava77:
There is one small difference in a jet b-tag discriminant variable
```

WF # | Differences
-- | --
all_OldVSNew_RunJetHT2016EreMINIAODwf136p7611 | 1

```
It's hard to say which one (I only plot by the vector index); it may even be one of the mxnet-based cases.
@hqucms we may need some further investigation.

@smuzaffar please clarify if this change (alone, or at least without much else related) can be made available in the DEVEL branch IBs.
Thank you.

smuzaffar:
@slava77 , DEVEL IBs are broken due to tensorflow update. I can include it in gcc8 or root 618 IBs

hqucms:
> There is one small difference in a jet b-tag discriminant variable
> 
> ```
> 
> WF # | Differences
> -- | --
> all_OldVSNew_RunJetHT2016EreMINIAODwf136p7611 | 1
> ```
> 
> It's hard to say which one (I only plot by the vector index); it may even be one of the mxnet-based cases.
> @hqucms we may need some further investigation.

@slava77 Is this difference in AK4 jets or AK8 jets? Only AK8 jets have MXNet-based taggers.

slava77:
> @slava77 Is this difference in AK4 jets or AK8 jets? Only AK8 jets have MXNet-based taggers.

in AK8

hqucms:
> > @slava77 Is this difference in AK4 jets or AK8 jets? Only AK8 jets have MXNet-based taggers.
> 
> in AK8

OK I will take a look.

hqucms:
@slava77 I confirm that there are some small numerical changes on the MXNet-based tagger outputs after changing the BLAS library. The differences are small enough so they should not have any impacts on the physics, but will unfortunately affect the reproducibility of existing NanoAOD workflows.

slava77:
@hqucms 
thank you for checking.

This sounds "good enough" for 11_X then. 
Perhaps if all is good after a relval cycle we could consider a backport.

smuzaffar:
+externals
@fabiocos , would you like to get it in 11.0.0.pre3?

fabiocos:
@smuzaffar @slava77 @hqucms how urgent is this/ i have already started the build, of course we can integrate, launch an IB and restart in parallel, if this is felt as urgent

slava77:
> @smuzaffar @slava77 @hqucms how urgent is this/ i have already started the build, of course we can integrate, launch an IB and restart in parallel, if this is felt as urgent

In the context of potential backports and the next pre potentially falling into August and slow response from validators (in August), it would be nice to have it in pre3.

fabiocos:
+1

@smuzaffar I would suggest to start a build with package integrated, and restart the 11_0_0_pre3 build with it

smuzaffar:
sure @fabiocos , I am triggering an IB now.

fabiocos:
@smuzaffar thanks

PR #5068, opened 2019-07-03
[DEVEL] Tensorflow build recipe cleanup

smuzaffar:
- fix eigen flags in toolfile: tensorflow is build with these extra flags
- added extra pip packages dependencies for tensorflow
- build py2-tensorflow using normal build-with-pip system to make sure that all of its deps are available
- build tensorflow and protobuf with c++1z
- combine tensoflow py2 and py2 build recipes
- Make use of cms protobuf to build tensorflow


smuzaffar:
please test

smuzaffar:
please test

PR #5080, opened 2019-07-08
Add onnxruntime

hqucms:
As discussed in https://github.com/cms-sw/cmssw/issues/27458.

@davidlange6 This builds only the C API -- but the python package might also be helpful if analyzers want to run some ML inference in their python-based analyses.

### DIFF
--- a/onnxruntime.spec
+++ b/onnxruntime.spec
@@ -13,6 +13,7 @@ BuildRequires: cmake ninja zlib python3
 rm -rf ../build; mkdir ../build; cd ../build
 
 cmake ../%{n}-%{realversion}/cmake -GNinja \
+   -DPYTHON_EXECUTABLE=${PYTHON3_ROOT}/bin/python3 \
    -DCMAKE_BUILD_TYPE=Release \
    -DCMAKE_INSTALL_PREFIX="%{i}" \
    -Donnxruntime_BUILD_SHARED_LIB=ON \
--- a/onnxruntime.spec
+++ b/onnxruntime.spec
@@ -4,7 +4,7 @@
 %define github_user microsoft
 Source: git+https://github.com/%{github_user}/%{n}.git?obj=%{branch}/v%{tag}&export=%{n}-%{realversion}&submodules=1&output=/%{n}-%{realversion}.tgz
 
-BuildRequires: cmake ninja
+BuildRequires: cmake ninja zlib python3
 
 %prep
 %setup -n %{n}-%{realversion}
--- a/onnxruntime.spec
+++ b/onnxruntime.spec
@@ -1,10 +1,11 @@
-### RPM external onnxruntime 0.4.0
-%define tag %{realversion}
-%define branch master
-%define github_user microsoft
-Source: git+https://github.com/%{github_user}/%{n}.git?obj=%{branch}/v%{tag}&export=%{n}-%{realversion}&submodules=1&output=/%{n}-%{realversion}.tgz
+### RPM external onnxruntime 0.5.0
+%define tag 2a8b02097514ca90ac7471054460dcdc47b69e1b
+%define branch cms/v%{realversion}
+%define github_user cms-externals
+Source: git+https://github.com/%{github_user}/%{n}.git?obj=%{branch}/%{tag}&export=%{n}-%{realversion}&submodules=1&output=/%{n}-%{realversion}.tgz
 
 BuildRequires: cmake ninja zlib python3
+Requires: eigen protobuf
 
 %prep
 %setup -n %{n}-%{realversion}
@@ -34,13 +35,16 @@ cmake ../%{n}-%{realversion}/cmake -GNinja \
    -Donnxruntime_USE_EIGEN_THREADPOOL=OFF \
    -Donnxruntime_USE_TENSORRT=OFF \
    -Donnxruntime_CROSS_COMPILING=OFF \
-   -Donnxruntime_USE_FULL_PROTOBUF=OFF \
+   -Donnxruntime_USE_FULL_PROTOBUF=ON \
    -Donnxruntime_DISABLE_CONTRIB_OPS=OFF \
-   -Donnxruntime_BUILD_UNIT_TESTS=OFF
+   -Donnxruntime_BUILD_UNIT_TESTS=OFF \
+   -Donnxruntime_USE_PREINSTALLED_EIGEN=ON \
+   -Deigen_SOURCE_PATH=$EIGEN_ROOT/include/eigen3 \
+   -Donnxruntime_USE_PREINSTALLED_PROTOBUF=ON \
+   -Dprotobuf_INSTALL_PATH=${PROTOBUF_ROOT}
 
 ninja -v %{makeprocesses} -l $(getconf _NPROCESSORS_ONLN)
 
 %install
 cd ../build
 ninja -v %{makeprocesses} -l $(getconf _NPROCESSORS_ONLN) install
-
--- a/onnxruntime.spec
+++ b/onnxruntime.spec
@@ -1,18 +1,18 @@
 ### RPM external onnxruntime 0.4.0
-Source: https://github.com/microsoft/onnxruntime/archive/v%{realversion}.tar.gz
+%define tag %{realversion}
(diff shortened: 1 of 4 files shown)


davidlange6:
please test

(the cmake update is indeed required - that will rebuild lots of things I suspect)

onnxruntime.spec
```diff
@@ -0,0 +1,46 @@
...
+   -Donnxruntime_ENABLE_MICROSOFT_INTERNAL=OFF \
+   -Donnxruntime_USE_BRAINSLICE=OFF \
+   -Donnxruntime_USE_NUPHAR=OFF \
+   -Donnxruntime_USE_EIGEN_THREADPOOL=OFF \
+   -Donnxruntime_USE_TENSORRT=OFF \
+   -Donnxruntime_CROSS_COMPILING=OFF \
+   -Donnxruntime_USE_FULL_PROTOBUF=OFF \
+   -Donnxruntime_DISABLE_CONTRIB_OPS=OFF \
+   -Donnxruntime_BUILD_UNIT_TESTS=OFF \
+   ../cmake
```
slava77: I do not see any specifics for the build architecture.
What does this default to? We still need to restrict to SSE4 (or is it SSE3?)

hqucms: It seems that it supports dynamic architecture -- I built it once on a machine with AVX and it runs fine on another machine without it.

hqucms:
I cannot reproduce the error -- igprof builds fine with the cmake update on my side.

smuzaffar:
please test

smuzaffar:
@hqucms , the build error is this. May be you have unwind installed on the system that is why cmake did not fail.

```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
UNWIND_INCLUDE_DIR
```

smuzaffar:
During PR tests we create symlink for packages which are not changed. In this case we create symlink for unwwind. I think problem is here
https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_11_0_X/gcc700/igprof.spec#L16
which just rsync symlink instead of contents of unwind and atoim_ops. I am looking in to this now.

hqucms:
> During PR tests we create symlink for packages which are not changed. In this case we create symlink for unwwind. I think problem is here
> https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_11_0_X/gcc700/igprof.spec#L16
> which just rsync symlink instead of contents of unwind and atoim_ops. I am looking in to this now.

@smuzaffar Thank you for looking into this! Indeed when I built igprof, unwind was downloaded rather than symlinked, which is probably why it built successfully on my machine.

smuzaffar:
please test

hqucms:
@smuzaffar What's the actual error? I suspect that we need to add `zlib python3` to `BuildRequires` -- they were pointing to system libs in my environment.

smuzaffar:
@hqucms , to see the actual error , you can go to  https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-c31479/1386/summary.html  and the click on the "External Build Logs" link and then go to external -> onnxruntime -> 0.4.0 -> log
Yes looks like( https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-c31479/1386/external/onnxruntime/0.4.0/log ) it did pick up some things from system. 
By the way, does it only build for python3? 

```
-- Found ZLIB: /usr/lib64/libz.so (found version "1.2.7") 
-- Found PNG: /usr/lib64/libpng.so (found version "1.5.13") 
CMake Error at /build/cmsbld/jenkins/workspace/ib-run-pr-tests/testBuildDir/slc7_amd64_gcc700/external/cmake/3.14.5/share/cmake-3.14/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find PythonInterp: Found unsuitable version "2.7.14", but
  required is at least "3.4" (found
  /build/cmsbld/jenkins/workspace/ib-run-pr-tests/testBuildDir/slc7_amd64_gcc700/external/python/2.7.14-pafccj/bin/python)
Call Stack (most recent call first):
  /build/cmsbld/jenkins/workspace/ib-run-pr-tests/testBuildDir/slc7_amd64_gcc700/external/cmake/3.14.5/share/cmake-3.14/Modules/FindPackageHandleStandardArgs.cmake:376 (_FPHSA_FAILURE_MESSAGE)
  /build/cmsbld/jenkins/workspace/ib-run-pr-tests/testBuildDir/slc7_amd64_gcc700/external/cmake/3.14.5/share/cmake-3.14/Modules/FindPythonInterp.cmake:160 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  CMakeLists.txt:205 (find_package)

```

davidlange6:
onnx-runtime only supports python3. That should be fine but it means having python3 in the spec rather than python (ah - but python is not there at all yet)

smuzaffar:
@hqucms , I also have updated the spec to use correct URL for source. We also need to make sure that the eigen, protobuf and other submodules used by onnxruntime are competible with the one used by CMS

davidlange6:
actually it likely is a problem if this is using python behind the scenes in the c(++) api. 

(but btw - can we do the cmake update separately to make these iterations faster?)

hqucms:
@smuzaffar Thank you very much for pointing me to the log files and for updating the spec (the `submodules` option is really nice to have)!

Regarding this:
> We also need to make sure that the eigen, protobuf and other submodules used by onnxruntime are competible with the one used by CMS

protobuf and eigen are not needed at run time (below is the the output of `ldd`). Could there still be compatibility issue then?

```
> ldd libonnxruntime.so 
        linux-vdso.so.1 =>  (0x00007ffd8c7ab000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00007f8999f23000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f8999d07000)
        libstdc++.so.6 => /cvmfs/cms.cern.ch/slc7_amd64_gcc700/external/gcc/7.0.0-pafccj/lib64/libstdc++.so.6 (0x00007f899a81d000)
        libm.so.6 => /lib64/libm.so.6 (0x00007f8999a05000)
        libgcc_s.so.1 => /cvmfs/cms.cern.ch/slc7_amd64_gcc700/external/gcc/7.0.0-pafccj/lib64/libgcc_s.so.1 (0x00007f899a804000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f8999638000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f899a7b0000)
```

davidlange6:
ok, it seems major surgery is needed to match the treatment of externals with the reset of the cmsdist stack

https://github.com/microsoft/onnxruntime/tree/master/cmake/external

hqucms:
> actually it likely is a problem if this is using python behind the scenes in the c(++) api. 

I think it doesn't call python in the C API -- it needs python at compilation time probably just for this:
https://github.com/microsoft/onnxruntime/blob/master/cmake/onnxruntime.cmake#L23.

> (but btw - can we do the cmake update separately to make these iterations faster?)

Good idea! Shall I do it or better leave it for @smuzaffar?

smuzaffar:
https://github.com/cms-sw/cmsdist/pull/5087 should update cmake.

hqucms:
> #5087 should update cmake.

Thanks @smuzaffar! I updated the spec to explicitly use the `python3` executable in the CMS software stack, hopefully it builds fine this time.

mrodozov:
please test
lets see

smuzaffar:
please test
just to have fresh results

smuzaffar:
+externals

smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
+externals

smuzaffar:
@hqucms , this is now available in CMSSW_11_0_X IBs. Note that occxruntime currently does not support PowerPC so that is why it is not available for ppc64le SCRAM ARCH IBs.

hqucms:
@smuzaffar Thank you very much! I will test occxruntime from the IBs and proceed from there.

PR #5085, opened 2019-07-10
fix igprof:properly use cmake to find unwind and pcre instead of copying their headers in igprof

smuzaffar:
- igprof failed to build when `--reference /local/repo` option is used e.g. this is used during PR tests. This PR make use of cmake to find unwind headers and libraries instead of copying them in the igprof
- libunwind is needed at runtime too so make it a `Require` instead of `BuildRequire`


smuzaffar:
please test

PR #5086, opened 2019-07-10
updated libunwind to 1.3.1 and drop libatomic_ops deps

smuzaffar:
- Updated libunwind to version 1.3.1
- Drop linatomic_ops: https://github.com/libunwind/libunwind/blob/master/NEWS#L71
```
** Dont link against libatomic_ops for now.  Due to a packaging bug on
   Debian, linking against this library causes libunwind.so to get
   a dependency on libatomic_ops.so, which is not at all what we want.
   Fortunately, we don't have to link against that library on x86 or
   ia64 since the library is strictly needed only for platforms with
   poor atomic operation support.  Once the libatomic_ops package is fixed,
   we can re-enable linking against libatomic_ops.
```


smuzaffar:
please test

smuzaffar:
+externals

PR #5094, opened 2019-07-16
Adjust root.spec according changes in ROOT

oshadura:

PR #5095, opened 2019-07-18
Revert "Adjust root.spec according changes in ROOT"

davidlange6:
Reverts cms-sw/cmsdist#5094
as the current root.spec and root version in cxxmodules branch are inconsistent.


davidlange6:
fyi @mrodozov - @oshadura and I are trying to fix the IB which is in an inconsistent state...

PR #5102, opened 2019-07-24
added libunwind-toolfile

smuzaffar:
libunwind tool file was missing and we were picking up libunwind.so from igprof. https://github.com/cms-sw/cmsdist/pull/5085 had fixed the igprof spec to not copy libunwind but it broke igprof 
```
cmsRun: error while loading shared libraries: libunwind.so.8: cannot open shared object file: No such file or directory
```


smuzaffar:
please test

smuzaffar:
@ianna , this should fix igprof

smuzaffar:
+externals

PR #5107, opened 2019-07-25
Rivet 3.0.1

mseidel42:
New major Rivet release, needs upgrade of RivetInterface: https://github.com/cms-sw/cmssw/pull/27610


mseidel42:
Oh, that comes unexpected. I would suggest to compile ThePEG without Rivet. There should not be any need for calling Rivet analyses directly from the the Herwig7 configuration, it can all be passed to the RivetInterface.

smuzaffar:
+externals

fabiocos:
+1

PR #5108, opened 2019-07-26
cleanup build rules

smuzaffar:
- drop obsolete capabilities support
- drop root's non-pcm based build support. cmssw based on root6 always build with PCMs


smuzaffar:
please test

PR #5111, opened 2019-07-27
Remove file, migration to ROOT6 was done.

vgvassilev:


smuzaffar:
@vgvassilev , looks like this file is still in use here
https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_11_0_X/gcc700/mctester.spec#L5
if this is not needed then please cleanup mctester.spec too.

vgvassilev:
I do not know if we need it in that case. What does this path exactly patch? Some external library?

vgvassilev:
@smuzaffar, I've removed the use but let's see if it works...

smuzaffar:
please test

smuzaffar:
@vgvassilev , looks like patch is needed to build mctester.

davidlange6:
hi @vgvassilev - i thought mctester-cling.patch is going in the needed direction? (eg, the patch we discussed last week was to do exactly rootcint -> rootling)

vgvassilev:
Yes, I got it now. I misunderstood the initial idea. Closing this one.

PR #5143, opened 2019-08-08
[BuildRules] Use c++ to generate precompiled headers

smuzaffar:
This should fix CommonTools/Utils  unit tests which are failing in CLANG IBs. Reason of failure is that PCH are generated using clang++ while the unit tests are using c++. This PR updates the build rules to always use c++ for generating PCH.


smuzaffar:
please test with cms-sw/cmssw#27723

smuzaffar:
+externals

PR #5150, opened 2019-08-13
Add mkFit as an external

makortel:
This PR adds the [mkFit](http://trackreco.github.io/) as an external package (the CMSSW side will follow).

Pardon my ignorance, but will the contents of a PR to `gcc700` propagate automatically to other compiler branches?

@slava77

### DIFF
--- a/mkfit.spec
+++ b/mkfit.spec
@@ -10,7 +10,7 @@ Requires: tbb
 %setup -q -n %{n}-%{realversion}
 
 %build
-TBB_PREFIX=$TBB_ROOT make
+make TBB_PREFIX=$TBB_ROOT VEC_GCC="-march=core2"
 
 %install
 mkdir %{i}/include %{i}/include/mkFit %{i}/Geoms
--- a/mkfit.spec
+++ b/mkfit.spec
@@ -1,0 +1,20 @@
+### RPM external mkfit 2.0.0
+%define tag V2.0.0-1+pr237
+%define branch devel
+%define github_user trackreco
+
+Source: git+https://github.com/%{github_user}/%{n}.git?obj=%{branch}/%{tag}&export=%{n}-%{realversion}&output=/%{n}-%{realversion}.tgz
+Requires: tbb
+
+%prep
+%setup -q -n %{n}-%{realversion}
+
+%build
+TBB_PREFIX=$TBB_ROOT make
+
+%install
+mkdir %{i}/include %{i}/include/mkFit %{i}/Geoms
+cp -a *.h %{i}/include
+cp -a mkFit/*.h %{i}/include/mkFit
+cp -a Geoms/*.so %{i}/Geoms
+cp -ar lib %{i}/lib
--- a/mkfit-toolfile.spec
+++ b/mkfit-toolfile.spec
@@ -1,0 +1,24 @@
+### RPM external mkfit-toolfile 2.0.0
+Requires: mkfit
+
+%prep
+
+%build
+
+%install
+
+mkdir -p %i/etc/scram.d
+cat << \EOF_TOOLFILE >%i/etc/scram.d/mkfit.xml
+<tool name="mkfit" version="@TOOL_VERSION@">
+  <lib name="MicCore"/>
+  <lib name="MkFit"/>
+  <client>
+    <environment name="MKFITBASE" default="@TOOL_ROOT@"/>
+    <environment name="INCLUDE" default="$MKFITBASE/include"/>
+    <environment name="LIBDIR" default="$MKFITBASE/lib"/>
+  </client>
+  <runtime name="MKFIT_BASE" value="$MKFITBASE"/>
+</tool>
+EOF_TOOLFILE
+
+## IMPORT scram-tools-post


makortel:
@cmsbuild, please test

(am I powerful enough also here?)

mkfit.spec
```diff
@@ -0,0 +1,20 @@
...
+%define github_user trackreco
+
+Source: git+https://github.com/%{github_user}/%{n}.git?obj=%{branch}/%{tag}&export=%{n}-%{realversion}&output=/%{n}-%{realversion}.tgz
+Requires: tbb
+
+%prep
+%setup -q -n %{n}-%{realversion}
+
+%build
+TBB_PREFIX=$TBB_ROOT make
```
slava77: can we make the target compiler and arch here explicit?

davidlange6:
No. That gets defined elsewhere (eg gcc.spec)


On 14 Aug 2019, at 00:22, Slava Krutelyov <notifications@github.com<mailto:notifications@github.com>> wrote:


@slava77 commented on this pull request.

________________________________

In mkfit.spec<https://github.com/cms-sw/cmsdist/pull/5150#discussion_r313637401>:

> @@ -0,0 +1,20 @@
+### RPM external mkfit 2.0.0
+%define tag V2.0.0-1+pr237
+%define branch devel
+%define github_user trackreco
+
+Source: git+https://github.com/%{github_user}/%{n}.git?obj=%{branch}/%{tag}&export=%{n}-%{realversion}&output=/%{n}-%{realversion}.tgz
+Requires: tbb
+
+%prep
+%setup -q -n %{n}-%{realversion}
+
+%build
+TBB_PREFIX=$TBB_ROOT make


can we make the target compiler and arch here explicit?

slava77:
> No. That gets defined elsewhere (eg gcc.spec)

please point me to the line where it's done.
I recall that in some other externals we had to set some flags manually (e.g. in TF and mxnet).

davidlange6:
Maybe you can be more specific. It doesn’t sound like you really mean compiler and arch but rather something else?




On 14 Aug 2019, at 08:55, Slava Krutelyov <notifications@github.com<mailto:notifications@github.com>> wrote:


No. That gets defined elsewhere (eg gcc.spec)

please point me to the line where it's done.
I recall that in some other externals we had to set some flags manually (e.g. in TF and mxnet).

slava77:
I meant compiler (gcc) and a specific target architecture (core2 or sse3 or avx etc)

davidlange6:
the compiler is defined by other spec files.

options are typically figured out by the configure/cmake scripts of external packages (and overridden via command line options passed to those scripts in the spec file when the default behavior is not grid computing compatible)

slava77:
how does the other spec file know about the fanciness of the mkFit/Makefile.config ?
About the arch, it looks like ```-mavx``` is the default.

davidlange6:
no other spec file should need to know about some internal makefile of mkFit. That said, Makefile.config looks pretty unfriendly for maintaining a build recipe from. is there a cmake or configure; make solution (of sorts) under development?

slava77:
> no other spec file should need to know about some internal makefile of mkFit

I do not quite understand what you mean.
My comment was about a line in mkfit.spec "make" call to add more than just ```TBB_PREFIX=$TBB_ROOT``` to the build command.

I'm so far reading your comments that all arguments specifying which make rule/setting to check is defined in the gcc.spec file and at the same time in no other file except mkfit.spec.

davidlange6:
maybe its worth having a look at some other spec files to see the typical interface (eg, root.spec) You'll see the way things like passing in the location of dependencies (gcc, TBB_ROOT) are typically parsed by software packages.

slava77:
> maybe its worth having a look at some other spec files to see the typical interface (eg, root.spec) You'll see the way things like passing in the location of dependencies (gcc, TBB_ROOT) are typically parsed by software packages.

Just to be clear, I'm making comments in the context that this is an external and this PR can not modify the files in the mkFit repository.
This is our standard approach in most of the cases of externals that I'm aware of.

slava77:
for minimal compliance with other externals, we need this built with sse3 or core2 setting, instead of avx.
This can be done by passing a corresponding flag in the mkfit.spec.

davidlange6:
sure - but in that sense mkfit needs to provide a documented and standard build recipe with flexibility needed by its users. It doesn't appear to do that at this point (thus my question of cmake/configure;make sort of build recipes, which are among the most common standards out there these days)

cms can of course hack mkfit by cloning the repository, but that would be particularly silly in this case...

slava77:
Code is the ultimate documentation.
To be more explicit, in this case I was asking to add ```VEC_GCC="-march=core2"``` to the make command line.

slava77:
BTW, the core2/sse3 is by far not the standard, it's a restriction that we have to apply for the choices that CMS made.
The standard recipe is indeed already in mkfit.spec file as implemented.

davidlange6:
> On Aug 14, 2019, at 4:11 PM, Slava Krutelyov <notifications@github.com> wrote:
> 
> BTW, the core2/sse3 is by far not the standard, it's a restriction that we have to apply for the choices that CMS made.
> The standard recipe is indeed already in mkfit.spec file as implemented.

the standard for what? 

but right - cms has to adapt to the resources and code it has.. today its not avx.(there is an r&d project to improve...) 

anyway if adding VEC_GCC="-march=core2" to the make command line does whats needed in the mkfit build rules, then you should add that to mkfit.spec.

makortel:
> anyway if adding VEC_GCC="-march=core2" to the make command line does whats needed in the mkfit build rules, then you should add that to mkfit.spec.

I'm working on that (thanks for pointing it out @slava77), update should come shortly.

makortel:
@cmsbuild, please test

Added `VEC_GCC="-march=core2"` as suggested above.

slava77:
@makortel 
thanks for the update.
I see that the standard tests passed.

@smuzaffar @mrodozov 
please consider merging this soon.
Thank you.

slava77:
@smuzaffar @mrodozov
please clarify on the status of the review of this external.
Thank you.

slava77:
> @smuzaffar @mrodozov
> please clarify on the status of the review of this external.
> Thank you.

ping

mrodozov:
+externals
I can merge this in the evening

PR #5160, opened 2019-08-16
Rivet3 gcc820

mseidel42:
@fabiocos I just built these packages using gcc820 without problem:
```
mseidel@cmsdev23 ~ % ls /build/mseidel/a/BUILD/slc7_amd64_gcc820/external/
herwigpp  herwigpp-toolfile  madgraph5amcatnlo  rivet  rivet-toolfile  thepeg  thepeg-toolfile  yoda
```

Can you start a test here, please, so that I can see where it goes wrong? Otherwise, I need the logs of the failed builds.


mrodozov:
please test

mrodozov:
I know where it was going wrong - didn't have fastjet contrib . The problem however started becoming more complex when the last cmsdist PR I tried didn't have cmssw IB to test against but the one from aug 12, which is without the rivet interface adjustments. Now the latest IB is with the adjustments so I hope this PR will build

fabiocos:
@mrodozov @intrepid42 for my understanding, if the problem is from missing headers why does this appear for gcc8 and gcc9 only, and we haven't broken the main gcc7 IB?

mseidel42:
Because the gcc700 PR https://github.com/cms-sw/cmsdist/pull/5107 had the correct `Requires` and got merged correctly (https://github.com/cms-sw/cmsdist/commit/fe7de339a285d9aa3dfe541f6160a30f710b0778 +66-12) but something went wrong in the gcc820/900 merges, note the different number of added/deleted lines: https://github.com/cms-sw/cmsdist/commit/61aa0ffb7387bbc2e70ea4fa65f399ce6c6d2ca8 https://github.com/cms-sw/cmsdist/commit/e127550b0452f53234e08dd52b2b50e2457c3cb0 (+65-10)

mrodozov:
Yes, the forward port merge failed for gcc8 and gcc9 in which case usually the .spec on gcc700 is not the same as on the branches it tries to auto merge. 
This is when we have patches for those not same specs in this case we had custom patches which are still there. 
The patch for gcc8 is still in use :
https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_11_0_X/gcc820/thepeg-2.1.1-gcc8.patch 

What we usually do is to keep the external as it was when it fails to auto merge which is what I tried but in this case also the cmssw interface changes and this doesn't work . 

And now I have to fix also the gcc9 external :)

fabiocos:
@mrodozov I understand that the comparison is made against the gcc7 build, this would explain the large number of differences. Apart for this it looks ok

mrodozov:
+externals
I agree, lets put it in to fix the IB

fabiocos:
+1

PR #5168, opened 2019-08-26
[FWLITE] Added py2-six deps for fwlite

smuzaffar:
FWLite builds are failing due to missing `six` python package. This PR include the needed python packages for fwlite releases.


smuzaffar:
these files are not used by PR tests so no point testing this. Building locally FWLite with these changes allowed tobuild fwlite release.

PR #5169, opened 2019-08-26
Add mkfit to cmssw-tool-conf

makortel:
Apparently #5150 was not sufficient, but `mkfit-toolfile` needs to be added to `cmssw-tool-conf`. Is anything else needed to get the external usable from CMSSW?

@slava77

### DIFF
--- a/cmssw-tool-conf.spec
+++ b/cmssw-tool-conf.spec
@@ -160,6 +160,7 @@ Requires: python_tools
 Requires: dasgoclient
 Requires: OpenBLAS-toolfile
 Requires: mxnet-predict-toolfile
+Requires: mkfit-toolfile
 
 # Only for Linux platform.
 %if %islinux


makortel:
@cmsbuild, please test

smuzaffar:
@makortel , every new external needs to go in to cmssw-tool-conf in order to get it in the distribution. 
By the way, I see that there is `$MKFIT_BASE/Geoms` directory with these two lib/plugins `CMS-2017.so,  CylCowWLids.so`. Do we need  this path be in LIBDIR or not?

makortel:
> every new external needs to go in to cmssw-tool-conf in order to get it in the distribution.

Right, thanks, I just didn't realize that until now.

> By the way, I see that there is `$MKFIT_BASE/Geoms` directory with these two lib/plugins `CMS-2017.so, CylCowWLids.so`. Do we need this path be in LIBDIR or not?

They are not needed in `LIBDIR`, because one of those `.so`'s is explicitly loaded with `dlopen() `by the `mkfit` external .

smuzaffar:
+externals

mrodozov:
@makortel any suggestion on what to hack since it doesn't build on arm ? I'm expecting the same story for ppc. I'll put it in arch condition for now

PR #5154, opened 2019-08-15
Updated root to tip of branch v6-18-00-patches

mrodozov:

PR #5171, opened 2019-08-26
Updated root to tip of branch v6-18-00-patches

smuzaffar:


smuzaffar:
please test

smuzaffar:
@slava77 (in response to https://github.com/cms-sw/cmsdist/pull/5154#issuecomment-523659677 ) no this error is not common and looks like some side effect of root update. I would like to merge this in to ROOT618 IBs to see what else is broken.  There are 4 reco comparison differences , can you please check if they are real?

slava77:
> There are 4 reco comparison differences , can you please check if they are real?

these are not real
In the comparisons:

all_OldVSNew_TTbar13TeVPU2017wf10224p0 | 4
-- | --

is related to the randomness in the pileup file opening

smuzaffar:
closing in favor of #5189

PR #5174, opened 2019-08-27
MkFit - add in architecture condition

mrodozov:
The external doesn't build for arm, expecting the same for ppc, put it in conditional statement until we have a proper fix


mrodozov:
please test

slava77:
@mrodozov 
in case the mkFit developers can make it work on these archs: which nodes can be used for testing?

@makortel 
@osschar

makortel:
@mrodozov Thanks. Would you happen to have pointers to the build logs on arm (and power) to get a glimpse on the failures? Eventually, as @slava77 wrote, they would need to be addressed by the developers.

I suppose the (upcoming) CMSSW lib and plugin need a similar arch protection as well, right? (as otherwise they couldn't link `mkfit` external)

mrodozov:
please test
this "fixes" arm in terms of "it compiles", for ppc we need a bit more

mrodozov:
> @mrodozov Thanks. Would you happen to have pointers to the build logs on arm (and power) to get a glimpse on the failures? Eventually, as @slava77 wrote, they would need to be addressed by the developers.

We'll think of some service account to give to developes

> I suppose the (upcoming) CMSSW lib and plugin need a similar arch protection as well, right? (as otherwise they couldn't link `mkfit` external)

If the external is not building/running on the non prod archs then yes.

mrodozov:
please test for slc7_ppc64le_gcc700

smuzaffar:
please test for slc7_ppc64le_gcc700

davidlange6:
nothing to do with the build error - but should this not be compiled with some optimization?

mrodozov:
update to use single patch is here https://github.com/mrodozov/cmsdist/tree/adjusting_mkfit 
I'll push it when the ppc tests are finished (not to ruin the relval results, the test is at the end)
@davidlange6 I'd left it to the devs first to propose suggestions

smuzaffar:
closing in favor of #5179 which uses a simple patch.

smuzaffar:
@davidlange6 , about optimization option. It does build with `-O3` . The `g++ -MM -MG -MQ ...` commands without any optimization options are for generating Make dependencies .

PR #5181, opened 2019-08-29
Bump DQMGUI to 9.3.7

schneiml:
This version should make DQMGUI able to understand some old links again.

The problem was reported recently by @mmusich affecting old links to plots with references such as this: https://goo.gl/u1rO51 .

PR #5184, opened 2019-08-30
Bump DQMGUI version to 9.3.8

schneiml:
This is the second version of the fix for broken links with references that where created before labeled references where introduced.

Fix for the fix in #5181.


h4d4:
+1 to f99ecddb53a128ae04eeca09df007dda7a611c96 on slc7_amd64_gcc630.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/5184/test_f99ecdd-20190830-142003.log" target="_blank">here</a>.

PR #5213, opened 2019-09-06
Update mkFit to 2.0.1

makortel:
This PR updates mkFit to 2.0.1, which itself is a code-wise small update to have the same version whose performance will be presented on TRK POG meeting next Monday (Sep 9).

FYI @slava77 @mtosi @JanFSchulte

### DIFF
--- a/mkfit-toolfile.spec
+++ b/mkfit-toolfile.spec
@@ -1,4 +1,4 @@
-### RPM external mkfit-toolfile 2.0.1
+### RPM external mkfit-toolfile 2.0.0
 Requires: mkfit
 
 %prep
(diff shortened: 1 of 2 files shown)


makortel:
@cmsbuild, please test

mkfit-toolfile.spec
```diff
@@ -1,4 +1,4 @@
-### RPM external mkfit-toolfile 2.0.0
+### RPM external mkfit-toolfile 2.0.1
```
smuzaffar: toolfile version is independent of tool itself. We only change it when the tool definition changes (e.g. add a new libdir, include or lib), So this change is not needed.

makortel: Thanks, I restored the version.

smuzaffar:
please test

smuzaffar:
+externals

fabiocos:
+1

PR #5215, opened 2019-09-09
[OpenLoops][ARM64] build loop with -Os optimization

smuzaffar:
Openloops failed[a] to build on ARM64. Looks like we had such issue in past https://sourceware.org/bugzilla/show_bug.cgi?id=18668 
May be size of openloops v2.1.0 is again causing this issue. For now we try to use `-Os` optimization

```
process_obj/ppaajjj/virtual_225_ppaajjj_ddxaaggg_1.os: In function `__ol_vamp_225_ppaajjj_ddxaaggg_1_dp_MOD_vamp_225':
virtual_225_ppaajjj_ddxaaggg_1.f90:(.text+0x170): relocation truncated to fit: R_AARCH64_CALL26 against symbol `malloc@@GLIBC_2.17' defined in .text section in /lib64/libc.so.6
virtual_225_ppaajjj_ddxaaggg_1.f90:(.text+0x244): relocation truncated to fit: R_AARCH64_CALL26 against symbol `memcpy@@GLIBC_2.17' defined in .text section in /lib64/libc.so.6
virtual_225_ppaajjj_ddxaaggg_1.f90:(.text+0x2b4): relocation truncated to fit: R_AARCH64_CALL26 against symbol `memcpy@@GLIBC_2.17' defined in .text section in /lib64/libc.so.6
virtual_225_ppaajjj_ddxaaggg_1.f90:(.text+0x310): relocation truncated to fit: R_AARCH64_CALL26 against symbol `malloc@@GLIBC_2.17' defined in .text section in /lib64/libc.so.6
virtual_225_ppaajjj_ddxaaggg_1.f90:(.text+0x35c): relocation truncated to fit: R_AARCH64_CALL26 against symbol `memcpy@@GLIBC_2.17' defined in .text section in /lib64/libc.so.6
virtual_225_ppaajjj_ddxaaggg_1.f90:(.text+0x3c4): relocation truncated to fit: R_AARCH64_CALL26 against symbol `malloc@@GLIBC_2.17' defined in .text section in /lib64/libc.so.6
virtual_225_ppaajjj_ddxaaggg_1.f90:(.text+0x410): relocation truncated to fit: R_AARCH64_CALL26 against symbol `memcpy@@GLIBC_2.17' defined in .text section in /lib64/libc.so.6
virtual_225_ppaajjj_ddxaaggg_1.f90:(.text+0x44c): relocation truncated to fit: R_AARCH64_CALL26 against symbol `__ol_merging_dp_MOD_ol_merge_tensors' defined in .text section in lib/libopenloops.so
virtual_225_ppaajjj_ddxaaggg_1.f90:(.text+0x4ac): relocation truncated to fit: R_AARCH64_CALL26 against symbol `free@@GLIBC_2.17' defined in .text section in /lib64/libc.so.6
virtual_225_ppaajjj_ddxaaggg_1.f90:(.text+0x4e8): relocation truncated to fit: R_AARCH64_CALL26 against symbol `free@@GLIBC_2.17' defined in .text section in /lib64/libc.so.6
```

### DIFF
--- a/openloops.spec
+++ b/openloops.spec
@@ -18,7 +18,7 @@ fortran_compiler = gfortran
 gfortran_f90_flags = -ffixed-line-length-0 -ffree-line-length-0
 generic_optimisation = -O2
 born_optimisation = -O2
-loop_optimisation = -O0
+loop_optimisation = -Os
 link_optimisation = -O2
 EOF
 export SCONSFLAGS="-j %{compiling_processes}"


smuzaffar:
please test for slc7_aarch64_gcc700

smuzaffar:
FYI @fabiocos @SiewYan

smuzaffar:
@jonaslindert while building openloops 2.1.0 on ARM64 we noticed this build error https://github.com/cms-sw/cmsdist/pull/5215#issue-315533789 . Have you seen it already and is there any fix? I found on google that `-Os`  helped in some cases but I am not sure if this is correct way to avoid this.

smuzaffar:
assign generators
`-Os` takes too much time to compile. I am afraid we are stuck we openloops on ARM64. Would you suggest to drop openloops for ARM64 (just like PowerPC)?

fabiocos:
@qliphy @agrohsje could you please follow up with the authors about this issue?

agrohsje:
Hi @fabiocos, @SiewYan will check with the authors.

SiewYan:
@smuzaffar , thanks for follow up on this issue. Contacting OpenLoops author is underway. From a perspective of user, https://github.com/cms-sw/cmsdist/blob/3fc90a0836ff2b8f364da42555c91b3f20a8aad4/openloops.spec#L26
download ~15 GB of processes collection in proclib folder. In the past openloop 2.0.0 does not have such issue...

jonaslindert:
@smuzaffar This looks very much like an error we know on x64 when the address space is to small for a given library, i.e. when in the "small code model" a library gets bigger than 2 GB. For ppaajjj on x64 this is not the case but we don't have an ARM64 machine to test.

Note: Similar limitations on ARM64 have been discussed in https://sourceware.org/bugzilla/show_bug.cgi?id=18668. They state that ARM64 has more limits than x64, e.g. for the size of a function body. If you hit that, you need to write smaller functions (this is why we split the process code into so many small files).

The solution on x64 to prevent the mentioned problem is to set 

common_flags = -mcmodel=medium

or

common_flags = -mcmodel=large

in the openloops.cfg file and recompile everything. However, again we can't test if this works also works on ARM64. Also, this will come with a slight performance loss. So the question clearly is: do you really need the ppaajjj library? In any case you should avoid installation all available process libraries, as we have been adding various very complex ones recently. 
  
Also note: indeed, -Os optimises for small code size, which in some cases may push you below the limit. So you can also/additionally try "loop_optimisation = -Os" (default is -O0; anything else may take ages and use a lot of RAM during compilation).

Please keep us updated here.

smuzaffar:
@jonaslindert , thanks for looking in to this and your suggestion. As our ARM64 IBs are broken due to this, so we have decided to disabled openloops for ARM64 (just like we have this disabled for PowerPC IBs). This will allow us to get ARM IB working again.

Mean while I will try your suggestion and see if `-mcmodel` flags resolve this issue. If it does then we will re-enable openloops for ARM64.

PR #5219, opened 2019-09-10
python package updates

davidlange6:


davidlange6:
please test

smuzaffar:
@davidlange6 , why not add importlib-metadata in python_tools or add it as py2-pluggy deps?

davidlange6:
i didn't update pluggy yet.. (due to the problems we discussed). lets sort that out tomorrow

davidlange6:
please test

smuzaffar:
@davidlange6 , you need newer setuptool. I am able to build pluggy properly with newer setuptools
```
diff --git a/py2-setuptools.spec b/py2-setuptools.spec
index b3db05b..1e7d3ad 100644
--- a/py2-setuptools.spec
+++ b/py2-setuptools.spec
@@ -1,14 +1,16 @@
-### RPM external py2-setuptools 28.3.0
+### RPM external py2-setuptools 41.2.0
 ## INITENV +PATH PYTHON27PATH %{i}/${PYTHON_LIB_SITE_PACKAGES}
 ## INITENV +PATH PYTHON3PATH %{i}/${PYTHON3_LIB_SITE_PACKAGES}
 
-Source: https://pypi.python.org/packages/6b/dd/a7de8caeeffab76bacf56972b3f090c12e0ae6932245abbce706690a6436/setuptools-28.3.0.tar.gz
+Source: https://github.com/pypa/setuptools/archive/v%{realversion}.tar.gz
+
 Requires: python python3
 
 %prep
 %setup -n setuptools-%{realversion}
 
 %build
+python3 bootstrap.py
 python3 setup.py build
 python setup.py build
 
```

smuzaffar:
@davidlange6 , in order to move to latest versions on these python packages we have to update setuptools. This mean every py2-* package will be rebuilt. We can take this opportunity and update build-with-pip.file is needed. Do you see any thing missing in it which can make pip packaging easier?
One thing I can think of is the patching of sources. For now, if we have to patch it then we do something like this

```
PipPreBuild tar -xzf xgboost-%{realversion}.tar.gz && pushd xgboost-%{realversion} && for pch in %{patches} ; do patch -p1 < ${pch} ; done && popd && rm -f xgboost-%{realversion}.tar.gz && tar czf xgboost-%{realversion}.tar.gz  xgboost-%{realversion}
```

We can do this automatically for packages which define  `PatchN:`

davidlange6:
please test

davidlange6:
hi @smuzaffar - a belated thanks for the detective work... our setuptools is indeed pretty old (maybe its possible to include that in the pip packages without a circular dependency...).

I don't think we should patch anything - at least nothing that I'm aware of is urgent - lets just find a good time to update setuptools. Maybe we can discuss if there is anything else to do in the core meeting (it could be we should use the moment we add the infrastructure for different py2 and py3 versions to update setuptools as well)?

smuzaffar:
@davidlange6 , we already have few packages whcih we patch (due to aarch, ppc build). So addeding patching support (specially when we have to re-build every pip package) does not hurt. We are also going to add support for python relocation so that package files are minimal. @tulamor is already working on it.

packaging setuptools using pip packaging scheme is not trivial.

smuzaffar:
@davidlange6 , agree,  we should discuss this in tomorrow's core sw meeting. I will add it to agenda

davidlange6:
i guess the log harvester differences are now "expected" differences?

smuzaffar:
closing in favor of https://github.com/cms-sw/cmsdist/pull/5240

PR #5249, opened 2019-09-25
Update tag for RecoHGCal-TICL to V00-01-00

smuzaffar:
Move RecoHGCal-TICL data to new tag, see
cms-data/RecoHGCal-TICL#1


smuzaffar:
this has been already tested via cms-external PR, so can be merged directly.

PR #5272, opened 2019-10-10
Add frontier client to py3 path

mrodozov:
Add the %{i}/python/lib to PY3 path to have the py pkg in PY3 IBs.


mrodozov:
please test

PR #5281, opened 2019-10-17
include  py2-pycuda fir not ppc64

smuzaffar:
PowerPC Ibs are failing to due missing cuda. This PR include pycuda only for non ppc64le archs


smuzaffar:
please test

PR #5308, opened 2019-10-28
Update cms-git-tools.spec

smuzaffar:
this includes https://github.com/cms-sw/cms-git-tools/pull/107 which shoudl fix the issue mentioned https://github.com/cms-sw/cms-git-tools/pull/106#issuecomment-546894038


smuzaffar:
I have tested this locally and it works. I am merging it now and will force build an IB so that 23h IB can be a patch release to test the new changes

PR #5309, opened 2019-10-29
Update millepede to version V04-06-00

### LINKED ISSUE #5307: Update millepede version
Dear experts, 
[cc @mmusich and @dmeuser]
We would like to update to a more recent release of millepede than is currently included as an external. As the last time an update was made was several years ago, and the person who took care of it at the time has since left CMS, we're not 100% sure of the procedure. Is all there is to it to update this file: https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_11_0_X/master/millepede.spec with the version of the programme that we want to include? Is there a way for us to test locally that the correct version is being included, or do we fully rely on the CI tests once a PR is made? Any pointers would be appreciated.

mrodozov:
Resolves #5307

### DIFF
--- a/millepede.spec
+++ b/millepede.spec
@@ -1,4 +1,4 @@
-### RPM external millepede V04-03-08
+### RPM external millepede V04-06-00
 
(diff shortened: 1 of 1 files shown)


mrodozov:
please test

mrodozov:
+externals

fabiocos:
+1

PR #5327, opened 2019-11-04
[OpenLoops-2.1.0]Topping up Vtt+jets and Htt+jets libraries processes

SiewYan:
lhc.coll does not contain all P-P collision processes. Topping up Vtt+Jets and Htt+jets loop processes.

### DIFF
--- a/openloops.spec
+++ b/openloops.spec
@@ -24,6 +24,7 @@ EOF
 export SCONSFLAGS="-j %{compiling_processes}"
 ./openloops update --processes generator=0
 ./openloops libinstall lhc.coll
+./openloops libinstall ppatt,ppatt_ew,ppattj,pplltt,pplltt_ew,ppllttj,ppllttj_ew,pplntt,pplnttj,ppwtt,ppwtt_ew,ppwttj,ppwttj_ew,ppztt,ppztt_ew,ppzttj,pphtt,pphtt_ew,pphttj
 
 %install
 mkdir %i/{lib,proclib}


smuzaffar:
In past we had issue ( https://github.com/cms-sw/cmsdist/pull/5215#issuecomment-530008641 ) for aarch64. We need to check if enabling these extra libs will cause the same big library issue.

@SiewYan , is it possible to build multiple small shared libs for openloops?

SiewYan:
@smuzaffar , yes, the LHC.coll is a pre-defined collection of small shared libs, and it should work, if not, I will have to pre-define a txt file with .coll extension specifying each process's library names.

smuzaffar:
please test workflow 534.0,536.0

smuzaffar:
please test workflow 534.0,536.0

smuzaffar:
+externals

fabiocos:
+1

SiewYan:
Hello @smuzaffar , I have a followup question on the Openloops's libraries processes. There is a growing interest in using other libraries processes, one way is to manually include all the libraries processes by defining a txt file and take in as an argument during compilation https://github.com/SiewYan/cmsdist/blob/cac85098983d7c2f0ff15444d71054f60c534deb/openloops.spec#L26 
It is possible to do so in cmsdist, creating a txt file in cmsdist and compile?
Thanks

fabiocos:
@SiewYan do you want to provide the txt as an external input? I can imagine it could be hardcoded into the spec file itself. One question I have is: how big would the proclib repository become? It is already larger than 3 GB

SiewYan:
@fabiocos , yes, to manually defined problib.
It is a lengthy list if one hardcorded it.  
I have installed user.coll defined for backport on 93X [*], it will take ~ 13 GB.

[*] https://github.com/cms-sw/cmsdist/pull/5369

PR #5346, opened 2019-11-12
Update ONNX Runtime to v1.0.0

hqucms:
Based on the [v1.0.0 release](https://github.com/microsoft/onnxruntime/releases/tag/v1.0.0), plus the modification in https://github.com/cms-sw/cmsdist/pull/5259.

v1.0.0 is not compatible with the eigen version we have -- given that eigen is a header-only library without any binaries, there is no runtime dependency on eigen, I think it should be fine to let ONNX Runtime use a different version (shipped as a git submodule). 

@smuzaffar Could you please create a new branch in https://github.com/cms-externals/onnxruntime/?

### DIFF
--- a/onnxruntime.spec
+++ b/onnxruntime.spec
@@ -1,11 +1,12 @@
-### RPM external onnxruntime 0.5.0
-%define tag 2824909ae569932d9aee1462049ff0da1e766989
-%define branch cms/master/9f633c5b
-%define github_user cms-externals
+### RPM external onnxruntime 1.0.0
+%define tag 12f9ce57d0b9cb82611284a435020de9b39e1107
+%define branch cms/v1.0.0
+#%define github_user cms-externals
+%define github_user hqucms
 Source: git+https://github.com/%{github_user}/%{n}.git?obj=%{branch}/%{tag}&export=%{n}-%{realversion}&submodules=1&output=/%{n}-%{realversion}.tgz
 
 BuildRequires: cmake ninja zlib python3
-Requires: eigen protobuf
+Requires: protobuf
 
 
 %prep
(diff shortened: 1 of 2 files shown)


onnxruntime.spec
```diff
@@ -1,11 +1,11 @@
...
-%define branch cms/master/9f633c5b
+### RPM external onnxruntime 1.0.0
+%define tag dfec41d1371a7b1233d25db55561ce4021fc7ceb
+%define branch cms/v1.0.0
 %define github_user cms-externals
 Source: git+https://github.com/%{github_user}/%{n}.git?obj=%{branch}/%{tag}&export=%{n}-%{realversion}&submodules=1&output=/%{n}-%{realversion}.tgz
 
 BuildRequires: cmake ninja zlib python3
-Requires: eigen protobuf
+Requires: protobuf
```
smuzaffar, line 8: @hqucms , yes it is not needed as onnxruntime uses eigen internally i.e. non of the onnxruntime public headers include eigen. So removing it is correct. We need to be careful when onnxruntime public headers starts using eigen. In that case we have to make sure that onnxruntime and cmssw use the same eigen version.

hqucms: @smuzaffar Thank you very much for confirming that!

hqucms:
@smuzaffar I think it would be helpful to have the python package available to CMS users as well, but I have not figured out how to build the python wheel in the CMS environment. Would you mind taking a look, or point me to some examples?

slava77:
this PR passed run time checks combined with cms-sw/cmssw#28112.
I think that the external could be merged now to simplify the signoff of the cmssw PR.

@smuzaffar 
please check and merge if this sounds OK.
Thank you.

smuzaffar:
@slava77 , is it safe to merge the external without cmssw PR? I mean are there any interface changes which could break cmssw?

slava77:
> @slava77 , is it safe to merge the external without cmssw PR? I mean are there any interface changes which could break cmssw?

AFAIK, there is no code using this external except for that PR cms-sw/cmssw#28112

slava77:
@cmsbuild please test
anyways, we can test this by itself to be sure that nothing is broken.

hqucms:
@smuzaffar Could you please merge https://github.com/cms-externals/onnxruntime/pull/3 and update the commit tag? It is to address https://github.com/cms-sw/cmssw/pull/28112#discussion_r345980162.

slava77:
> @smuzaffar Could you please merge [cms-externals/onnxruntime#3](https://github.com/cms-externals/onnxruntime/pull/3) and update the commit tag? It is to address [cms-sw/cmssw#28112 (comment)](https://github.com/cms-sw/cmssw/pull/28112#discussion_r345980162).

I'm not sure about the connection between my comment and the external interface update, I'll try to elaborate in the original place

smuzaffar:
please test

smuzaffar:
+externals

slava77:
similar to the earlier version of this PR, I think that it can be merged first (not coupled with the CMSSW PR).

PR #5383, opened 2019-11-27
[BuildRules] New rules to avoid picking externals libs for release/external/lib path

smuzaffar:


smuzaffar:
+externals 
this looks good, it was suppose to fail

fabiocos:
@smuzaffar is this PR supposed to address issues like those we recently had?

smuzaffar:
yes @fabiocos

PR #5398, opened 2019-12-03
use -mlong-double-64 to avoid link error on ppc64le

smuzaffar:
For `ppc64le`, we use  `-mlong-double-64` ( https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_11_1_X/master/gcc-toolfile.spec#L191 ) to compile cmssw. `fftjet` also uses long double but with 128 bits. That causes link error for `ppc64le` IBs [a]

By default fftjey is build in debug mode (`-g -O2`), we explicitly set CXXFLAGS to `-O2` to avoid debug build.

[a]

```
ld: RecoJets/FFTJetAlgorithms/src/EtaDependentPileup.cc.o uses 64-bit long double, libfftjet.a(libfftjet_la-LinearInterpolator2d.o) uses 128-bit long double
ld: failed to merge target specific data of file libfftjet.a(libfftjet_la-LinearInterpolator2d.o)
```

### DIFF
--- a/fftjet.spec
+++ b/fftjet.spec
@@ -24,7 +24,11 @@ chmod +x ./config.{sub,guess}
 touch pkg-config ; chmod +x pkg-config
 ./configure $PLATF_CONF_OPTS --disable-dependency-tracking --enable-threads \
             --prefix=%i F77="$F77" CXX="$CXX" DEPS_CFLAGS=-I$FFTW3_ROOT/include \
+%ifarch ppc64le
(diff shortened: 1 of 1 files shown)


smuzaffar:
please test for slc7_ppc64_gcc820

smuzaffar:

smuzaffar:
please test

smuzaffar:
+externals

PR #5412, opened 2019-12-06
Hydjet integration as external

wouf:
Hydjet 1.9.1 integration as external. Update pyquen.

PR cmssw: https://github.com/cms-sw/cmssw/pull/30185

### DIFF
--- a/hydjet.spec
+++ b/hydjet.spec
@@ -17,4 +17,4 @@ cmake --build . --clean-first -- %{makeprocesses}
 
 %install
 
-cmake --build . --target install --clean-first -- %{makeprocesses}
+cmake --build . --target install -- %{makeprocesses}
--- a/hydjet.spec
+++ b/hydjet.spec
@@ -1,0 +1,20 @@
+### RPM external hydjet 1.9.1
+
+Source: http://cern.ch/lokhtin/hydro/%{n}-%{realversion}.tar.gz
+
+BuildRequires: cmake
+
+Requires: pyquen pythia6 lhapdf
+
+
+%prep
+%setup -q -n %{n}-%{realversion}
+
+%build
+
+cmake . -DCMAKE_INSTALL_PREFIX=%i -DCMAKE_BUILD_TYPE=Release -DPYQUEN_DIR=${PYQUEN_ROOT} -DPYTHIA6_DIR=${PYTHIA6_ROOT} -DLHAPDF_ROOT_DIR=${LHAPDF_ROOT}
+cmake --build . --clean-first -- %{makeprocesses}
+
+%install
+
+cmake --build . --target install --clean-first -- %{makeprocesses}
--- a/hydjet.spec
+++ b/hydjet.spec
@@ -2,7 +2,7 @@
 
 Source: http://cern.ch/lokhtin/hydro/%{n}-%{realversion}.tar.gz
 
-BuildRequires: cmake
+BuildRequires: cmake gmake
 
 Requires: pyquen pythia6 lhapdf
 
--- a/cmssw-tool-conf.spec
+++ b/cmssw-tool-conf.spec
@@ -45,6 +45,7 @@ Requires: hepmc-toolfile
 Requires: heppdt-toolfile
 Requires: herwig-toolfile
 Requires: herwigpp-toolfile
+Requires: hydjet-toolfile
 Requires: ittnotify-toolfile
 Requires: jemalloc-toolfile
 Requires: jemalloc-debug-toolfile
--- a/hydjet-toolfile.spec
+++ b/hydjet-toolfile.spec
@@ -1,0 +1,23 @@
+### RPM external hydjet-toolfile 1.0
+Requires: hydjet
+%prep
+
+%build
+
+%install
+
+mkdir -p %i/etc/scram.d
+cat << \EOF_TOOLFILE >%i/etc/scram.d/hydjet.xml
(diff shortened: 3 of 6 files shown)


smuzaffar:
assign generators

hydjet.spec
```diff
@@ -0,0 +1,20 @@
+### RPM external hydjet 1.9.1
+
+Source: http://cern.ch/lokhtin/hydro/%{n}-%{realversion}.tar.gz
+
+BuildRequires: cmake
```
smuzaffar: please also add build dependency on `gmake` otherwise it will pick system gmake.

hydjet.spec
```diff
@@ -0,0 +1,20 @@
...
+%setup -q -n %{n}-%{realversion}
+
+%build
+
+cmake . -DCMAKE_INSTALL_PREFIX=%i -DCMAKE_BUILD_TYPE=Release -DPYQUEN_DIR=${PYQUEN_ROOT} -DPYTHIA6_DIR=${PYTHIA6_ROOT} -DLHAPDF_ROOT_DIR=${LHAPDF_ROOT}
+cmake --build . --clean-first -- %{makeprocesses}
+
+%install
+
+cmake --build . --target install --clean-first -- %{makeprocesses}
```
smuzaffar: please remove `--clean-first` here otherwise it will clean and re-build

smuzaffar:
in order to get this in cmssw environment please update `cmssw-tool-conf.spec` and add dependency on `hydjet-toolfile`

hydjet.spec
wouf: Thanks for advice. But, does it matter?

hydjet.spec
wouf: Yes, that was the goal (to clean in case the previous build found). What's wrong? The same behaviour was realized for pyquen.

wouf:
Thanks, I added hydjet-toolfile to cmssw-tool-conf.spec in patch-7

hydjet.spec
smuzaffar: there is no need to do it twice. It has been done in the %build section already

wouf: I see, thanks!

hydjet.spec
smuzaffar: yes it does, some times we end up in env (e.g. docker containers with minimal package installation) where gmake is not available via system

wouf: I mean does it matter gmake, make, or ninja tool is used?

smuzaffar: it depends,  if you have run cmake to use ninja then  you need to add dependency on ninja otheriwse default is `make` (which is available via our gmake.spec) and for that you need to add dependency on gmake.

wouf: Thank You for Your explanation!

smuzaffar:
please test workflow 300.0

fabiocos:
@wouf @hqucms @agrohsje I understand that this PR moves hydjet, currently within CMSSW, outside as an external. I imagine therefore that there should be a corresponding update of https://cmssdt.cern.ch/lxr/source/GeneratorInterface/HydjetInterface/ please clarify.

Is there a validation of the correctness of this backport wrt the original implementation?

wouf:
@fabiocos sure we tested it with the new GeneratorInterface/HydjetInterface/. I was going to update it soon, but I realised that new (as well as existing, please see https://github.com/cms-sw/cmssw/blob/master/GeneratorInterface/HydjetInterface/src/HydjetHadronizer.cc#L195) interface has an issue: we trying to read empty edm::event using BaseHadroniser::getEDMEvent function. To correct it we have to call BaseHadroniser::setEDMEvent somewhere, but I can't to figure out how to access edm::event (from HiMixGEN) in GeneratorInterface. We need to fix it in the new HydjetInterface. Do You have any ideas on this? Example of workflow which is needs it: 

`cmsDriver.py Hydjet_Quenched_MinBias_5020GeV_cfi --conditions auto:phase1_2018_realistic --scenario HeavyIons -n 1 --era Run2_2018 --eventcontent RAWSIM --relval 2000,1 -s GEN,SIM --datatier GEN-SIM --beamspot Realistic25ns13TeVEarly2018Collision --geometry DB:Extended --pileup HiMixGEN --pileup_input file:Hydjet_Quenched_B12_5020GeV_cfi_GEN_SIM.root --fileout step1_GEN_SIM_PU.root --no_exec`

The corresponding issue #28636 was created.

silviodonato:
@wouf do you have any news on this PR?

wouf:
I added the interface but we still working on [issue 28636](https://github.com/cms-sw/cmssw/issues/28636)

qliphy:
Andrey gave a talk at Generator meeting with local tests look good.
https://indico.cern.ch/event/932445/contributions/3935235/attachments/2073197/3480861/HJ1.pdf

qliphy:
+1

mrodozov:
please test
@wouf could you please provide description and documentation (if possible) in the hydjet/doc.txt file like in here:
https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_11_2_X/master/fmt/doc.txt
we want to keep track who added what, why and some basic description of the external.
besides that this one can go in (from externals perspective)

mrodozov:
please test

wouf:
@mrodozov I added for pyquen as well

mrodozov:
please test

qliphy:
+1

silviodonato:
merge

PR #5418, opened 2019-12-12
Revert to CUDA 10.1 Update 1 (10.1.168), add support for Power

fwyzard:
CUDA 10.1 Update 2 and later show a problem with running under MPS (NVIDIA Multi Process Server) or using CUDA Dynamic Parallelism within CMSSW.
While awaiting feedback from NVIDIA, the only solution seems to revert to the latest working version, which was 10.1 Update 1.

Drop the Nsight Compute and Nsigh System tools from the CUDA package, because they are released much more often as external packages.

Add support for CUDA on IBM Power architecture (ppc64le) on Linux.

### DIFF
--- a/cmssw-tool-conf.spec
+++ b/cmssw-tool-conf.spec
@@ -169,13 +169,13 @@ Requires: codechecker-toolfile
 Requires: gcc-checker-plugin-toolfile
 Requires: openldap-toolfile
 Requires: gperftools-toolfile
+Requires: cuda-toolfile
+Requires: cub-toolfile
+Requires: cuda-api-wrappers-toolfile
 
 %if %isnotppc64le_be
 Requires: libunwind-toolfile
 Requires: igprof-toolfile
-Requires: cuda-toolfile
-Requires: cub-toolfile
-Requires: cuda-api-wrappers-toolfile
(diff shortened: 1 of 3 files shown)


fwyzard:
please test

fwyzard:
please test

fwyzard:
@mrodozov @smuzaffar how do I ask to run the tests on Power ?

fwyzard:

fwyzard:

fwyzard:
enable gpu

fwyzard:

fwyzard:
The CUDA part on Power seems to have completed fine ?

smuzaffar:
Yes, cuda part of cmssw build fine for ppc64le

smuzaffar:
+externals
@fwyzard , so cuda builds for all archs (x86_64, aarch64, ppc64le) but cuda gcc combination only works for x86_64 and ppc64le. aarch64 cuda version 10.0 does not support gcc8.

fwyzard:
> aarch64 cuda version 10.0 does not support gcc8.

I know... but there is no newer version officially available, yet :-(

PR #5432, opened 2019-12-16
[OpenLoop-2.1.0] patch up missing lib for 10_6_X 

SiewYan:
continue the Openloop lib sage part2


smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
+externals

openloops.spec
```diff
@@ -23,7 +23,8 @@ link_optimisation = -O2
 EOF
 export SCONSFLAGS="-j %{compiling_processes}"
 ./openloops update --processes generator=0
-./openloops libinstall all.coll
+./openloops libinstall lhc.coll
```
fabiocos, line 26: @SiewYan why this line is still present? This is not in the master version, and it is not strictly needed at this point I would say. Please clarify

SiewYan: Yes, it is not needed. Sorry.

fabiocos: @SiewYan ok, but according to your previous answer it does not harm either, right?

fabiocos:
+1

openloops.spec
SiewYan: yes, it does not though, in the future i will be removed.

PR #5466, opened 2020-01-14
install crab in common; remove crab deps

smuzaffar:
- Install crab in common arwa
- Do not unset cmssw env
- Remove crab deps and let it use pycurl and cjson from cmssw env


smuzaffar:
please test

smuzaffar:
+externals

PR #5471, opened 2020-01-15
Disable AVX2 for pytables

smuzaffar:
PyTables enables avx2 instructions set if built on a machines with avx2 . This causes unit tests to fail if run on  machine without avx2. PR disable AVX2 for PyTables so make sure we can run it on any of our build nodes.

[a]
https://github.com/PyTables/PyTables/blob/master/setup.py#L888


smuzaffar:
please test

smuzaffar:
FYI @silviodonato , this shoudl fix the random unit tests failure we see in IBs and PR tests. Unit tests only fails when the py2-tables external was build on newer hardware (e.g. our cmsbuildXX machines) with AVX2 and run QA job on old hardware (e.g our SLC6 vocms machine) without AVX2

smuzaffar:
+externals

PR #5472, opened 2020-01-16
Fix CUDA device code used in Eigen for compatibility with MPS

fwyzard:
Use hard-coded cache size parameters on CUDA devices, instead of trying to read
the L2 cache size at runtime, which is not supported under CUDA MPS.

Update the upstream repository to https://gitlab.com/libeigen/eigen.git .

### DIFF
--- a/eigen.spec
+++ b/eigen.spec
@@ -1,14 +1,14 @@
-### RPM external eigen e4c107b451c52c9ab2d7b7fa4194ee35332916ec
+### RPM external eigen d812f411c3f9
 ## INITENV +PATH PKG_CONFIG_PATH %{i}/share/pkgconfig
 ## INITENV SETV EIGEN_SOURCE %{source0}
 ## INITENV SETV EIGEN_STRIP_PREFIX %{source_prefix}
 ## NOCOMPILER
-%define tag %{realversion}
+%define tag b20a61c3a0dc9a79790cd258130a99b574662272
 
 # These are needed by Tensorflow sources
 # NOTE: Never apply any patch in the spec file, this way tensorflow gets the exact same sources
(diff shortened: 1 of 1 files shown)


fwyzard:
@cmsbuild, please test

fwyzard:
@cmsbuild, please test

fwyzard:
@cmsbuild, please test

fwyzard:
@cmsbuild, please test

fwyzard:
@silviodonato, please consider this for CMSSW_11_1_X IBs and CMSSW_11_1_0_pre2

eigen.spec
```diff
@@ -1,14 +1,14 @@
-### RPM external eigen e4c107b451c52c9ab2d7b7fa4194ee35332916ec
+### RPM external eigen b20a61c3a0dc
```
smuzaffar: better to use original eigen short hash as real version e..g for this PR it shoudl be `d812f411c3f9` . This shows us which eigen commit we started from.

fwyzard: I don't understand - then what happens when we push more changes to the same branch ?

fwyzard: if we want some reference to the original branching point, I think it would be better to use a date; this way we can easily seen which is the newer (original) version

smuzaffar: then we only update the `tag` ( https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_11_1_X/master/eigen.spec#L6 ) to point to the tip of branch. cmsBuild will then build it with version `<eigen-short-hash>-<tagN>`

fwyzard: done

fwyzard:
@cmsbuild, please test

smuzaffar:
+externals

smuzaffar:
@silviodonato , feel free to merge it for next IB

silviodonato:
+1

PR #5511, opened 2020-01-30
Update libpng to 1.6.37

mrodozov:
Update libpng for what tensorflow 2.1 uses now
https://github.com/tensorflow/tensorflow/blob/r2.1/tensorflow/workspace.bzl#L271


mrodozov:
please test

mrodozov:
+externals
it's described above why we need this update

PR #5512, opened 2020-01-30
Include the CUDA Profiling Tools Interface (CUPTI)

fwyzard:


fwyzard:
@cmsbuild, please test

fwyzard:
@smuzaffar could you include this in the next IBs ?

cuda-toolfile.spec
```diff
@@ -152,6 +152,21 @@ cat << \EOF_TOOLFILE >%{i}/etc/scram.d/cuda-nvrtc.xml
...
+  <lib name="cupti"/>
+</tool>
+EOF_TOOLFILE
+
+cat << \EOF_TOOLFILE >%{i}/etc/scram.d/nvperf.xml
+<tool name="nvperf" version="@TOOL_VERSION@">
+  <info url="https://docs.nvidia.com/cupti/Cupti/index.html"/>
+  <lib name="nvperf_host"/>
+  <lib name="nvperf_target"/>
+</tool>
```
smuzaffar, line 177: @fwyzard  do we need these new tools? A tool with out INCLUDE and LIBDIR directory is of no use. If someone adds a dependency in BuildFile then scram does not know what should be `-L/path` for these libs.

fwyzard: the include and libdir should be the same as cuda; should I add them explicitly ?

smuzaffar: did you check if these libs are linked against `libcuda.so`? if then then just add dependency on `cuda` otherwise I would suggest to add a new `cuda_interface` tool file e.g. something like [a] and depend on it 

[a]
```
cat << \EOF_TOOLFILE >%{i}/etc/scram.d/cuda_intrface.xml
<tool name="cuda_interface" version="@TOOL_VERSION@">
  <info url="https://docs.nvidia.com/cuda/index.html"/>
  <client>
    <environment name="CUDA_INTERFACE_BASE" default="@TOOL_ROOT@"/>
    <environment name="LIBDIR"          default="$CUDA_INTERFACE_BASE/lib64/stubs"/>
    <environment name="INCLUDE"         default="$CUDA_INTERFACE_BASE/include"/>
  </client>
</tool>
EOF_TOOLFILE
```

fwyzard: No, according to the output of `ldd` the new libraries do not depend on other cuda libraries.

For the moment I'll just add the include and lib64 explicitly, we can revise it once we understand a bit better the dependencies.

fwyzard:
@cmsbuild, please test

fwyzard:
@smuzaffar can you merge this ?
I would like to include the fix in the next Patatrack release

smuzaffar:
+externals

fwyzard:
@silviodonato could you approve this for the coming IBs ?

smuzaffar:
merging it as this looks good and it just adds new tools, so should not break any thing.

PR #5520, opened 2020-02-04
Update rivet-toolfile.spec

mseidel42:
For unknown reason, the Rivet paths need to be set more carefully now.

Fixes the following problem:
```
cmsRun: RivetYODA.cc:52: typename T::Ptr Rivet::Wrapper<T>::active() const [with T = YODA::Counter; typename T::Ptr = std::shared_ptr<YODA::Counter>]: Assertion `false && "No activ e pointer set. Was this object booked in init()?"' failed.
```

How to reproduce the problem:
CMSSW_11_0_0_pre7/gcc700 -> `rivet --list-analyses` working
CMSSW_11_0_0_pre7/gcc820 (and newer) -> not working

### DIFF
--- a/rivet-toolfile.spec
+++ b/rivet-toolfile.spec
@@ -16,7 +16,8 @@ cat << \EOF_TOOLFILE >%i/etc/scram.d/rivet.xml
 <environment name="INCLUDE" default="$RIVET_BASE/include"/>
 </client>
 <runtime name="PATH" value="$RIVET_BASE/bin" type="path"/>
-<runtime name="RIVET_ANALYSIS_PATH" value="$RIVET_BASE/lib" type="path"/>
+<runtime name="RIVET_ANALYSIS_PATH" value="$RIVET_BASE/lib/Rivet" type="path"/>
+<runtime name="RIVET_DATA_PATH" value="$RIVET_BASE/share/Rivet" type="path"/>
 <runtime name="PDFPATH" default="$RIVET_BASE/share" type="path"/>
 <runtime name="ROOT_INCLUDE_PATH" value="$INCLUDE" type="path"/>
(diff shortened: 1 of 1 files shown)


mrodozov:
please test

mrodozov:
+externals

smuzaffar:
@intrepid42 , can you please a unit tests to in cmssw to make sure that `rivet --list-analyses` runs. A simple unit test like
```
<test name="test-revet-list" command="rivet --list-analyses"/>
```
in GeneratorInterface/RivetInterface/test/BuildFile.xml  should be enough.

mseidel42:
@smuzaffar Yes, I would be happy to add that. Is it also possible to do a cmsRun on an example config? We would just need to make sure that it does not crash.

mseidel42:
Instead could we check that one of the analyses is actually loaded?
The command would be `rivet --list-analyses | grep MC_GENERIC`.
Can I put that in the xml? How can I test it locally?

smuzaffar:
for bit complex things , it is better to add a shell script and run it e.g
```
<test name="test-revet-list" command="test-revet-list.sh"/>
```
in side the script you can run cmsRun and/or rivet

smuzaffar:
in order to test, just run `scram build runtests`

mseidel42:
Thank you, that works.

It would be nice to test GenParticles2HepMCConverter also.
Do you know if there are any persistent AODSIM or MINIAODSIM files that we can use for that?

smuzaffar:
@intrepid42 I have no idea.
@fabiocos , @silviodonato do you have any idea of any persistent AODSIM or MINIAODSIM?

mseidel42:
Ah never mind, I think I can generate some Pythia events, convert to genParticles, convert back to HepMC, and use that as input for Rivet, to test the whole chain.

PR #5527, opened 2020-02-09
cms-git-tools: use no-renames option with git diff during sparse-checout

smuzaffar:
see https://github.com/cms-sw/cms-git-tools/pull/108


smuzaffar:
this has been tested via cms-sw/cms-git-tools#108

PR #5529, opened 2020-02-10
[XrootD] Get the latest release 4.11.2

mrodozov:
please test
we are not putting any patches on top of xrootd, we better get it from its own source
build requires libuuid now (it complained it needs it I just added it)


mrodozov:
reopen when not so busy

PR #5540, opened 2020-02-12
[11.1.X] Use OpenBlas instead of GSL Blas implementation

### LINKED ISSUE #5528: MXNet slowdown -- OpenBLAS hidden by another BLAS library again?
Similar as https://github.com/cms-sw/cmssw/issues/25230#issuecomment-506422754. MXNet is slow in the latest IB while using `LD_PRELOAD` to load OpenBLAS recovers the speed.

A quick search in cmsdist seems to point to https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_11_1_X/master/photospline.spec#L19-L20 and https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_11_1_X/master/suitesparse.spec#L10. Could we change them to build w/ OpenBLAS?

Actually should we consider removing `lapack`? OpenBLAS contains a full lapack package w/ some of the functions optimized, so it seems that the plain lapack package should not be needed anymore.

smuzaffar:
Resolves https://github.com/cms-sw/cmsdist/pull/5528

Drop unused specs
-  photospline
-  suitesparse

and to make sure that we do not use lapack blas implementation we propose to drop it in favor of openblas.

Move gslcblas library in to separate directory so that no one can link against it.


smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
+externals
unit test failure has nothing to do with this update.
we will merge this after 11.1.0.pre3 is built.

mrodozov:
right. for 10_6 and 2 wont work wit the jenkins job :)

silviodonato:
Backport
10_2_X #5550
10_6_X #5551
11_0_X #5549

PR #5560, opened 2020-02-18
update crab client verions. remove needs for PandaServerInterface

belforte:

PR #5561, opened 2020-02-18
remove dependancy from CRABServer/PandaServiceInterface

belforte:
may make crab-prod and crab-pre fail, but I guss it is the only way to test.
Is there a way to mark this as "do in IB only and do not update master release in standard CVMFS" ?

### DIFF
--- a/crab-build.file
+++ b/crab-build.file
@@ -5,7 +5,7 @@
   %define wmcore_packages PSetTweaks Utils WMCore
 %endif
 %if "%{?crabserver_packages:set}" != "set"
-  %define crabserver_packages PandaServerInterface.py RESTInteractions.py ServerUtilities.py
+  %define crabserver_packages RESTInteractions.py ServerUtilities.py
 %endif
 %if "%{?dbs_packages:set}" != "set"
   %define dbs_packages Client/src/python/dbs PycurlClient/src/python/RestClient


smuzaffar:
please test

smuzaffar:
@belforte , please note that this file `crab-build.file` is used by all 3 flavors of crab i.e `dev, pre and prod`. So is it safe to remove it for all 3? 
I see that you also have opened #5560 to update `crab-dev` , I would suggest to first test this change in `crab-dev` only by adding the following line in `crab-dev.spec` 
```
%define crabserver_packages RESTInteractions.py ServerUtilities.py
```
with the above change you do not need to update `crab-build.file`

belforte:
OK. will do tomorrow. feel free to reject change to crab-build.file

smuzaffar:
-externals
this change should go in when all crab flavors are ready to drop `PandaServerInterface.py` dependency ( See https://github.com/cms-sw/cmsdist/pull/5561#issuecomment-587293336 )

belforte:
@smuzaffar  with ref. to https://github.com/cms-sw/cmsdist/pull/5561#issuecomment-587293336
I see that `crab-dev.spec` has `## IMPORT crab-build` at the bottom,
so should I put the modified  line
```
%define crabserver_packages RESTInteractions.py ServerUtilities.py
```
before, or after that ?

should I assume that the crab-build.spec is imported, then the whole crab-dev.spec parsed and the last instance of `%define  blah` wins ?

smuzaffar:
@belforte , `## IMPORT file` are always included at the end. although it does not matter where you add the `%define` but it is better to just add this line before the `## IMPORT crab-build` line

PR #5531, opened 2020-02-10
[ROOT master] Integrate new PyROOT with CMSSW

mrodozov:
please test
no updates from https://sft.its.cern.ch/jira/browse/ROOT-10545 but lets get the new commits

PR #5596, opened 2020-02-24
[ROOT master] Updated root to tip of branch master

mrodozov:
my update for master is stuck on the pip package (not the pyroot instalation which is in ROOT, but the pip package that depends on it, see https://github.com/cms-sw/cmsdist/pull/5531#issuecomment-590374644) 
once the 1500 today IB is done we can test this


mrodozov:
please test

mrodozov:
looks like the new pyroot is already in then

PR #5603, opened 2020-02-26
Updates for tensorflow on cc8

mrodozov:
Changes taken from the slc7 PR for tensorflow. 
The forward porting of cmsdist master to cc8 is not straightforward, so it had to be done by hand. 
To be tested when the next cc8 IB is available


mrodozov:
please test

mrodozov:
please test for cc8_amd64_gcc820

mrodozov:
test parameters:
- architecture = cc8_amd64_gcc8

mrodozov:
please test

mrodozov:
please test

mrodozov:
@slava77 could you have a look please there are comparison differences

slava77:
@mrodozov 
accodring to this https://cmssdt.cern.ch/SDT/jenkins-artifacts/ib-baseline-tests/CMSSW_11_1_X_2020-02-27-2300/
the baseline is slc7_amd64_gcc820,
compared to this PR arch cc8_amd64_gcc8.
I do not expect bitwise agreement between the two, even if the CMSSW code is the same.

Is there a way to set the baseline to be from the same IB/arch that is used for the (special) tests like this?

mrodozov:
I have to extend my understanding about published baselines, from what I see it's only the production architecture. which explains the differences. 

> Is there a way to set the baseline to be from the same IB/arch that is used for the (special) tests like this?

there is, I think what we did in the past was to produce the baseline by hand for the arch and publish it, but I've never done it myself. we did it at least once for this PR:
https://github.com/cms-externals/onnxruntime/pull/4

mrodozov:
+externals
fixes the cc8 IBs, differences in the comparison are to be expected

PR #5611, opened 2020-02-28
Fix 8_0 slc7 IB

mrodozov:
Currently the 8_0_X IB of slc7 is failing because of curl library that needs to be updated in the 
bootstrap-bundle, and I still don't know how.
adding curl in build require is working hack to stitch it until I find how to fix the lib in the bundle.


mrodozov:
please test for slc7_amd64_gcc530

mrodozov:
+externals

silviodonato:
+1

PR #5617, opened 2020-03-02
sync up with rootmaster branch

davidlange6:
we no longer need the CXX modules specific patches in root, so we can use the same branch/commit of root in the modules branch as in the root master branch. At the same time I've cleaned up some unnecessary differences between these branches in root.spec.

We can use this for the next modules IB (Monday night I think) - I've tested it via a full build locally.

@vgvassilev @mrodozov @oshadura @smuzaffar

### DIFF
--- a/root.spec
+++ b/root.spec
@@ -2,8 +2,8 @@
 ## INITENV +PATH PYTHON27PATH %{i}/lib
 ## INITENV +PATH PYTHON3PATH %{i}/lib
 ## INITENV SET ROOTSYS %{i}
-%define tag c534ce081bfbb3eb7fee2b0a08f2e9c8cb0a7e10
-%define branch cmscxx/master/a8d257c
+%define tag 83199012fe44af9d7a8edfb0329e265fdfbdaf51
+%define branch cms/master/a98c539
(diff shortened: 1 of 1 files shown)


mrodozov:
so we wait for the IB to build, since it has to pick the latest changes for DD4hep fix, and when it's built we test this against it. In short - tomorrow morning ? 
Or merge it since it builds, so the IB will include it ?

davidlange6:
We can do at the same time. It’s already tested...

Cheers,
David


On 2 Mar 2020, at 04:30, Mircho Rodozov <notifications@github.com> wrote:

﻿

so we wait for the IB to build, since it has to pick the latest changes for DD4hep fix, and when it's built we test this against it. In short - tomorrow morning ?

mrodozov:
+externals

mrodozov:
abort
@silviodonato if we test this it will fail because the last modules IB doesn't include a change for DD4Hep, so no point of testing it. The code includes it now (both cmssw and cmsdist) but the IB did not have one part (the cmsdist one)

mrodozov:
get this in the next modules IB

PR #5525, opened 2020-02-06
[TF] Update TF to version 2.1 using externals from cms distribution

smuzaffar:
For building Tensorflow, following externals are now used from cms distribution. We can also use `eigen` from cms externals but currently the `eigen` version we have in cmsdist is bit newer then the one needed by TF (so TF failed to build due to some missing headers). We need to sync these versions so that TF can use our verison of `eigen`
 
 - png
 - jpeg
 - zlib
- curl
 - protobuf
 - pcre
 - gif
 - sqlite3
 - swig
 - cython
 - py2-functools32
 - py2-enum34
 - py2-astor
 - py2-six
 - py2-absl_py
 - py2-termcolor
 - py2-keras_applications
 - py2-pasta
 - py2-wrapt
 - py2-gast
 - py2-backports_weakref
 - py2-opt_einsum

PR #5220, opened 2019-09-10
use -msse3 in mkfit.spec to match gcc-toolfile.spec

slava77:
This is a followup to tests and discussion in https://github.com/cms-sw/cmssw/pull/27895#issuecomment-530019102

@makortel @osschar

PR #5623, opened 2020-03-05
march=core2 -> -msse3 for tensorflow x86_64 to keep old AMDs running

slava77:
based on observations summarized in https://github.com/cms-sw/cmsdist/pull/5525#issuecomment-594976041

Similar to  #5220 and to align with gcc.spec, switch to using msse3


slava77:
@cmsbuild please test

smuzaffar:
+externals

PR #5641, opened 2020-03-13
Add Niels Lohmann's JSON for Modern C++ header-only library

fwyzard:
Add an external and a tool file for  [Niels Lohmann's JSON for Modern C++](https://github.com/nlohmann/json) header-only library, from https://github.com/nlohmann/json .

Can be used in CMSSW by adding to a BuildFile.xml:
```xml
<use name="json"/>
```

and to a source or header file:
```c++
#include <nlohmann/json.hpp>
using json = nlohmann::json;
```

or, to include only the forward-declation:
```c++
#include <nlohmann/json_fwd.hpp>
using json = nlohmann::json;
```

See https://github.com/nlohmann/json/blob/develop/README.md for some documentation.

### DIFF
--- a/cmssw-tool-conf.spec
+++ b/cmssw-tool-conf.spec
@@ -50,6 +50,7 @@ Requires: ittnotify-toolfile
 Requires: jemalloc-toolfile
 Requires: jemalloc-debug-toolfile
 Requires: jimmy-toolfile
+Requires: json-toolfile
 Requires: ktjet-toolfile
 Requires: lhapdf-toolfile
 Requires: libhepml-toolfile
--- a/json-toolfile.spec
+++ b/json-toolfile.spec
@@ -1,0 +1,19 @@
+### RPM external json-toolfile 1.0
+Requires: json
+
+%prep
+
+%build
+
+%install
+mkdir -p %i/etc/scram.d
+cat << \EOF_TOOLFILE > %i/etc/scram.d/json.xml
+<tool name="json" version="@TOOL_VERSION@">
+  <client>
+    <environment name="JSON_BASE" default="@TOOL_ROOT@"/>
+    <environment name="INCLUDE"   default="$JSON_BASE/include"/>
+  </client>
+</tool>
+EOF_TOOLFILE
+
+## IMPORT scram-tools-post
--- a/json.spec
+++ b/json.spec
@@ -1,0 +1,16 @@
+### RPM external json 3.7.3
+## NOCOMPILER
+
+Source: https://github.com/nlohmann/%{n}/releases/download/v%{realversion}/include.zip
+
+%prep
+%setup -c
+
+%build
+
+%install
+mkdir -p %{i}/include/nlohmann
+cp -a include/nlohmann/json_fwd.hpp     %{i}/include/nlohmann/
+cp -a single_include/nlohmann/json.hpp  %{i}/include/nlohmann/
+
+%post


smuzaffar:
@fwyzard , why not RapidJson ( https://github.com/Tencent/rapidjson/ ) which has better performance  https://github.com/miloyip/nativejson-benchmark#conformance-1

fwyzard:
I picked this one because it's what ROOT uses internally: https://github.com/root-project/root/blob/master/io/io/res/json.hpp .

I don't mind a library with better performance, but I'll have to see how to use it.

fwyzard:
For simple cases (like cms-sw/cmssw#29201) rapidjeson might be usable, but the syntax is pretty ugly.

Common part:
```c++
  std::string type = "FedRawDataInputSource";
  std::string label = "source";
  uint32_t events = 1;
  float time_thread = 4.2;
  float time_real = 4.2;
  float allocated = 3065;
  float deallocated = 1532;
```

Here is a simple test case with nlohmann:
```c++
  json j = {
    { "type", type },
    { "label", label },
    { "active", events },
    { "time_thread", time_thread },
    { "time_real", time_real },
    { "mem_alloc", allocated },
    { "mem_feee", deallocated }
  };

  std::cout << std::setw(2) << j << std::endl;
```

Here is the same based on rapidjson's [simplewriter](https://github.com/Tencent/rapidjson/blob/master/example/simplewriter/simplewriter.cpp):
```c++
  StringBuffer j;
  PrettyWriter<StringBuffer> writer(j);
  writer.StartObject();
  writer.Key("type");
  writer.String(type);
  writer.Key("label");
  writer.String(label);
  writer.Key("active");
  writer.Uint(events);
  writer.Key("time_thread");
  writer.Double(time_thread);
  writer.Key("time_real");
  writer.Double(time_real);
  writer.Key("mem_alloc");
  writer.Double(allocated);
  writer.Key("mem_free");
  writer.Double(deallocated);
  writer.EndObject();

  std::cout << j.GetString() << std::endl;
```

And here is something a bit less awkward I managed to put together after reading some more examples:
```c++
  StringBuffer j;
  PrettyWriter<StringBuffer> writer(j);
  Document document(kObjectType);
  auto & allocator = document.GetAllocator();
  document.AddMember("type", type, allocator);
  document.AddMember("label", label, allocator);
  document.AddMember("active", events, allocator);
  document.AddMember("time_thread", time_thread, allocator);
  document.AddMember("time_real", time_real, allocator);
  document.AddMember("mem_alloc", allocated, allocator);
  document.AddMember("mem_feee", deallocated, allocator);
  document.Accept(writer);

  std::cout << j.GetString() << std::endl;
```

I'd say nlohmann's sytnax is much better, but I could agree to use the last one from rapijson if we have serious concerns about speed or memory.

However, we might as well just use `std::stringstream` directly ...
```c++
  std::ostringstream j;
  j << "{\n";
  j << "  { \"type\", " << type << " },\n";
  j << "  { \"label\", " << label << " },\n";
  j << "  { \"active\", " << events << " },\n";
  j << "  { \"time_thread\", " << time_thread << " },\n";
  j << "  { \"time_real\", " << time_real << " },\n";
  j << "  { \"mem_alloc\", " << allocated << " },\n";
  j << "  { \"mem_feee\", " << deallocated << " }\n";
  j << "}";

  std::cout << j.str() << std::endl;
```

smuzaffar:
+externals

silviodonato:
+1

PR #5642, opened 2020-03-13
Remove tfaas from comp repository and spec due to outdated Go TensorFlow stack

vkuznet:
I found that Go TensorFlow stack is broken due to backward incompatible changes in ops module. This affects our ability to move forward with Go compiler version. Therefore I (temporary) remove tfaas from comp repository to allow to update Go compiler and upstream packages, such as DAS, exporters, etc. I'll revisit tfaas spec with newer compiler and once it is ready I'll put it back to comp repository.


vkuznet:
@muhammadimranfarooqi , we need to merge this first. Then I'll create another PR for new Go compiler.

muhammadimranfarooqi:
+1 to fc2a417dfbff9509dca526475c65ee53b46cb86d on slc7_amd64_gcc630.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/5642/test_fc2a417-20200313-194503.log" target="_blank">here</a>.

PR #5648, opened 2020-03-15
Python 3.8.2: updates (git, pcre) and cleanups (qt, pyqt, thriftpy) 

smuzaffar:
Updated python3 version to 3.8.2. As changing python means rebuilding nearly half of externals, so we propose to integrate some cleanup from CC8 branch which includes
- Delete unused specs/files
- Cleanup for geting config.gues and config.sub files
- Use build in PRM macros instead of custom one
- Remove explicit generation of dependencies-setup.*sh files
- Fixing autotools to use included xml instead of looking at system headers
- Make sure to fail is pip can not find a installed package name
- Backports:
  - cmsos changes
  - git version 2.23
  - pcre version 8.43
- Drop Qt and PyQt ( as part of Qt cleanup https://github.com/cms-sw/cmssw/issues/28436 )
- Unify Zlib build for x86_64 and non-x86_64 archs
- Drop thriftpy it has been deprecated and does not build with new python 3.8. Although there is thriftpy2 but as non of our externals require it so we propose to remove it.


smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
please test with cms-sw/pkgtools#157

smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
+externals

smuzaffar:
@silviodonato , any objection on merging this.

smuzaffar:

cmsswdata.spec
```diff
@@ -66,15 +63,12 @@ Requires: data-RecoHGCal-TICL
 Requires: data-SimG4CMS-HGCalTestBeam
 Requires: data-SimPPS-PPSPixelDigiProducer
 
-%if %isnotonline
```
fabiocos, line 69: @smuzaffar out of curiosity, what is the reason of this change? For sure it makes life simpler, as there is no more distinction between what is "online" and the rest, but I wonder why this was in place there earlier at this point

smuzaffar: in past we had architectures like `slc5onl_amd64_gcc*` for special online builds. We are not building any more for those archs, so its time to clean up.

smuzaffar:
https://github.com/cms-sw/cmsdist/pull/5648#issuecomment-601219705 is known build error for `ppc64le` . externals build fine which is good news.

cmsswdata.spec
fabiocos: @smuzaffar ok, so no reason to make it more complex than it needs to be now

silviodonato:
+1

Issue #5627, opened 2020-03-07
LLVM distribution


fwyzard:
While working with the Intel fork of LLVM, I've been trying to use as much as possible the same flags that we use to build LLVM for CMSSW:
https://github.com/cms-sw/cmsdist/blob/1cb8199fb63b14de5de873c56b0bfb5afcdf6fa5/llvm.spec#L40-L50

This `-DLLVM_LIBDIR_SUFFIX:STRING=64` has been an issue, but after discussing with the developers should be getting fixed (intel/llvm#1228, intel/llvm#1261).

This `-DBUILD_SHARED_LIBS:BOOL=ON` has been an issue, after reporting it was fixed (intel/llvm#1259) at least for the LLVM/clang *build*, but not for the *installation* (see intel/llvm#1229).
I was about to start working on the installation part, but reading the LLVM documentation I found a couple of mentions that `BUILD_SHARED_LIBS` should not be used when building a distribution:
   - from [CMake.html](https://llvm.org/docs/CMake.html):
     > **Note**
     > `BUILD_SHARED_LIBS` is only recommended for use by LLVM developers. If you want to build LLVM as a shared library, you should use the `LLVM_BUILD_LLVM_DYLIB` option.

   - from [BuildingADistribution.html](https://llvm.org/docs/BuildingADistribution.html#general-distribution-guidance):
     > **Warning**
     > One very important note: Distributions should never be built using the `BUILD_SHARED_LIBS` CMake option. That option exists for optimizing developer workflow only. Due to design and implementation decisions, LLVM relies on global data which can end up being duplicated across shared libraries resulting in bugs. As such this is not a safe way to distribute LLVM or LLVM-based tools.

   - more from [BuildingADistribution.html](https://llvm.org/docs/BuildingADistribution.html#special-notes-for-library-only-distributions):
     >*Special Notes for Library-only Distributions*
     > One of the most powerful features of LLVM is its library-first design mentality and the way you can compose a wide variety of tools using different portions of LLVM. Even in this situation using `BUILD_SHARED_LIBS` is not supported. If you want to distribute LLVM as a shared library for use in a tool, the recommended method is using `LLVM_BUILD_LLVM_DYLIB`, and you can use `LLVM_DYLIB_COMPONENTS` to configure which LLVM components are part of libLLVM.

At this point, before trying to fix the SYCL installation when `BUILD_SHARED_LIBS` is used, I'm left wondering: why do we use it at all ?

---

[...]

PR #5656, opened 2020-03-17
[LLVM] Test functionality removing static libs

mrodozov:
Following https://github.com/cms-sw/cmsdist/issues/5627 latest comments

### DIFF
--- a/llvm.spec
+++ b/llvm.spec
@@ -79,5 +79,8 @@ rm -f %{i}/bin/FileRadar.scpt %{i}/bin/GetRadarVersion.scpt
 # Avoid dependency on /usr/bin/python, Darwin + Xcode specific
 rm -f %{i}/bin/set-xcode-analyzer
 
+# Remove all static libs except libomptarget-nvptx.a
+find %{i}/lib64 -name "*.a" -and -not -name "libomptarget-nvptx.a" | xargs rm -f
+
 %post
 %{relocateConfig}include/llvm/Config/llvm-config.h
--- a/llvm.spec
+++ b/llvm.spec
@@ -80,7 +80,7 @@ rm -f %{i}/bin/FileRadar.scpt %{i}/bin/GetRadarVersion.scpt
 rm -f %{i}/bin/set-xcode-analyzer
 
 # Remove all static libs except libomptarget-nvptx.a
-find %{i}/lib64 -name "*.a" -and -not -name "libomptarget-nvptx.a" | xargs rm -f
+find %{i}/lib64 -name "*.a" -and -not -name "libomptarget-*.a" -delete
 
 %post
 %{relocateConfig}include/llvm/Config/llvm-config.h
--- a/py2-dxr-llvm901.patch
+++ b/py2-dxr-llvm901.patch
@@ -1,0 +1,14 @@
+diff --git a/dxr/plugins/clang/makefile b/dxr/plugins/clang/makefile
+index 7f8fc29..60e83c3 100644
+--- a/dxr/plugins/clang/makefile
++++ b/dxr/plugins/clang/makefile
+@@ -1,8 +1,4 @@
+-CLANG_LIBS := -lclangFrontendTool -lclangFrontend -lclangDriver \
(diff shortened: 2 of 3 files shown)


mrodozov:
please test
not sure what test I had to run locally so lets just test it to see if we can catch an error

mrodozov:
If I list the lib64 with find it only returns .a static libs.
`  find /cvmfs/cms.cern.ch/slc7_amd64_gcc820/external/llvm/9.0.1//lib64 -name "*.a" -and -not -name "libomptarget-nvptx.a"`
not sure why this .so libs are not there. they should've not been touched

smuzaffar:
idea is that we do not have these small .so files but one big LLVM.so. Looks like dxr makefile needs updates to link against the big library instead of small one.

mrodozov:
please test

mrodozov:
py2-dxr is building, but the codechecker is using the .cmake files in
```
llvm/9.0.1-cms/lib64/cmake/clang
llvm/9.0.1-cms/lib64/cmake/llvm
```
and this files contain the references for all this libs we delete. We better find a way how not to build the libs in first place, from the cmake config. Hopefully this will fix also the non existent reference problem

llvm.spec
```diff
@@ -79,5 +79,8 @@ rm -f %{i}/bin/FileRadar.scpt %{i}/bin/GetRadarVersion.scpt
 # Avoid dependency on /usr/bin/python, Darwin + Xcode specific
 rm -f %{i}/bin/set-xcode-analyzer
 
+# Remove all static libs except libomptarget-nvptx.a
+find %{i}/lib64 -name "*.a" -and -not -name "libomptarget-nvptx.a" | xargs rm -f
```
fwyzard:
```suggestion
find %{i}/lib64 -name "*.a" -and -not -name "libomptarget-*.a" -delete
```

This should be slightly safer:
  - `find -delete` should handle any non-printable characters and race conditions better than `xargs`
  - `libomptarget-*.a` is more future-proof, in case new backends are added

fwyzard:
> the codechecker is using the .cmake files ...

Where/how do we build the code checker ?
`LLVM_LINK_LLVM_DYLIB=ON` should tell the LLVM tools to link with the big shared library; maybe we need a similar flag for the code checker ?

fwyzard:
> We better find a way how not to build the libs in first place, from the cmake config. Hopefully this will fix also the non existent reference problem

I don't know if that is possible.

mrodozov:
> > We better find a way how not to build the libs in first place, from the cmake config. Hopefully this will fix also the non existent reference problem
> 
> I don't know if that is possible.

It is possible to some extent if you see this: 
https://github.com/llvm-mirror/llvm/blob/master/docs/BuildingADistribution.rst
the section "Building distribution"
https://github.com/llvm-mirror/llvm/blob/master/docs/BuildingADistribution.rst#difference-between-install-and-install-distribution
If one specifies the target list in 
LLVM_DISTRIBUTION_COMPONENTS="LLVM-9;clang;clang-cpp" etc 
in which case we try to get all relevant targets (i.e. whatever remains after we | xargs rm -f) 
where I got stuck on which targets should I put for the \*omp\* libs and the nvptx.a lib alone
The other way I'm considering was to patch the LLVMConfig.cmake and ClangConfig.cmake after the installation so that the codechecker doesn't search for target libs that are not there

fwyzard:
> If one specifies the target list in
> LLVM_DISTRIBUTION_COMPONENTS="LLVM-9;clang;clang-cpp" etc
> in which case we try to get all relevant targets

Good point, I missed that.

mrodozov:
listing the targets doesn't change the cmake configs anyhow so I should fix clang/llvm cmake somehow

smuzaffar:
I would say that do not mess with LLVM configuration too much :-) May be better to patch dxr to use shared library.

mrodozov:
I'm closing this, it is for llvm9 and it is the codechecker's cmake needs to be edit/redone 
I'll open it for llvm10 when we merge it in any branch

PR #5662, opened 2020-03-18
Add mxnet python package

mrodozov:
please test

### DIFF
--- a/pip/mxnet.file
+++ b/pip/mxnet.file
@@ -1,0 +1,2 @@
+Requires: py2-requests py2-graphviz py2-numpy
+%define PipDownloadSourceType none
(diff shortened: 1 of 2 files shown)


smuzaffar:
@mrodozov , did you try using INSTALL_PYTHON_VERSIONS to directly build mxnet python ?

mrodozov:
@smuzaffar ok I'll see both

smuzaffar:
@mrodozov , I am fine with using pip package but just make sure that they do not install onnxrumtime libs or dependnecies e.g protobuf etc. If they are just installing python packages then I have no objections using these pip packages.

Can you build them locally and share the list of files they install?

smuzaffar:
@mrodozov , you need to add py2-mxnet in python_tools.spec too

mrodozov:
There is another PR to substitute this one

mrodozov:
I'll check what onnxruntime installs because we might need the pip install there. 
Enabling the flag doesn't build the python part :/

PR #5668, opened 2020-03-18
Add onnxruntime python package

mrodozov:
please test
add also libpng and zlib for onnxruntime, it was picking the system one.
py3-onnxruntime creates a wheel that results in only lib/python3.6/site-packages/ 
looks safe to be used

### DIFF
--- a/pip/requirements.txt
+++ b/pip/requirements.txt
@@ -75,7 +75,6 @@ GitPython==2.1.15 ; python_version<'3.0'
 GitPython==3.0.4 ; python_version>'3.0'
 google-common==0.0.1
 google-pasta==0.1.8
-graphviz==0.13.2
 grpcio==1.25.0
 h5py-cache==1.0
 h5py==2.10.0
@@ -128,7 +127,6 @@ more-itertools==5.0.0 ; python_version<'3.0'
 more-itertools==7.2.0 ; python_version>'3.0'
 mpld3==0.3
 mpmath==1.1.0
-mxnet==1.6.0
 nbconvert==5.6.1
 nbdime==1.1.0
 nbformat==4.4.0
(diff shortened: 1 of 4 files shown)


pip/requirements.txt
```diff
@@ -127,6 +128,7 @@ more-itertools==5.0.0 ; python_version<'3.0'
 more-itertools==7.2.0 ; python_version>'3.0'
 mpld3==0.3
 mpmath==1.1.0
+mxnet==1.6.0
```
smuzaffar: do we need this?

pip/requirements.txt
```diff
@@ -75,6 +75,7 @@ GitPython==2.1.15 ; python_version<'3.0'
 GitPython==3.0.4 ; python_version>'3.0'
 google-common==0.0.1
 google-pasta==0.1.8
+graphviz==0.13.2
```
smuzaffar: and this?

pip/requirements.txt
```diff
@@ -146,6 +148,7 @@ numpy==1.15.1 ; python_version<'3.0' ; platform_machine=='aarch64'
 numpy==1.15.1 ; python_version>'3.0' ; platform_machine=='aarch64'
 oamap==0.12.4
 onnx==1.6.0
+onnxruntime==1.2.0 ; python_version>'3.0'
```
smuzaffar, line 149: shouldn't we also update python_tools.spec (only for non-ppc64 )

pip/requirements.txt
mrodozov: I though I removed this

pip/requirements.txt
mrodozov: this one yes, it complained it needs it

smuzaffar: it might be needed for mxnet, as I do not see this as a dependency in onnxruntime.file

mrodozov: it wasn't , its also remnant from the mxnet pip requirements

smuzaffar:
pleae test

mrodozov:
please test

smuzaffar:
@mrodozov , can you please also add some unit tests for onnxruntime python?

smuzaffar:
+externals

PR #5675, opened 2020-03-20
OnnxRuntime: enable python bindings

smuzaffar:
- Enable python bindings
- Due to bug in onnxruntime build, we need to enable unit tests to get the python build
- Use `lib` instead of lib64 so that shared lib and python bindings live in same path


smuzaffar:
please test with cms-sw/cmssw#29246

mrodozov:
I tried to build it with enabling the flag, it didn't build anything new, that's why I decide to do it with the pip package, did it worked ?

mrodozov:
>     * Enable python bindings
> 
>     * Due to bug in onnxruntime build, we need to enable unit tests to get the python build
> 
>     * Use `lib` instead of lib64 so that shared lib and python bindings live in same path

ah ok then

smuzaffar:
yes it worked on aarch64

mrodozov:
please test with cms-sw/cmssw#29246

smuzaffar:
+externals

PR #5682, opened 2020-03-22
python tool updates

davidlange6:
Additional python package updates - nearly caught up with what is in pypi now.

Not really for this PR - it seems we need to revisit what ends up in python_tools.spec. It doesn't seem we are maintaining the py3-*specs there.  Perhaps this file can be automatically derived from requirements.txt plus the list of exceptions?


davidlange6:
please test

davidlange6:
please test

davidlange6:
please test

smuzaffar:
+externals

PR #5696, opened 2020-03-24
Remove mkFit as an external

fwyzard:
Please re-submit it as a part of CMSSW, following the usual (lengthy and through) integration procedure.


mrodozov:
> Please re-submit it as a part of CMSSW

You are suggesting we make mkFit part of cmssw instead of being external because ... ?

fwyzard:
... because it implements a part of the physics reconstruction software for CMS (track reconstruction), written by members of CMS.

smuzaffar:
@makortel , what do you think? I guess we need a cmssw PR to go with it .... right?

makortel:
Right, a CMSSW PR would be needed.

smuzaffar:
closing it, reopen for latest release cycle along with CMSSW PR

PR #5698, opened 2020-03-26
[ROOT master] Get latest commits to adjust to new libs

mrodozov:
please test
lets see what needs to be adjusted


smuzaffar:
did you try building locally?

mrodozov:
now I'm trying it locally, after seeing how many things they've changed

mrodozov:
please test

mrodozov:
please test

mrodozov:
please test

root.spec
```diff
@@ -56,6 +57,7 @@ cmake ../%{n}-%{realversion} \
   -Dhdfs=OFF \
   -Dqt=OFF \
   -Dtmva=ON \
+  -DPython3_EXECUTABLE="${PYTHON3_ROOT}/bin/python3" \
```
smuzaffar, line 60: why we are building with python3 here while CPYY_BACKEND_LIBRARY points to libcppyy_backend2_7.so?

mrodozov: it uses both installations. the build fails if I don't provide the python3 executable location. reading the new cmake configs it seems the idea is to 'find out which python are available' 2, 3 or both and, use them nevertheless when I used only python2 it complained right away for incomplete cmake config in here:
https://github.com/root-project/root/blob/master/cmake/modules/SearchInstalledSoftware.cmake#L1490
In general I wanted to have a working build and then ask questions 'can we only use python2 if we want to and how'

smuzaffar: ah ok, I thought it was only building for python3. If it is build for both then this looks good. 
Can we build yoda also for both python version?

mrodozov: I changed the CPYY_BACKEND_LIBRARY to libcppyy_backend3_8 
and few other things to try. We can see also if pyroot builds with py3 (which is available for py3 and should work)

smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
closing in favor of https://github.com/cms-sw/cmsdist/pull/5724

PR #5704, opened 2020-03-30
[GCC 8.4] Update GCC to 8.4.0

mrodozov:
the recipe builds (at least)


smuzaffar:
@mrodozov , can you plese check if any of these 
https://github.com/cms-sw/cmsdist/pull/5704/files#diff-a9cd42e7697d89896582fdc58f96b627L18-L22
need updates?

smuzaffar:
try updaing these versions 
https://formulae.brew.sh/formula/gcc@8

smuzaffar:
I see you have updated gmp, can you update the gmp-static.spec too? Pls check if we are using gmp.spec or not ? Do we have any other external spec which should be updated?

smuzaffar:
same for gcc9 branch

mrodozov:
both 8.4 and 9.3 built. three tools can be updated but also require further digging islVersion to 0.22 binutilsVersion to 2.34 and elfutilsVersion to 0.179

gmp.spec
```diff
@@ -1,2 +1,2 @@
-### RPM external gmp 6.1.2
-Source: ftp://mirrors.kernel.org/gnu/%n/%n-%realversion.tar.xz
+### RPM external gmp 6.2.0
```
smuzaffar: do we have any thing using this? I mean any package depends on gmp or all of our packages use gmp-static?

mrodozov: only in this file:
cmssw-queue-override.file
```
%if "%(case %realversion in (*_ICC_X*) echo true ;; (*) echo false ;; esac)" == "true"                                                                                    
%define gitcommit       %(echo %realversion | sed -e 's|_ICC_X|_X|')                                                                                                      
%define scram_compiler  icc                                                                                                                                               
%define extra_tools     mpfr gmp icc-cxxcompiler icc-f77compiler icc-ccompiler                                                                                            
%endif 
```
icc is using it somehow
also there is elfutils.spec not used by anything.

smuzaffar: this `gmp` tool is defined in gmp-static ( https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_11_1_X/master/gmp-static-toolfile.spec ) , so I think no one is using gmp.spec. So better to delete it.

smuzaffar:
please test

smuzaffar:
test parameters
 - full_cmssw = true

smuzaffar:
please test

smuzaffar:
@slava77 , can you please look at these comparison differences https://github.com/cms-sw/cmsdist/pull/5704#issuecomment-607483906 . Note here we are trying to move to a newer version of GCC (a long with few updates from other externals)

slava77:
the differences appear to be compatible with small changes due to numerical precision differences. 
Most differences in the DQM plots are in profile/fit plots; what remains appears to be from some bin migrations.
It's somewhat usual for compiler changes to change order of floating point operations, considering our optimization settings.
So, overall, I did not find anything suspicious in the comparisons.

smuzaffar:
Thanks a lot @slava77 for the analysis.

smuzaffar:
please test
refresh the test results

smuzaffar:
please test
refresh the test results

silviodonato:
+1
I merged https://github.com/cms-sw/cmssw/pull/29738
@smuzaffar  feel free to merge this when you prefer.
Sorry for being so late!

smuzaffar:
thanks @silviodonato . As I wrote to you privately ,  let wait for 23h00 IB reuslts (with cmssw changes only), Tomorrow morning I will merge GCC and force build an IB (I will make sure tag cmsdist before merging gcc).

smuzaffar:
+externals
GCC 8.4 was merged in DEVEL IBs for full scale testing. DEVEL IBs looks good and we are ready to include this for normal IBs now. So that we can get GCC 8.4 in 11.1.0.pre7. A cmsdist tag `IB/CMSSW_11_1_X/master/GCC83` has been created to point to the cmsdist without this update.

PR #5592, opened 2020-02-23
[10.6.X] Update MXNet to 1.5.0

mrodozov:
please test
redo of #5535

PR #5717, opened 2020-04-04
[10.6.X] apply RecoBTag-Combined=V01-03-00 for 106X

slava77:
The initial goal here is to propagate cms-data/RecoBTag-Combined#24 and cms-data/RecoBTag-Combined#26.
However, all changes since 106X, in addition to what is needed to be backported are pure additions https://github.com/cms-data/RecoBTag-Combined/compare/V01-01-04...V01-03-00
A more targeted backport will require making a new branch and out-of-order tag for cms-data/RecoBTag-Combined repo.

So, for 10_6_X, I think that it's more practical to just pick the current master version


slava77:
@cmsbuild please test

slava77:
> * Reco comparison results: 4 differences found in the comparisons
> * DQMHistoTests: Total failures: 20
> 

these are all apparently random, not related to this PR.

Note that either #5592 or #5718 (both merged by now) now should trigger a warning about older version of MXNet file, which is fixed in cms-data/RecoBTag-Combined#24 , which is a part of this PR update.

@smuzaffar please check and sign when you have a chance.
Next step would be to add it to the release.

smuzaffar:
+externals

smuzaffar:
@slava77 , can this be safely merged without https://github.com/cms-sw/cmssw/pull/29062

silviodonato:
+1

PR #5591, opened 2020-02-23
[10.2.X] Update MXNet to 1.5.0

mrodozov:
please test
redo of #5534 which needs also cms-sw. I merged it without reading this detail

PR #5719, opened 2020-04-04
[10.2.X][RecoBTag-Combined]DNN models for the ParticleNet tagger

smuzaffar:
back ported https://github.com/cms-data/RecoBTag-Combined/pull/26 for 10.2.X. Should be tested with https://github.com/cms-sw/cmssw/pull/29063

### DIFF
--- a/data-RecoBTag-Combined.spec
+++ b/data-RecoBTag-Combined.spec
@@ -1,4 +1,4 @@
-### RPM cms data-RecoBTag-Combined V01-01-03
+### RPM cms data-RecoBTag-Combined V01-01-03-01
 
 %prep
 


smuzaffar:
+externals

silviodonato:
@smuzaffar  can this PR be merged independently from cms-sw/cmssw#29063 ?

smuzaffar:
@silviodonato , as this conatians modified files  ( https://github.com/cms-data/RecoBTag-Combined/pull/29/files) so it should be merged with cms-sw/cmssw#29063.

slava77:
the modified files are actually needed to fix the issue which is already in the IBs, it was introduced by #5591
Well, it's also in the files that correspond to taggers that are not running in jenkins tests.

So, I think that this can be merged by itself

PR #5725, opened 2020-04-06
use current cx-oracle

belforte:
No idea why it was locked to 5.1. in https://github.com/cms-sw/cmsdist/commit/276421ac450b4e5d1042cc5d111a899b197c3f22
Maybe it was meant as a temporary thing, currently crabserver would be the only app in cmsweb using py2-cx-oracle51 instead of py2-cx-oracle 
Anyhow 5.1 does not work anymore with sqlplus creating problems in building/changing the DB


muhammadimranfarooqi:
+1 to bc4867f39d3c6f98f65c51453317f556b8d44efc on slc7_amd64_gcc630.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/5725/test_bc4867f-20200406-215503.log" target="_blank">here</a>.

belforte:
merge

PR #5727, opened 2020-04-07
Do not allow to link agaisnt PyROOT library

smuzaffar:
This can bring in a different python version, so do not allow any product to link against it. We can drop the `rootpy.xml` all together.


smuzaffar:
please test

smuzaffar:
+externals

PR #5734, opened 2020-04-09
Include Eigne include directory in ROOT_INCLUDE_PATH

smuzaffar:
As reported here https://github.com/cms-sw/cmssw/issues/29444 , now we have eigen dependency in dataformats and phase2 workflows workflows are failing as root is not able to find eigen headers. This PR propose to add eigen include directory in ROOT_INCLUDE_PATH which is used by root for finding headers at runtime.


smuzaffar:
please test

smuzaffar:
+externals
I am merging it so that it can go in next IB. This should fix the relvals issues we have seen in last IB.

PR #5743, opened 2020-04-14
[ONNX runtime] Update for powerpc

mrodozov:
please test
patch is taken from here https://github.com/onnx/onnx/pull/2482

PR #5744, opened 2020-04-14
[Sonic] Add grpc recipe

mrodozov:
this recipe needs additional work.
the original recipe checks out git submodules and uses
`make `
but we need protobuf, zlib and libunwind from externals. 
problem is that
`-DgRPC_PROTOBUF_PROVIDER=package \`
for example is searching for cmake config in the protobuf install location, which doesn't exist. 
for onnxruntime we have the same problem which is solved like this:
https://github.com/cms-externals/onnxruntime/commit/dfec41d1371a7b1233d25db55561ce4021fc7ceb
same is (probably) true for libunwind. 
I'll try to get opencv and then return here.
@kpedro88 that's on your request.

### DIFF
--- a/grpc-toolfile.spec
+++ b/grpc-toolfile.spec
@@ -46,8 +46,6 @@ cat << \EOF_TOOLFILE >%i/etc/scram.d/grpc.xml
   <lib name="absl_utility"/>
   <lib name="absl_meta"/>
   <lib name="cares"/>
-  <lib name="ssl"/>
-  <lib name="crypto"/>
   <lib name="address_sorting"/>
   <client>
     <environment name="GRPC_BASE" default="@TOOL_ROOT@"/>
@@ -55,6 +53,7 @@ cat << \EOF_TOOLFILE >%i/etc/scram.d/grpc.xml
     <environment name="LIBDIR" default="$GRPC_BASE/lib"/>
   </client>
   <use name="protobuf"/>
+  <use name="openssl"/>
 </tool>
 EOF_TOOLFILE
 
--- a/grpc-toolfile.spec
+++ b/grpc-toolfile.spec
@@ -1,0 +1,61 @@
+### RPM external grpc-toolfile 1.0
+Requires: grpc
+
+%prep
+
+%build
+
+%install
+mkdir -p %i/etc/scram.d
+cat << \EOF_TOOLFILE >%i/etc/scram.d/grpc.xml
+<tool name="grpc" version="@TOOL_VERSION@">
+  <info url="https://github.com/grpc/grpc"/>
+  <lib name="grpc"/>
+  <lib name="grpc++"/>
+  <lib name="grpc++_reflection"/>
+  <lib name="absl_algorithm"/>
+  <lib name="absl_atomic_hook"/>
+  <lib name="absl_bad_optional_access"/>
+  <lib name="absl_base"/>
+  <lib name="absl_base_internal"/>
+  <lib name="absl_bits"/>
+  <lib name="absl_civil_time"/>
+  <lib name="absl_compressed_tuple"/>
+  <lib name="absl_config"/>
+  <lib name="absl_core_headers"/>
+  <lib name="absl_dynamic_annotations"/>
+  <lib name="absl_endian"/>
+  <lib name="absl_errno_saver"/>
+  <lib name="absl_inlined_vector"/>
+  <lib name="absl_inlined_vector_internal"/>
+  <lib name="absl_int128"/>
+  <lib name="absl_log_severity"/>
+  <lib name="absl_memory"/>
+  <lib name="absl_optional"/>
+  <lib name="absl_raw_logging_internal"/>
+  <lib name="absl_span"/>
+  <lib name="absl_spinlock_wait"/>
+  <lib name="absl_str_format"/>
+  <lib name="absl_str_format_internal"/>
+  <lib name="absl_strings"/>
+  <lib name="absl_strings_internal"/>
+  <lib name="absl_throw_delegate"/>
+  <lib name="absl_time"/>
+  <lib name="absl_time_zone"/>
+  <lib name="absl_type_traits"/>
+  <lib name="absl_utility"/>
+  <lib name="absl_meta"/>
+  <lib name="cares"/>
+  <lib name="ssl"/>
+  <lib name="crypto"/>
+  <lib name="address_sorting"/>
+  <client>
+    <environment name="GRPC_BASE" default="@TOOL_ROOT@"/>
+    <environment name="INCLUDE" default="$GRPC_BASE/include"/>
+    <environment name="LIBDIR" default="$GRPC_BASE/lib"/>
+  </client>
+  <use name="protobuf"/>
+</tool>
+EOF_TOOLFILE
+
+## IMPORT scram-tools-post
--- a/grpc-toolfile.spec
+++ b/grpc-toolfile.spec
@@ -7,12 +7,20 @@ Requires: grpc
 
 %install
 mkdir -p %i/etc/scram.d
+
 cat << \EOF_TOOLFILE >%i/etc/scram.d/grpc.xml
 <tool name="grpc" version="@TOOL_VERSION@">
   <info url="https://github.com/grpc/grpc"/>
-  <lib name="grpc"/>
-  <lib name="grpc++"/>
-  <lib name="grpc++_reflection"/>
+  <client>
+    <environment name="GRPC_BASE" default="@TOOL_ROOT@"/>
+    <environment name="INCLUDE" default="$GRPC_BASE/include"/>
+    <environment name="LIBDIR" default="$GRPC_BASE/lib"/>
+  </client>
+</tool>
+EOF_TOOLFILE
+
+cat << \EOF_TOOLFILE >%i/etc/scram.d/grpc-absl.xml
+<tool name="grpc-absl" version="@TOOL_VERSION@">
   <lib name="absl_algorithm"/>
   <lib name="absl_atomic_hook"/>
   <lib name="absl_bad_optional_access"/>
@@ -45,16 +53,23 @@ cat << \EOF_TOOLFILE >%i/etc/scram.d/grpc.xml
   <lib name="absl_type_traits"/>
   <lib name="absl_utility"/>
   <lib name="absl_meta"/>
+  <use name="grpc"/>
+</tool>
+EOF_TOOLFILE
+
+cat << \EOF_TOOLFILE >%i/etc/scram.d/grpc-lib.xml
+<tool name="grpc-lib" version="@TOOL_VERSION@">
+  <lib name="grpc"/>
+  <lib name="grpc++"/>
+  <lib name="grpc++_reflection"/>
   <lib name="cares"/>
   <lib name="address_sorting"/>
-  <client>
-    <environment name="GRPC_BASE" default="@TOOL_ROOT@"/>
-    <environment name="INCLUDE" default="$GRPC_BASE/include"/>
-    <environment name="LIBDIR" default="$GRPC_BASE/lib"/>
-  </client>
+  <use name="grpc"/>
+  <use name="grpc-absl"/>
   <use name="protobuf"/>
   <use name="openssl"/>
   <use name="pcre"/>
+  <use name="grpc-absl"/>
 </tool>
 EOF_TOOLFILE
 
--- a/grpc-toolfile.spec
+++ b/grpc-toolfile.spec
@@ -54,6 +54,7 @@ cat << \EOF_TOOLFILE >%i/etc/scram.d/grpc.xml
   </client>
   <use name="protobuf"/>
   <use name="openssl"/>
+  <use name="pcre"/>
 </tool>
 EOF_TOOLFILE
 
--- a/grpc-toolfile.spec
+++ b/grpc-toolfile.spec
@@ -11,65 +11,17 @@ mkdir -p %i/etc/scram.d
 cat << \EOF_TOOLFILE >%i/etc/scram.d/grpc.xml
 <tool name="grpc" version="@TOOL_VERSION@">
   <info url="https://github.com/grpc/grpc"/>
+  <lib name="grpc"/>
+  <lib name="grpc++"/>
+  <lib name="grpc++_reflection"/>
   <client>
     <environment name="GRPC_BASE" default="@TOOL_ROOT@"/>
     <environment name="INCLUDE" default="$GRPC_BASE/include"/>
     <environment name="LIBDIR" default="$GRPC_BASE/lib"/>
   </client>
-</tool>
-EOF_TOOLFILE
-
-cat << \EOF_TOOLFILE >%i/etc/scram.d/grpc-absl.xml
-<tool name="grpc-absl" version="@TOOL_VERSION@">
-  <lib name="absl_algorithm"/>
-  <lib name="absl_atomic_hook"/>
-  <lib name="absl_bad_optional_access"/>
-  <lib name="absl_base"/>
-  <lib name="absl_base_internal"/>
-  <lib name="absl_bits"/>
-  <lib name="absl_civil_time"/>
-  <lib name="absl_compressed_tuple"/>
-  <lib name="absl_config"/>
-  <lib name="absl_core_headers"/>
-  <lib name="absl_dynamic_annotations"/>
-  <lib name="absl_endian"/>
-  <lib name="absl_errno_saver"/>
-  <lib name="absl_inlined_vector"/>
-  <lib name="absl_inlined_vector_internal"/>
-  <lib name="absl_int128"/>
-  <lib name="absl_log_severity"/>
-  <lib name="absl_memory"/>
-  <lib name="absl_optional"/>
-  <lib name="absl_raw_logging_internal"/>
-  <lib name="absl_span"/>
-  <lib name="absl_spinlock_wait"/>
-  <lib name="absl_str_format"/>
-  <lib name="absl_str_format_internal"/>
-  <lib name="absl_strings"/>
-  <lib name="absl_strings_internal"/>
-  <lib name="absl_throw_delegate"/>
-  <lib name="absl_time"/>
-  <lib name="absl_time_zone"/>
-  <lib name="absl_type_traits"/>
-  <lib name="absl_utility"/>
-  <lib name="absl_meta"/>
-  <use name="grpc"/>
-</tool>
-EOF_TOOLFILE
-
-cat << \EOF_TOOLFILE >%i/etc/scram.d/grpc-lib.xml
-<tool name="grpc-lib" version="@TOOL_VERSION@">
-  <lib name="grpc"/>
-  <lib name="grpc++"/>
-  <lib name="grpc++_reflection"/>
-  <lib name="cares"/>
-  <lib name="address_sorting"/>
-  <use name="grpc"/>
-  <use name="grpc-absl"/>
   <use name="protobuf"/>
   <use name="openssl"/>
   <use name="pcre"/>
-  <use name="grpc-absl"/>
 </tool>
 EOF_TOOLFILE
 
--- a/grpc.spec
+++ b/grpc.spec
@@ -21,6 +21,7 @@ cmake ../%{n}-%{realversion} \
     -DgRPC_INSTALL=ON \
     -DCMAKE_BUILD_TYPE=Release \
     -DBUILD_SHARED_LIBS=ON \
+    -DCMAKE_INSTALL_LIBDIR=lib \
     -DgRPC_ABSL_PROVIDER=module \
     -DgRPC_CARES_PROVIDER=module \
     -DgRPC_PROTOBUF_PROVIDER=package \
@@ -28,7 +29,7 @@ cmake ../%{n}-%{realversion} \
     -DgRPC_ZLIB_PROVIDER=package \
     -DZLIB_ROOT=${ZLIB_ROOT} \
     -DCMAKE_INSTALL_PREFIX=%{i} \
-    -DCMAKE_PREFIX_PATH="${LIBUNWIND_ROOT};${PROTOBUF_ROOT};${ZLIB_ROOT};${OPENSSL_ROOT}"
+    -DCMAKE_PREFIX_PATH="${PCRE_ROOT};${PROTOBUF_ROOT};${ZLIB_ROOT};${OPENSSL_ROOT}"
 
 ninja -v %{makeprocesses} -l $(getconf _NPROCESSORS_ONLN)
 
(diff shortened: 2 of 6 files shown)


kpedro88:
@mrodozov I have this working now: https://github.com/kpedro88/cmsdist/tree/add_grpc_as_external

here's the saga for future reference:
1. tried to change protobuf spec to install using cmake, in hopes it would preserve the cmake config for later use. That attempt is in https://github.com/kpedro88/cmsdist/tree/protobuf_cmake; mostly worked, but for some reason `make install` failed, and it seems like protobuf primarily maintains their cmake setup for MSVC, so abandoned this path.
2. manually specified cmake definitions needed to get zlib and protobuf. this is kind of brute force, but a little easier to maintain than hacking the cmake files of every external that needs protobuf.
3. encountered this weird issue: https://github.com/grpc/grpc/issues/20909 even though /usr/include/linux/ptrace.h on my system definitely defines `PTRACE_O_EXITKILL`. https://github.com/grpc/grpc/issues/20909#issuecomment-585170200 had a workaround using `-DBUILD_SHARED_LIBS=ON`, which succeeded for me.

hopefully this result will make the Triton setup a bit easier... I may try to tackle that tomorrow.

mrodozov:
We changed the protobuf config to cmake to fix onnxruntime as well, see 
https://github.com/cms-sw/cmsdist/pull/5743
I was hoping it will also help for grpc 
I'll try your recipe using ninja, with the new protobuf and update the recipe if it goes well

kpedro88:
@mrodozov excellent, then I can simplify my branch. I tested with the updated protobuf.spec from #5743 and removed the manual cmake args for protobuf, and it still works correctly. https://github.com/kpedro88/cmsdist/tree/add_grpc_as_external is updated accordingly.

mrodozov:
it found protobuf:
`Found Protobuf: /build/mrodozov/Sonic_externals/build_grpc/slc7_amd64_gcc820/external/protobuf/3.11.3/lib/libprotobuf.so;-lpthread (found version "3.11.3")`
with the updated config, libunwind now

kpedro88:
right now it says:
> -- Checking for module 'libunwind-generic'
> --   No package 'libunwind-generic' found
> libunwind not found. Disabling unwind tests.

if it's only used for tests, is this a problem?

mrodozov:
I'm installing go, because that's where it fails with error

kpedro88:
@mrodozov the install seems to complete when I run it. What error do you see?

mrodozov:
it breaks complaining it doesn't have Go. fails with 
```
--   No package 'libunwind-generic' found
libunwind not found. Disabling unwind tests.
CMake Error at third_party/boringssl/CMakeLists.txt:51 (message):
  Could not find Go
```
can you paste the the details of how you build it , I took
https://github.com/kpedro88/cmsdist/tree/add_grpc_as_external
and just updated the protobuf spec
Maybe your system has Go

kpedro88:
That is exactly how I build it, and I don't see any message about Go (at all) in the log.

You're right, for some reason I have it here:
> which go
> /usr/kerberos/bin/go

smuzaffar:
by the way we now have go for amd64 in cmsdist. If you need to use go to build then add `BuildRequires: go` 

Note, it is only availble for amd64. @mrodozov , please check if there are aarch64 and ppc64le distribution of go available. If yes then please update https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_11_1_X/master/go.spec

kpedro88:
update: I'm now trying to install triton-inference-server on top of gRPC. It looks like it needs to link to static libraries from abseil-cpp, which is included in grpc as a third_party dependency, and therefore not distributed by itself. So we probably need a spec file for that one (and maybe the other two: boringssl and cares). I'll try to work more on that later...

kpedro88:
So the ninja install actually copies the absl static libraries into the lib folder:
```
-- Installing: /data/pedrok/sonic/CMSSW_11_1_0_pre5/build/tmp/BUILDROOT/9c1ea7238130191342694691f51e395a/opt/cmssw/slc7_amd64_gcc820/external/grpc/1.27.3/lib/libabsl_bad_optional_access.a
```
but then they don't appear in the final rpm. Is there a way to keep them?

smuzaffar:
@kpedro88 , you need to add
```
%define keep_archives true
```
in the spec e.g. https://github.com/cms-sw/cmsdist/search?q=keep_archives&unscoped_q=keep_archives

kpedro88:
I now have https://github.com/cms-sw/cmsdist/compare/IB/CMSSW_11_1_X/master...kpedro88:add_triton_externals which succeeds in installing triton using all the needed dependencies. Not the most elegant, but it works. (This branch includes several of the open cmsdist PRs.)

@mrodozov where should Go be added in grpc.spec?

mrodozov:
it's like this:
`BuildRequires: cmake ninja go`
It builds with this changes :)

smuzaffar:
@mrodozov , did you update go.spec for aarch64 and ppc64le?

mrodozov:
There are, yes,
https://golang.org/doc/install?download=go1.13.8.linux-ppc64le.tar.gz
https://golang.org/doc/install?download=go1.13.8.linux-amd64.tar.gz
I'm adding conditions for the other archs

smuzaffar:
are these part of our go.spec or not?

mrodozov:
Not yet, I've added a condition to change the source for every arch

pip/requirements.txt
```diff
@@ -79,6 +79,7 @@ GitPython==3.1.0 ; python_version>'3.0'
 google-common==0.0.1
 google-pasta==0.2.0
 grpcio==1.27.2
+grpcio-tools==1.27.2
```
kpedro88: should we update this version (and the above grpcio version) to be consistent with 1.28.1?

mrodozov:
please test

mrodozov:
please test

grpc-toolfile.spec
```diff
@@ -0,0 +1,61 @@
...
+  <lib name="absl_strings_internal"/>
+  <lib name="absl_throw_delegate"/>
+  <lib name="absl_time"/>
+  <lib name="absl_time_zone"/>
+  <lib name="absl_type_traits"/>
+  <lib name="absl_utility"/>
+  <lib name="absl_meta"/>
+  <lib name="cares"/>
+  <lib name="ssl"/>
+  <lib name="crypto"/>
```
smuzaffar: why `ssl` and `crypto` libs here? these are part of openssl? if grpc is generating its own ssl and crypto libs then we should  see how to build it using external oepnssl.

grpc.spec
```diff
@@ -0,0 +1,37 @@
...
+
+cmake ../%{n}-%{realversion} \
+    -G Ninja \
+    -DgRPC_INSTALL=ON \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DBUILD_SHARED_LIBS=ON \
+    -DgRPC_ABSL_PROVIDER=module \
+    -DgRPC_CARES_PROVIDER=module \
+    -DgRPC_PROTOBUF_PROVIDER=package \
+    -DgRPC_SSL_PROVIDER=module \
```
smuzaffar: please use package for ssl and add dependency on openssl.

mrodozov:
please test

kpedro88:
I've updated my working branch with a tensorflow-serving.spec (depends on grpc): https://github.com/cms-sw/cmsdist/commit/f363d1dafb98ac68a475b4c811697cfbfb9c7b5f

The build doesn't quite finish because:
```
/lib64/libstdc++.so.6: version `CXXABI_1.3.9' not found (required by /data/pedrok/sonic/CMSSW_11_1_0_pre5/build/BUILD/slc7_amd64_gcc820/external/tensorflow-serving/2.1.0/build/9d6bcae215bc84787d86b62f7e8781b8/execroot/tf_serving/bazel-out/host/bin/external/swig/swig)
```

However, the correct version of swig from pkgtools is `/data/pedrok/sonic/CMSSW_11_1_0_pre5/build/slc7_amd64_gcc820/external/swig/3.0.10-pafccj/bin/swig`. It links to /data/pedrok/sonic/CMSSW_11_1_0_pre5/build/slc7_amd64_gcc820/external/gcc/8.2.0-pafccj/lib64/libstdc++.so.6 rather than the system lib. I don't know how to make bazel use the desired version of swig. Has this been encountered before?

smuzaffar:
Check our tenserflow spec. I remember we had to patch it so that bazel use our swig

kpedro88:
I included `--action_env=SWIG_LIB=${SWIG_LIB}` in the `BAZEL_OPTS` in my spec file.

I see swig is also included in `TF_SYSTEM_LIBS`, but tensorflow-serving doesn't recognize that variable.

kpedro88:
@smuzaffar @mrodozov let me know if you have more thoughts on tensorflow-serving or the other externals
it would be great to start getting things merged

mrodozov:
@kpedro88  maybe you want to share with us your latest tensorflow-serving stuff, so we can have look at it. Push it somewhere I'll make another PR with it. The PRs for opencv and grpc can go in and then we continue

mrodozov:
never mind I found it

kpedro88:
I've decided I dislike Bazel so much that I'd rather fix up the custom cmake installer. Points in favor:
* Bazel build takes 2+ hrs, while cmake build takes a couple minutes (we just need one shared library and some headers)
* As noted before, tensorflow/serving doesn't recognize `TF_SYSTEM_LIBS`; what I now realize is that all the dependencies in https://github.com/tensorflow/serving/blob/master/tensorflow_serving/workspace.bzl would need to be manually redirected
  * and a lot of those dependencies aren't necessary for the products we need anyway

So I've made a separate set of spec files ("tfserving" rather than "tensorflow-serving"): https://github.com/cms-sw/cmsdist/commit/b6263b11ae0bc139dfb550e2a9804c0a08d78ab6

This seems to build fine. Please let me know if this strategy is acceptable as a starting point.

davidlange6:
This doesn’t look like something I’d want to commit to supporting in production..

On May 11, 2020, at 2:18 AM, Kevin Pedro <notifications@github.com<mailto:notifications@github.com>> wrote:



I've decided I dislike Bazel so much that I'd rather fix up the custom cmake installer. Points in favor:

  *   Bazel build takes 2+ hrs, while cmake build takes a couple minutes (we just need one shared library and some headers)
  *   As noted before, tensorflow/serving doesn't recognize TF_SYSTEM_LIBS; what I now realize is that all the dependencies in https://github.com/tensorflow/serving/blob/master/tensorflow_serving/workspace.bzl would need to be manually redirected
     *   and a lot of those dependencies aren't necessary for the products we need anyway

So I've made a separate set of spec files ("tfserving" rather than "tensorflow-serving"): b6263b1<https://github.com/cms-sw/cmsdist/commit/b6263b11ae0bc139dfb550e2a9804c0a08d78ab6>

This seems to build fine. Please let me know if this strategy is acceptable as a starting point.

kpedro88:
@davidlange6 it's not my first choice, either. But if we're going to have to hack into the build scripts for tensorflow/serving anyway, it's a lot easier for me to use a tool that I actually understand. Either way, we incur some maintenance burden.

The inception_cmake repository really just contains CMakeLists.txt and a few "Find*.cmake" files, about 400 lines. (By distributing tensorflow .proto files in our external, I've now eliminated the need to clone the tensorflow repo in the serving install: https://github.com/cms-sw/cmsdist/commit/915d7effd883c79afc6d260881497537b80db8ee.)

kpedro88:
ping @smuzaffar @mrodozov is there anything holding up the integration of this PR and #5747? It would be good to move on to reviewing the triton spec files

mrodozov:
If @smuzaffar agrees we can get them in.

smuzaffar:
please test
lets renew the tests so that I can also look in to the log files

grpc-toolfile.spec
```diff
@@ -0,0 +1,60 @@
...
+  <lib name="absl_strings"/>
+  <lib name="absl_strings_internal"/>
+  <lib name="absl_throw_delegate"/>
+  <lib name="absl_time"/>
+  <lib name="absl_time_zone"/>
+  <lib name="absl_type_traits"/>
+  <lib name="absl_utility"/>
+  <lib name="absl_meta"/>
+  <lib name="cares"/>
+  <lib name="address_sorting"/>
```
smuzaffar: @mrodozov , we do not need these non grpc libs in here. we should only have grpc libs defined here (they are linked to these linabsl_* libs and dynamic loader will take care of loading abls when needed).

@kpedro88 , there are few libgrps* libs. how are you planing to use them [a]. Should they all be in one tool file? If you know that you only need `grpc++` or `grpc` then we can start with those.  

[a]
libgrpc++_alts.so  libgrpc++_error_details.so  libgrpcpp_channelz.so    libgrpc.so    libgrpc_unsecure.so
libgrpc_cronet.so  libgrpc_plugin_support.so   libgrpc++_reflection.so  libgrpc++.so  libgrpc++_unsecure.so

grpc-toolfile.spec
```diff
@@ -0,0 +1,60 @@
...
+  <client>
+    <environment name="GRPC_BASE" default="@TOOL_ROOT@"/>
+    <environment name="INCLUDE" default="$GRPC_BASE/include"/>
+    <environment name="LIBDIR" default="$GRPC_BASE/lib"/>
+  </client>
+  <use name="protobuf"/>
+  <use name="openssl"/>
```
smuzaffar, line 23: @mrodozov , looks like grpc also link against pcre [a], so please add 'Requires: pcre` in grpc.spec and add here `<use name="pcre"/>`

[a]
>ldd libgrpc.so | grep pcre
    libpcre.so.1 => /lib64/libpcre.so.1 (0x00007fab25bc3000)

grpc.spec
```diff
@@ -0,0 +1,37 @@
+### RPM external grpc 1.28.1
+
+Source: git+https://github.com/grpc/grpc.git?obj=master/v%{realversion}&export=%{n}-%{realversion}&submodules=1&output=/%{n}-%{realversion}.tgz
+
+BuildRequires: cmake ninja go
+Requires: protobuf zlib libunwind openssl
```
smuzaffar: why libunwind?

grpc-toolfile.spec
kpedro88: @smuzaffar we need all of these libraries (from abseil and grpc) for downstream products such as Triton.

We can try to make a separate spec for abseil - I mentioned this earlier in the thread, but did not get a response.

grpc.spec
kpedro88: I think it's used in a test that grpc runs during the build process

grpc-toolfile.spec
smuzaffar: having these libs in grpc distribution is fine but I think they should not be part of grpc tool file. 
@mrodozov , please create a separate tool file.

kpedro88: @smuzaffar that is much easier, thanks for the suggestion
is it fine to keep them in the same install area (GRPC_BASE)?

grpc.spec
smuzaffar: Can you please re-check this dependency? I built grpc without libunwind (and libunwind was not installed on system) and it did not fail.

kpedro88: Indeed, the build does not fail without it. I had remarked on this previously: https://github.com/cms-sw/cmsdist/pull/5744#issuecomment-618684272

smuzaffar: I do not see any such message in my build log.  `grep unwind w/BUILD/slc7_amd64_gcc820/external/grpc/1.28.1/log` returns nothing.

smuzaffar: I think it was needed when you were building the build-in ssl. As we now use cms openssl, so no need to add this dep.

grpc-toolfile.spec
mrodozov: I separated only absl into different toolfile. I left the cares and address_sorting in . Or should I separate them ? And can we keep the install area as suggested ?

grpc-toolfile.spec
kpedro88: I tried adding `Requires: pcre` in grpc.spec and I get:
```
ERROR! File not found: /data/pedrok/sonic/CMSSW_11_1_0_pre5/cmsdist/prce.spec
```

kpedro88: Ah, typo, never mind.

grpc-toolfile.spec
smuzaffar: @kpedro88 , yes it is find to keep them in the grpc/lib
@mrodozov , all these absl, c-ares and address_sorting are externals which grpc build internal. I will suggest to move all of them out  in separate toolfiles e.g. abseil-cpp.xml, c-ares.xml, address_sorting.xml

smuzaffar: checking the build area again, I am not sure if we need toolfile for abseil or address_sorting? The headers files of these tools are not exported (copied to grpc/include. So @kpedro88 I do not understand how you are going to use absl libs? Looks like these are internall used by grpc (and grpc libs are linked to them) , so there is no need to have them in toolfile.

kpedro88: When I build Triton, it fails unless the absl libs are kept.

kpedro88: The underlying problem is that `build/slc7_amd64_gcc820/external/grpc/1.28.1-cms/lib/cmake/grpc/gRPCTargets.cmake` includes the absl libs as targets (as well as cares and address_sorting)

smuzaffar: this is fine, we will keep the libs in distribution (we have to as grpc libs are linked to them). I am only suggesting to drop them from scram toolfile.

kpedro88: Okay, I'll run the Triton installation that way and make sure it works.

kpedro88: @smuzaffar @mrodozov the Triton installation works, so we can simplify by removing third party libs from the grpc toolfile.

mrodozov:
please test

smuzaffar:
@mrodozov , I just realized that we do not have openssl for CC8. Can you try building `grpc` for `cc8_amd64_gcc8`. Either keep openssl conditioanlly for `slc7` or remove openssl dependency and let grpc pick it up from system (e.g. see python, curl and many other specs in cc8 branch). For now just make sure that it builds for cc8_amd64_gcc8  too.

kpedro88:
> we do not have openssl for CC8

Really? This seems like a bigger problem...

smuzaffar:
no it is not a problem. For CC8, we decided to use openssl from system instead of we build and distribute it. CC8 boost strap requires to have `openssl openssl-devel openssl-libs` installed.  So every thing should be fine (all our specs for cc8 branch use ssl from system already). We just need to update grpc spce to only use our  openssl for slc7

kpedro88:
Ah, I see. Okay, that should not hurt anything.

mrodozov:
please test for cc8_amd64_gcc8

mrodozov:
test parameters:
- arch = cc8_amd64_gcc8
- release = CMSSW_11_1_X_2020-05-17-2300

mrodozov:
please test

mrodozov:
abort

mrodozov:
please test

mrodozov:
please test

smuzaffar:
@mrodozov , did you test it for cc8?

mrodozov:
Yes I did, it builds.

smuzaffar:
PR does not build on cc8 in its current state , so how have you built it? please share receipe. for me `%{centos}` not working

mrodozov:
I have a cmsdist cc8 folder on cmsdev22, in 
`/build/mrodozov/Sonic_externals/cc8`
with 
`./pkgtools/cmsBuild -i build_grpc_cc8 -c /build/mrodozov/Sonic_externals/cc8/ --repository cms.week1 --arch cc8_amd64_gcc8 -j 16 build grpc`
with pkgtools 33 tag and cmssw-cc8 image 
the build dir is in 
`/build/mrodozov/Sonic_externals/build_grpc_cc8/BUILD/cc8_amd64_gcc8/external/grpc`
and the external is also there in 
`/build/mrodozov/Sonic_externals/build_grpc_cc8/cc8_amd64_gcc8/external/grpc`
I tried it few times until it was sure it gets ssl for centos7 but it doesn't for 8

smuzaffar:
the following segment [a] is not working. `%{centos}` is alway empty. Currently it is working on centos7 as one of the dependent spec brings in openssl and for centos8 it works as non of the dependent spec include openssl. But this is not correct behaviour.  Basically in this case [a] is useless, if we want to use a condition then we should make sure that it is working

[a]
```
%if 0%{?centos} == 7
Requires: openssl
%endif
```

mrodozov:
I'm fixing the condition

smuzaffar:
wait @mrodozov , we need some differnet condition. I am going though all the `IB/CMSSW_11_1_X/*` branches and noticed that we have dropped openssl for gcc9 and gcc10 (although those are build for slc7). 
For now I would say that remove `openssl` from `Requires` and add ${OPENSSL_ROOT} in CMAKE_PREFIX_PATH if it is set.

mrodozov:
please test
indeed even without openssl dependency the OPENSSL_ROOT variable is set, can be seen in the cmake config. it's not set for cc8. the rpm macroses both return the proper values if I run 
`rpm --eval '%{centos_ver}'` for both cc7 and 8
I don't understand why the condition was not working, they should've  
anyway ....

smuzaffar:
> `rpm --eval '%{centos_ver}'` for both cc7 and 8
> I don't understand why the condition was not working, 

note that we use our own rpm and not the one from system. It is not working using our rpm.

kpedro88:
Don't we need openssl in the requires for sl7, in order to make sure it builds against our rpm and not the system version? (For example, if I'm running a test on my development machine...)

mrodozov:
> the following segment [a] is not working. `%{centos}` is alway empty. Currently it is working on centos7 as one of the dependent spec brings in openssl and for centos8 it works as non of the dependent spec include openssl. But this is not correct behaviour. Basically in this case [a] is useless, if we want to use a condition then we should make sure that it is working
> 
> [a]
> 
> ```
> %if 0%{?centos} == 7
> Requires: openssl
> %endif
> ```

@kpedro88 this explains how it will work, it's available trough the required externals. if you see the cmake log:
https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-29288d/6418/external/grpc/1.28.1-29288d/log
the OPENSSLROOT envvar is set.

smuzaffar:
@kpedro88 , yes we need ssl for slc7 (but only for slc7/gcc8 and not for slc7/gcc9-10). Currently we have two sets of specs
1. with openssl from our externals
2. With openssl from system (and these spec do not contain any ref to openssl)

so the current implementation of grpc.spec should work for both as it is getting openssl indirectly from on of its Requirements. So in case of (1), OPENSSL_ROOT will be set and we pass that to cmake. And for (2) , OEPNSSL_ROOT is not net and cmake picks it up from system.

kpedro88:
Thanks for the explanation. Are we ready to merge? Once #5747 is also merged, I can submit the PR for triton.

smuzaffar:
+externals
but lets merge it after 11.1.0.pre8 build.

kpedro88:
yes, fine to wait till after pre8

PR #5746, opened 2020-04-15
updated vecgeom patch for arm64: remove mm_free from test

smuzaffar:
VecGeom 1.1.6 fails to build on aarch64 due to build error[a]. For now we propose to patch it  by removing the usage of _mm_free for non x86_64 archs. We have open a JIRA issue https://sft.its.cern.ch/jira/browse/VECGEOM-545 .

This change only effects non x86_64 archs, so it is safe to include it once it builds.


smuzaffar:
please test

smuzaffar:
+externals
merging it as it will fix non x86_64 IBs

PR #5747, opened 2020-04-15
[Sonic] Add OpenCV external

mrodozov:
please test
@kpedro88


mrodozov:
hm, right. there is no existing reference. I'll add the toolfile
no I didn' build it on arm and ppc yet but it's simple enough and I'm not expecting any trouble there.

davidlange6:
how does this work with python3?

mrodozov:
let me post the cmake config for short:
```
-- Found PythonInterp: /build/mrodozov/Sonic_externals/build_grpc/slc7_amd64_gcc820/external/python/2.7.15-bcolbf/bin/python2.7 (found suitable version "2.7.15", minimum required is "2.7")           
-- Found PythonLibs: /build/mrodozov/Sonic_externals/build_grpc/slc7_amd64_gcc820/external/python/2.7.15-bcolbf/lib/libpython2.7.so (found suitable exact version "2.7.15+")                           
ImportError: No module named numpy.distutils                                                                                                                                                           
-- Found PythonInterp: /usr/bin/python3 (found suitable version "3.6.8", minimum required is "3.2")                                                                                                    
-- Could NOT find PythonLibs (missing: PYTHON_LIBRARIES PYTHON_INCLUDE_DIRS) (Required is exact version "3.6.8")                                                                                       
ModuleNotFoundError: No module named 'numpy'
```
there are cmake vars that can be added to the config to make it find both py2 and 3 , but because the config from @kpedro88 aimed to build opencv libs I didn't add them. 
If the concern is 'how do we make this work only with py3' I actually don't know. 
It only needed to find py2 to build it so I didn't check

davidlange6:
seems more cmake options are needed to pick up the right python environment.

mrodozov:
please test
I added all libs I found in the lib dir to the toolfile.
works for both py2 and 3 with `import cv2` to use it

mrodozov:
please test

mrodozov:
please test

mrodozov:

kpedro88:
https://github.com/NVIDIA/triton-inference-server/blob/master/docs/client_example.rst#download-from-github says that Triton needs libopencv-dev, libopencv-core-dev. Based on these lists:
https://packages.debian.org/stretch/amd64/libopencv-dev/filelist
https://packages.debian.org/stretch/amd64/libopencv-core-dev/filelist

it seems like only libopencv_core is needed.

I think it's good to keep a commit in the history showing all the available libs, in case somebody needs another one in the future. But I'll try setting up Triton with just libopencv_core and see how it goes.

mrodozov:

mrodozov:

mrodozov:
please test

mrodozov:
abort
we need the ppc IB first

mrodozov:

kpedro88:
comparison never arrived?

smuzaffar:
please test
lets test for amd64.
@kpedro88 , for ppc64le/aarch64 comparison job does not work. We normally run the pr tests for ppc64le and aarch64 to make sure external builds successfully.

kpedro88:
please test
(just to refresh)

smuzaffar:

smuzaffar:
@mrodozov , please add opencv in to python_tools.spec too so that `PYTHON*PATH` are also properly setup in cmssw.

opencv.spec
```diff
@@ -0,0 +1,37 @@
+### RPM external opencv 4.3.0
+## INITENV +PATH PYTHON27PATH %{i}/${PYTHON_LIB_SITE_PACKAGES}
+## INITENV +PATH PYTHON3PATH %{i}/${PYTHON3_LIB_SITE_PACKAGES}
+%define tag 01b2c5a77ca6dbef3baef24ebc0a5984579231d9
```
davidlange6: can we not use a traditionally tagged release version?

smuzaffar: +1
01b2c5a77ca6dbef3baef24ebc0a5984579231d9 anyway points to tag 4.3.0. So we can just use `%{realversion}` instead of tag here.

smuzaffar: done

smuzaffar:
please test

smuzaffar:
+externals
we will merge it after cmssw 11.1.0.pre8

smuzaffar:
please test

PR #5750, opened 2020-04-16
cms monitoring tools spec

vkuznet:
New spec for cms monitoring tools. So far I disabled `Requirements: go` because go spec only exists in comp branch, but since go is pre-installed on lxplus I think the build will be fine.

@smuzaffar please inspect the spec and make a suggestion on how to adjust it correctly If it will be required.

### DIFF
--- a/cmsmon-tools.spec
+++ b/cmsmon-tools.spec
@@ -1,11 +1,11 @@
-### RPM cms cmsmon-tools 0.3.8
+### RPM cms cmsmon-tools 0.3.9
 ## INITENV +PATH PYTHONPATH %i/${PYTHON_LIB_SITE_PACKAGES}
 
 %define pkg CMSMonitoring
 %define ver %realversion
 Source0: https://github.com/dmwm/%pkg/archive/%ver.tar.gz
 
-#Requires: go
+Requires: go
 
 # RPM macros documentation
 # http://www.rpm.org/max-rpm/s1-rpm-inside-macros.html
@@ -13,6 +13,7 @@ Source0: https://github.com/dmwm/%pkg/archive/%ver.tar.gz
 %setup -D -T -b 0 -n %pkg-%ver
 
 %build
+export GOCACHE=%{_builddir}/gocache
 cd ..
 cd %pkg-%ver
 echo "build $PWD"
@@ -46,13 +47,11 @@ cd ../%pkg-%ver
 echo "### current dir: $PWD"
 cp src/go/MONIT/monit %i/
 cp src/go/NATS/nats-sub %i/
-
-# Generate dependencies-setup.{sh,csh} so init.{sh,csh} picks full environment.
-%addDependency
+cp src/wrappers/cms-monit %i/
+cp src/wrappers/cms-nats-sub %i/
 
 %post
-%{relocateConfig}etc/profile.d/dependencies-setup.*sh
-
-%files
-%{installroot}/%{pkgrel}/monit
-%{installroot}/%{pkgrel}/nats-sub
+cp $RPM_INSTALL_PREFIX/%{pkgrel}/cms-monit $RPM_INSTALL_PREFIX/common
+ln -sf $RPM_INSTALL_PREFIX/common/cms-monit cms-monit
+cp $RPM_INSTALL_PREFIX/%{pkgrel}/cms-nats-sub $RPM_INSTALL_PREFIX/common
+ln -sf $RPM_INSTALL_PREFIX/common/cms-nats-sub cms-nats-sub
--- a/cmsmon-tools.spec
+++ b/cmsmon-tools.spec
@@ -4,6 +4,7 @@
 
 %define pkg CMSMonitoring
 %define ver %realversion
+%define cmsmon_commands nats-sub nats-pub nats-exitcodes-termui dbs_vm
 Source0: https://github.com/dmwm/%pkg/archive/%ver.tar.gz
 
 Requires: go
@@ -51,22 +52,13 @@ cd -
 cd ../%pkg-%ver
 echo "### current dir: $PWD"
 cp src/go/MONIT/monit %i/
-commands="nats-sub nats-pub nats-exitcodes-termui dbs_vm"
-for cmd in $commands; do
+for cmd in %cmsmon_commands; do
+if [ -f src/go/NATS/$cmd ]; then
 cp src/go/NATS/$cmd %i/
+fi
 done
 
 %post
-commands="monit nats-sub nats-pub nats-exitcodes-termui dbs_vm"
-for cmd in $commands; do
-cat > $RPM_INSTALL_PREFIX/%{pkgrel}/cms-$cmd << EOF
-#!/bin/bash -e
-eval \$(scram unsetenv -sh)
-THISDIR=\$(dirname \$0)
-SHARED_ARCH=\$(cmsos)
-TOOL=\$(ls -d \${THISDIR}/../\${SHARED_ARCH}_*/cms/cmsmon-tools/*/$cmd 2>/dev/null | sort | tail -1)
-[ -z $TOOL ] && >&2 echo "ERROR: Unable to find command '$cmd' for '\$SHARED_ARCH' architecture." && exit 1
-\$TOOL "\$@"
-EOF
-chmod +x $RPM_INSTALL_PREFIX/%{pkgrel}/cms-$cmd
+for cmd in %cmsmon_commands; do
+cp $RPM_INSTALL_PREFIX/%{pkgrel}/$cmd $RPM_INSTALL_PREFIX/common
 done
--- a/cmsmon-tools.spec
+++ b/cmsmon-tools.spec
@@ -1,5 +1,6 @@
 ### RPM cms cmsmon-tools 0.3.8
 ## INITENV +PATH PYTHONPATH %i/${PYTHON_LIB_SITE_PACKAGES}
+## NOCOMPILER
 
 %define pkg CMSMonitoring
 %define ver %realversion
@@ -13,6 +14,7 @@ Source0: https://github.com/dmwm/%pkg/archive/%ver.tar.gz
 %setup -D -T -b 0 -n %pkg-%ver
 
 %build
+export GOCACHE=%{_builddir}/gocache
 cd ..
 cd %pkg-%ver
 echo "build $PWD"
@@ -46,13 +48,3 @@ cd ../%pkg-%ver
 echo "### current dir: $PWD"
 cp src/go/MONIT/monit %i/
 cp src/go/NATS/nats-sub %i/
-
-# Generate dependencies-setup.{sh,csh} so init.{sh,csh} picks full environment.
-%addDependency
-
-%post
-%{relocateConfig}etc/profile.d/dependencies-setup.*sh
-
-%files
-%{installroot}/%{pkgrel}/monit
-%{installroot}/%{pkgrel}/nats-sub
--- a/cmsmon-tools.spec
+++ b/cmsmon-tools.spec
@@ -15,10 +15,7 @@ BuildRequires: go
 %setup -D -T -b 0 -n %pkg-%ver
 
 %build
-cd ..
-cd %pkg-%ver
-echo "build $PWD"
-ls
+cd ../%pkg-%ver
 mkdir -p gopath/bin
 export GOPATH=$PWD/gopath
 export GOCACHE=%{_builddir}/gocache
@@ -37,26 +34,22 @@ go get github.com/nats-io/nats.go
 go get github.com/gizak/termui/v3
 
 # build monit tools
-cd src/go/MONIT
-go build %flags monit.go
-cd -
+pushd src/go/MONIT
+  go build %flags monit.go
+popd
+
 # build NATS tools
-cd src/go/NATS
-for cmd in %cmsmon_commands; do
-if [ -f $cmd.go ]; then
-go build %flags $cmd.go
-fi
-done
-cd -
+pushd src/go/NATS
+  for cmd in %cmsmon_commands; do
+    go build %flags $cmd.go
+  done
+popd
 
 %install
 cd ../%pkg-%ver
-echo "### current dir: $PWD"
 cp src/go/MONIT/monit %i/
 for cmd in %cmsmon_commands; do
-if [ -f src/go/NATS/$cmd ]; then
-cp src/go/NATS/$cmd %i/
-fi
+  cp src/go/NATS/$cmd %i/
 done
 
 #####################################################
--- a/cmsmon-tools.spec
+++ b/cmsmon-tools.spec
@@ -48,3 +48,18 @@ cd ../%pkg-%ver
 echo "### current dir: $PWD"
 cp src/go/MONIT/monit %i/
 cp src/go/NATS/nats-sub %i/
+
+%post
+cat << \END_SCRIPT >$RPM_INSTALL_PREFIX/common/cms-monit
+#!/bin/bash -e
+eval $(scram unsetenv -sh)
+THISDIR=$(dirname $0)
+TOOLNAME=$(basename $0 | sed 's|^cms-||')
+SHARED_ARCH=$(cmsos)
+TOOL=$(ls -d ${THISDIR}/../${SHARED_ARCH}_*/%{pkgcategory}/%{pkgname}/*/$TOOLNAME 2>/dev/null | sort | tail -1)
+[ -z $TOOL ] && >&2 echo "ERROR: Unable to find command '$TOOLNAME' for '$SHARED_ARCH' architecture." && exit 1
+$TOOL "$@"
+END_SCRIPT
+
+chmod +x $RPM_INSTALL_PREFIX/common/cms-monit
+ln -sf cms-monit $RPM_INSTALL_PREFIX/common/cms-nats-sub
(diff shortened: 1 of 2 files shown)


smuzaffar:
@vkuznet , I propose few changes here https://github.com/cms-sw/cmsdist/pull/5750/commits/937c48371caeb7c2eee40c4e20ead62634e11c93

So after installation of cmsmon-tools , you will have /cvmfs/cms.cern.ch/cmsmon/slc7_amd64_gcc820/cms/cmsmon-tools/version . Does this sound good? I would suggest to create some simple wrapper script which can find the latest version and run the executable from there.

vkuznet:
that's fine. I'll work on wrappers later, my understanding that we'll need them in /cvmfs/cms.cern.ch/common area similar to dasgoclient, right?

smuzaffar:
We can deploy the wrapper under /cvmfs/cms.cern.ch/common (which is by default in user's PATH) but what are the chances that we accidently override any existing`monit` or `nats-sub` command?

vkuznet:
So, what are the options then:
- we may have distinguished names, e.g. cms-monit, cms-nats-sub
- we will provide wrappers for monit and nats-sub and let users decided on a path

smuzaffar:
option 1 sounds good. We only need wrappers to have `cms-` prefix. Internally wrapper can run the actual `monit` or `nats-sub` commands. 
We only need one wrapper e.g. `cms-monit` and 2nd could be a symlink. Check `/cvmfs/cms.cern.ch/common/crab-*`   scripts for example.

smuzaffar:
go.spec from comp looks a trivial one to be added in cms. Can you please include that too here? Drop https://github.com/cms-sw/cmsdist/blob/comp_gcc630/go.spec#L15-L28 lines though

vkuznet:
ok, how about this wrapper:
```
#!/bin/bash -e

tool="monit"

# Clean-up CMSSW environment
if [ -f /cvmfs/cms.cern.ch/common/scram ] ; then
  eval `/cvmfs/cms.cern.ch/common/scram unsetenv -sh`
fi
# find latest version
SHARED_ARCH=`/cvmfs/cms.cern.ch/common/cmsos`
VERSION=`cd /cvmfs/cms.cern.ch; ls ${SHARED_ARCH}_*/cms/cmsmon-tools/v*/$tool | sed 's|.*/cms/cmsmon-tools/||' | sort | tail -1`
TOOL=`ls /cvmfs/cms.cern.ch/${SHARED_ARCH}_*/cms/cmsmon-tools/${VERSION} | sort | tail -1`
$TOOL "$@"
```

If it is ok, where it should be placed in a spec?

smuzaffar:
You can do something like this.
1. Add script [a] in your github repo 
2. In %install section copy cms-monit to %i/
2. Add %post section, something like this 
```
%post
cp $RPM_INSTALL_PREFIX/%{pkgrel}/cms-monit $RPM_INSTALL_PREFIX/common
chmod +x $RPM_INSTALL_PREFIX/common/cms-monit
ln -sf cms-monit $RPM_INSTALL_PREFIX/common/cms-nats-sub
```

[a] 
cms-monit
```
#!/bin/bash -e
eval $(scram unsetenv -sh)
THISDIR=$(dirname $0)
TOOLNAME=$(basename $0 | sed 's|^cms-||')
SHARED_ARCH=$(cmsos)
TOOL=$(ls -d ${THISDIR}/../${SHARED_ARCH}_*/cms/cmsmon-tools/*/$TOOLNAME 2>/dev/null | sort | tail -1)
[ -z $TOOL ] && >&2 echo "ERROR: Unable to find command '$TOOLNAME' for '$SHARED_ARCH' architecture." && exit 1
$TOOL "$@"
```

smuzaffar:
thinking about it more, as this wrapper script actually needed for distributiuon so please directly create it from the spec file . e.g. in %post section you can do something like
```
%post
cat << \EOF_SCRIPT >$RPM_INSTALL_PREFIX/common/cms-monit
CONTENTS OF SCRIPT
EOF_SCRIPT
```

smuzaffar:
Ok I have added the wrappers here https://github.com/cms-sw/cmsdist/pull/5750/commits/d509adeb847dc813c518b6924fbaaa147b87b1ab . So no action is needed on your side now. I will also add the go.spec

vkuznet:
no, please wait

vkuznet:
I'm adding more tools and different wrappers, your wrapper incorrectly link different tools (monit != nats-sub), therefore we need one wrapper per tool.

smuzaffar:
ok but if all tools are under same path then you need one wrapper and the rest could be symlink. `$0` is the name of wrapper when user runs. So in case `cms-monit`, the TOOLNAME will be monit and for `cms-nats-sub` it will be `nats-sub`  . So I do not see any reaspn to have one wrapper per tool

smuzaffar:
if you have to add many more tools then how about , we have one script e.g. `cmsmon` and one runs it something like `cmsmon monit <options>` or  `cmsmon nats-sub <options>` and the `cmsmon` search for the correct tool and run it?

vkuznet:
I would like to avoid having an additional prefix wrapper for every tool. It makes things more confusing since all these tools are independent and can run without a wrapper.

Please review PR now, I added wrappers for all tools in a concise way which will allow to expand tools as we will produce more of them.

cmsmon-tools.spec
```diff
@@ -0,0 +1,72 @@
+### RPM cms cmsmon-tools 0.3.10
+## INITENV +PATH PYTHONPATH %i/${PYTHON_LIB_SITE_PACKAGES}
+## NOCOMPILER
+
+%define pkg CMSMonitoring
+%define ver %realversion
+Source0: https://github.com/dmwm/%pkg/archive/%ver.tar.gz
```
smuzaffar, line 8: define commands here once e.g.
```
%define cmsmon_commands nats-sub nats-pub nats-exitcodes-termui dbs_vm
```
and then use %{cmsmon_commands} where needed

vkuznet:
I also think that having `cms-` prefix for wrappers make things more confusing since the tool I run on my laptop, e.g. `nats-sub`, on lxplus will be called `cms-nats-sub`, therefore it breaks the tool names.

So, I think I can remove `cms-` prefix and wrappers will only refer to the latest tool for given platform/architecture.

Let me know your opinion.

cmsmon-tools.spec
```diff
@@ -0,0 +1,72 @@
...
+echo "### current dir: $PWD"
+cp src/go/MONIT/monit %i/
+commands="nats-sub nats-pub nats-exitcodes-termui dbs_vm"
+for cmd in $commands; do
+cp src/go/NATS/$cmd %i/
+done
+
+%post
+commands="monit nats-sub nats-pub nats-exitcodes-termui dbs_vm"
+for cmd in $commands; do
```
smuzaffar: I still do not think that we need to create seperate scripts for each command. symlinks will do the trick. For example in my local builds 
```
>ls -l w/common/cms-*
-rwxr-xr-x. 1 muzaffar zh 353 Apr 16 22:02 w/common/cms-monit
lrwxrwxrwx. 1 muzaffar zh   9 Apr 16 22:02 w/common/cms-nats-sub -> cms-monit
>./w/common/cms-monit --help  |& head -1
Usage: monit [options]
>./w/common/cms-nats-sub  --help | & head -1
Usage: nats-sub [-s server] [-creds file] [-t] <subject>
```

cmsmon-tools.spec
```diff
@@ -0,0 +1,72 @@
...
+cat > $RPM_INSTALL_PREFIX/%{pkgrel}/cms-$cmd << EOF
+#!/bin/bash -e
+eval \$(scram unsetenv -sh)
```
smuzaffar: do you need to escape `$`?

go.spec
```diff
@@ -0,0 +1,13 @@
+### RPM external go 1.13.8
+
```
smuzaffar, line 3: Add no compiler to avoid gcc dependency.
```
## NOCOMPILER
```

go.spec
```diff
@@ -0,0 +1,13 @@
...
+Provides: /bin/rc
+
+%prep
+%setup -n go
+
+%build
+
+%install
+cp -r * %i
```
smuzaffar: why not use
```
rsync -a ./ %i/
```

smuzaffar:
As I wrote eralier, what are the chances that these tools override any existing tool in user path? As `/cvmfs/cms.cern.ch/common` is in user PATH so adding any thing in it means that it will override any system command. 

To be on safe side, may be we can copy these tools under /cvmfs/cms.cern.ch/cmsmon diecotry and those who wants to use these tools can explicitly add /cvmfs/cms.cern.ch/cmsmon in PATH.

> I also think that having `cms-` prefix for wrappers make things more confusing since the tool I run on my laptop, e.g. `nats-sub`, on lxplus will be called `cms-nats-sub`, therefore it breaks the tool names.
> 
> So, I think I can remove `cms-` prefix and wrappers will only refer to the latest tool for given platform/architecture.
> 
> Let me know your opinion.

cmsmon-tools.spec
vkuznet: done

go.spec
vkuznet: done

cmsmon-tools.spec
```diff
@@ -0,0 +1,66 @@
...
+cp src/go/MONIT/monit %i/
+for cmd in %cmsmon_commands; do
+if [ -f src/go/NATS/$cmd ]; then
+cp src/go/NATS/$cmd %i/
+fi
+done
+
+%post
+for cmd in %cmsmon_commands; do
+cp $RPM_INSTALL_PREFIX/%{pkgrel}/$cmd $RPM_INSTALL_PREFIX/common
```
smuzaffar: I would suggest to not do this? This will break if we ever build this for aarch64 /ppc64le.  keep the wrapper script logic.

smuzaffar:
what do you think about https://github.com/cms-sw/cmsdist/pull/5750#issuecomment-614870273 ? I would prefer to copy these tools under `/cvmfs/cms.cern.ch/cmsmon` . This way these tools will not be in default PATH and there is no change that they will break any thing. Those who need these can explicitly add this in their PATH.

vkuznet:
Shahzad, I'm totally fine to copy tools to `/cvmfs/cms.cern.ch/cmsmon`.

vkuznet:
please review the new changes:
- I put back wrappers
- I made wrappers available in `/cvmfs/cms.cern.ch/cmsmon` area

cmsmon-tools.spec
```diff
@@ -0,0 +1,76 @@
+### RPM cms cmsmon-tools 0.3.10
+## INITENV +PATH PYTHONPATH %i/${PYTHON_LIB_SITE_PACKAGES}
+## NOCOMPILER
+
+%define pkg CMSMonitoring
+%define ver %realversion
+%define cmsmon_commands nats-sub nats-pub nats-exitcodes-termui dbs_vm
+%define flags -ldflags="-s -w -extldflags -static"
```
smuzaffar: please add `-p %{compiling_processes}` so that we have control over the build processes. Default is to use number of CPUs which we do not want as cmsBuild also build other packages in parallel which might need cpus.

cmsmon-tools.spec
```diff
@@ -0,0 +1,76 @@
+### RPM cms cmsmon-tools 0.3.10
+## INITENV +PATH PYTHONPATH %i/${PYTHON_LIB_SITE_PACKAGES}
```
smuzaffar: there is nothing python related in this package , so why to set PYTHONPATH? Please remove

cmsmon-tools.spec
```diff
@@ -0,0 +1,76 @@
...
+## NOCOMPILER
+
+%define pkg CMSMonitoring
+%define ver %realversion
+%define cmsmon_commands nats-sub nats-pub nats-exitcodes-termui dbs_vm
+%define flags -ldflags="-s -w -extldflags -static"
+Source0: https://github.com/dmwm/%pkg/archive/%ver.tar.gz
+
+Requires: go
```
smuzaffar: Change this to `BuildRequires: go`

cmsmon-tools.spec
```diff
@@ -0,0 +1,76 @@
...
+go get github.com/shirou/gopsutil/mem
+go get github.com/shirou/gopsutil/load
+go get github.com/shirou/gopsutil/process
+go get github.com/go-stomp/stomp
+go get github.com/nats-io/nats.go
+go get github.com/gizak/termui/v3
+
+# build monit tools
+cd src/go/MONIT
+go build %flags monit.go
```
smuzaffar: you are building monit but not copying it in the distribution.

cmsmon-tools.spec
```diff
@@ -0,0 +1,76 @@
...
+for cmd in %cmsmon_commands; do
+if [ -f src/go/NATS/$cmd ]; then
+cp src/go/NATS/$cmd %i/
+fi
+done
+
+%post
+mkdir -p $RPM_INSTALL_PREFIX/cmsmon
+for cmd in %cmsmon_commands; do
+cat > $RPM_INSTALL_PREFIX/cmsmon/$cmd << EOF
```
smuzaffar: You just need to escape EOF here e.g. change it to `\EOF` and then you do not need to escape every `$` in the script itself.

smuzaffar: Again, why to create a seperate script for each if symlinks will work? Having  on script means in case we have to fix the script by have in /cvmfs then it is only one place to fix and not N scripts.

cmsmon-tools.spec
```diff
@@ -0,0 +1,76 @@
...
+#!/bin/bash -e
+eval \$(scram unsetenv -sh)
+THISDIR=\$(dirname \$0)
+SHARED_ARCH=\$(cmsos)
```
smuzaffar: just add `cmd=$(basename $0)` and you will be fine.

cmsmon-tools.spec
vkuznet: done

cmsmon-tools.spec
vkuznet: done

cmsmon-tools.spec
vkuznet: done

cmsmon-tools.spec
vkuznet: it is copied here:
https://github.com/cms-sw/cmsdist/pull/5750/files#diff-c13d4ca5c238805a5e42ba8af94d9b5bR55

cmsmon-tools.spec
vkuznet: Shahzad, you lost me in these discussions. My understanding that you want wrapper scripts because we can build on different architectures, e.g. x86 vs ppc. That's why I put wrappers back. But I said that I don't want one script/wrapper since I think it is confusing. If we should have wrappers script the script name should match actual executable and we will have one wrapper per one executable such that from a user usage point of view it will look like a command, regardless if it is run on lxplus or local laptop.

Please adjust this part is you think is appropriate based on the aforementioned arguments.

cmsmon-tools.spec
vkuznet: done

cmsmon-tools.spec
vkuznet: replaced

cmsmon-tools.spec
smuzaffar: Please see my original commit https://github.com/cms-sw/cmsdist/pull/5750/commits/d509adeb847dc813c518b6924fbaaa147b87b1ab . Yes I want wrapper but as the contents of all the wrappers are same that is why I was suggested to create one wrapper and then create symlinks points to the wrapper. At the end, for each command there will be one wrapper (it do not matter it that is actual script of symlink to script :-) ).

Do not worry I will update the spec.

cmsmon-tools.spec
smuzaffar: this is copying in %i. We need to create a wapper for this in %post too. Anyway, I will fix it.

cmsmon-tools.spec
vkuznet: ok, now I understand. Please proceed with the change I think I've done everything I need from my side and you may finish with the single wrapper and different links.

smuzaffar:
@vkuznet , I think I am down with the changes.

cmsmon-tools.spec
```diff
@@ -0,0 +1,86 @@
...
+go get github.com/gizak/termui/v3
+
+# build monit tools
+cd src/go/MONIT
+go build %flags monit.go
+cd -
+# build NATS tools
+cd src/go/NATS
+for cmd in %cmsmon_commands; do
+if [ -f $cmd.go ]; then
```
smuzaffar: by the way, why this check? Shouldn't we know what command we have to build and if they are missing then we should fail ... right?

vkuznet: because cmsmon_commands lists scripts from different areas, e.g. monit.go comes from src/go/MONIT area, while nats-sub.go comes from src/go/NATS. This is why I check that actual file we intend to compile exists. I think we should keep this structure if we want to use  common cmsmon_commands which lists names from different areas of our repository.

vkuznet: if you agree with my comment here, I think everything else is ready and can be merged.

vkuznet:
@smuzaffar , is there anything else holding for merge in this PR?

smuzaffar:
@vkuznet , cmsmon tools are now available under /cvmfs/cms-ib.cern.ch/latest/cmsmon . In first week of MAY, when we will build CMSSW_11_1_0_pre7 then, these will be available via our production CVMFS area /cvmfs/cms.cern.ch/cmsmon .

PR #5751, opened 2020-04-16
Warn about classes with only inline virtual functions

fwyzard:
From the GCC documentation:
> C++ virtual functions are implemented in most compilers using a lookup table,
> known as a vtable. The vtable contains pointers to the virtual functions
> provided by a class, and each object of the class contains a pointer to its
> vtable (or vtables, in some multiple-inheritance situations).
> If the class declares any non-inline, non-pure virtual functions, the first
> one is chosen as the “key method” for the class, and the vtable is only
> emitted in the translation unit where the key method is defined.

If no such "key method" exists, the vtable is emitted in all translation units
that include the header file.


smuzaffar:
please test for CMSSW_11_1_CLANG_X

smuzaffar:
Test parameters
 -  full_cmssw = true

smuzaffar:
please test for CMSSW_11_1_CLANG_X

fwyzard:
@smuzaffar any idea how we ended up passing the `-Wweak-vtables` to GCC ?

smuzaffar:
@fwyzard , because of this bug https://github.com/cms-sw/cmsdist/pull/5751/commits/831baec1c0b2a3d1fb54f449502d77c256c46cd6 .

smuzaffar:
please test for CMSSW_11_1_CLANG_X

fwyzard:
Mhm,  OK, with other **120k** compilation warnings (many in ROOT or in Boost), we probably don't want to enable this...

makortel:
> Mhm, OK, with other **120k** compilation warnings (many in ROOT or in Boost), we probably don't want to enable this...

Can we then think of adding such a check in the static analyzer? @gartung

gartung:
@makortel This would involve importing the code that does the check to produce the warning and then limiting it to source in CMSSW or excluding headers. I thought there might be a flag to exclude warnings from headers but I have not found it yet.

gartung:
@makortel  https://clang.llvm.org/docs/UsersManual.html#controlling-diagnostics-in-system-headers 
We can set a clang compiler flag telling it to treat certain headers as system header and skip the warnings for those.

gartung:
For instance you could ignore any headers in external.

smuzaffar:
please test for CMSSW_11_2_CLANG_X

mrodozov:
please test for CMSSW_11_2_CLANG_X

smuzaffar:
Summary
-  Over 35K warnings from externals (723 unique warnings)
- Over 129K warnings from CMSSW (2767 unique warnings)

smuzaffar:
closing it, reopen for latest release cycle if we still need this flag

PR #5763, opened 2020-04-20
[ROOT modules] Get latest from master

mrodozov:
please test


mrodozov:
please test

davidlange6:
this may or may not be a real error. will integrate and find out tomorrow .

davidlange6:
hi @mrodozov, @smuzaffar  - indeed it seems that the above was a fake error - but it looks like pyroot doesn't work in the modules branch. Can you have a look if something is out of sync (as the mainstream was ok)?

PR #5800, opened 2020-05-04
[ROOT master] Get latest commits

mrodozov:
please test


mrodozov:
please test with cms-sw/cmssw#29740

mrodozov:
seems there are some major changes in builtin glew 
https://github.com/root-project/root/pull/5458
which is fine, it's only resulting in libGLEW being static now and not being installed, also the headers

mrodozov:
please test with cms-sw/cmssw#29740

mrodozov:
abort

mrodozov:
please test with cms-sw/cmssw#29740

mrodozov:
please test with cms-sw/cmssw#29740

mrodozov:
the cmake finds glew, sets it's include dir and then it doesn't use it :/

mrodozov:
ROOT team are fixing this, lets wait for the fix before hacking the recipe

oshadura:
@mrodozov I opened PR https://github.com/root-project/root/pull/5571, I will ping you when it will be merged.

oshadura:
@mrodozov  fix is merged in the master couple of minutes ago, please try and let me know.

mrodozov:
great I'll wrap it in the morning

mrodozov:
please test

mrodozov:
please test with cms-sw/cmssw#29740

mrodozov:
abort

mrodozov:
please test with cms-sw/cmssw#29740

smuzaffar:
@mrodozov , why external glew? what was wrong with root internal glew? Have they dropped it from root?

smuzaffar:
I think here we should only update the root branch/tag and every thing else should remain same (unless root build system has changed and needs external glew). If that does not work then report thr issue to root team instead of hacking

oshadura:
@smuzaffar glew is external package and we moved it to builtins (it will allow people to build against newer versions of glew and update easier glew inside ROOT) as it is done with lz4/zstd/xxhash and etc. If you want to use Glew provided by ROOT just select CMake option ```-Dbuiltin_glew=ON```

smuzaffar:
humm strange as we were already building with `-Dbuiltin_glew=ON`  ( https://github.com/cms-sw/cmsdist/pull/5800/files#diff-81217efc115a8de6b75eeef2a6324e47L82 ) .

oshadura:
Do you have ```opengl=ON```? Glew depends on opengl...

Can you please try to build again with builtin_glew=ON after my changes in master? 

Basically now you need to treat glew the same way you treat lz4 or xxhash, so I believe approach of Mircho could be correct.

mrodozov:
I decided to get glew as external after ROOT changed their config and did not distribute it any more. I can retry with using the builtin_glew=ON but that's how I started - ROOT is using it, but not distribute it. I don't know if thats still the case with the latest ROOT changes

smuzaffar:
Normally we add an external if it is also used by other package. If `glew` is only used by `root` so there is no point building and distributing it as external.

mrodozov:
CMSSW are also using glew as rootglew I also had to change that, in Fireworks/Core . I decided to get it as external after ROOT stopped distributing it. I'll retry with the latst config

smuzaffar:
@mrodozov , rootglew is the ROOT library. So it should stay as it is

smuzaffar:
just remove all your changes and only update root branch/tag

oshadura:
@smuzaffar @mrodozov So if you are using glew only inside ROOT you need to have -Dbuiltin_glew=ON,  but if you are relying on glew functionality outside of ROOT, you probably should have it build as external dependency.

I am not sure what is rootglew, we don't have such ROOT library. If it is libGLEW.so then it is simple Glew library that was shipped in the middle of of other ROOT libraries, very old and not updated  and now since we move to mechanism of builtins, it is not distributed anymore. I am trying to understand if  builtins  (xz, lz4,zstd, glew and etc.) actually should be distributed... (cc: @Axel-Naumann)

smuzaffar:
@oshadura , are you planning to drop GLEW from root? if yes then sure we need to add it as cms external.

smuzaffar:
only usage of glew in cmssw is https://github.com/cms-sw/cmssw/search?q=glew&unscoped_q=glew . So I think if root does not need glew then may be we also do not need it in cmssw.

mrodozov:
> just remove all your changes and only update root branch/tag

Sure, but what I was saying is ROOT kept having it as builtin, but didn't distribute it in the ROOT/include and ROOT/lib anymore, thats when I had to add it (as external, since it wasn't present in the ROOT installation anymore)
I'm building ROOT with the builtin option ON, if the libs and headers are installed of course I'll keep it as it was.

smuzaffar:
I understand that @mrodozov . I would suggest to first check with root team to make sure that the change is on purpose. It could that due to a bug they might have changed something and then we have to patch it to make it work.  If root team statement is that they do not want to make GLEW visible then good enough and we need to tell firework teams to update cmmsw.

root.spec
```diff
@@ -3,8 +3,8 @@
 ## INITENV +PATH PYTHON3PATH %{i}/lib
 ## INITENV SET ROOTSYS %{i}
 
-%define tag 384e06fed3d7305d8d438c34399d6e6e327256de
-%define branch cms/master/73ae672
+%define tag a2847019732b1b02e0cb1e437b4b9f4c913ddee1
+%define branch cms/master/e2c9e16
```
smuzaffar, line 7: please open a new PR with only this change. All of other changes are needed now ... right?

mrodozov: I'll keep the branch if ROOT decide to ditch glew, apart from that yes no need to keep this one open. This is not merged yet so it also needs it https://github.com/root-project/root/pull/5583

mrodozov:
Closed in favor of https://github.com/cms-sw/cmsdist/pull/5813

PR #5802, opened 2020-05-06
[DEVEL][GCC 8.4] Update GCC to 8.4.0

smuzaffar:
same as https://github.com/cms-sw/cmsdist/pull/5704 but for devel IBs. As IBs itself are not in good state, so we propose to intergate GCC 8.4 only in DEVEL now. If DEVEL does not show any errors due to GCC update then we can include https://github.com/cms-sw/cmsdist/pull/5704 in normal IBs.


smuzaffar:
FYI @silviodonato and @makortel

smuzaffar:
merging it for 11h00 DEVEL IBs

PR #5815, opened 2020-05-11
[ROOT 6-20] Updated root to tip of branch v6-20-00-patches

mrodozov:
please test


smuzaffar:
hold
we will move to root 6.20.06 once out. Till then we wait for the CMSSW_11_1_ROOT620 release validation.

PR #5823, opened 2020-05-15
Update jemalloc to version 5.2.1

fwyzard:
This release is primarily about Windows. A critical virtual memory leak is resolved on all Windows platforms.
The regression was present in all releases since 5.0.0.

Bug fixes:
  - Fix a severe virtual memory leak on Windows. This regression was first released in 5.0.0.
  - Fix size 0 handling in posix_memalign(). This regression was first released in 5.2.0.
  - Fix the prof_log unit test which may observe unexpected backtraces from compiler optimizations. The test was first added in 5.2.0.
  - Fix the declaration of the extent_avail tree. This regression was first released in 5.1.0.
  - Fix an incorrect reference in jeprof. This functionality was first released in 3.0.0.
  - Fix an assertion on the deallocation fast-path. This regression was first released in 5.2.0.
  - Fix the TLS_MODEL attribute in headers. This regression was first released in 5.0.0.

Optimizations and refactors:
  - Implement opt.retain on Windows and enable by default on 64-bit.
  - Optimize away a branch on the operator delete path.
  - Add format annotation to the format generator function.
  - Refactor and improve the size class header generation.
  - Remove best fit.
  - Avoid blocking on background thread locks for stats.

### DIFF
--- a/jemalloc.spec
+++ b/jemalloc.spec
@@ -1,4 +1,4 @@
-### RPM external jemalloc 5.2.0
+### RPM external jemalloc 5.2.1
 
 %define tag %{realversion}
 %define branch master


smuzaffar:
please test

smuzaffar:
+externals

fwyzard:
FYI, here is a comparison of the full-node HLT performance using jemalloc 5.2.0 (current the version in CMSSW and standalone), jemalloc 5.2.1, and the latter with some additional options:

image

We might say that 5.2.1 is slightly better than 5.2.0, but it doesn't seem statistically significant.

Enabling "heap profiling" (even if we don't use it) has a slight hit.
My "peal memory" patches (https://github.com/jemalloc/jemalloc/pull/1176, rebased on 5.2.1) don't seem to give any degradation in performance.

PR #5825, opened 2020-05-18
[SCRAM] New tag which support replacement of runtime env via hooks

smuzaffar:
This change will allow to dynamically set/replace the runtime env for multi vector libraries support.


smuzaffar:
please test

smuzaffar:
+externals

PR #5841, opened 2020-05-21
[crab][dbs] expose dbs python api and examples via crab distribution

smuzaffar:
- expose `dbs` and `RestClient` python api via crab distribution
- crab distribution to include `dbs/Client/utils` (as `examples`)
- `crab-setup.*sh` scripts to set `DBS3_CLIENT_ROOT` pointing to latest crab/dbs distribution

### DIFF
--- a/crab-build.file
+++ b/crab-build.file
@@ -34,6 +34,8 @@ rsync -a %{_builddir}/CRABClient/etc/  %{i}/etc/
 
 #List of CRAB python pakcages for which we need to create ProxyPackage symlink
 ls %{_builddir}/CRABClient/src/python/*/__init__.py | sed 's|/__init__.py$||;s|.*/||' > %{i}/etc/crab_py_pkgs.txt
+echo dbs >> %{i}/etc/crab_py_pkgs.txt
+echo RestClient >> %{i}/etc/crab_py_pkgs.txt
 
 #Copy WMCore
 for pkg in %{wmcore_packages} ; do
@@ -61,6 +63,7 @@ for pkg in %{dbs_packages} ; do
     cp %{_builddir}/DBS/${pkg} %{i}/lib/${pkg}
   fi
 done
+rsync -a %{_builddir}/DBS/Client/utils/ %{i}/examples/
 
 if [ $(grep '_UseCrab' %{i}/etc/crab-bash-completion.sh | wc -l) -gt 0 ] && [ $(grep ' filenames  *crab' %{i}/etc/crab-bash-completion.sh |wc -l) -gt 0 ] ; then
   sed -i -e 's|_UseCrab|_UseCrab_%{crab_type}|g;s| filenames  *crab| filenames crab-%{crab_type}|' %{i}/etc/crab-bash-completion.sh
@@ -72,6 +75,6 @@ fi
 %post
 cd ${RPM_INSTALL_PREFIX}
 mkdir -p share/%{pkgdir}
-for dir in bin lib etc ; do 
+for dir in bin lib etc examples; do
   rsync -a %{pkgrel}/${dir}/ share/%{pkgdir}/${dir}/
 done
--- a/crab/crab-proxy-package.file
+++ b/crab/crab-proxy-package.file
@@ -1,4 +1,4 @@
-#CMSDIST_FILE_REVISION=2
+#CMSDIST_FILE_REVISION=3
 import sys
 from os.path import join, exists
 CMSPath="@CMS_PATH@/share/cms"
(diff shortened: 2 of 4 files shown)


smuzaffar:
please test

smuzaffar:
@belforte and @yuyiguo , once this is merged then `crab-setup.*sh` script should
- set `DBS3_CLIENT_ROOT` env variable pointing to `crab-(pre|propd|dev)` installation where dbs/examples will be available.
- `dbs` and `RestClient` python api will also be available and there is no need to do `import CRABClient` first to get dbs.

yuyiguo:
Thanks @smuzaffar !

smuzaffar:
+externals

belforte:
@smuzaffar @yuyiguo I see that this was used in last IB since using that [1] I now see the new
env. var and the expanded pythonpath
DBS3_CLIENT_ROOT=/cvmfs/cms-ib.cern.ch/nweek-02629/share/cms/crab-prod/3.3.2005
and 
belforte@lxplus757/TC3> ls $PYTHONPATH 
CRABAPI/  CRABClient/  RestClient/  dbs/

But I do not see the DBS3 examples around:
belforte@lxplus757/TC3> find $DBS3_CLIENT_ROOT -name 'DBS3*py'
/cvmfs/cms-ib.cern.ch/nweek-02629/share/cms/crab-prod/3.3.2005/lib/WMCore/Services/DBS/DBS3Reader.py
belforte@lxplus757/TC3> 

Do we need to touch crab-*.spec files to trigger a new build ?

[1]
https://twiki.cern.ch/twiki/bin/view/CMSPublic/CMSCrabClient#Using_CRABClient_from_nightly_CM

smuzaffar:
No, no need to touch crab-*.spec. The problem is with finding the lastest version of crab. I have tried to fix it here https://github.com/cms-sw/cmsdist/pull/5847/files .

smuzaffar:
@belforte , there are two issues with finding the latest version of crab client

1. `.rcN` versions 
2. Automatic suffix `-<some-string>` (added by cmsBuild) whenever you change WMCore, dbs or CRABServer tags.

To fix (1),  I would suggest to change the crab version naming scheme. As there are no more monthly releases and there is no need to have release candidates so I would suggest that crab client tags in guthub could be either of form 3.3.YYMMDD (e.g. 3.3.200522) or v3.YYMMDD  (e.g. v3.200522).

To fix (2), I would suggest that we add a numeric suffix for the installation e.g. initial version v3.200522 could be installed as v3.200522.00 and later if you have to update dbs, WMCore or CRABServers tags then the build can be installed in v3.200522.01, v3.200522.02 and so on. So basically every new tag of crabclient will be installed as v3.YYMMDD.00 and every re-build (due to change in WMCore, dbs, CRABServer) will be installed in v3.YYMMDD.NN. This will simplify the crab latest version logic a lot.

What do you think?

belforte:
thans @smuzaffar very nice ideas. We are not in a hurry. Let me think a bit over this and run the idea in the (small) CRAB group and converge. Will track in a CRAB issue, do no like to add much to a closed PR.

belforte:
if a new version of crab client can help in the short tern, I can surely make one in order to test things while sticking with current scheme, right ? I.e. hopefully you do not need the (hard to read) logic in https://github.com/cms-sw/cmsdist/pull/5847/files.

Rather I need to ask about (2): who will add the numeric suffix ? where/how will it be visible ? what will crab version return ?

smuzaffar:
@belforte , no, currently there is no need to have new tag. my latest changes to get the latest crab version are now in and available under /cvmfs/cms-ib.cern.ch. So dbs/examples should be available now under `$DBS3_CLIENT_ROOT`.

For (2), the numeric suffix needs to be set/modified by hand. For e.g. `crab-*.spec` files will look like this

```
%define crab_client_tag 3.200522
#For new crab_client_tag set the version suffix (in next line) to "00" otherwise for any other change just increament it
### RPM cms crab-prod %{crab_client_tag}.00
```
So whenever you need to update crab_client_tag then make sure the suffix is `00` and for any other change e.g. updating WMCore, dbs etc., just increment it.

belforte:
I understand. Thanks.,
And I confirm that DBS examples are now available in
$DBS3_CLIENT_ROOT/examples is now visible and filled
I will change again the documentation for CRAB users :-)

belforte:
@smuzaffar I understand the idea, and all in all I like this schema and would like to adopt for crabserver/crabcacge/crabtaskworker as well. 
But have a question on the technicality.
We now have in crab-prod.spec (for the crab client):
```
### RPM cms crab-prod 3.3.2005
```
which tells causes versin 3.3.2005 to be checked out from github and built. Looking at crab-build.spec I would unserstand that `%{realversion}` is defined as the righmost word in ### RPM statement. WHen I put `### RPM cms crab-prod %{crab_client_tag}.00`  what will be pulled from github ? Do we also need to change
https://github.com/cms-sw/cmsdist/blob/f6d48c13b822b38e7dfd5d414f33a7b99b203809/crab-build.file#L14
so that it uses `%{crab_client_tag}` instead of ` %{realversion}` ?

Anyhow.. can we give it a try ? What should I do ?

PR #5842, opened 2020-05-22
[GCC10] Fix scipy and sckikit-learn for gcc10

mrodozov:
please test


mrodozov:
the py2 version is unchanged, it only adds a fortran flag to deal with the new defaults in gcc10. 
the change for py3 comes from here:
https://github.com/scipy/scipy/pull/11842
which is above 1.4.1 and not an official release. The question is - is that ok ? 
I tried to backport the same change for py2 but 1.2.3 is so much different/older than 1.4.x that this backport became too much of an effort and had to leave it.
Many python packages are having either no support for py2 since given version or some support that comes from old releases, like this one, which becomes a serious problem when one have to deal with the fortran part.

davidlange6:
Ah - I think we didn’t consider the gcc support aspect of the end of python2 support issue.. Interesting..

Does the same flag solve the errors on scipy 1.4.1 as with 1.2.3? Perhaps we should live with that change rather than picking up an unreleased version of scipy? I don’t have a strong opinion, is just a thought.

On May 22, 2020, at 11:51 AM, Mircho Rodozov <notifications@github.com<mailto:notifications@github.com>> wrote:



the py2 version is unchanged, it only adds a fortran flag to deal with the new defaults in gcc10.
the change for py3 comes from here:
scipy/scipy#11842<https://github.com/scipy/scipy/pull/11842>
which is above 1.4.1 and not an official release. The question is - is that ok ?
I tried to backport the same change for py2 but 1.2.3 is so much different/older than 1.4.x that this backport became too much of an effort and had to leave it.
Many python packages are having either no support for py2 since given version or some support that comes from old releases, like this one, which becomes a serious problem when one have to deal with the fortran part.

mrodozov:
I didn't think of this because the py3 solution is surgical and the py2 one is not: adds a flag globally to everything, but now it doesn't sound so bad ... 
Then remove it with the next release for py3, seems the child solution ate its parent

mrodozov:
please test

smuzaffar:
+externals

PR #5852, opened 2020-05-24
Extend jmalloc to monitor the peak memory usage of each thread (11.2.x)

fwyzard:
See https://github.com/jemalloc/jemalloc/pull/1176 and https://github.com/jemalloc/jemalloc/issues/1850 for the discussion with the jemalloc development team.

### DIFF
--- a/jemalloc.spec
+++ b/jemalloc.spec
@@ -1,8 +1,8 @@
-### RPM external jemalloc 5.2.1
+### RPM external jemalloc 5.2.1-peakmem
 
 %define tag %{realversion}
-%define branch master
-%define github_user jemalloc
+%define branch 52x-peakmem
+%define github_user fwyzard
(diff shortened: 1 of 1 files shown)


fwyzard:
@cmsbuild, please test

jemalloc.spec
```diff
@@ -1,8 +1,8 @@
-### RPM external jemalloc 5.2.1
+### RPM external jemalloc 5.2.1-peakmem
 
 %define tag %{realversion}
-%define branch master
-%define github_user jemalloc
+%define branch 52x-peakmem
+%define github_user fwyzard
```
makortel: If we go with the patch, we should probably fork `jemalloc` under https://github.com/cms-externals/

fwyzard: Sure - this is just a way to show what the changes would be, and make sure it doesn't break anything in CMSSW.

smuzaffar:
please test

smuzaffar:
+externals

PR #5857, opened 2020-05-25
[GCC 10] Onnxruntime, update to tag 1.3.0

mrodozov:
please test
1.2.0 was missing a header, since the powerpc PR is part of 1.3.0 and it is building lets just get it

### DIFF
--- a/onnxruntime.spec
+++ b/onnxruntime.spec
@@ -1,7 +1,7 @@
-### RPM external onnxruntime 1.2.0
+### RPM external onnxruntime 1.3.0
 ## INITENV +PATH PYTHON3PATH %{i}/${PYTHON3_LIB_SITE_PACKAGES}
-%define tag 8679719d81d0d77762d642d00717b189990b4a88
-%define branch cms/v1.2.0_plus_ppc_update_pb31130
+%define tag 9294d790219aa0d1007d918d7d736d7f5d3e82d8
+%define branch cms/v1.3.0
 %define github_user cms-externals
 Source: git+https://github.com/%{github_user}/%{n}.git?obj=%{branch}/%{tag}&export=%{n}-%{realversion}&submodules=1&output=/%{n}-%{realversion}.tgz
 


smuzaffar:

smuzaffar:

smuzaffar:
+externals

smuzaffar:
@slava77 , we would like to update onnxruntime to version 1.3.0 (where ppc64 support is included). Can you please check the comparison results here https://github.com/cms-sw/cmsdist/pull/5857#issuecomment-633661597 and let us know if these look good?

slava77:
> @slava77 , we would like to update onnxruntime to version 1.3.0 (where ppc64 support is included). Can you please check the comparison results here [#5857 (comment)](https://github.com/cms-sw/cmsdist/pull/5857#issuecomment-633661597) and let us know if these look good?

From these:

> * You potentially added 7365 lines to the logs
These are full of messages like 
```
2020-05-25 15:54:35.755287261 [W:onnxruntime:, graph.cc:814 Graph] Initializer jnet0_densenetdropout0_dense3_bias appears in graph inputs and will not be treated as constant value/weight. This may fail some of the graph optimizations, like const folding. Move it out of graph inputs if there is no need to override it, by either re-generating the model with latest exporter/converter or with the tool onnxruntime/tools/python/remove_initializer_from_input.py.
```
I think that this is not desired. @hqucms please take a look ; we may need to have the models files updated or some way of silencing the messages.

> 
> * Reco comparison results: 4 differences found in the comparisons
> * DQMHistoTests: Total failures: 17

these are false-positives, unrelated to this PR

hqucms:
@slava77 We can suppress these warnings -- I made a PR: https://github.com/cms-sw/cmssw/pull/29989.

slava77:
@cmsbuild please test with cms-sw/cmssw#29989

slava77:
@cmsbuild please test with cms-sw/cmssw#29989

slava77:
compared to https://github.com/cms-sw/cmsdist/pull/5857#issuecomment-633661597
in the new test https://github.com/cms-sw/cmsdist/pull/5857#issuecomment-634382689
we have 
> No significant changes to the logs found

so, this is apparently good to go, assuming we have cms-sw/cmssw#29989 merged at about the same time

hqucms:
@mrodozov @smuzaffar 
Can we have ONNXRuntime integrated for 10_6_X (only the C++ API is needed)? We plan to have a new version of the ParticleNet tagger for UL and will be using ONNXRuntime for inference there.

mrodozov:
when both are merged I'll backport them

hqucms:
Thank you very much, @mrodozov!

smuzaffar:
no idea why cmsbuild closed this PR.

mrodozov:
Hm there is probably a way to tell it if one got merged to merge also the other one
It's because the first comment in  cms-sw/cmssw#29989 says
this should 
fix #prnumber  https://github.com/cms-sw/cmsdist/pull/5857
and then fix is underlined because it's a keyword that 'fixes' the reference that is followed by closing it, as it does for issues

smuzaffar:
I was looking for exactly that logic in process_pr.py but could not find any thing. Message is very strange `cmsbuild closed this in cms-sw/cmssw#29989` . I think github might have triggered this  due to the user comment in the PR

smuzaffar:
yes, looks like due to the user comment, github decided to close this PR too. I will ask github support about this feature/bug

hqucms:
Can this be due to https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword? 

I wrote `fix cms-sw/cmsdist#5857 (comment)` in the PR description where `fix` is one of the keywords.

mrodozov:
The latest onnxruntime wont be straightforward as the default protobuf is 3.5.2 and getting the new one will also lead to tensorflow rebuild and creating new .pb files for it etc.
Is 0.5.0 onnxrt ok ? it's is building fine (same as in 11_0) and doesn't rebuild tensorflow

hqucms:
@mrodozov I think we need the new version. In this case would it be possible to build onnxruntime with its own protobuf (i.e., from its submodules)?

smuzaffar:
@hqucms , if I remember correctly , onnxruntime build with internal protobuf exposes protobuf symbols and that can cause runtime issues in cmssw. I would not suggest to go down that direction.

hqucms:
@smuzaffar According to https://github.com/microsoft/onnxruntime/issues/2812#issuecomment-580901286, onnxruntime only exposes its C/C++ API in the shared library. The protobuf symbols are local so it should be OK I guess?

```
> nm -C libonnxruntime.so|grep ' T '     
00000000000cd080 T OrtGetApiBase
0000000000177960 T OrtSessionOptionsAppendExecutionProvider_CPU
```

```
> nm -C libonnxruntime.so|grep 'google::protobuf'
...
00000000006f0270 t google::protobuf::SimpleDtoa[abi:cxx11](double)
00000000006ee0e0 t google::protobuf::SimpleFtoa[abi:cxx11](float)
00000000006ec1c0 t google::protobuf::SimpleItoa[abi:cxx11](int)
...
```

smuzaffar:
@mrodozov , we can give it a try. Just build the c++ apl part (drop all the python dependies)

PR #5861, opened 2020-05-25
test new CRAB tag naming scheme

belforte:
as suggested in https://github.com/cms-sw/cmsdist/pull/5841#issuecomment-632713671
@smuzaffar can you please do other changes to this spec file as you suggested ? I tried to do this for crab* spec files in COMP branch, but could not find the way, so I'd like to start from a working example.

### DIFF
--- a/crab-build.file
+++ b/crab-build.file
@@ -11,12 +11,16 @@
   %define dbs_packages Client/src/python/dbs PycurlClient/src/python/RestClient
 %endif
 
-Source0: git://github.com/dmwm/CRABClient.git?obj=master/%{realversion}&export=CRABClient&output=/CRABClient-%{realversion}.tar.gz
+Source0: git://github.com/dmwm/CRABClient.git?obj=master/%{crabclient_version}&export=CRABClient&output=/CRABClient-%{crabclient_version}.tar.gz
 Source1: git://github.com/dmwm/WMCore.git?obj=master/%{wmcore_version}&export=WMCore&output=/WMCore-%{wmcore_version}.tar.gz
 Source2: git://github.com/dmwm/CRABServer.git?obj=master/%{crabserver_version}&export=CRABServer&output=/CRABServer-%{crabserver_version}.tar.gz
 Source3: git://github.com/dmwm/DBS.git?obj=master/%{dbs_version}&export=DBS&output=/DBS-%{dbs_version}.tar.gz
 
 %prep
+if [ "%{v}" != "%{realversion}" ] ; then
+  echo "ERROR: %{v} not same as %{realversion}. Please increment version suffix in %{n}.spec file."
+  exit 1
+fi
 %setup -D -T -b 0 -n CRABClient
 %setup -D -T -b 1 -n WMCore
 %setup -D -T -b 2 -n CRABServer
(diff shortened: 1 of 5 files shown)


smuzaffar:
thanks @belforte  for the new tag, I will update the specs to use it properly

smuzaffar:
@belforte , I have updated the spec. local build of crab-* shows correct results. One thing you might want to fix is that `crab --version` command now returns `vv3.xxx` [a] I think one extra `v` is added by crabclient itself. I would suggest that either fix the crab code to not add the extra `v` or do not include `v` in the tag e.g. just create `3.200525` tag.

[a]
```
muzaffar@cmsdev31.cern.ch>w/common/crab-prod --version
CRAB client v3.3.2005
muzaffar@cmsdev31.cern.ch>w/common/crab-pre --version
CRAB client v3.3.2004
muzaffar@cmsdev31.cern.ch>w/common/crab-dev --version
CRAB client vv3.200525
```

belforte:
thanks @smuzaffar . I like it a lot to be able to give different names to builds with same githgub tags but different dependencies, but am still having hard time to change crab spec files in comp branch. Is there some documentation for these spec files to help me so that I do not bug you ?

smuzaffar:
@belforte , our spec files are basically a very simplified RPM spec file. you can find many doc about RPM spec on internet e.g. https://rpm-packaging-guide.github.io/#hello-world . In our can one needs to only provide the `%prep, %build, %install and %post` sections. Every thing else is auto generated by cmsBuild using the special comments in spec e.g. following s just a comment for RPM spec but cmsBuild parse these comments to auto generate the spec header
```
### RPM cms crab-dev %{crabclient_version}.%{version_suffix}
```
also 
```
## IMPORT crab-build
```
is again a comment for RPM but cmsBuild parses it and basically include the contents of crab-build.file in to this SPEC.

belforte:
thanks @smuzaffar . I had figured out the metacomments indeed and after I tried things for a couple of days I know more now. I was trying to make e.g. https://github.com/cms-sw/cmsdist/blob/comp_gcc630/crabtaskworker.spec use the new tag scheme with build iteration suffix and was getting odd errors. Those were actually due to pkgtoolks v00-30 vs v00-32. I had to change things in a way I do not fully understand (ppossibly the original crabserver.spec has more -%ver% than needed) and then made it pass the %build step, than got stuck inside https://github.com/cms-sw/cmsdist/blob/4b1edca6571b8d2db8d4e2ac56fe6982cf120602/crabtaskworker.spec#L42
which is not "build" but yet depends on understanding the directory structure, which maybe is now different from when Brian initially coded that. I can surely live w/o the build iteration suffix in crabserver/cache/taskworker but am getting worried that something changes and we can't maintain those build procedures in the next years. I guess I should make sure my stuff in COMP branch works with pkgtool V00-32 w/o changed. Or am I worrying w/o good reasons. I suspect I am only wasting time if I try to understand these details.
Advice welcome, but of course you are a busy man, and I'll manage even w/o it.
Thanks for all you did.

belforte:
@smuzaffar can this go ahead so we get CRA v3.200525 in crab-dev and validate new build files ?

smuzaffar:
@belforte what about the issue i mentioned https://github.com/cms-sw/cmsdist/pull/5861#issuecomment-633896420 ? New crab client tag prints out an extra `v` in the version

belforte:
OK. Made a new tag for that and have updated the spec file.

smuzaffar:
Please test

belforte:
@smuzaffar I am assuming that this failure is something at your end. Let me know if instead it is for me to fix something.

smuzaffar:
please test

smuzaffar:
+externals
yes @belforte , error was on my end. This has been fixed

silviodonato:
+1

PR #5862, opened 2020-05-26
make pcms for tinyxml and clhep

davidlange6:


davidlange6:
@vgvassilev @smuzaffar 
this would need the on-going IB to test - I would plan to merge it tonight (its just in the modules branch...)

pcm_util.spec
```diff
@@ -0,0 +1,30 @@
...
+rm -f dummy_dict.cc
+rm -f libDummy.so
+rootcling dummy_dict.cc -s ./libDummy.so -moduleMapFile=dummy.modulemap -cxxmodule -m tinyxml2 -mByproduct tinyxml2  -I ${TINYXML2_ROOT}/include/ empty.h
+
+
+%install
+
+mkdir %{i}/include
+cp clhep.pcm %{i}/include/.
+cp tinyxml2.pcm %{i}/include/.
```
smuzaffar: Mostly PCM resides in the lib directory, so I would suggest to copy them to lib directory and use LIBDIR in toolfile too.

@vgvassilev , If I remember correctly, there was some restrictions about having libNAME.so and NAME.pcm in same directory (may be that is only for root dictionaries). So in case of these (clhep and tinyxml) PCMs do we need a library with same name in lib directory?

vgvassilev: No, they should not be connected to any library/dictionary.

smuzaffar: also I think llvm finds PCMs via CLING_PREBUILT_MODULE_PATH ... right @vgvassilev ? I guess we need to set CLING_PREBUILT_MODULE_PATH in toolfile too

davidlange6:
Mostly PCM resides in the lib directory, so I would suggest to copy them to lib directory and use LIBDIR in toolfile too.

@vgvassilev<https://github.com/vgvassilev> , If I remember correctly, there was some restrictions about having libNAME.so and NAME.pcm in same directory (may be that is only for root dictionaries). So in case of these (clhep and tinyxml) PCMs do we need a library with same name in lib directory?


Ah - My confusion. I will change to lib. @vgvassilev will have to answer the rest (in which case we may have some circular dependencies to worry about)

pcm_util.spec
vgvassilev: Good point, usually we do. In those particular case tinyxml2 and clhep depend only on libc and std (afaik) which will be loaded by default by rootcling -- we should be safe without that variable I think.

smuzaffar: without the variable how will llvm know about these pcms?

smuzaffar: for cmssw, we set CLING_PREBUILT_MODULE_PATH (pointing to `CMSSW_BASE/lib/arch:CMSSW_RELEASE/BASE/lib/arch` ) so that already build pcms are visible to clang.

davidlange6: @smuzaffar - how do I add a PATH like variable to the tool file? Or does scram magically know that this variable is a PATH like variable?

vgvassilev: > without the variable how will llvm know about these pcms?

If you mean how rootcling will find libc.pcm and std.pcm -- you are right in that case we need that variable pointing to the location of those.

smuzaffar: > @smuzaffar - how do I add a PATH like variable to the tool file? Or does scram magically know that this variable is a PATH like variable?

just like https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_11_2_X/master/clhep-toolfile.spec#L20 i.e. add `type="path"`

I think you do not need LIBDIR or LD_LIBRARY_PATH for this tool. Just add
```
<runtime name="CLING_PREBUILT_MODULE_PATH" value="$PCM_UTIL_BASE/lib" type="path"/>
```

davidlange6:
I pushed changes that should address the feedback. What did I miss?

pcm_util-toolfile.spec
```diff
@@ -0,0 +1,20 @@
...
+%build
+
+%install
+
+mkdir -p %i/etc/scram.d
+cat << \EOF_TOOLFILE >%i/etc/scram.d/pcm_util.xml
+<tool name="pcm_util" version="@TOOL_VERSION@">
+  <client>
+    <environment name="PCM_UTIL_BASE" default="@TOOL_ROOT@"/>
+    <runtime name="CLING_PREBUILT_MODULE_PATH" value="$PCM_UTIL_BASE/lib" type="path"/>
```
smuzaffar: @davidlange6 this should be outside the `<client></client>` block :-)

davidlange6: trying again

smuzaffar:
test parameters
 - full_cmssw = true

smuzaffar:
please test
modules IBs is available now

PR #5863, opened 2020-05-26
[GCC10] blackhat patch

mrodozov:
please test
I think we better start using this external github repo from cms-externals. 
with every gcc change we add a patch for it :/


smuzaffar:
@mrodozov , we have https://github.com/cms-externals/blackhat available. There is already  a cms/v0.9.9 branch  with few patches. Can you please check if that has the right contents and push the extra patches there too?

mrodozov:
please test

smuzaffar:
+externals

PR #5865, opened 2020-05-26
[GCC10] classlib - turn error to warning

mrodozov:
please test
seems
`-Werror=cast-function-type`
is default for gcc10 resulting in an error, that's the fastest trick to avoid it.


mrodozov:
please test
looks unrelated (but LLVM last update related somehow). lets try again if we catch something inbetween :/

mrodozov:
please test

smuzaffar:
+externals

PR #5867, opened 2020-05-26
[GCC10] distcc - add flag for gcc10

mrodozov:
this flag is suggested as easy way for porting code to gcc10
https://gcc.gnu.org/gcc-10/porting_to.html
the proper way is also described , add `extern` for global variables
this explains other link failures we have with gcc10, at least in frontier_client


mrodozov:
please test

smuzaffar:
+externals

PR #5868, opened 2020-05-26
Triton inference server

kpedro88:
This PR is the next step after #5744 and #5747 for fast inference as a service.

Notes:
* The spec file does *not* use the top-level CMakeLists.txt in the Triton repository, because this sets up a lot of unneeded external projects. Instead, it uses the CMakeLists.txt for the client package, which is the only part that is needed for CMSSW.
* A few of the tweaks to the CMake setup, currently implemented using `sed`, will hopefully be imported to the official repository at some point (work in progress).
* the curl version is upgraded because of a missing def in the old version, needed by Triton.

### DIFF
--- a/triton-inference-server.spec
+++ b/triton-inference-server.spec
@@ -21,7 +21,6 @@ rm -rf ../build
 mkdir ../build
 cd ../build
 
-# -Wno-deprecated-declarations because of https://github.com/grpc/grpc/issues/21603
 cmake ../%{n}-%{realversion}/build/trtis-clients \
     -DCMAKE_INSTALL_PREFIX="%{i}" \
     -DCMAKE_INSTALL_LIBDIR=lib \
--- a/triton-inference-server.spec
+++ b/triton-inference-server.spec
@@ -13,9 +13,9 @@ Requires: openssl opencv protobuf grpc curl python py2-wheel py2-setuptools py2-
 %build
 
 # remove config required in cmake
-sed -i 's/find_package(CURL CONFIG REQUIRED)/find_package(CURL REQUIRED)/' ../triton-inference-server-1.12.0/src/clients/c++/library/CMakeLists.txt
+sed -i 's/find_package(CURL CONFIG REQUIRED)/find_package(CURL REQUIRED)/' ../%{n}-%{realversion}/src/clients/c++/library/CMakeLists.txt
 # remove perf_client which requires rapidjson
-sed -i 's/add_subdirectory(perf_client)//' ../triton-inference-server-1.12.0/src/clients/c++/CMakeLists.txt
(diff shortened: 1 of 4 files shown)


kpedro88:
please test

smuzaffar:
@kpedro88 , please add triton-inference-server-toolfile dependency in cmssw-tool-conf

triton-inference-server.spec
```diff
@@ -0,0 +1,43 @@
...
+Requires: openssl opencv protobuf grpc curl python py2-wheel py2-setuptools py2-grpcio-tools
+
+%prep
+
+%setup -n %{n}-%{realversion}
+
+%build
+
+# remove config required in cmake
+sed -i 's/find_package(CURL CONFIG REQUIRED)/find_package(CURL REQUIRED)/' ../triton-inference-server-1.12.0/src/clients/c++/library/CMakeLists.txt
```
smuzaffar: can you please replace `triton-inference-server-1.12.0` with `%{n}-%{realversion}` here and in next line too

smuzaffar:
please test

mrodozov:
please test

smuzaffar:
@mrodozov , looks like one of gcc 10 updates broke sherpa for ppc64. Can yo uplease try to build latest cmsdist on ppc64 fix it?

mrodozov:
Yes I'm checking it

kpedro88:
are these unit test failures spurious? if not, i propose rolling back the curl version, since the previous tests were successful.

smuzaffar:
@kpedro88 , these are for aarch64 IBs and we do have such errors in IBs too. So please ignore these. The amd64 tests with latest curl are still running

mrodozov:
I think it's not curl but the ppc machine is having trouble. Or at least looks like it. I'll be able to tell in 10 minutes

mrodozov:
yes sherpa built with curl 7.70 on ibmminsky-2 the test failed on 1. I'll restart the test on 2

mrodozov:

mrodozov:

smuzaffar:

kpedro88:
@smuzaffar maybe we need a procedure to automate testing externals w/ all architectures? otherwise the process seems to be very manual and onerous

smuzaffar:
@kpedro88 we do not have enough resources (power/aarch) to test every external PR. So currently we are manually testing these for all new externals

mrodozov:
the actual errors were first a malfunctioning machine and then a typo in madgraph .spec 
apart from that we can test by default all three and we don't need build+utests+relvals for all archs, but mostly build, to make sure it doesn't break when we merge it in non prod archs.
we thought about it

smuzaffar:
please test
looks good for power and aarch

smuzaffar:
+externals

PR #5873, opened 2020-05-27
[GCC10] fix pyquen from pythia6

mrodozov:
please test
the issues was in pythia, pyquen is fine without changes


smuzaffar:
+externals

PR #5889, opened 2020-05-29
[Geant4] Update to verion new patched version 10.6.2

smuzaffar:
New Geant4 patch02 version:  http://cern.ch/geant4-data/ReleaseNotes/Patch4.10.6-2.txt
FYI @civanch , let me know if we need any CMS patches to go with it or if any geant4 data needs to be updated. Do we need to build G4Py (only available for python3 with Boost.Python version 1.72)?

### DIFF
--- a/geant4.spec
+++ b/geant4.spec
@@ -1,5 +1,5 @@
-### RPM external geant4 10.6.1
-%define tag b14d73978beaee6f5b6694ddbfe0dd54c7cdcf19
+### RPM external geant4 10.6.2
+%define tag 0f6c7dd7bcf054e57fdc34da9b4696fce5c6ac90
 %define branch cms/v%{realversion}
 %define github_user cms-externals
 Source: git+https://github.com/%github_user/%{n}.git?obj=%{branch}/%{tag}&export=%{n}.%{realversion}&output=/%{n}.%{realversion}-%{tag}.tgz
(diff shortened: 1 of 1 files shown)


smuzaffar:
please test

smuzaffar:
please test

civanch:
@smuzaffar , no modification in data files are required for this patch. We never use G4Py in CMSSW. Not clear if we will use in future.

Unfortunately, differences in regression are visible, so ordinary 11_2_0_pre1 validation will be required. 

I do not understand why ECAL unit test fails in reading of DB, likely it is not connected with this PR?

smuzaffar:
@civanch thanks.
unit test failure is not related to this change , it is already failing in IB ( https://github.com/cms-sw/cmssw/issues/30049 )
So so we first integrate in a dedicated GEANT4 IB and build CMSSW_11_2_0_pre1_GEANT4 for validation?

civanch:
@smuzaffar , yes, let us start from GEANT4 IB. I am not sure if it will be needed CMSSW_11_2_0_pre1_GEANT4 or test within CMSSW_11_2_0_pre1. Let us ask PPD.

civanch:
@smuzaffar , should this PR be re-tested or GEANT4 IB may be updated with this patch?

smuzaffar:
@civanch , I just have added GEANT4 IB and will include this in that.

PR #5898, opened 2020-06-02
Patch frontier_client for gcc10

mrodozov:
please test


smuzaffar:
@mrodozov , do we still need `-fcommon` flag in this case?

mrodozov:
right ill fix it missed it somehow

mrodozov:
please test

smuzaffar:
+externals

DrDaveD:
@mrodozov @smuzaffar Why did this PR remove the python bindings for frontier client?  There's no explanation of that and doesn't fit the title.

Back when the python bindings were first written there was talk of including them in the frontier client itself, but then the author of them stopped communicating with me and it never happened.  I didn't realize that CMS had been applying it in a patch all this time, up until a year ago when it was removed by this PR.  Now ATLAS is inquiring about python bindings for the frontier client.

mrodozov:
we didn't removed it, we just use it as a commit:
https://github.com/cms-externals/frontier_client/commit/72241af53b9d2d5bdaa1539bdafd03087cffb98c
we use the top of this branch 
https://github.com/cms-externals/frontier_client/compare/cms/2.9.1
which also includes the patch instead of having it as additional file

DrDaveD:
So ... it's everything in the python subdirectory, right?  Why don't I just include that in the frontier client package itself?   Isn't there anything that needs to be in a Makefile to build anything?

smuzaffar:
@DrDaveD , yes every thing in python sub-directory (except couple of `.orig and .rej` files in python/lib) is needed. It would be great if you can include it directly in frontier client package. Currently we just do `cp -r python <install-dir>` to include these. May be you can update the Makefile to copy the python directory when `make dist` target is run.

DrDaveD:
Ok I made an [issue](https://github.com/fermitools/frontier/issues/37) for that and I'll include it in the next frontier client release.

PR #5899, opened 2020-06-02
[GCC10] fix mxnet and update to 1.6.0 tag

mrodozov:
please test
super weird bug, it was always there but gcc10 made it fail
got lucky that the author was also using gcc10 I guess


smuzaffar:
@hqucms any objections on moving to mxnet 1.6?
@slava77 , the comparison difference are in L1 and message logger. Can you please give it a look?

slava77:
>  @slava77 , the comparison difference are in L1 and message logger. Can you please give it a look?

these are apparently false-positives.
I do not see anything changing related to mxnet. We run it with particleNet tagger in the standard workflows, which means that the update at least did not change the behavior for the existing model.

smuzaffar:
+externals

PR #5900, opened 2020-06-02
use build suffix to tag version

belforte:


smuzaffar:
Do you really want to do this for these packages? With long list of dependencies it will be a nightmare to keep the suffix up to date

muhammadimranfarooqi:
+1 to 8a868aa52232ac212a2a9f5d0c54329538054524 on slc7_amd64_gcc630.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/5900/test_8a868aa-20200603-014002.log" target="_blank">here</a>.

belforte:
good point. Thanks @smuzaffar 

> Do you really want to do this for these packages? With long list of dependencies it will be a nightmare to keep the suffix up to date

As many things, it seemed a good idea, but of course we can change. Indeed if this means that Imran will need to edit the suffix every time he rebuilds HGsomething it is a no go.
But from my experience so far we very seldom change the dependencies, other than wmcore is less than one line per year. I would like to avoid using a new crabserver tag only to track that we build with a different wmcore version. It has always been tricky to say which wmcore version was used in a server which simply says "I am running crabserver v3.3.1903".

Of course it is quite possible that I have not understood how this works in practice !
Though surely I finally start to understand how this procedure works.

We will not merge now.

belforte:
unless you mean not the dependencies listed in the spec file that the amzingly long list of externals in the build log, I surely do not want to bother when spidermonkey or xrootd minor version number change and not  even if valgrind or cms-workqueue do.

belforte:
well.. this seems a good reason not to do what I have tried here, the suffix I introduced clashes with compX one. Building a service only is OK, but building "COMP" yields:
```
+ echo 'ERROR: 3.200531.01-comp not same as 3.200531.01. Please increment version suffix in crabtaskworker.spec file.'
```

smuzaffar:
> unless you mean not the dependencies listed in the spec file that the amzingly long list of externals in the build log, I surely do not want to bother when spidermonkey or xrootd minor version number change and not even if valgrind or cms-workqueue do.

Yes I meant the dependencies (direct or indirect) list in the package's `*Requires:`. For example crabtaskworker dependents on 79 externals tools [a]. If one has to update any of these then he/she also has to update crabtaskworker. And then imagine if there are different PRs to update few of these .... so this tag suffix is not suitable for these packages.

[a]
```
autotools boost bz2lib cherrypy condor curl db4 db6 dbs-client dbs3-client
dbs3-pycurl-client dls-client expat gcc gdbm jemalloc libffi libtool libuuid libxml2
ncurses openldap openssl p5-archive-tar p5-extutils-makemaker p5-io-zlib 
p5-package-constants p5-time-hires pcre py2-argcomplete py2-asn1crypto 
py2-bcrypt py2-boto py2-boto3 py2-botocore py2-bz2file py2-certifi py2-cffi 
py2-chardet py2-cjson py2-cryptography py2-docutils py2-dogpile-cache py2-enum34
py2-futures py2-httplib2 py2-idna py2-ipaddress py2-ipython py2-jinja py2-jmespath
py2-nose py2-paramiko py2-progressbar2 py2-pycparser py2-pycurl py2-pygments
py2-pynacl py2-pysftp py2-python-dateutil py2-python-magic py2-python-utils
py2-pytz py2-pyxml py2-requests py2-retry py2-rucio-clients py2-s3transfer
py2-setuptools py2-six py2-sphinx py2-tabulate py2-urllib3 py2-zsi python
python-ldap readline sqlite zlib
```

belforte:
OK. Will close and review.
@ddaina please be aware

belforte:
@sharmaprajesh

PR #5901, opened 2020-06-03
remove cmssw/bin/arch from PYTHONPATH

### LINKED ISSUE #17200: Broken output when using python's warnings module
When using python's `warnings` module, the output of `warnings.warn()` contains some seemingly random characters at the end. I have completely reduced a configuration (`test_cfg.py`) to
```
import warnings
warnings.warn("this is a warning")
```
i.e. not even containing a process object. Executing this gives:
```
$ cmsRun test_cfg.py 
cmsRun:2: UserWarning: this is a warning
  � 
      )L��"E
----- Begin Fatal Exception 18-Jan-2017 11:54:00 CET-----------------------
An exception of category 'ConfigFileReadError' occurred while
   [0] Processing the python configuration file named test_cfg.py
Exception Message:
python encountered the error: <type 'exceptions.NameError'>
name 'process' is not defined
----- End Fatal Exception -------------------------------------------------
```
The strange characters are the same in every execution and on different machines, but depend on the string passed to `warn()`. They also depend on the length of the python file: adding comments *before* the `warn()` changes the output.
You can also add a process object + input source to get rid of the exceptions/errors, but the random characters still appear.
Executing this with pure python leads to the expected output:
```
$ python test_cfg.py 
test_cfg.py:2: UserWarning: this is a warning
  warnings.warn("this is a warning")
```

Adding a `print warnings` shows that the same path is used by `cmsRun` and pure python:
```
<module 'warnings' from '/cvmfs/cms.cern.ch/slc6_amd64_gcc530/external/python/2.7.11-ikhhed/lib/python2.7/warnings.pyc'>
```

I'm using `CMSSW_8_0_20_patch1` and also tested this in `CMSSW_7_6_3_patch2`.

For this case it might not be an important issue, but I don't know what the cause of this behavior could imply for important modules. What is `cmsRun` doing on top of python that could cause this?

smuzaffar:
There is no need to hadd CMSSW/bin/arch in PYTHONPATH for patch releases. If it is not needed for Full cmssw releases then there is no need to have it for cmssw patch releases, Side effect of this is  reported https://github.com/cms-sw/cmssw/issues/17200 .
Resolves https://github.com/cms-sw/cmssw/issues/17200

### DIFF
--- a/cmssw-toolfile.spec
+++ b/cmssw-toolfile.spec
@@ -19,9 +19,7 @@ cat << \EOF_TOOLFILE >%i/etc/scram.d/cmssw.xml
   <runtime name="@OS_RUNTIME_LDPATH_NAME@" value="$CMSSW_BASE/biglib/$SCRAM_ARCH" type="path"/>
   <runtime name="@OS_RUNTIME_LDPATH_NAME@" value="$CMSSW_BASE/lib/$SCRAM_ARCH" type="path"/>
   <runtime name="PATH"       value="$CMSSW_BINDIR" type="path"/>
-  <runtime name="PYTHON27PATH" value="$CMSSW_BINDIR" type="path"/>
   <runtime name="PYTHON27PATH" value="$LIBDIR" type="path"/>
-  <runtime name="PYTHON3PATH" value="$CMSSW_BINDIR" type="path"/>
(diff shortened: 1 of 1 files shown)


smuzaffar:
please test

smuzaffar:
+externals

PR #5906, opened 2020-06-04
[ROOT 6-22] Get 6-22 branch for root next

mrodozov:
please test

### DIFF
--- a/root.spec
+++ b/root.spec
@@ -1,9 +1,10 @@
-### RPM lcg root 6.20.01
+### RPM lcg root 6.21.01
 ## INITENV +PATH PYTHON27PATH %{i}/lib
 ## INITENV +PATH PYTHON3PATH %{i}/lib
 ## INITENV SET ROOTSYS %{i}
-%define tag 60ac037020707345f4c935b6f842a2506d0a067b
-%define branch cms/v6-20-00-patches/7fcce62
+
+%define tag 8014e908bf35e2c03a7dd9f4dfaead8d31db5b3a
+%define branch cms/v6-22-00-patches/539eada
 %define github_user cms-sw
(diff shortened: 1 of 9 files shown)


smuzaffar:
@mrodozov , I think its better to use root.spec from rootmaster here. There we have enabled pyroot and build for both python versions

smuzaffar:
also check what other changes we have in rootmaster branch and if we have to get them back in here e.g. yoda might need updates too

root.spec
```diff
@@ -39,7 +40,7 @@ export CXXFLAGS=-D__ROOFIT_NOBANNER
 
 cmake ../%{n}-%{realversion} \
   -G Ninja \
-  -DCMAKE_BUILD_TYPE=Release \
+  -DCMAKE_BUILD_TYPE=Debug \
```
smuzaffar: @mrodozov , I would suggest to build it in Release mode

mrodozov: i was going to ask if the new cmake was fixing this 😅 it's also need to be changed in master

smuzaffar: No, for master we need to build in Debug mode.

mrodozov:
please test

root.spec
smuzaffar: issue with cmake was that it was not allowing us to build with `-DCMAKE_BUILD_TYPE=Debug -DLLVM_BUILD_TYPE=Release` (i.e. one with Release and other with Debug). For root master we want to build with `-DCMAKE_BUILD_TYPE=Debug`  bit for normal IBs and rootnext we want to be in `Release` mode (i.e. close to the normal IBs)

smuzaffar:
test parameters
- full_cmssw = true

smuzaffar:
@mrodozov , you should test this in full release mode :-)

mrodozov:
please test

mrodozov:
abort

mrodozov:
please test

smuzaffar:
+externals

PR #5908, opened 2020-06-04
update gdb and build it with python for PY3 IBs

smuzaffar:
GDB currently is linking against python2 which makes it impossible to run cmsRun or cms tests under GDB. Here we propose to build GDB with python3 for PY3 IBs.


smuzaffar:
please test

smuzaffar:
+externals

PR #5920, opened 2020-06-08
fix tinyxml2 pcm and add the type_trait boost pcm

davidlange6:


davidlange6:
@smuzaffar - this fixes the tinyxml2 issue you noticed (the clhep one is a warning so its ignored for now)

PR #5928, opened 2020-06-10
Update cmssw-ccX script which now prefer unpacked.cern.ch over cms-ib repo

smuzaffar:


smuzaffar:
there is nothing in PR tests which tests it, so directly merging it

PR #5930, opened 2020-06-10
Update PYQUEN core

wouf:
Update PYQUEN core up to version 1.5.4


wouf:
Sorry it's already in cms-sw/cmsdist#4512

PR #5955, opened 2020-06-16
FWLite: Added zstd dependency due to root update

smuzaffar:


smuzaffar:
there are no tests for fwlite, so lets merge it and restart the fwlite build

PR #5994, opened 2020-06-26
added gcc-fixincludes spec

smuzaffar:


smuzaffar:
Add new package to fix gcc headers

PR #6010, opened 2020-06-30
[Geant4] Update Geant4 branch to use 10.7.0 Beta

mrodozov:
please test
Resolves: https://github.com/cms-externals/geant4/issues/47
both are building but that wasn't enough the last time, lets see
change the geant repo to their official github (they have one now)
The tag 10.7.0.beta was tagged from master, should be fine


smuzaffar:
+externals

PR #6055, opened 2020-07-09
[BuildRules] when -k is used then make sure to not break if library build failed

smuzaffar:


smuzaffar:
test parameters
- full_cmssw = true

smuzaffar:
please test for CMSSW_11_2_CXXMODULE_X

PR #6078, opened 2020-07-15
Update rucio-clients to 1.23.0, and bring all its dependencies

amaltaro:
The new rucio-clients (and 1.22.8) depends on `jsonschema`, which by itself depends on many other python packages.
Some of those python packages don't even seem to support python2 (by a quick look at their home page), even though rucio-clients is supposed to work on both py2 and py3 versions...

This PR also updates the py2-six spec, reason why it rebuilds most of our python-based applications.

### DIFF
--- a/py2-attrs.spec
+++ b/py2-attrs.spec
@@ -1,0 +1,2 @@
+### RPM external py2-attrs 19.3.0
+## IMPORT build-with-pip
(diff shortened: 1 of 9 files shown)


muhammadimranfarooqi:
+1 to 150779771ebd829a0094073982d7bf04663968a8 on slc7_amd64_gcc630.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/6078/test_1507797-20200715-120503.log" target="_blank">here</a>.

amaltaro:
@muhammadimranfarooqi I have just deployed WMAgent depending on all these new packages and it seems to start everything properly. Could you please merge it as soon as possible? Thanks

PR #6085, opened 2020-07-15
hot patch for CrabClient

belforte:
addresses https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/5732.html in a very ad hoc way

### DIFF
--- a/crab-dev.spec
+++ b/crab-dev.spec
@@ -3,7 +3,7 @@
 #For any other change, increment version_suffix
 ##########################################
 %define version_suffix 02
-%define crabclient_version v3.200531
+%define crabclient_version v3.200531.patch1
 ### RPM cms crab-dev %{crabclient_version}.%{version_suffix}
 %define wmcore_version     1.3.3
 %define crabserver_version v3.200531


belforte:
please test

crab-dev.spec
```diff
@@ -3,7 +3,7 @@
 #For any other change, increment version_suffix
 ##########################################
 %define version_suffix 02
-%define crabclient_version v3.200531
+%define crabclient_version v3.200531.patch1
```
smuzaffar, line 6: why adding a patch suffix? can not we just increment the date? version checking scheme will fill to find the latest version  if version is properly followed

belforte: I was curious if this could just work. I do not want it to become the latest version. It is effectively a dead branch for a limited scope (help one user our of troubles). I have already a latest tag in the standard numbering which have not queued for build yet becasue I was away. In the latest tag myproxy credential handling is different and this patch can't be appliet. I have to decide it to completely redo the myproxy delegation stuff or consider this an isolated incident and forget about it.

belforte: anyhow myproxy.cern.ch admins solved the current problem already. So this patch may not be needed at all. I'd like to know if it could have worked (no problem in missing up crab-dev for a couple days in case). But almost certainly I will roll back the tag right after.

belforte: rolling back

PR #6118, opened 2020-07-28
Remove codechecker external

mrodozov:
please test

-

the codechecker external is part of llvm but it was copied from older version of llvm and now every time 
we change llvm we need also to adjust it, although we already are using it from the llvm for cmssw code-checks and format.
last time this adjustment took me long enough until I figured this is already part of llvm. 
this externals is also the only reason (IIRC) why we can't conclude https://github.com/cms-sw/cmsdist/issues/5627 - it has custom cmake config to make it a standalone tool (although it's a part of llvm) which needs to be redone entirely so to work with llvm if we conclude https://github.com/cms-sw/cmsdist/issues/5627

no other external is using it nor any part of cmssw, so it should not break anything


smuzaffar:
@gartung , can you please confirm if we are using it and if there are any objections on dropping it?

gartung:
It is needed for https://github.com/cms-externals/CMSCodeChecker.
You need to check with @makortel if we should drop this since it was never used.

makortel:
In general I'm fine with dropping stuff that was never used. Is the (intended) functionality of CMSCodeChecker essentially included in the static analyzer (that is inside CMSSW)?

@Dr15Jones, do you have any thoughts?

gartung:
CMSCodeChecker was a clang-formatter that rewrites getByToken with get in place. I had it working for every case in the code base. It was never integrated because there was an objection (from David Lange?) to keeping a modified copy of the clang code-checker subsystem.

gartung:
If the problem is fixing it for new versions of clang I can do this.

davidlange6:
I can’t recall objecting, but maybe?… but indeed externals in a random person’s GitHub area are probably not what we want?

Anyway, does clang-format/tidy not support plugins for this sort of thing?

On Jul 28, 2020, at 5:17 PM, Patrick Gartung <notifications@github.com<mailto:notifications@github.com>> wrote:



CMSCodeChecker was a clang-formatter that rewrites getByHandle in place. I had it working for every case in the code base. It was never integrated because there was an objection (from David Lange?) to keeping a modified copy of the clang code-checker subsystem.

gartung:
Unfortunately no. It is like the custom attributes that we add to clang. The only way to add it is to add the source file.

gartung:
It was only ~2 months development time on my part so feel free to abandon it.

mrodozov:
If the thing is used I'm fine with keeping it there two problems with it:

- needs interface adjustment, which was frustrating until I figured out it's part of clang
- it's cmake config needs to be redone so that it works with shared libs llvm requested here: #5627 for which I have to rebuild it again and see what was not working (it was something that cmake that describes what llvm libs, headers, structure etc *should* be so that it would config itself, based on the llvm cmake config)

There was the benefit of reading it and trying to understand how it works that once you start understand it you can write your own checks. Not that I learned it but I recall that was the general idea of it

davidlange6:
I think we are saying the same thing in different ways.  Perhaps one should look at different integration mechanisms for including this?

On Jul 28, 2020, at 5:28 PM, Patrick Gartung <notifications@github.com<mailto:notifications@github.com>> wrote:


Unfortunately no. It is like the custom attributes that we add to clang. The only way to add it is to add the source file.

makortel:
I believe the functionality of rewriting specific functions calls could be handy in the future.

(did we actually run the `getByToken()` -> `get()` on the CMSSW? I can't remember)

gartung:
Looking at the changes @mrodozov made for llvm9, it seems like it might be better to just add the one source file to the llvm/clang tree and build it there.
https://github.com/cms-externals/CMSCodeChecker/compare/master...mrodozov:changes-for-llvm9

davidlange6:
For example, yes. This should be pretty stable and perhaps something to expand in the future.

On Jul 28, 2020, at 5:41 PM, Patrick Gartung <notifications@github.com<mailto:notifications@github.com>> wrote:



Looking at the changes @mrodozov<https://github.com/mrodozov> made for llvm9, it seems like it might be better to just add the one source file to the llvm/clang tree and build it there.
cms-externals/CMSCodeChecker@master...mrodozov:changes-for-llvm9<https://github.com/cms-externals/CMSCodeChecker/compare/master...mrodozov:changes-for-llvm9>

gartung:
That was what @Dr15Jones asked me to implement. It was never run and committed. I had it running on every example in the code at the time. Is the getByToken() -> get() migration happening now?

gartung:
@mrodozov please ask if you have problems with the interface changes. I could have saved you some frustrations.

makortel:
> That was what @Dr15Jones asked me to implement. It was never run and committed. I had it running on every example in the code at the time. Is the getByToken() -> get() migration happening now?

No, there is no migration going on (little point in doing it manually). I guess the this checker fell through the cracks, and we should think it again. Ideally the tool should be used exactly along clang-tidy (in PR checks and automated code migrations).

gartung:
Looking at the tests I see some of the warnings expected from the checker
```
warning: direct call of function getByToken(EDGetTokenT<>&, Handle<>&) is deprecated and should be replaced here with handleVar = iEvent.getHandle(token)
warning: bool return call of function getByToken(EDGetTokenT<>&, Handle<>&) is deprecated and should be replaced here with bool(m_handle = iEvent.getHandle(m_token))
warning: direct call of function getByToken(EDGetTokenT<>&, Handle<>&) is deprecated and should be replaced here with handleVar4 = iEvent.getHandle(*ptoken)
```
This code check will rewrite the code as suggested --> getByToken is replace by getHandle

gartung:
If we can figure out how to add this source to our fork of llvm/clang, I am OK with removing the stand alone checker and toolfile.

mrodozov:
abort

gartung:
It looks like there was a major change to how the clang-tidy checks are handled between v7.x and later versions. I will submit a pull request to cms-externals/llvm-project once I figure out how to add this checker.

gartung:
I made a pull request to cms-externals/llvm-project with the CMS HandleCheck module added and the build of llvm completed. Just waiting for all of the llvm dependent packages to build
https://github.com/cms-externals/llvm-project/pull/9

gartung:
@mrodozov go ahead and merge this to remove the standalone clang-tidy.

gartung:
please test

smuzaffar:
+externals

mrodozov:
> @mrodozov go ahead and merge this to remove the standalone clang-tidy.

sorry I saw this too late. anyway, it's merged

PR #6122, opened 2020-07-29
Add spec and new tag for SimG4CMS-Forward big file

### LINKED ISSUE #6077: do we need three castorShowerLibraries
-rw-r--r--. 1 dlange zh 570090084 Jul 15 09:47 CastorShowerLibrary_CMSSW500_Standard.root
-rw-r--r--. 1 dlange zh 165866508 Jul 15 09:48 castorShowerLibrary_QFBE.root
-rw-r--r--. 1 dlange zh 142193172 Jul 15 09:47 castorShowerLibrary.root

only the first (sadly the biggest) is used it seems..

mrodozov:
this adds the external without github repo because the file is too big (100mb max per file)
the other approach as with Fireworks-Geometry requires additionally to the .file in /data to have also a github repo with that name OR the distributed file to be a tar.gz archive. 
This one works the only problem is the tag needs to be updated in two files. With respect to having an empty github repo it's nearly the same 
Resolves: #6077 and needs https://github.com/cms-sw/cmssw/pull/30946 to be merged together

### DIFF
--- a/data/cmsswdata.txt
+++ b/data/cmsswdata.txt
@@ -94,7 +94,6 @@ SimTransport-HectorProducer=V01-00-01
 
 #Never update any package here. Always move it to default section.
 [data-cmssw-package-build]
-SimG4CMS-Forward=V02-03-18
 GeneratorInterface-ReggeGribovPartonMCInterface=V00-00-02
 
 #Never update any package here. Always move it to default section.
@@ -105,3 +104,4 @@ L1Trigger-RPCTrigger=V00-15-00
 RecoParticleFlow-PFBlockProducer=V02-04-02
 SimG4CMS-Calo=V03-01-00
 Validation-Geometry=V00-07-00
+SimG4CMS-Forward=V02-03-19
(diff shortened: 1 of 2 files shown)


smuzaffar:
@mrodozov , please build it the way we build https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_11_2_X/master/data/data-Fireworks-Geometry.file

mrodozov:
I tried and it requires to have a github repo with that name on cms-data
The Fireworks-Geometry.file was added to distribute additionally a file that was too big 
and there is a repo https://github.com/cms-data/Fireworks-Geometry/ that keeps the rest of them.
If we want to do it as in https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_11_2_X/master/data/data-Fireworks-Geometry.file we need an empty repo.

smuzaffar:
you just need to define `source0`
I think a simple data-SimG4CMS-Forward.file with one line to define source0 should be enough

smuzaffar:
closing it in favor of #6124

PR #6144, opened 2020-08-04
[LLVM] Remove static libs

### LINKED ISSUE #5627: LLVM distribution

mrodozov:
please test
Resolves: https://github.com/cms-sw/cmsdist/issues/5627

### DIFF
--- a/llvm.spec
+++ b/llvm.spec
@@ -78,6 +78,8 @@ rm -f %{i}/bin/FileRadar.scpt %{i}/bin/GetRadarVersion.scpt
 
 # Avoid dependency on /usr/bin/python, Darwin + Xcode specific
 rm -f %{i}/bin/set-xcode-analyzer
+# Remove all static libs except libomptarget-nvptx.a
+find %{i}/lib64 -name "*.a" -and -not -name "libomptarget-*.a" -delete
 
 %post
 %{relocateConfig}include/llvm/Config/llvm-config.h
--- a/py2-dxr-llvm-shared.patch
+++ b/py2-dxr-llvm-shared.patch
@@ -1,0 +1,14 @@
+diff --git a/dxr/plugins/clang/makefile b/dxr/plugins/clang/makefile
+index 7f8fc29..60e83c3 100644
+--- a/dxr/plugins/clang/makefile
++++ b/dxr/plugins/clang/makefile
+@@ -1,8 +1,4 @@
+-CLANG_LIBS := -lclangFrontendTool -lclangFrontend -lclangDriver \
+-              -lclangSerialization -lclangCodeGen -lclangParse \
+-              -lclangSema -lclangRewriteFrontend -lclangRewrite \
+-              -lclangAnalysis -lclangEdit -lclangAST -lclangLex \
+-              -lclangBasic
++CLANG_LIBS := -lclang
+ LLVM_CONFIG ?= llvm-config
+ LLVM_LDFLAGS := $(shell ${LLVM_CONFIG} --ldflags --libs --system-libs) $(CLANG_LIBS)
+ CXXFLAGS := $(shell ${LLVM_CONFIG} --cxxflags) -Wall -Wno-strict-aliasing \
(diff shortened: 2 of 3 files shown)


llvm.spec
```diff
@@ -78,6 +78,8 @@ rm -f %{i}/bin/FileRadar.scpt %{i}/bin/GetRadarVersion.scpt
 
 # Avoid dependency on /usr/bin/python, Darwin + Xcode specific
 rm -f %{i}/bin/set-xcode-analyzer
+# Remove all static libs except libomptarget-nvptx.a
+find %{i}/lib64 -name "*.a" -and -not -name "libomptarget-*.a" -delete
```
smuzaffar, line 82: @mrodozov , Instead of this you need to remove https://github.com/cms-sw/cmsdist/blob/bdfcb03effcafcd041143d9681899dd0f6fa5a42/llvm.spec#L19 line

smuzaffar: ah no, you want to keep `libomptarget-*.a` , ok ignore my previous comment

mrodozov:
please test

smuzaffar:
@mrodozov , I guess you did not read my 2nd comment :-)

llvm.spec
smuzaffar: as you want to keep `libomptarget-*.a` so your original change was correct.

smuzaffar: @mrodozov , FYI, please read the above comments

mrodozov: yes I just figured it from the issue. so yes we go back one commit

mrodozov:
please test

mrodozov:
please test

mrodozov:
ah ok the old test is running

smuzaffar:
test parameters
- full_cmssw = true

smuzaffar:
please test

smuzaffar:
+externals

PR #6149, opened 2020-08-05
update cms common revision to 1210

smuzaffar:
New cms-common which makes scram V2 as default


smuzaffar:
please test

smuzaffar:
+externals

PR #6152, opened 2020-08-05
HDF5 - Add requirement on openmpi

mrodozov:
please test
see: https://github.com/cms-sw/cmssw/pull/30312

### DIFF
--- a/hdf5-toolfile.spec
+++ b/hdf5-toolfile.spec
@@ -11,9 +11,7 @@ cat << \EOF_TOOLFILE >%i/etc/scram.d/hdf5.xml
 <tool name="hdf5" version="@TOOL_VERSION@">
   <info url="https://support.hdfgroup.org/HDF5/"/>
   <lib name="hdf5"/>
-  <lib name="hdf5_cpp"/>
   <lib name="hdf5_hl"/>
-  <lib name="hdf5_hl_cpp"/>
   <client>
     <environment name="HDF5_BASE" default="@TOOL_ROOT@"/>
     <environment name="LIBDIR" default="$HDF5_BASE/lib"/>
--- a/hdf5-toolfile.spec
+++ b/hdf5-toolfile.spec
@@ -11,6 +11,7 @@ cat << \EOF_TOOLFILE >%i/etc/scram.d/hdf5.xml
 <tool name="hdf5" version="@TOOL_VERSION@">
   <info url="https://support.hdfgroup.org/HDF5/"/>
   <lib name="hdf5"/>
+  <lib name="hdf5_hl"/>
   <client>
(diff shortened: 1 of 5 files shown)


smuzaffar:
@mrodozov , you also need to update hdf5-toolfile.spec and add update the toolfile to depened on openmpi

mrodozov:
abort

mrodozov:
please test

smuzaffar:
@mrodozov , please check why hdf5 is not building hdf5_cpp and hdf5_hl_cpp libs?

smuzaffar:
may be because we have dropped `--enable-cxx` ? do you know why it was dropped ... do they say that it does not work with parallel build support?

mrodozov:
this is strange. the installation we have:
ls /cvmfs/cms.cern.ch/slc7_amd64_gcc820/external/hdf5/1.10.6/lib/
shows the libs are there. but not in this PR, which doesn't change flags. 
So (probably) it's not the flag but I have to find what is it

smuzaffar:
but not in ls /cvmfs/cms-ib.cern.ch/latest/slc7_amd64_gcc820/external/hdf5/1.10.6-ninmko/lib/
which was build withotu --enable-cxx may be?

mrodozov:
10.6 was built always without the --enable-cxx because it contradict to --enable-parallel,
but there is another flag to build with both. im tring it . I'm checking the time when 
/cvmfs/cms.cern.ch/slc7_amd64_gcc820/external/hdf5/1.10.6/lib/
was installed and by what config

smuzaffar:
1.10.6 was initially build without parallel options ( https://github.com/cms-sw/cmsdist/blob/REL/CMSSW_11_2_0_pre3/slc7_amd64_gcc820/hdf5.spec ) .

mrodozov:
right. there is this option "enable-unsupported" which allows both, but in any case I have to see what both provide and why the one disallows the other.

smuzaffar:
so hdf5 build system tells that
```
configure: error: --enable-cxx and --enable-parallel flags are incompatible
```
I think this is due to threadsafety issue. Do we need these _cpp libs? can we just use the c libs which support thread safety?

mrodozov:
>  Do we need these _cpp libs?

I don' know but I didn't wanted to assume it doesn't . the test will show, I was trying to find out reading it

smuzaffar:
all these hdf5_* are same libs build for different interfaces e.g. C, C++ , fortran and High Level (_hl). We do not need to have all of these generated. Looks like only C (default ) one supports threadsafety and allows parallel option. So lets only build the default lib. It was a mistake to have all the hdf5_* libs defined in one tool file. If needed then one should have separate tool files and only one of these should be available to user (linking cmsw with both hdf5 and hdf5_hl or hdf5_cpp is wrong).

PR #6159, opened 2020-08-06
update cmsswdata to RecoBTag-Combined to V01-06-00

slava77:
backport of #6156

it looks like it's most direct to just advance to the master version instead of making a targeted split tag
https://github.com/cms-data/RecoBTag-Combined/compare/V01-04-00...V01-06-00


smuzaffar:
+externals
only new files , so should be safe to merge

silviodonato:
+1

PR #6167, opened 2020-08-11
roll back to PREvious CRAB client version.

belforte:
@mrodozov this will allow users to run crab client on older CMSSW w/o the hack described in
https://hypernews.cern.ch/HyperNews/CMS/get/cernCompAnnounce/1444/1/1/2/1/1.html
they will have to use crab from IB, but that's rather straightforward
https://twiki.cern.ch/twiki/bin/view/CMSPublic/CMSCrabClient#Using_CRABClient_from_nightly_CM

For crab-prod I will rather change the code to have both this fix and the fixes from current (alas broken) version.

Sorry for the extra work,

PR #6170, opened 2020-08-15
move forward to 0812 tag to fix httplib import

belforte:
rolling back was not a good idea (see https://github.com/cms-sw/cmsdist/pull/6167#issuecomment-673645488  )
so move forward to tag with fix and align with current crab-prod


mrodozov:
please test

PR #6191, opened 2020-08-21
Update XrootD to latest tag 5.0.1

mrodozov:
wait for all root pr's to be merged and then test and merge this

### DIFF
--- a/xrootd-toolfile.spec
+++ b/xrootd-toolfile.spec
@@ -10,7 +10,7 @@ mkdir -p %i/etc/scram.d
 cat << \EOF_TOOLFILE >%i/etc/scram.d/xrootd.xml
 <tool name="xrootd" version="@TOOL_VERSION@">
   <lib name="XrdUtils"/>
-  <lib name="XrdClient"/>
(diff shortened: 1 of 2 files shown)


mrodozov:
please test with https://github.com/cms-sw/cmsdist/pull/6189

mrodozov:
please test

mrodozov:
please test
XrootD XrdClient library was never used by CMSSW, specific libs were added to be used in this commit:
https://github.com/cms-sw/cmssw/commit/e266506ab744f8ecdfb56ba871b67aa679325545
There is a mismatch between whats in the toolfile and in the BuildFile of the only package using specific XrootD libs in CMSSW:
https://github.com/cms-sw/cmssw/tree/master/Utilities/XrdAdaptor
There is another package that uses it, not specifiying libs:
IORawData/DTCommissioning
https://github.com/cms-sw/cmssw/blob/e0abed7c2d7f244cd4b400853ac78d27a502aab7/IORawData/DTCommissioning/plugins/RawFile.cc#L38

PR #4103, opened 2018-06-09
Python package updates / addition of lz4 

davidlange6:
adding lz4 instead of picking up from inside of root
adding python interfaces to lz4 and lzma
updating to latest version of a number of python packages

PR #6227, opened 2020-09-13
Added Lz4 in to cmssw-tool-conf

smuzaffar:
LZ4 dependency was added (https://github.com/cms-sw/cmsdist/pull/4103 )  but its toolfile was not part of cmssw that is why at runtime we pick lz4 from system. 

This should be backported to 10.2 and above


smuzaffar:
please test

smuzaffar:
+externals

PR #6243, opened 2020-09-15
Remove page size flag for non intel archs

mrodozov:
The jemalloc manual says for this option: 
"this option need not be specified unless the system page size may change between configuration and execution, e.g. when cross compiling"
https://github.com/jemalloc/jemalloc/blob/5.2.1/INSTALL.md
The last tests we were running showed 20 to 30 % improvement in max virtual memory used, and we are using 2^16 which is the default for our arm and ppc distros and it doesn't have to be specified. 
Also after few IBs we don't have to run relvals by hand for comparisons :)

### DIFF
--- a/jemalloc.spec
+++ b/jemalloc.spec
@@ -11,14 +11,8 @@ BuildRequires: autotools
 %setup -n %{n}-%{realversion}
 
 %build
-XOPTS=""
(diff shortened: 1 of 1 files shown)


mrodozov:
please test

mrodozov:
+externals

PR #6263, opened 2020-09-23
Added new cmssw-wm-tools tool

smuzaffar:
New cmssw-wm-tools ( https://github.com/cms-sw/cmssw-wm-tools)  tool is added for tweaking CMSSW job configurations.
This will install `cmssw-wm-tools/bin` under `CMS_PATH/share/overrides/bin` and `cmssw-wm-tools/python` under `CMS_PATH/share/overrides/python` . SCRAM then adds these in PATH and PYTHON*PATH after `cmsenv`. This should work for all CMSSW releases (5.3X and above).

FYI @davidlange6

### DIFF
--- a/cmssw-tool-conf.spec
+++ b/cmssw-tool-conf.spec
@@ -3,6 +3,7 @@
 # With cmsBuild, change the above version only when a new tool is added
 
 Requires: crab
+Requires: cmssw-wm-tools
 Requires: google-benchmark-toolfile
 Requires: catch2-toolfile
 Requires: starlight-toolfile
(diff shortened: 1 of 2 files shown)


smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
+externals
it is a new tool for deploying the cmssw tweaking scripts for WM.

davidlange6:
hi @smuzaffar - I had expected this to get installed with the 10_2_24 release that got uploaded. That didn't seem to happen - do we need an 11_2_x build instead?

smuzaffar:
No, this is only part of 11.2.X branch so should already work in 11.2.X IBs env

davidlange6:
Ah - I had missed that the IB area had a different overrides area..  Cool - thanks!

PR #6272, opened 2020-09-25
Update python27 to last available tag 2.7.18

mrodozov:
python2 has newer tag https://github.com/python/cpython/tree/v2.7.18
I'm not changing the version because this has to change also in tensorflow at least. lets see if integrates.
it builds on amd and ppc

### DIFF
--- a/python.spec
+++ b/python.spec
@@ -11,8 +11,8 @@ Requires: zlib sqlite
 
 # FIXME: readline, crypt 
 # FIXME: gmp, panel, x11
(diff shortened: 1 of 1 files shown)


mrodozov:
please test

smuzaffar:
hold
if we have to re-build python (means every pip package) then check if we can update pip and setuptools

smuzaffar:
not needed any more. This was suppose to fix the ppc64le unit test which was failing as python was not built with readline support. As we moved to gcc9, so ppc64le_gcc9 tools were all rebuilt (including python) and unit test is now fixed.

PR #6279, opened 2020-09-29
[BuildRule] Updates scram build rules

smuzaffar:
- Use system cuda only if its version is >= to the version shipped with cmssw
- for PR tests also disable usage of release/externals/bin path


smuzaffar:
please test

PR #6283, opened 2020-09-30
added cmsLHEtoEOSManager gcc-fixincludes install only deps for cmssw

smuzaffar:
- Added cmsLHEtoEOSManager package which should deploy latest `GeneratorInterface/LHEInterface/scripts/cmsLHEtoEOSManager.py` in to CMS-PATH/share/override/bin so that all releases can pick it up.
- Added support for virtual packages which can have dynamic version based on date of some file in git
- make use of `## INSTALL_DEPENDENCIES <deps>` feature of `cmsBuild` to add dependency on packages without rebuilt of package even if `<deps>` are re-build


smuzaffar:
please test

smuzaffar:
+externals

PR #6288, opened 2020-10-02
Update tag for RecoBTag-Combined to V01-07-00 (106X backport #6282)

silviodonato:
Backport 106X of #6282 (Update tag for RecoBTag-Combined to V01-07-00)

Move RecoBTag-Combined data to new tag, see
cms-data/RecoBTag-Combined#35


silviodonato:
merge

PR #6290, opened 2020-10-02
Use git clone instead of source tar ball to fetch code with full git history as it requires to assign proper version of the executable

vkuznet:


muhammadimranfarooqi:
+1 to 2099829d213df06cd00fe7d3a64dec1b7a672c1e on slc7_amd64_gcc630.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/6290/test_2099829-20201002-190503.log" target="_blank">here</a>.

PR #6299, opened 2020-10-10
Revert "Revert "added back cmsLHEtoEOSManager gcc-fixincludes install deps""

smuzaffar:
Reverts cms-sw/cmsdist#6298
This should work now. It was issue with cmsBuild where it failed when a reference repository is used


smuzaffar:
please test

PR #6335, opened 2020-10-26
move tested version from crab-dev to crab-prod

belforte:


smuzaffar:
please test

smuzaffar:
please test

belforte:
@smuzaffar I checked the log but could not figure out the reason for the failure. Is this something I can/should fix at my end, or a glitch at yours ?  All I can say is that dependencies in the spec file are the same as used [here](https://github.com/cms-sw/cmsdist/blob/2fe267ef30b77e8af4ed5ce36508e10608fea4fc/crab-dev.spec) which worked

smuzaffar:
@belforte , it failed to download the sources [a]. Are you sure the WMCore tag is correct?

[a]
```
* The action "fetch-cms+crab-prod+v3.200818.00" was not completed successfully because Unable to download git://github.com/dmwm/WMCore.git?obj=master/1.3.6_crab1&export=WMCore&output=/WMCore-1.3.6_crab1.tar.gz
```

belforte:
well spotted ! Thanks @smuzaffar Indeed I put a '_' where there should have been a '.'

smuzaffar:
please test

smuzaffar:
+externals

PR #6161, opened 2020-08-08
Update to CUDA 11.0 Update 1, and update the list of supported architectures

fwyzard:
Update CUDA to version 11.0 Update 1:
  * CUDA version 11.0.228
  * NVIDIA drivers version 450.51.06
    
See [What's New in CUDA 11.0 Update 1](https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#cuda-whats-new-11Upd1).

Replace the Compute Sanitizer wrapper with a symlink, to work as a CMSSW external.

Drop support for Kepler, which is deprecated in CUDA 11, and add explicit support for Turing (e.g. Tesla T4).

Note that one can check the compute capabilities of the GPUs on the local machines with:
```bash
cudaComputeCapabilities
```
and re-build all CUDA code for the compute capabilities of the GPUs on the local machines with:
```
cmsCudaSetup.sh
cmsCudaRebuild.sh
```

PR #6342, opened 2020-10-27
[GCC10][CUDA] Use -std=c++14 to build cupla

smuzaffar:
FYI @fwyzard 
This is follow up on https://github.com/cms-sw/cmsdist/pull/6161#issuecomment-703491631


smuzaffar:
please test
but it will fail as there is no IB for gcc10

PR #6347, opened 2020-10-30
WMCore 1.4.2.pre7 release candidate for HG2011; drop py2-mysqldb

amaltaro:
In addition to bringing WMCore services to `1.4.2.pre7`, which has a few bug fixes; I also dropped the WMAgent dependency on `py2-mysql`, such that we can make sure our code no longer needs it.

TODO (in a different PR): T0 still needs to have that dependency dropped; and CRABServer needs to be investigated.


muhammadimranfarooqi:
+1 to 8408189d8c8c1b6620a1f6330646f814042bcf5b on slc7_amd64_gcc630.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/6347/test_8408189-20201030-095503.log" target="_blank">here</a>.

PR #6349, opened 2020-10-30
Update millepede to V04-08-00

vbotta:
Updating MillePede to the latest version. 

cc @mmusich @connorpa @adewit

### DIFF
--- a/millepede.spec
+++ b/millepede.spec
@@ -1,4 +1,4 @@
-### RPM external millepede V04-06-00
+### RPM external millepede V04-08-00
 
 Source: svn://svnsrv.desy.de/public/MillepedeII/tags/%{realversion}/?scheme=http&module=%{realversion}&output=/%{n}-%{realversion}.tgz
 Requires: zlib


mrodozov:
please test

smuzaffar:
@vbotta where are we using millepede in CMSSW? I do not see any CMSSW package using it, so any reason to keep it in the cmssw software stack?

mmusich:
> where are we using millepede in CMSSW? I do not see any CMSSW package using it, so any reason to keep it in the cmssw software stack?

@smuzaffar a part from this entire CMSSW package: https://github.com/cms-sw/cmssw/tree/master/Alignment/MillePedeAlignmentAlgorithm, it's necessary to be in the software stack as it is the algorithm used by the pixel alignment Prompt Calibration Loop at Tier-0. The configuration is here:

- https://github.com/cms-sw/cmssw/blob/master/Alignment/CommonAlignmentProducer/python/ALCARECOPromptCalibProdSiPixelAli_cff.py  (`mille` part)
- https://github.com/cms-sw/cmssw/blob/master/Alignment/CommonAlignmentProducer/python/AlcaSiPixelAliHarvester_cff.py (`pede` part).

FYI, it's also exercised in IB and PR relvals via worfkows `1001.0`.

smuzaffar:
+externals
ah OK, it is just one executable `pede` which is provided by this tool. I was looking for `<use name="millepede"/>` in BuildFile.xml which of course is missing as there are no headers and libraries. That also explains why bot did not checkout any cmssw package ( https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-7f926b/10428/build-logs/)  to test this.

PR #6364, opened 2020-11-03
Update 6-22 on cmsdist master with one fix

mrodozov:
please test with https://github.com/cms-sw/root/pull/145

### DIFF
--- a/root.spec
+++ b/root.spec
@@ -3,8 +3,8 @@
 ## INITENV +PATH PYTHON3PATH %{i}/lib
 ## INITENV SET ROOTSYS %{i}
 
-%define tag c8fedfbe7f2179e4bcccbc20be7bcdc1f49e2a8f
-%define branch cms/v6-22-00-patches/d6156de
+%define tag 4a7085f8b9785306fe0f528b7e9f8f51a187bb79
+%define branch cms/v6-22-00-patches/50e52ba43c
 %define github_user cms-sw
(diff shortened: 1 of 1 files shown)


mrodozov:
abort

mrodozov:
please test with https://github.com/cms-sw/root/pull/145, cms-sw/cmssw#32000
to test the fix for ROOT and cmssw altogether

mrodozov:
please test with cms-sw/cmssw#32000

root.spec
```diff
@@ -3,16 +3,15 @@
 ## INITENV +PATH PYTHON3PATH %{i}/lib
 ## INITENV SET ROOTSYS %{i}
 
-%define tag c8fedfbe7f2179e4bcccbc20be7bcdc1f49e2a8f
-%define branch cms/v6-22-00-patches/d6156de
+%define tag 4a7085f8b9785306fe0f528b7e9f8f51a187bb79
+%define branch cms/v6-22-00-patches/50e52ba43c
```
smuzaffar: @mrodozov , look like you are testing tip of root 6.22 . I though you will only include one commit pn top of our branch

mrodozov:
abort

root.spec
mrodozov: yes, wrong branch. I should've noticed the branch name doesn't need to change.
I was doing something else (but it seems I was looking into the last changes on 6-22 at the same time)

mrodozov:
please test with cms-sw/cmssw#32000

mrodozov:
please test with cms-sw/cmssw#32000

mrodozov:
since the fix is not enough lets close this and open another one with a solution

PR #6387, opened 2020-11-09
New dasgoclient version

vkuznet:
This version includes switch to using machine based port (8443) on cmsweb services and bug fix for missing key values used in DAS filters

### DIFF
--- a/dasgoclient-binary.spec
+++ b/dasgoclient-binary.spec
@@ -1,12 +1,15 @@
-### RPM cms dasgoclient-binary v02.04.00
+### RPM cms dasgoclient-binary v02.04.02
 Source0: git+https://github.com/dmwm/dasgoclient?obj=master/%{realversion}&export=dasgoclient&output=/dasgoclient.tar.gz
 Source1: git+https://github.com/dmwm/cmsauth?obj=master/e9fca92e3335252a5f71d8e6d09c64012f7d3c0c&export=github.com/dmwm/cmsauth&output=/cmsauth.tar.gz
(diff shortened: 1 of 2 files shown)


smuzaffar:
please test

smuzaffar:
+externals

smuzaffar:
@vkuznet , this broke our AARCH and PPC64 . Are you not building dasgoclient for these arch any more?

mrodozov:
the binary is called dasgoclient_linux for all arch now, so if we use only the _linux suffix it works 
https://github.com/cms-sw/cmsdist/pull/6387/files#diff-db2426b62a0cd487a4242ca023baa09dcb5a6b64ec1f1117d81a1adb8da3b2fbL30
and then fails in the installation part (on both arm and power) :/

smuzaffar:
humm, I wondor how one can run same binary on all arch?

mrodozov:
assuming (wrong) the binary was created. I wanted to rush for the 1100 IB, but didn't read what it was doing.
this uses privately owned sources, for example I can't list whats in https://cern.ch/valya/dasgoclient/
if you hit 
https://cern.ch/valya/dasgoclient/cms+dasgoclient-binary+v02.04.02-1-1.slc7_amd64_gcc820.rpm
for example you'll get it because it exists but then you have to guess if other archs are there
and then I'll assume whatever is in slc7_amd64_gcc820.rpm is also the binaries for arm and ppc, which are created by
dasgoclient-binary.spec (following the "use build_all" comment).
in short lets prevent the same for the next time :/

vkuznet:
sorry again, too many tickets, I closed original PR and make a new one here https://github.com/cms-sw/cmsdist/pull/6401 which builds dasgoclient for all archs.

PR #6394, opened 2020-11-10
move crab-prod to v3.201026

belforte:
Makes crab getoutput and getlog work after PhEDEx is turned off.  Was tested as crab-dev in IB.
This will be a fall back in case I don't get a newer version (which also fixes crab checkwrite) working by Friday

### DIFF
--- a/crab-prod.spec
+++ b/crab-prod.spec
@@ -3,10 +3,10 @@
 #For any other change, increment version_suffix
 ##########################################
 %define version_suffix 00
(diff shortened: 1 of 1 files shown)


smuzaffar:
please test

belforte:
surely the unit test failure in the  MagneticFieldEngineTestDriver can't be due to CRABClient !

smuzaffar:
+externals
yes unit test failure has nothing to do with crab update

PR #6397, opened 2020-11-11
Possible 7_6_X update to openssl and friends

davidlange6:
see discussion in https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/5832/1.html

@belforte


smuzaffar:
please test

smuzaffar:
ah we are not building IBs for 76X any more

belforte:
thanks @davidlange6 for letting me know of your efforts to address this, your perseverance encourages me ! But I will not try to understand this PR :-)

davidlange6:
no, we do not… I checked that all of tool-conf builds after these changes (well, once I figured that that I should not have an env-var called “ARCH” set when building the version of Pythia8 that is in 76x..)

davidlange6:
Stefano- the issue reported is in the end having a too-old version of openssl. I cherry-picked new commits from our master branch to update openssl. I updated curl as that was the initial thing reported as being wrong. These changes broke the git build, so I updated that (again cherry-picking most of it)

belforte:
yeah.. I had a look and could tell that much. But will not comment on the specifics. Thanks for looking into this.

smuzaffar:
closing this, please open a cms-sw/cmssw issue if we still need to update curl/openssl in 76X. Though I am afraid it will take too much of effort to update it

belforte:
thanks, CRAB Client can work around this via using this in forked subprocesses which use curl
```
source /cvmfs/cms.cern.ch/slc6_amd64_gcc700/external/curl/7.59.0/etc/profile.d/init.sh
```
 so it is indeed not needed.

PR #6401, opened 2020-11-11
New version of dasgoclient

vkuznet:
Put back build_all

### DIFF
--- a/dasgoclient-binary.spec
+++ b/dasgoclient-binary.spec
@@ -1,11 +1,11 @@
-### RPM cms dasgoclient-binary v02.04.02
+### RPM cms dasgoclient-binary v02.04.04
 Source0: git+https://github.com/dmwm/dasgoclient?obj=master/%{realversion}&export=dasgoclient&output=/dasgoclient.tar.gz
 Source1: git+https://github.com/dmwm/cmsauth?obj=master/e9fca92e3335252a5f71d8e6d09c64012f7d3c0c&export=github.com/dmwm/cmsauth&output=/cmsauth.tar.gz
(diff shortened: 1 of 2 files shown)


mrodozov:
please test

mrodozov:
and there is one thing I don't quite understand - how does the dasgoclient-binary.rpm ends up in https://cern.ch/valya/dasgoclient/
the dasgoclient-binary.spec creates the rpm and dasgoclient.spec uses it but there is nothing that uploads it in between.
I uncommented the build_all and the spec built the rpm, but then I had to upload it manually so that dasgoclient.spec to get it from
https://github.com/mrodozov/dasgoclient_rpms

vkuznet:
Mircho, I follow procedure established together with Shahzad some time ago. When I build an RPM for dasgoclient-binary I manually copy it to /eos/user/v/valya/www/dasgoclient/ area. Then we build dasgoclient which picks it up from there. I'm all in favor of more clear build instructions but that what was setup originally and we follow this path.

mrodozov:
ok now I know :) thanks. I was thinking maybe I'm missing something.

vkuznet:
Is there any way to look at unit test log outside of browser, it seems to me it is so huge that I can't even load it in a browser. I also have doubt that Magnetic Field unit test has anything to do with dasgoclient.

mrodozov:
it doesn't . it fails in IBs. If the relvals are fine we'll get it in today before 2300 (to fix the IBs)

smuzaffar:
+externals

smuzaffar:
@mrodozov , can you please start eithe rarm or ppc64 IB now

smuzaffar:
just call it 11h00 IB

PR #6408, opened 2020-11-11
move latest version to prod

belforte:
We managed to validate the latest crab client in dev. So this will move it to production and allow to remove all dependencies from PhEDEx in the next (Friday's ?) release build.

### DIFF
--- a/crab-prod.spec
+++ b/crab-prod.spec
@@ -3,10 +3,10 @@
 #For any other change, increment version_suffix
 ##########################################
 %define version_suffix 00
-%define crabclient_version v3.201026
+%define crabclient_version v3.201110
 ### RPM cms crab-prod %{crabclient_version}.%{version_suffix}
-%define wmcore_version     1.3.6.crab2
-%define crabserver_version v3.201026
+%define wmcore_version     1.3.6.crab4
+%define crabserver_version v3.201104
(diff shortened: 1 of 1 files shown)


smuzaffar:
please test

belforte:
@smuzaffar a question about procedure: this crab client version was merged in latest release and is now in production
```
belforte@lxplus714/TC3> /cvmfs/cms.cern.ch/common/crab --version
CRAB client v3.201110
```
Thanks for that ! But it did not make it to the IB:
```
belforte@lxplus714/TC3> /cvmfs/cms-ib.cern.ch/nweek-02654/common/crab-prod --version
CRAB client v3.201026
```
in this particular case, we appreciate the speed and all is fine, but in general I would have expected that things need to be in IB for some time (one day?) before making it to the main area. Maybe I was simply assuming wrongly ?

smuzaffar:
@belforte , we started to build 11.2.0.pre9 release before the IBs , that is why the externals for the release get deployed even before the IB :-)

belforte:
@smuzaffar sure. I could guess so. So it sounds like this "just can happen" and in case I am not so sure I better avoid updating crab-prod.spec close to a latest-release build (since the date is not firmly defined). That's all.

PR #6409, opened 2020-11-12
upgrade DBS to 3.15.1.

yuyiguo:
This DBS release fixed a bug that allowed run_num=1 to pass in when the run range is used.  Add more unit tests and fixed some because of fixing the bug. This also removed phys01/02 in  the view_instances_mg.json so we no longer need a special k8s view.


muhammadimranfarooqi:
+1 to cceb19bd04c2e75b952ded5512dba3ec797e73f5 on slc7_amd64_gcc630.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/6409/test_cceb19b-20201112-021503.log" target="_blank">here</a>.

PR #6420, opened 2020-11-13
Remove bootstrap bin and lib paths during the build stage

smuzaffar:
We noticed that during the build of a package, `cmsBuild` source the env of rpm (so that it can run `rpmbuild` commands). One side effect of this is that it sets `PATH` and `LD_LIBRARY_PATH` of `bootstrap-bundle` package. Libraries (e.g openssl , xml2, zma etc) and binaries from these paths can be picked up by the up stream packages and can override system libs. For example , if one wants to use `openssl` from system then `openssl` library from bootstrap bundle will hide it.

The change in rpm-preamble.file does not re-build any thing. It will be only use when a package (for some other reason) needs a rebuilt. In orde rto test this fully, I have changed gcc.spec to re-build all packages to see if this has any side effects,

### DIFF
--- a/dpm.spec
+++ b/dpm.spec
@@ -22,7 +22,7 @@ Provides: libdpm.so%{libsuffix}
 
 %prep
 rm -f %_builddir/DPM-%{downloadv}.src.tar.gz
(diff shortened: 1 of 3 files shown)


smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
please test

PR #6421, opened 2020-11-13
download dasgoclient binaries directly from github

smuzaffar:
Thanks to @vkuznet , now the dasgoclient binaries are directly available via dasgoclient github repository. This change proposed to use the new binaries directly from there,

Note that we are referrening all the archs sources here so make sure that all of them are available when we test PRs.


smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
please test

PR #6432, opened 2020-11-18
[Rivet][OpenMP] Disable OpenMP for aarch64

smuzaffar:
We noticed many Relvals fails for aarch64 due to OpenMP cannot allocate memory in static TLS block. This is happening as Rivet library is dynamically linked to gomp which CMSSW plugin manager loads using dlopen. We only see such crashes for aarch64 releases, so this PR suggests to disable openmp for aarch64.

### DIFF
--- a/rivet.spec
+++ b/rivet.spec
@@ -38,6 +38,12 @@ case %{cmsplatf} in
 esac
 
 autoreconf -fiv
+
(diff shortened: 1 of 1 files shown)


smuzaffar:
please test workflow 1331.0 for slc7_aarch64_gcc820

smuzaffar:
looks good for aarch. There nothing changed for other archs.

PR #6441, opened 2020-11-23
bump python versions to pip packages

davidlange6:
@smuzaffar - fwiw, it looks like we are quite behind in setup tools.. I think it blocks updating Tqdm (maybe I've misunderstood the errors..)


davidlange6:
please test

smuzaffar:
@davidlange6 , yes we are far behind, setup tools and pip versions. we should be able to update setup tools though

smuzaffar:
+externals

PR #6455, opened 2020-11-26
[OpenLoops] do not build pplljj_ew for aarch64 as it again fails to build

smuzaffar:
With GCC9 on aarch64, we again have openloops failing to build due to large function size. Removing another big object for aarch64 allow us to build

### DIFF
--- a/openloops.spec
+++ b/openloops.spec
@@ -24,9 +24,16 @@ born_optimisation = -O2
 loop_optimisation = -O0
 link_optimisation = -O2
 EOF
+
+%ifarch aarch64
+grep -v '^pplljj_ew$' %{_sourcedir}/openloops-user.coll > openloops-user.coll
+%else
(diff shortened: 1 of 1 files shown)


smuzaffar:
please test

smuzaffar:
@SiewYan , @jonaslindert openloops once again failing for aarch64 with GCC9.3. Error is same as we had seen in past https://github.com/cms-sw/cmsdist/pull/5215. 
`common_flags = -mcmodel=large` flag does not work with -fPIC on arm this means we can not build shared libraries. `-Os` takes too much time (I had killed the job after 16 hours and it was still building).

Looks like removing `pplljj_ew` from the user fed list of libs allows it to build. So for aarch64 only, I have removed `pplljj_ew`. No idea if this will break something at runtime or not.

SiewYan:
@smuzaffar , thanks for bringing this up. I am looking for an alternative solution to decouple the libraries installation from cmssw.

PR #6460, opened 2020-11-27
BuiidRule: Allow to set flags for header checks via BuildFile

smuzaffar:


smuzaffar:
PR are tests are not going to run the new change. I am merging it and building an IB with it

PR #6461, opened 2020-11-27
[BuildRuiel] Do not drop optimization flags for header checks

smuzaffar:
GCC seems to work with optimization flags for header checks


smuzaffar:
test parameters:
- full_cmssw = true

smuzaffar:
please test

Issue #5324, opened 2019-11-03
Herwig 7 migration to python3


fabiocos:
Looking at the failures in the PY3 IB I see that test workflow 511.0 (Herwig7 + external ME) fails due to a trivial issue in an internal python script:

```
&gt;&gt;&gt; Compiling MadGraph amplitudes. This may take some time -- please be patient.
&gt;&gt;&gt; In case of problems see ./Herwig-scratch/Build/MadGraphAmplitudes/MG.log for details.

  File "/cvmfs/cms-ib.cern.ch/nweek-02600/slc7_amd64_gcc820/external/herwigpp/7.1.4-fdffeh2/bin/mg2herwig", line 63
    if it.replace("-","").isdigit():
                                   ^
TabError: inconsistent use of tabs and spaces in indentation
Error: Failed to initialize amplitude 'MadGraph'
```

happening in https://phab.hepforge.org/source/herwighg/browse/herwig-7-1/MatrixElement/Matchbox/External/MadGraph/mg2herwig.py.in$63

A closer inspection shows that there are other issues (for instance print statement syntax). There are not so many python scripts in this generator:

```
07:50 cmsdev25 561> find ./Herwig-7.1.4 -name "*.py" -print
./Herwig-7.1.4/Models/Feynrules/python/ufo2peg/particles.py
./Herwig-7.1.4/Models/Feynrules/python/ufo2peg/vertices.py
./Herwig-7.1.4/Models/Feynrules/python/ufo2peg/helpers.py
./Herwig-7.1.4/Models/Feynrules/python/ufo2peg/lorentzparser.py
./Herwig-7.1.4/Models/Feynrules/python/ufo2peg/__init__.py
./Herwig-7.1.4/Models/Feynrules/python/ufo2peg/collapse_vertices.py
./Herwig-7.1.4/Models/Feynrules/python/ufo2peg/write_paramcard.py
./Herwig-7.1.4/Models/Feynrules/python/ufo2peg/converter.py
./Herwig-7.1.4/Models/Feynrules/python/ufo2peg/check_lorentz.py
./Herwig-7.1.4/Models/Feynrules/exampleInputs/particles.py
./Herwig-7.1.4/Models/Feynrules/exampleInputs/couplings.py
./Herwig-7.1.4/Models/Feynrules/exampleInputs/vertices.py
./Herwig-7.1.4/Models/Feynrules/exampleInputs/write_param_card.py
[...]

PR #6462, opened 2020-11-27
Upgrade Herwig to v7.2.1, and add patch to fix crash when reading lhe…

Dominic-Stafford:
Update Herwig v7.2.0->v7.2.1, and add patch which fixes crash when reading LHE files with comments in weight groups.

Tested by checking read-in of LHE files with problematic comments from Madgraph now runs without problems, and using runTheMatrix workflow 535 and Matchbox test command from the Herwig twiki

### DIFF
--- a/FxFxFileReader.patch
+++ b/FxFxFileReader.patch
@@ -1,0 +1,63 @@
+--- Herwig-7.2.1/MatrixElement/FxFx/FxFxFileReader.cc	2020-10-11 13:40:21.000000000 +0300
++++ FxFxFileReader_fix.cc	2020-10-11 17:02:37.000000000 +0300
+@@ -527,6 +527,14 @@
+        */
+       if(readingInitWeights_sc && !cfile.find("</weightgroup") && !cfile.find("<weightgroup")) {
+ 	hs = cfile.getline();
++	//cout << "hs=" << hs << endl;
++	//cout << "weightinfo= " << weightinfo << endl;
++	//fix for potential new lines:
++	if(!cfile.find("<weight") and !cfile.find("</weightgroup")) {
++	  weightinfo = weightinfo + hs;
++	  //cout << "weightinfo fixed= " << weightinfo << endl;
++	  continue;
++	}
+ 	istringstream isc(hs);
+ 	int ws = 0;
+ 	/* get the name that will be used to identify the scale 
+@@ -544,10 +552,10 @@
+ 	unsigned firstLim = hs.find(startDEL); //find start of delimiter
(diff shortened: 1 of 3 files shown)


smuzaffar:
please test

smuzaffar:
assign generators

smuzaffar:
+externals

SiewYan:
The upgrade looks all good to me. Before signing off, @Dominic-Stafford , may i know if the version upgrade is anticipated with the external patch? Or the patch is not part of the version upgrade but private fix? Thanks

Dominic-Stafford:
Yes, I received this patch from the authors, and it assumes 7.2.1, so I decided to do the upgrade to keep things consistent. I believe the only differences between 7.2.0 and 7.2.1 are bug fixes, but the recommendation from the authors is to stay with the latest version. Thanks

SiewYan:
@Dominic-Stafford ,Perfect, thanks

SiewYan:
+1

silviodonato:
+1

mrodozov:
After this PR,
a relval is failing due to Herwig not fully compatible with python3
https://cmssdt.cern.ch/SDT/cgi-bin/logreader/slc7_amd64_gcc900/CMSSW_11_3_PY3_X_2020-12-14-2300/pyRelValMatrixLogs/run/511.0_QCD_Pt-30_13TeV_aMCatNLO_herwig7+QCD_Pt-30_13TeV_aMCatNLO_herwig7+HARVESTGEN/step1_QCD_Pt-30_13TeV_aMCatNLO_herwig7+QCD_Pt-30_13TeV_aMCatNLO_herwig7+HARVESTGEN.log#/
, for which we have also this opened:
https://github.com/cms-sw/cmsdist/issues/5324
Would you please provide a patch for Herwig or should we go back to the prev version ?

Dominic-Stafford:
Hi @mrodozov. I'm surprised this issue only appeared with this pull request, as python3 is so far not supported in any versions of Herwig (though I think workflow 511.0 may at one point have been removed from the standard tests, so maybe this is why). I have asked the Herwig authors when they will be able to issue a python3 compatible release- this is something they are currently working on, so hopefully this should be sorted soon.

agrohsje:
Hi @mrodozov. The patch and so this PR is c++ code. The problem is not related to it and reverting will not help. We need to migrate H7 to Python3. Let's see what feedback Dominic gets from the authors. We can also disable this workflow until the H7 problem is fixed. If H7 involves only very little python we could also think about fixing with additional patches. @Dominic-Stafford can you check how much code would be affected?

PR #6482, opened 2020-12-03
bazel: download the linux binary installer 

gartung:
Update to 3.7.1. Download the binary installer. Bazel would not build on SL7 with newer java version.


smuzaffar:
@gartung , what about non-x86_64 archs?

gartung:
You are correct. There are only x86_64 binaries for linux, macos and windows

https://github.com/bazelbuild/bazel/releases/tag/3.7.1

PR #6487, opened 2020-12-07

mrodozov:
please test

### DIFF
--- a/root.spec
+++ b/root.spec
@@ -3,8 +3,8 @@
 ## INITENV +PATH PYTHON3PATH %{i}/lib
 ## INITENV SET ROOTSYS %{i}
 
-%define tag 33c382d4bebfbc1d59b6536d36ff9f246e83e4a5
-%define branch cms/master/ec0bb25271
+%define tag a2571656abd79f3c44e2d4e6d9016fc1f392310f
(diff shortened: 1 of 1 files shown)


mrodozov:
please test

mrodozov:
ROOT changed Minuit logging here in master:
root-project/root#6705
and it's only in master, which means either we ask them to put it in 6-22 and fix it once or device something else.
To use the new logging nearly but not exactly straightforward so authors of
RecoLocalCalo/HcalRecAlgos should have to setup the MnPrint:: SetLevel -> Verbose and some other printing

davidlange6:
I realize that we have copied and modified some root code here.. The trouble with that being that now we should in theory re-copy and re-modify it.. 

Practically some easy alternatives are
a) use the cms message logger for printouts (advantage - will behave like other cms code)
b) add the ifdefs for these - eg, from here:
https://github.com/root-project/root/pull/6705/files#diff-0dee475476c5a425f99aedd676a201955d915d7622abcfe256a2d1fcccc36872L105

mrodozov:
I substituted all calls with LogInfo since they were just printouts anyway 
https://github.com/cms-sw/cmssw/pull/32440

mrodozov:
please test

PR #6488, opened 2020-12-07
Updated root module to tip of branch master

mrodozov:
please test

### DIFF
--- a/root.spec
+++ b/root.spec
@@ -3,8 +3,8 @@
 ## INITENV +PATH PYTHON3PATH %{i}/lib
 ## INITENV SET ROOTSYS %{i}
 
-%define tag 02fe854d6a44ced526c06e23e011b66142c4bf6d
-%define branch cms/master/ed07d0a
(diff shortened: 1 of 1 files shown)


mrodozov:
please test

mrodozov:
ROOT changed Minuit logging here in master:
https://github.com/root-project/root/pull/6705
and it's only in master, which means even we ask them to put it in 6-22 and fix it once or device something else.
To use the new logging nearly but not exactly straightforward so authors of 
RecoLocalCalo/HcalRecAlgos should have to setup the MnPrint:: SetLevel -> Verbose and some other printing

davidlange6:
hi @mrodozov - I will merge this as its the way forward to add a needed llvm patch (llvm inside of root) for gcc9

mrodozov:
its fine its just now it will also needs the cmssw change for minuit

davidlange6:
Yes, If the error in RecoLocalCalo becomes the worst problem in the modules ib, I’ll be happy!

PR #6491, opened 2020-12-08
[10.6] Update RecoEgamma-ElID tag to V01-01-04

mrodozov:
see https://github.com/cms-data/RecoEgamma-ElectronIdentification/pull/17


silviodonato:
merge

PR #6492, opened 2020-12-09
Update CLHEP to 2.4.4.0

mrodozov:
please test
the external builds with g4 and vecgeom 117, 
checking if this version will cause comparison differences etc.
it's the shortest way to get new geant4 branch changes, see
https://github.com/cms-sw/cmssw/pull/32420


mrodozov:
abort
forgot the cmssw pr

PR #6506, opened 2020-12-13
update crab-pre to same version as prod and dev. Ol once does not wor…

belforte:
…k after phedex retirement anyhow


smuzaffar:
please test

PR #6508, opened 2020-12-15
make crab-pre point to previous tag of crabserver

belforte:


belforte:
this makes CRAB Client to use again cmsweb.cern.ch for talking to crab server in place of cmsweb-k8s-prod.cern.ch. The latter works finely and is fully validated for submission etc.but the crabserver/UI javascript has some issues, in case we don't manage to fix them soon, this will be a fallback for those (few users) who relies on it.

smuzaffar:
please test

smuzaffar:
+externals
merging it as externals build fine. There is nothing in cmssw which actually use crab, so no need to wait for rest of the tests

PR #6526, opened 2021-01-04
[cms-common] default setting for the new SITECONFIG_PATH

smuzaffar:
Includes https://github.com/cms-sw/cms-common/pull/2 which now sets new SITECONFIG_PATH env variable


smuzaffar:
please test

PR #6529, opened 2021-01-04
Add h5cpp as hdf5 cpp interface

mrodozov:
Following this thread:
https://hypernews.cern.ch/HyperNews/CMS/get/swDevelopment/3582/2/1/1/1/2/2/1.html
@grasph

### DIFF
--- a/cmssw-tool-conf.spec
+++ b/cmssw-tool-conf.spec
@@ -37,6 +37,7 @@ Requires: glimpse-toolfile
 Requires: gmake-toolfile
 Requires: gsl-toolfile
 Requires: highfive-toolfile
+Requires: h5cpp-toolfile
 Requires: hector-toolfile
 Requires: hepmc-toolfile
 Requires: heppdt-toolfile
--- a/h5cpp-toolfile.spec
+++ b/h5cpp-toolfile.spec
@@ -1,0 +1,24 @@
+### RPM external h5cpp-toolfile 1.0
+Requires: h5cpp
+%prep
+
+%build
+
+%install
+
+mkdir -p %i/etc/scram.d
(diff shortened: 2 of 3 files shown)


mrodozov:
please test

grasph:
Thanks @mrodozov. How can I test it ? I usually test my code within a CMSSW release environment. How can I set it up to use one from this branch ?

In parallel, we have been working in porting the code to use directly the C-API for the longer-term solution.

Philippe.

mrodozov:
get this scripts:
https://github.com/cms-sw/cms-bot/blob/master/test-prs.sh
and run it on a cmsdev* machine with this arguments:
`./cms-bot/test-prs.sh -c IB/CMSSW_11_3_X/master -a slc7_amd64_gcc900 cms-sw/cmsdist#6529 -r CMSSW_11_3_X_2021-01-04-1100`
or any  release for that arch.
This will setup the library and setup it as available tool for the release. Then you get into the CMSSW area, cmsenv
and add a package that you want to use the lib with, add the tool in the BuildFIle.xml by name, it's h5cpp. 
Then it should be as described on their github page - you add a(the?) header like:
`#include <h5cpp/hdf5.hpp>`
and the lib should be available , you can try `scram tool info h5cpp` to check it.
Let us know if you find it useful, so we'd know if we have to integrate it

grasph:
Hello @mrodozov,

While testing the library I realize that I misunderstood the PR. I thought it was the HDF5 library compiled with the c++ option.

We have followed a different path and switched to the c-api, considering the lack of visibility on which C++-API will take the lead.

Therefore, for this work we won't use h5cpp. If it 's kept in CMSSW releases,  we may eventually use it as the c++ api might be more convenient than the C one.

Thanks a lot for your effort.

Philippe.

mrodozov:
I'll se what I can do

PR #6533, opened 2021-01-06
added new root tools for histpainter and imt

smuzaffar:
This PR added two new root tools
- HistPainter
- Imt
CMSSW UBSAN IBs failed to build due to missing linkage to HistPainter library. We propose to add these tools and then update cmssw/CondCore/SiStripPlugins/plugins to depend on HistPainter


smuzaffar:
please test
FYI @Dr15Jones  and @makortel

smuzaffar:
+externals

PR #6543, opened 2021-01-11
Update to cms-data/RecoEgamma-ElectronIdentification/pull/16

silviodonato:
See https://github.com/cms-data/RecoEgamma-ElectronIdentification/pull/16
and https://github.com/cms-sw/cmssw/pull/32391


silviodonato:
merge

PR #6547, opened 2021-01-12
remove ldscript from triton build to keep typeinfo

kpedro88:
Thanks to the quick response from the Triton team (https://github.com/triton-inference-server/server/issues/2404), a minor modification to the `CMakeLists.txt` to remove lines relating to an `ldscript` restores the `typeinfo` in the shared library. (It increases the shared library size by 25%, but it's still only 4MB, so I don't think we care very much.)


kpedro88:
resolves https://github.com/cms-sw/cmssw/issues/32630

kpedro88:
please test

smuzaffar:
+externals

PR #6549, opened 2021-01-13
Update Pythia8 to 230 in CMSSW_7_1_X

silviodonato:
This should be the minimal change to update Pythia8 to 230 in CMSSW_7_1_X

### DIFF
--- a/pythia8.spec
+++ b/pythia8.spec
@@ -6,9 +6,14 @@ Source: git+https://github.com/%github_user/%{n}.git?obj=%{branch}/%{tag}&export
 
 Requires: hepmc lhapdf
 
+%if "%{?cms_cxxflags:set}" != "set"
+%define cms_cxxflags -std=c++0x
+%endif
+
 %prep
 %setup -q -n %{n}-%{realversion}
 
+export USRCXXFLAGS="%cms_cxxflags"
 ./configure --prefix=%i --enable-shared --with-hepmc2=${HEPMC_ROOT} --with-lhapdf6=${LHAPDF_ROOT}
 
 %build
--- a/pythia8.spec
+++ b/pythia8.spec
@@ -1,22 +1,15 @@
-### RPM external pythia8 226
-
-Requires: hepmc lhapdf
-Requires: boost
-
-%define tag 7ea2c89a92fdfe2f110a1a669c0f3b185ec2978b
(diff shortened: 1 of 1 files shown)


silviodonato:
please test

qliphy:
assign @cms-sw/generators-l2   @cms-sw/pdmv-l2

qliphy:
please test

qliphy:
@mkirsano @cms-sw/generators-l2  We are building a special 7_1_X release with Pythia8.230, under the requirement from pdmv.  
Would you please have a look at the above compilation error message?  Thanks.

mkirsano:
What I see for the moment is that the author's version compiles, the one from cms git - not.

SiewYan:
Hi, I did some digging as well, it looks like the defined type shared_ptr is not defined in cmsexternal's version of pythia8.

`include/Pythia8Plugins/LHAPDF6.h:26:5: error: 'shared_ptr' in namespace 'std' does not name a type
`
The author version defines shared_ptr in SharedPointers.h [1]


[1] http://home.thep.lu.se/~torbjorn/doxygen83/SharedPointers_8h_source.html

mkirsano:
I remember now that story. The compiler in 7_1_X is too old (gcc 4.8). For this reason the upgrade to pythia8 230, patched against memory leak in lhapdf interface, was not backported to 7_1_X. The patch did not compile by 4.8.

mkirsano:
Upgrade to a higher version of pythia8 could be a simplest solution.

qliphy:
Thanks @mkirsano @SiewYan ,  so we can go for pythia8.303 directly , right?

mkirsano:
If higher version is OK, I would suggest 240. pythia8 303 means many changes in the interface and many differences in physics

agrohsje:
Thanks for the investigations. I will ping Alp to see if they are ok with 240.

alpakpinar:
Hi @agrohsje , all, thanks a lot for the work. I think we should be okay with Pythia 8 240, I'll quickly check with the analysis team and I'll let you know soon.

Thanks,
Alp

alpakpinar:
Hi @agrohsje , all, Pythia 8 240 would also work for us, thank you!

smuzaffar:
closing, this is not needed any more.

PR #6554, opened 2021-01-14
Update DD4Hep to latest on master

mrodozov:


mrodozov:
please test

mrodozov:
@cvuosalo @ianna this needs cmssw update

ianna:
@mrodozov - yes, thanks!
@cvuosalo - the recent changes to testDDFilteredView should be removed - the reference values are already expecting a correct result from an evaluator:
```diff
Get attl from hf as double values:
attl 8.09653e-05 == 0.000809653
F

DDFilteredView.cppunit.cc:100:Assertion
Test name: testDDFilteredView::checkFilteredView
assertion failed
- Expression: abs(i - refdattl_[count]) < 10e-6

Failures !!!
Run: 1   Failure total: 1   Failures: 1   Errors: 0

---> test testDD4hepFilteredView had ERRORS

```

cvuosalo:
@mrodozov How is CMSSW being compiled against this version of DD4hep? When CMSSW code includes `DD4hep/DDParsers/include/Evaluator/DD4hepUnits.h`, the `HAVE_GEANT4_UNITS` compiler directive needs to be set. Is that what happens in this PR?

cvuosalo:
@ianna In the line:
```
attl 8.09653e-05 == 0.000809653
```
The value on the left is the calculated value, and on the right it is the reference value. Here, the calculated value is too small. The current code does this:
```
 i *= dd4hep::cm;  // Convert DD4hep units to /cm
```
For Geant4 units `dd4hep::cm` = 10, but I don't think CMSSW is including the Geant4 version of `DD4hepUnits.h`. I think CMSSW is getting the ROOT version of  `DD4hepUnits.h`, `dd4hep::cm` = 1, and the value returned by `DDFilteredView::get()` is multiplied by 1 instead of 10 and turns out too small.

mrodozov:
test parameters:
- full_cmssw = true

mrodozov:
please test

mrodozov:
@cvuosalo this flag
`-DDD4HEP_USE_GEANT4_UNITS=1`
is set:
https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-e91374/12291/external/dd4hep/v01-15x-c618cc8da6965545fc12580382847581/log

smuzaffar:
@mrodozov , please update https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_11_3_X/master/dd4hep-toolfile.spec and add something like
```
<flags cppdefines="DD4HEP_USE_GEANT4_UNITS=1"/>
```

add this line here https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_11_3_X/master/dd4hep-toolfile.spec#L24

smuzaffar:
lets fix the toolfile and re-run the tests and see if we unit tests still fail

davidlange6:
> @mrodozov , please update https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_11_3_X/master/dd4hep-toolfile.spec and add something like
> 
> ```
> <flags cppdefines="DD4HEP_USE_GEANT4_UNITS=1"/>
> ```
> 
> add this line here https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_11_3_X/master/dd4hep-toolfile.spec#L24

having this defined in two places seems a plan for problems. Why can't dd4hep define this internally in a header according to the way make is run?

smuzaffar:
This is a build time configuration parameter. So either we can patch dd4hep (in our spec) to set it in side one of the dd4hep header (which will avoid the duplication) or we have to use the scram toolfiles

smuzaffar:
ah right ( https://github.com/cms-sw/cmsdist/pull/6554/files#diff-71d575042aa0c5063c7a3b3456d09216a283b3805936d7ee23702fac27a752e9R27 ), I did not noitce that it is set via cmake ( i thought we were setting it at make time). Yes dd4hep should set it in one of its header

mrodozov:
actually the macro to be setup is different from the cmake flag. I made the same mistake :) The C++ macro to be defined is called 
`HAVE_GEANT4_UNITS`
while the cmake variable is called 
`DD4HEP_USE_GEANT4_UNITS`
HAVE_GEANT4_UNITS in the toolfile then ? 
I have a local setup to tst until it works, then I'm adding it to the toolfile

smuzaffar:
@mrodozov , I copuld not find any ref to HAVE_GEANT4_UNITS but I see that DD4HEP_USE_GEANT4_UNITS is the marco used in few dd4hep files e.g. https://github.com/AIDASoft/DD4hep/blob/master/DDParsers/include/Evaluator/DD4hepUnits.h#L128 
I would suggest that for now update the toolfile to set the DD4HEP_USE_GEANT4_UNITS cppdefine so that we can test the dd4hep changes

Currently dd4hep record this marco in the its generated cmake files 
```
./cmake/DD4hepConfig-targets.cmake:  INTERFACE_COMPILE_DEFINITIONS "BOOST_SPIRIT_USE_PHOENIX_V3;DD4HEP_USE_GEANT4_UNITS=1"
```
but as cmssw do not use cmake so that is no use of us. I would suggest that ask dd4hep developers to define this macro in their header file if dd4hep is build with DD4HEP_USE_GEANT4_UNITS ON. this will allow projects without cmake support to use it without re-definiing it in their build system

mrodozov:
Yes, I just found it was renamed. I'll right an issue for the macro to be defined

smuzaffar:
please test

mrodozov:
I tried that last night (the test was still failing)
```
diff --git a/dd4hep-toolfile.spec b/dd4hep-toolfile.spec
index 0941f7c..b4b3f71 100644
--- a/dd4hep-toolfile.spec
+++ b/dd4hep-toolfile.spec
@@ -21,6 +21,7 @@ cat << \EOF_TOOLFILE >%i/etc/scram.d/dd4hep-core.xml
   </client>
   <runtime name="PATH" value="$DD4HEP_CORE_BASE/bin" type="path"/>
+  <flags cppdefines="DD4HEP_USE_GEANT4_UNITS=1"/>
   <use name="root_cxxdefaults"/>
   <use name="root"/>
   <use name="boost"/>
```
anyway, the tests are reseting the env properly, lets see

mrodozov:
actually some tests are running this way, but fails for another reason.
there is this warning:
`DD4hep           WARN  ++ Using globally Geant4 unit system (mm,ns,MeV)`
meaning the macro is set (scram also shows the -DDD4HEP_USE_GEANT4_UNITS=1 when ran with -v)
at the end the test says
OK (1)
which I think it's the internal "i ran fine" flag
and fails after that

for example with the flag this test:
OK (1)
---> test testDD4hepDDSolid succeeded
runs in my local area

but in the prev test (without the flag) this test is failing first in the list

mrodozov:
hm , could it be related to this:
https://github.com/cms-sw/cmssw/issues/32181#issuecomment-763117890
when I run this test by hand on cmsdev24 with only changing dd4hep-toolfile some of the tests run some doesn't but they doesn't fail with the xml parser
this tests fail only with the xml parser

mrodozov:
please test
full cmssw was on, lets try with a patch only

mrodozov:
@cvuosalo @ianna the flag is properly set and failures has nothing to do with it, please advise. thank you

cvuosalo:
@ianna @civanch Advice from @ianna and the DD4hep team is needed. I sent email about the question. It's not clear to me whether the fault is in our unit test or in DD4hep.

cvuosalo:
After discussion, we decided that the failing unit tests need a fix. I will work on that.

cvuosalo:
@mrodozov PR #32721 should fix the `DDFilteredView` error. Could you please test with it?

cvuosalo:
@mrodozov I don't understand the `GeometryDTGeometryBuilderTestDriver` error. I tried to test with this PR by using `test-prs.sh` but did not succeed. Could you please give me simple instructions for how I can test `GeometryDTGeometryBuilderTestDriver` using this PR?

mrodozov:
please test

mrodozov:
please test

mrodozov:
want to check with IB later than this PR was merged 
https://github.com/cms-sw/cmssw/pull/32721

PR #6563, opened 2021-01-19
[11.1.X] Retrained particle ID model and energy regression for HLT TDR

smuzaffar:
backport of #6562

### DIFF
--- a/data/cmsswdata.txt
+++ b/data/cmsswdata.txt
@@ -15,6 +15,7 @@ RecoMET-METPUSubtraction=V01-01-00
 EgammaAnalysis-ElectronTools=V00-02-00
 PhysicsTools-NanoAOD=V01-01-00
 RecoBTag-Combined=V01-03-00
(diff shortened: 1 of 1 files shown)


smuzaffar:
please test

smuzaffar:
@slava77 , I assume that comparison diff here are same as we have seen for https://github.com/cms-data/RecoHGCal-TICL/pull/2 ? If you are OK with these then we can merge it for 11.1.X too.

silviodonato:
urgent
(for HLT TDR)

slava77:
> @slava77 , I assume that comparison diff here are same as we have seen for [cms-data/RecoHGCal-TICL#2](https://github.com/cms-data/RecoHGCal-TICL/pull/2) ? If you are OK with these then we can merge it for 11.1.X too.

yes, they look quite similar; it should be OK to go ahead wrt reco.

smuzaffar:
+externals
thanks @slava77  for the confirmation

PR #6568, opened 2021-01-20
Upgrade cx-Oracle from 5.2.1 to 7.3.0

amaltaro:
This PR brings the following changes:
* upgrade py2-cx-oracle from 5.2.1 to 7.3.0
* remove old py2-cx-oracle51.spec spec
* move ASO and CRABServer to the latest cx-oracle version (7.3.0)

This has been (or is being) discussed in this HN thread:
https://hypernews.cern.ch/HyperNews/CMS/get/webInterfaces/1763.html

however, people didn't raise any strong concerns not to upgrade it and proceed with further tests in cmsweb-testbed.

It's already written in that email thread, but here is the list of services affected by this upgrade:
PHEDEX-web.spec
confdb.spec
crabcache.spec
dbs3-migration.spec
dbs3.spec
frontend.spec
overview.spec
popdbweb.spec
sitedb.spec
t0.spec
t0wmadatasvc.spec
victorweb.spec
webtools.spec
wmagent.spec

Release notes can be found here: https://cx-oracle.readthedocs.io/en/latest/release_notes.html#version-7-3-december-2019

And big thanks to @smuzaffar and @belforte for getting this spec fixed.
FYI @yuyiguo .


muhammadimranfarooqi:
+1 to ded23ad7d532ddbbf7b2cd373b3379efe3bbd330 on slc7_amd64_gcc630.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/6568/test_ded23ad-20210120-085003.log" target="_blank">here</a>.

belforte:
@amaltaro we also need to change crab*.spec from py2-cx-oracle51 back to the common py2-cx-oracle

amaltaro:
Thanks Stefano. I have pushed in a second commit with those changes (and deleted the old cx-oracle51 spec). Can you please check again?

muhammadimranfarooqi:
+1 to eb709b8644aec7e202caf3c5269f602912a676df on slc7_amd64_gcc630.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/6568/test_eb709b8-20210120-095503.log" target="_blank">here</a>.

belforte:
thanks @amaltaro all OK. Sorry for delay, car shop was late
.BTW we can delete asyncstageout.spec. I can make and ad hoc PR later. So you can push this in testbed now.
Side note: can you fstop Jenkins from trying to build on slc6 ? So checks do not stay pending forever.

amaltaro:
>  Side note: can you fstop Jenkins from trying to build on slc6 ? So checks do not stay pending forever.

I'm afraid this change has to be done by Imran. I do not know how that has been setup.

amaltaro:
Now that we have services deployed in cmsweb-testbed, we noticed that:
* confdb fails to start up
* DBS reports the following warning - but it seems to be working okay - :
```
/data/srv/beHG2102a/sw/slc7_amd64_gcc630/external/py2-sqlalchemy096/0.9.6/lib/python2.7/site-packages/sqlalchemy/engine/strategies.py:78: SAWarning: cx_Oracle is compiled under Python 2.xx using the WITH_UNICODE flag.  Consider recompi
ling cx_Oracle without this flag, which is in no way necessary for full support of Unicode. Otherwise, all string-holding bind parameters must be explicitly typed using SQLAlchemy's String type or one of its subtypes,or otherwise be pa
ssed as Python unicode.  Plain Python strings passed as bind parameters will be silently corrupted by cx_Oracle.
```

I don't remember seeing this warning when testing WMAgent. I fail to spot it in the CRABServer logs as well.

So I suspect it's a mix of the old SQLAlchemy version with this new cx-oracle.
We will have to make sure the DBS functionality isn't affected by this; Yuyi said to be able to validate it from Friday on.

belforte:
CRABServer lists a depenency from SQLAlchemy but it does not use it. I heard from Marco Mascheroni the other day about this, he's not sure why we have that dependency, maybe to please some WMCore functionality that we do not use, maybe we can remove it and things will still work.
OTOH the message about Unicode for DBS is suspiciously close to the known problem of Unicode in query outputs passed again as query input. If there's a way to make cx-oracle return ASCII strings, we may certainly benefit !

PR #6573, opened 2021-01-20
Remove spec file for SiteDB

amaltaro:
I guess it's been almost a year that SiteDB service has been shutdown, so we better stop building its RPMs as well.
The cleaner we can get cmsdist/COMP, the easier it is to move forward with services that are still very much relevant to CMS.


muhammadimranfarooqi:
+1 to 1513844e7005d29062c9c1a26de834f79b3dab8f on slc7_amd64_gcc630.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/6573/test_1513844-20210120-214503.log" target="_blank">here</a>.

PR #6574, opened 2021-01-20
Remove spec file for Overview and xRegexp

amaltaro:
I don't think this Overview service was ever put in production and, latest (and only) change to the dmwm/overview repository date back to 2012, so we better stop building its RPMs as well.
The cleaner we can get cmsdist/COMP, the easier it is to move forward with services that are still very much relevant to CMS.


muhammadimranfarooqi:
+1 to abd85e42ce19c308ba83127e131cb12339154508 on slc7_amd64_gcc630.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/6574/test_abd85e4-20210120-221855.log" target="_blank">here</a>.

PR #6575, opened 2021-01-20
Remove spec file for ConfDB and its dependencies

amaltaro:
As confirmed with the Trigger Coordinator, this service is no longer maintained and we can remove it from the CMSWEB umbrella, such that we can make cmsdist/COMP cleaner for future changes.


muhammadimranfarooqi:
+1 to 384fb6fe62e55d860d65887daf263e36f3f734b4 on slc7_amd64_gcc630.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/6575/test_384fb6f-20210120-220504.log" target="_blank">here</a>.

muhammadimranfarooqi:
@amaltaro 
I am wondering why you deleted spec files for obselete services?  Can't we leave them and only delete them from comp.spec?

amaltaro:
Imran, just replicating here what we talked over gtalk.
Yes, we could stop building them by removing them from the comp.spec, but it's still an spec file (and its dependencies) that we somehow have to maintain in our COMP branch, which might just make large changes/migrations difficult in the future.

Thanks for merging it btw.

PR #6557, opened 2021-01-16
Update setuptools for py2 and py3 and pip for py3

davidlange6:
Separate the py2 and py3 versions of setuptools to allow us to follow the last year of development after py2 support was dropped. These are now separate spec files.

This pr includes #6556 as all of those packages would be rebuilt anyway.

There are some new dependencies included. One to note is a number of new dependencies on wheel, which others have encountered when bumping setuptools. I just added a build requires in each case.

From discussion in this PR, pip is now updated for python3. While current pip releases nominally support python2, I quickly ran into dependency issues that I didn't want to solve for an obsolete stack. So it is updated for python3 only.

The one oddity is with scipy (as commented below). I don't think there is any regression, just a newly known dependency.

PR #6590, opened 2021-01-22
[CC8] Update setuptools for py2 and py3 and pip for py3

smuzaffar:
same as #6557 but for cc8 testing


smuzaffar:
please test for cc8_amd64_gcc9

PR #6611, opened 2021-02-02
Update t0.specs for release 2.2.2

germanfgv:
This release uses WMCore 1.4.3.patch2

### DIFF
--- a/t0.spec
+++ b/t0.spec
@@ -1,11 +1,11 @@
-### RPM cms t0 2.2.1
+### RPM cms t0 2.2.2
 ## INITENV +PATH PATH %i/xbin
 ## INITENV +PATH PYTHONPATH %i/${PYTHON_LIB_SITE_PACKAGES}
 ## INITENV +PATH PYTHONPATH %i/x${PYTHON_LIB_SITE_PACKAGES}
 
 %define webdoc_files %{installroot}/%{pkgrel}/doc/
 
-%define wmcver 1.4.1.patch4
+%define wmcver 1.4.3.patch2
 %define wmcpkg WMCore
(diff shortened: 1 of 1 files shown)


amaltaro:
@germanfgv perhaps we should hold off with this upgrade(?) I'm expecting to have a new WMAgent production release by the beginning of the next week (1.4.5.patchX).

germanfgv:
@amaltaro Our production node is using release 2.2.1 (WMCore 1.4.1.patch4). We were planning to upgrade to 2.2.2 this week, in preparation for next week's MWGR. If we wait for 1.4.5.patchX, we'd have to use 2.2.1 for MWGR. I have no problem with that, but I would like to know if we are missing some critical WMCore update by not using 1.4.3.patch2

amaltaro:
The only change that looks important to T0 is the one you made:

>     Removing file status condition from GetUnsubscribedDatasets query #10067

other than that, I think it would be okay to remain in wmcore 1.4.1 for another week or two.

Otherwise you will have two large validations to run, one based on wmcore 1.4.3 and another on 1.4.5.

Can you please remind me what are the dates for the MWGR?

germanfgv:
Ok, then we will use 2.2.1 for now and upgrade after 1.4.5.patchX is available. MWGR will take place from February 10th to 12th.

muhammadimranfarooqi:
+1 to aa81c535d31ae230670f4f69a5d7012161f4b22a on slc7_amd64_gcc630.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/6611/test_aa81c53-20210202-093504.log" target="_blank">here</a>.

PR #6638, opened 2021-02-10
add SYSTEM_INCLUDE to tensorflow

dan131riley:
When TensorFlow is compiled with optimization off, e.g. with `USER_CXXFLAGS="-g -O0"`, it fails to compile with several "integer expressions of different signedness" diagnostics that we promote to errors.  This PR adds the `SYSTEM_INCLUDE` flag so that these diagnostics are suppressed.  Detailed discussion is in https://github.com/cms-sw/cmssw/pull/32852


smuzaffar:
please test

smuzaffar:
+externals

PR #6639, opened 2021-02-10
update RecoJets-JetProducers to V05-13-00

slava77:
this is a backport from 11_X
the change https://github.com/cms-data/RecoJets-JetProducers/compare/V05-12-00...V05-13-00 adds new files on top of V05-12-00 and should be straightforward to merge


slava77:
@cmsbuild please test

to check that this works by itself and introduces no changes to the defaults

slava77:
> * Reco comparison results: 0 differences found in the comparisons
> * DQMHistoTests: Total failures: 1

These worked as expected.
I propose to merge soon.

mrodozov:
+externals

PR #6652, opened 2021-02-16
[SCRAMV3] do not fail if failed to create runtime cache file on read-only filesystem e.g. cvmfs

smuzaffar:


smuzaffar:
please test

PR #6656, opened 2021-02-17
ROOT next [6.22] Update to latest on 6.22

mrodozov:
include latest DD4Hep updates

### DIFF
--- a/root.spec
+++ b/root.spec
@@ -3,8 +3,8 @@
 ## INITENV +PATH PYTHON3PATH %{i}/lib
 ## INITENV SET ROOTSYS %{i}
 
-%define tag c1b93b0ceb48b2be280fd43c1eeaebaaac3f864f
-%define branch cms/v6-22-00-patches/ab1b7ca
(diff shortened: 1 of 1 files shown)


mrodozov:
please test

mrodozov:
please test

mrodozov:
@slava77 considering https://github.com/cms-sw/cmsdist/pull/6657 would you please have a look also this one
usually when I see non-master branch differences I attribute it to the differences in externals/OS, but this is very close to what we have in cmsdist/master (excluding dd4hep changes)
I think we have to go back also here :/

slava77:
> @slava77 considering #6657 would you please have a look also this one

Apart for 
> You potentially added 8146090 lines to the logs

the reco differences look OK
-  there are a bunch in 140.53 due to the DAS issue of getting an inconsistent input file compared to the baseline (transient?)
- - and a few logError/Warnings in workflows with mixing which can be ignored.

mrodozov:
the #6657 also looks ok on it's own so it's probably what's in dd4hep

PR #6659, opened 2021-02-19
add correctionlib and dependencies. 

davidlange6:
Also adds python-rapidjson and pydantics as dependencies.

Currently pip installs only the python interface. There is a C++ one as well that hopefully we get from a future pypi release

@nsmith-

### DIFF
--- a/pip/requirements.txt
+++ b/pip/requirements.txt
@@ -54,6 +54,7 @@ cachy==0.3.0 ; python_version>'3.0'
 cleo==0.8.1 ; python_version>'3.0'
 clikit==0.6.2 ; python_version>'3.0'
 cmsml==0.1.1
+correctionlib==1.1.0 ; python_version>'3.0'
 crashtest==0.3.1 ; python_version>'3.0'
 certifi==2020.12.5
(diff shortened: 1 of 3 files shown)


davidlange6:
please test

pip/requirements.txt
```diff
@@ -251,6 +253,7 @@ python-cjson==1.2.2;python_version<'3.0'
 python-daemon==2.2.4
 python-dateutil==2.8.1
 python-ldap==3.3.1
+python-rapidjson==1.0;python_version>'3.0'
```
nsmith-, line 256: Does this need to be there to build? The rapidjson headers should have been shipped with the correctionlib sdist, and we don't use rapidjson on the python side at all.

davidlange6:
Possibly not. Seemed otherwise useful to have.

smuzaffar:
@davidlange6 , please add `py3-correctionlib` to python_tools ? Can you please also add simple unit tests in cmssw for these new tools?

smuzaffar:
please test

smuzaffar:
+externals

PR #6703, opened 2021-03-07
New dasgoclient version

vkuznet:
Fix rare issue with socket too many open files which happen on queries with large amount of HTTP calls, see discussion in [1]

[1] https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/5945/1/1/1/2/1/2/1/1/1/1/1/1/2.html


smuzaffar:
please test

PR #6708, opened 2021-03-09
Upgrade to Herwig 7.2.2

Dominic-Stafford:
Upgrade Herwig to version 7.2.2 (which mainly contains bug fixes, but should also be python3 compatible). Also added patch for LesHoucheFileReader, which would crash if running over madgraph lhe files without merging (the corresponding issue with the FxFx reader was fixed in 7.2.2).

Tested with workflow 535 and matchbox test command command from Herwig twiki, and with two madgraph + Herwig examples: one with merging, one without.


smuzaffar:
please test workflow 535.0

smuzaffar:
assign generators

smuzaffar:
+externals

agrohsje:
+generators

qliphy:
+1

PR #6499, opened 2020-12-11
Rename herwigpp tool to herwig7

Dominic-Stafford:
Rename herwigpp.spec to herwig7.spec and similarly change the tool name so that it can be found by "scram tool info herwig7" and gen checking script. Tested with runTheMatrix workflow 535 and test command from Herwig twiki.

PR #6709, opened 2021-03-09
Rename herwigpp to herwig7, and upgrade to Herwig 7.2.2

Dominic-Stafford:
Backports  #6499  (Rename herwigpp tool to herwig7 so that it can be found by scram tool info and gen checking script) and #6708 (Upgrade Herwig to version 7.2.2, and include patch for madgraph lhe files without merging).

Tested with workflow 535 and matchbox test command command from Herwig twiki, and with two madgraph + Herwig examples: one with merging, one without.


mrodozov:
please test

mrodozov:
this reminds me you need something like this for 10_6 then 
https://github.com/cms-sw/cmssw/pull/32468

Dominic-Stafford:
Sorry, I've just made that, it's here: cms-sw/cmssw#33111

mrodozov:
and it also reminds me we are having trouble with one relval because madgraph and herwig needs to be python3 compatible

Dominic-Stafford:
Herwig 7.2.2 should be python3 compatible, though I haven't checked this yet

mrodozov:
+externals
look ok to me

qliphy:
+1

PR #6717, opened 2021-03-10
Updated root to tip of branch master (need to pick a fix)

mrodozov:
getting a fix for arm, see https://github.com/cms-sw/cmssw/issues/31123#issuecomment-795870758


mrodozov:
please test

smuzaffar:
+externals
but this is not going to fix aarch64 Ibs as root master is only built for amd64

PR #6731, opened 2021-03-15
Update to new version; add new prometheus version; add aps tools

vkuznet:


mrodozov:
please test

cmsmon-tools.spec
```diff
@@ -60,6 +61,14 @@ chmod +x stern_linux_amd64
 cp stern_linux_amd64 %i/stern
 cd -
 
+# install token-manager
+cd %{_builddir}
+curl -ksLO https://github.com/vkuznet/auth-proxy-server/releases/download/%apsver/auth-proxy-tools.tar.gz
```
smuzaffar: @vkuznet , why not add this to sources?

vkuznet: Shahzad, honestly I don't remember RPM spec command syntax, and I find it very difficult to find the appropriate docs for them. The setup syntax I obviously don't remember, e.g. how it untar/unzip all possible combinations, and it is much easier to call curl instead of using sources/setup. If you remember proper syntax feel free to let me know how to do it (I can't do man rpm synax, and googling rpm setup does not lead anywhere close to how RPM commands are structured and which options to use for different use-cases).

smuzaffar:
please test

cmsmon-tools.spec
smuzaffar: I mean you already have 4 sources defined here https://github.com/cms-sw/cmsdist/pull/6731/files#diff-15398a788af5b8c0e165e8cca9a44fb7637daebb4850bf66571da9512729a925L11-L14 , so why not just and this as a `Source4` and extract is just like you extract other sources https://github.com/cms-sw/cmsdist/pull/6731/files#diff-15398a788af5b8c0e165e8cca9a44fb7637daebb4850bf66571da9512729a925L24

smuzaffar: by the way the RPM doc here https://rpm-packaging-guide.github.io/ looks good. See the `%setup` section for details how to expand

vkuznet:
@smuzaffar I added sources, but for Source4 which is not a tarball I have no idea how to deal with it in rpm macro world. I also don't know how to refer to it if I call curl. Once again RedHat docs are not helpful and search for "how to download source file in rpm" or similar queires does not lead to what I need to know. If you know how to force in prep step a download of given source please let me know and I'll be happy to adjust the spec, otherwise for stern tool (which is just executable we need to download) I left the curl approach.

vkuznet:
otherwise, the build is fine now in my local area. Please test it.

smuzaffar:
@vkuznet , I have fixed the source4. For such files you can use `%{_sourcedir}` to access the downloaded file

smuzaffar:
please test

vkuznet:
thanks, I'll keep this in mind for other specs as we may have more and more just static executables from go tools.

smuzaffar:
+externals

PR #6734, opened 2021-03-16
[LLVM] Added EventSetupRecord::get checker

smuzaffar:
See https://github.com/cms-externals/llvm-project/pull/10

### DIFF
--- a/llvm.spec
+++ b/llvm.spec
@@ -8,7 +8,7 @@ Requires: gcc zlib python python3
 Requires: cuda
 AutoReq: no
 
-%define llvmCommit 9ac43ce9cfea653af380f1ac550c1171dab9a7cf
+%define llvmCommit 6afd8d28e7729e1256cc832c60e4be6830f05dc6
 %define llvmBranch cms/release/11.x/1fdec59
 %define iwyuCommit 5db414ac448004fe019871c977905cb7c2cff23f
 %define iwyuBranch clang_11


smuzaffar:
please test

smuzaffar:
+externals
FYI @gartung , this should include your changes on top of llvm 11.1

smuzaffar:
@gartung , the PR tests use `-enable-checker threadsafety -enable-checker cms -disable-checker cms.FunctionDumper` options ( https://github.com/cms-sw/cms-bot/blob/master/pr_testing/test_multiple_prs.sh#L687 ). Do they look good ?

smuzaffar:
please test with cms-sw/cmssw#33189
lets test with one of cmssw PR to see what static checks report

gartung:
For the static analyzer that is correct. Wherever you enable the clang-tidy checks, you need to change/add cms.*.

gartung:
There does not appear to be any restrictions on clang-tidy checkers so the new one should just run.
https://github.com/cms-sw/cmssw-config/blob/ef9d787cad2b4af040256fb0e14bbe7540a55351/SCRAM/GMake/Makefile.coderules#L24

smuzaffar:
for `clang-tidy`, checks are picked up from https://github.com/cms-sw/cmssw/blob/master/.clang-tidy  where we have non of cms checks enabled.

smuzaffar:
> There does not appear to be any restrictions on clang-tidy checkers so the new one should just run.
> https://github.com/cms-sw/cmssw-config/blob/ef9d787cad2b4af040256fb0e14bbe7540a55351/SCRAM/GMake/Makefile.coderules#L24

the  `USER_CODE_CHECKS` is empty for code checks ( https://github.com/cms-sw/cmssw-config/blob/ef9d787cad2b4af040256fb0e14bbe7540a55351/SCRAM/GMake/Makefile.coderules#L15 ) , so it just runs default checks from the .clang-tidy file

gartung:
OK. I will submit a pull request to add cms.* to .clang-tidy.

makortel:
I'm now confused. Would changing `.clang-tidy` make those checks run at `code-checks` time? Without that change, will they get run at the "PR test" time? (for this check we want the latter, at least for now)

smuzaffar:
Yes, now after ` .clang-tidy` change it will run at code-checks time. 
If I understand correctly, this clang-tidy check only prints warning without any code change , so I think this will not break any thing. But this also means that no one will look in to those warnings ( unless we start reporting warnings in the code-checks summary results).

gartung:
If we want the CMS checks to run any time clang-tidy is run then this PR is needed https://github.com/cms-sw/cmssw/pull/33197
If we want the CMS checks to only run for pull request checks then the environment variable USER_CODE_CHECKS needs to be set.

gartung:
The Event::getByToken check will actually change the code because it has a fixit replacement. I test that all of the replacements produced code that could be compiled when I wrote that checker.

gartung:
Correction. The environment variable USER_CODE_CHECKS_ARGS needs to be set for pull request testing and the warnings recorded.

makortel:
is `clang-tidy` currently run at the PR test phase (in addition to code-checks)?

smuzaffar:
No, .clang-tidy is only used at the time of code-checks

gartung:
Only for code-checks as far as I can tell. Running the code-checks with the environment variable could be added as part of the static analyzer check here https://github.com/cms-sw/cms-bot/blob/master/pr_testing/test_multiple_prs.sh#L687 or added as its own check in the same script and the warnings recorded.

makortel:
Let's backtrack a bit.

What w'd want is *some* way to flag deprecated functionality in PRs with different levels of enforcement. I can think of
1. flag code in IBs without impacting the dashboard
2. flag code in PRs in all files in packages checked out with `git cms-checkdeps -a`
3. flag code only in user local area (during `scram b` or a separate checker)
4. flag code in PRs (not in IBs), limit only to files modified in the PR 
   a. list warnings in the PR test report (allowing L2s to ignore if necessary for good reason)
   b. hard failure of PR tests
5. warnings/failures in IBs, visible in the dashboard

Currently we do
* 1 with static analyzer and various clang reports
* 2 with static analyzer
* 4b with code checks (`clang-format` and `clang-tidy`) and compiler itself (warnings and errors must be fixed)
* 5 (and 3) with compiler warnings and errors  (`[[deprecated]]` suggested in https://github.com/cms-sw/cmssw/issues/31021 would also fall here)

I'm currently looking for 4a for the `EventSetupRecord::get()`, likely elevating to 4b at some point in the future (eventually to 5).

3 or 4a could be useful for legacy EDModules as well (as mentioned in #31021).

Does 4a make sense, or would it be better to wait until we can enforce with 4b (in which case the `clang-tidy` at the code-checks step would be fine)?

If the answer is "yes", what would be the best way to do that? Make use of static analyzer information, use `clang-tidy`, use C++ attributes, something else?

gartung:
I think only the files changed in a pull request have clang-tidy code format checks run on them. The CMS clang-tidy test could be run only on the files changed as part of the pull request tests.

makortel:
Actually I think the `Event::getByToken()` check (`cms-handle`)  could be enabled in code-checks phase because it always proposes a fix and therefore applying the fix is trivial for the developer (and I'd imagine performing that check later in the PR test phase would be more annoying for the developer).

I'm fine with running clang-tidy again in the PR test phase for reporting warnings of the `EventSetupRecord::get()` check. Would a custom clang-tidy check be then the way we want to deal with similar API deprecation in the future (like legacy EDModules)?

smuzaffar:
I enabled `cms-handle` check for full cmssw and it suggested 6.5K changes in cmssw ( https://cmssdt.cern.ch/SDT/jenkins-artifacts/jenkins-test-code-format/CMSSW_11_3_X_2021-03-16-1100/64/ ) . Yes we can enable it for code-checks but better to first run a campaign to first get most of these fixed

smuzaffar:
I agree that for warnings about EventSetupRecord::get() check , we can run clang-tidy and any other cms check during PR static analysis tests. @gartung , can you please open a cms-bot PR to do this ?

makortel:
> Yes we can enable it for code-checks but better to first run a campaign to first get most of these fixed

Good point. It could also probably be useful to submit PRs first one-to-few packages manually before launching bigger campaign (and I have one question to Patrick first).

gartung:
https://github.com/cms-sw/cms-bot/pull/1513

gartung:
Alternative based on running clang-tidy with custom CMS checkers enabled.
https://github.com/cms-sw/cms-bot/pull/1514

PR #6735, opened 2021-03-16
[XRootd] Update to version 5.1.1

smuzaffar:


smuzaffar:
please test

mrodozov:
please test
checking something

mrodozov:
this is weird I ran it locally with -t and all - it works  (all this failing wf's) with the new xrootd, with a cmake_build_type=Debug 
I don't think it's because of the debug build. I also checked the local files they are there

mrodozov:
please test

mrodozov:
lets check this update on devel, I suspect the problem is not with xrootd itself

PR #6739, opened 2021-03-18
add cuda to ROOT_INCLUDE_PATH

davidlange6:
(@smuzaffar - are you aware of some reason why ROOT_INCLUDE_PATH should differ from the list of all include paths of cmssw and its externals?)


davidlange6:
please test

smuzaffar:
@davidlange6 , in start we blindly start adding include directories to $ROOT_INCLUDE_PATH . this side effect of this is that root cling has to go through all these to find header files at runtime.
Ideally we should only have those include path in $ROOT_INCLUDE_PATH  which are part of our dictionaries.

davidlange6:
Ok- thanks for the clarification - maybe we have a good way to trimming that list with modules.

smuzaffar:
+externals

PR #6744, opened 2021-03-21

mrodozov:
this contains the Arm update we wanted

### DIFF
--- a/root.spec
+++ b/root.spec
@@ -3,8 +3,8 @@
 ## INITENV +PATH PYTHON3PATH %{i}/lib
 ## INITENV SET ROOTSYS %{i}
 
-%define tag 8390dd2a3e4dbadd6720dad96d231e6a11c0abb6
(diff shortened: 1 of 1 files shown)


mrodozov:
please test

smuzaffar:
+externals
comparison dif are due to change in input for wf  140.53

smuzaffar:
@mrodozov , did this fix any of the arm failures?

mrodozov:
as far as I can tell not really, the report was that it does quote 
https://github.com/cms-sw/cmssw/issues/31123#issuecomment-779375531
followed by https://github.com/cms-sw/cmssw/issues/31123#issuecomment-802323891
but I'll just ask for examples to check. It does fix a problem that should've not been there 
we can build an IB in non-threading mode to separate relvals that are not thread related and maybe fix them
first or at least to have a list with the 'normal failures'

PR #6759, opened 2021-03-25
add full rucio-client to crabserver

belforte:
needed for S3 transfers for new crabcache.
thake this chance to askk ipython as well to help when working in crabserver pod


belforte:
please merge

PR #6762, opened 2021-03-27
patching CLEO model for production studies

alberto-sanchez:
This simple patch in required in order to use this model in MC production.

### DIFF
--- a/0001-applyng-fix-FULSOM-Bryan-May-12-2015-Change-boost-L1.patch
+++ b/0001-applyng-fix-FULSOM-Bryan-May-12-2015-Change-boost-L1.patch
@@ -1,0 +1,13 @@
+diff --git a/1.6.0/src/EvtGenModels/EvtYmSToYnSpipiCLEO.cpp b/1.6.0/src/EvtGenModels/EvtYmSToYnSpipiCLEO.cpp
+index 1ba2b0e..401ab96 100644
+--- a/1.6.0/src/EvtGenModels/EvtYmSToYnSpipiCLEO.cpp
++++ b/1.6.0/src/EvtGenModels/EvtYmSToYnSpipiCLEO.cpp
+@@ -170,7 +170,7 @@ void EvtYmSToYnSpipiCLEO::decay( EvtParticle *p){
+       // the direction opposite the Y(mS) in the X rest frame. If the pions are pi0s, then
+       // choose the one where cos(theta) = [0:1].
+ 
+-      EvtVector4R P_YmS_X = boostTo(p->getP4(), P_X);
++      EvtVector4R P_YmS_X = boostTo(p->getP4Restframe(), P_X);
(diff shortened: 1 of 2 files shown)


alberto-sanchez:
please test workflow 541.0, 545.0

alberto-sanchez:
please test

alberto-sanchez:
Dear @qliphy, @silviodonato,  we are interested in having this patch in production asap. We will need to generate some MC samples using this decay model.

This is the distribution before the patch

BuggedInvarianteMass

and, this is the distribution after the patch

CorrectedInvMass

qliphy:
> Dear @qliphy, @silviodonato, we are interested in having this patch in production asap. We will need to generate some MC samples using this decay model.

@alberto-sanchez 
Thanks!  It looks good to me.  Adding @smuzaffar @mrodozov @cms-sw/generators-l2 
What is the time scale for the MC production?

smuzaffar:
looks good to me. @alberto-sanchez do we need a similar patch for evtgen 2.0.0 https://github.com/cms-externals/evtgen/blob/cms/2.0.0/src/EvtGenModels/EvtYmSToYnSpipiCLEO.cpp#L145 ?

alberto-sanchez:
> @alberto-sanchez
> What is the time scale for the MC production?

As soon as is available

alberto-sanchez:
> looks good to me. @alberto-sanchez do we need a similar patch for evtgen 2.0.0 https://github.com/cms-externals/evtgen/blob/cms/2.0.0/src/EvtGenModels/EvtYmSToYnSpipiCLEO.cpp#L145 ?

Yes, we will need something similar in evtgen 2.0
I wonder how is the best way to go for it. In principle the compilation of 2.0 is different than in 1.6, and we already have a 
repo for that in cms-external, probaly we could create a new branch (say cms/2.0.1) and make the change in that branch,
waht do you think?

smuzaffar:
@alberto-sanchez , the usual procedure for such fixes is that 
- fork the cms-external repo e.g in this case https://github.com/cms-externals/evtgen
- Add changes in your fork on top of the branch e.g. in this case add your changes for cms/2.0.0 branch
- Open a PR using your feature/bugfix branch to the https://github.com/cms-externals/evtgen

Once we have PR for https://github.com/cms-externals/evtgen then we can already start testing it

smuzaffar:
by the way, you can do the same for cms/1.6.0 branch (i.e instead of adding patch in to cmsdist, it is better to patch it directly in the cms-externals/evtgen repo cms/1.6.0 branch)

alberto-sanchez:
I have create the Pull request for 2.0, in 1.6 the way the source is downloaded is different, we actually are not using the cms repo, but the tarball distributed at that time.

smuzaffar:
+externals
>I have create the Pull request for 2.0, in 1.6 the way the source is downloaded is different, we actually are not using the cms repo, but the tarball distributed at that time. 

ah you are right @alberto-sanchez. This PR is good for 10.6.X then.

qliphy:
+1

PR #6767, opened 2021-03-29
 RucioClient is not needed. boto3 is enough

belforte:
adding full rucio client was due to a "wrong perception". Actual problem was elsewhere and is now fixed.


belforte:
merge

PR #6525, opened 2020-12-30
Update TBB to the master branch

fwyzard:
Update TBB to the HEAD of the master branch as of 2021.01.12 (oneapi-src/oneTBB@d86ed7fb).
This branch is newer than `v2021.1.1`, and fixes compilation warnings with GCC 9.

Building with the HEAD version of TBB is currently know to fail.
We can use this PR to check for progress as more packages support it.

PR #6792, opened 2021-04-07
Update TBB to 2021.2.0 and use cmake to build

mrodozov:
this is using existing PR + additional flag and new version


mrodozov:
please test

mrodozov:
@fwyzard do we need this patch and what it does (I didn't check what is actually doing), and also can I remove the -DCMAKE_BUILD_TYPE=RelWithDebInfo  because it's a debug build or you want to check what it's doing first ?

makortel:
```
.../FWCore/PluginManager/interface/PluginFactoryBase.h:64:58: error: 'zero_allocator' is not a member of 'tbb'
```
I see the `zero_allocator` has indeed been removed from oneTBB, but I wasn't able to find any information on why. @Dr15Jones

Dr15Jones:
The oneAPI documentation for std::concurrent_vector says "You must synchronize construction and access." Maybe they changed their policy and put a much greater burden on the users?

Dr15Jones:
@dan131riley thoughts?

dan131riley:
> The oneAPI documentation for std::concurrent_vector says "You must synchronize construction and access." Maybe they changed their policy and put a much greater burden on the users?

That language was in the docs before oneAPI.  Also strange, zero_allocator is [still in the oneAPI spec](https://spec.oneapi.com/versions/0.5.0/oneTBB/memory_allocation/zero_allocator.html) as is the [advice to use it for waiting on an element](https://docs.oneapi.com/versions/latest/onetbb/tbb_userguide/Advanced_Idiom_Waiting_on_an_Element.html).

I think the zero allocator is still the right thing to do there.  It isn't a complicated class, so we could just copy it into FWCore/Utilities and s/tbb/edm/.

I also [opened an issue](https://github.com/oneapi-src/oneTBB/issues/378).

fwyzard:
@mrodozov 
> @fwyzard do we need this patch and what it does (I didn't check what is actually doing)

If you mean the changes related to `tbbbind` and `hwloc`, yes, please keep them.

If you mean the part about
```diff
+# Patch tbb v2021.1.1 CMake file to support $HWLOC_ROOT
+# not needed in the tbb master branch
+Patch: tbb-2021.1.1-cmake_policy-CMP0074
```
then I think it can be dropped: according to [the documentation](https://github.com/oneapi-src/oneTBB/blob/v2021.2.0/cmake/README.md#tbbbind-library-configuration) we can use the same approach as in #6525 .


> can I remove the -DCMAKE_BUILD_TYPE=RelWithDebInfo because it's a debug build or you want to check what it's doing first ?

I actually think we should keep `RelWithDebInfo` - as far as I know it just means that TBB will be built with the `-g` flag, so we still get an optimised build, but we also get debug symbols.

tbb.spec
```diff
@@ -1,25 +1,35 @@
...
-%define branch tbb_2020
+%define tag    %{realversion}
+%define branch onetbb_2021
 %define github_user oneapi-src
-Source: git+https://github.com/%{github_user}/oneTBB.git?obj=%{branch}/%{tag}&export=%{n}-%{realversion}&output=/%{n}-%{realversion}-%{tag}.tgz
+%define github_repo oneTBB
+Source: git+https://github.com/%{github_user}/%{github_repo}.git?obj=%{branch}/%{tag}&export=%{n}-%{realversion}&output=/%{n}-%{branch}-%{tag}.tgz
+# Patch tbb v2021.1.1 CMake file to support $HWLOC_ROOT
+# not needed in the tbb master branch
+Patch: tbb-2021.1.1-cmake_policy-CMP0074
```
fwyzard:
```suggestion
```

tbb.spec
```diff
@@ -1,25 +1,35 @@
...
+Requires: hwloc
 BuildRequires: cmake
 
 %prep
 %setup -n %{n}-%{realversion}
+%patch -p1
```
fwyzard:
```suggestion
```

tbb.spec
```diff
@@ -1,25 +1,35 @@
...
 
-make %{makeprocesses} stdver=c++17 tbb_cpf=1
+cd %{_builddir}/build
+cmake ../%{n}-%{realversion} \
+  -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+  -DCMAKE_INSTALL_PREFIX=%{i} \
+  -DCMAKE_INSTALL_LIBDIR=lib \
+  -DTBB_CPF=ON \
+  -DCMAKE_CXX_FLAGS=-Wno-deprecated-copy \
+  -DHWLOC_ROOT=$HWLOC_ROOT
```
fwyzard:
```suggestion
  -DCMAKE_HWLOC_2_INCLUDE_PATH=$HWLOC_ROOT/include \
  -DCMAKE_HWLOC_2_LIBRARY_PATH=$HWLOC_ROOT/lib/libhwloc.so \
  -DTBB_CPF=ON
```

fwyzard (changes requested):
Update the spec file to use the syntax for using the hwloc library introduced in TBB v2021.2.0.

mrodozov:
please test
checking the ext build

tbb-toolfile.spec
```diff
@@ -13,8 +13,8 @@ cat << \EOF_TOOLFILE >%i/etc/scram.d/tbb.xml
   <lib name="tbb"/>
   <client>
     <environment name="TBB_BASE" default="@TOOL_ROOT@"/>
-    <environment name="LIBDIR" default="$TBB_BASE/lib"/>
-    <environment name="INCLUDE" default="$TBB_BASE/include"/>
+    <environment name="LIBDIR"   default="$TBB_BASE/lib"/>
```
gartung, line 16: You might need to change this to lib64.

gartung: Nevermind. I see you set CMAKE_INSTALL_LIBDIR=lib

makortel:
> I think the zero allocator is still the right thing to do there. It isn't a complicated class, so we could just copy it into FWCore/Utilities and s/tbb/edm/.

I'm leaning towards this option in order to make progress. (especially given that @gartung has some evidence that the update could help for the issue of profilers missing accounting)

tbb-toolfile.spec
mrodozov: we are having this policy of deliberately keeping the lib instead of lib64

gartung:
My promising results have vanished. I got different results running igprof on test_resumable_tasks when the system was under load.

gartung:
please test with https://github.com/cms-sw/cmssw/pull/33474

gartung:
please test with cms-sw/cmssw#33474

mrodozov:
@gartung this is due to using the 1100 IB from yesterday all this tests are fixed now in 2300

mrodozov:
please test with cms-sw/cmssw#33474

gartung:
This compiles. Why is it failing the header check?
```
>> Compiling src/TrackingTools/TrajectoryParametrization/interface/CurvilinearTrajectoryParameters.h 
src/FWCore/Utilities/interface/zero_allocator.h: In member function 'T* edm::zero_allocator<T, Allocator>::allocate(std::size_t, const void*)':
src/FWCore/Utilities/interface/zero_allocator.h:40:12: error: 'memset' is not a member of 'std'
   40 |       std::memset(static_cast<void *>(ptr), 0, n * sizeof(value_type));
      |            ^~~~~~
gmake: *** [config/SCRAM/GMake/Makefile.rules:1705: tmp/slc7_amd64_gcc900/src/FWCore/Utilities/src/FWCoreUtilities/zero_allocator.h.chk_header] Error 1
```

makortel:
> This compiles. Why is it failing the header check?

Maybe `tbb/tbb_allocator.h` does not `#include <cstring>` (even indirectly), but all source files including `zero_allocator.h` do (indirectly)?

gartung:
please test with cms-sw/cmssw#33474

mrodozov:
+externals
leaving this to the release manager to merge when 33474 is ready

gartung:
please test with cms-sw/cmssw#33474

gartung:
please test with cms-sw/cmssw#33474

makortel:
please test with cms-sw/cmssw#33474

makortel:
Is there a difference between starting tests from cmssw vs cvmsdist?

mrodozov:
the 'tests approved' label will appear only on the PR from where you are testing
before it did on both places but that was not optimal - for example you might have wanted to test the cmssw alone, or with the cmsdist + other PRs.
the test labels will appear only on the PR domain from which they were started
but the testing alone - no. there is no difference

makortel:
> but the testing alone - no. there is no difference

Ok, thanks. The last test launched from cmssw side just failed in a way that looked suspicious (like TBB library would be missing).

silviodonato:
+1

PR #6800, opened 2021-04-08
[10.6] update EgammaAnalysis-ElectronTools to V00-03-01

slava77:
backport of cms-data/EgammaAnalysis-ElectronTools#11

only new files were added. So, this should be straightforward to integrate.

This is needed in connection with cms-sw/cmssw#33359 , which does not trigger an existing workflow to use the new data files, but can be enabled in a local/analysis setup.


slava77:
@cmsbuild please test

smuzaffar:
+externals
only new files

PR #6813, opened 2021-04-15
[11.3] Build tensorflow to check mkldnn version

mrodozov:
see https://github.com/cms-sw/cmssw/issues/33442
we don't keep the build logs so let the bot builds it and see the log


mrodozov:
please test

PR #6836, opened 2021-04-23
Update professor2 patch for PPC

mrodozov:
professor2 patch needed to be updated. didn't check the yoda update for ppc and professor failed


mrodozov:

PR #6839, opened 2021-04-26
[11.1.X] Remove crab dependency

smuzaffar:
CMSSW 11.1.X release cycle has an old version of crab (3.3.2005). New crab version scheme is vN.N and that is already included in latest CMSSW release cycles ( 11.2 and above). We suggest to drop crab in 11.1.X. 

There is no need to build a full CMSSW_11_1 release now. Just include this in any future full cmssw 11.1 release.


smuzaffar:
FYI @belforte , @silviodonato , @qliphy

smuzaffar:
please test

PR #6845, opened 2021-04-27
build py2-Pygments which is needed by dxr; treat .icc as cpp files

smuzaffar:
DXR which needs py2-Pygments is only available for Python2. Python3 support seems unlikely ( https://github.com/mozilla/dxr/issues/704 ) . So for now, we add back py2-Pygments  and also patch dxr and Pygments  to treat our `.icc` files as `c++` files

### DIFF
--- a/Pygments-cpp-extension-fix.patch
+++ b/Pygments-cpp-extension-fix.patch
@@ -1,0 +1,13 @@
+diff --git a/pygments/lexers/_mapping.py b/pygments/lexers/_mapping.py
+index acb71ad..e8169a3 100644
+--- a/pygments/lexers/_mapping.py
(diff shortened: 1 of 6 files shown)


smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
please test

davidlange6:
so what is the longer term plan - or does dxr in cmssw end in a few months?

smuzaffar:
looks like porting it to py3 should be easy ( at least that what DXR developer say) but as DXR is not actively developed so I am thinking of replacing it with searchFox ( which is what mozilla is moving to)

PR #6846, opened 2021-04-27
drop  python2 packages without dependencies

davidlange6:
Start of process towards dropping python2 in 12_0_x. This will probably trip over some unit test - to be seen.


davidlange6:
please test

PR #6851, opened 2021-04-28
Keep the PTX in CUDA binaries, so it can be JIT'ted for newer devices

fwyzard:


fwyzard:
please test

fwyzard:
enable gpu

fwyzard:
please test

fwyzard:

fwyzard:

fwyzard:
```
===== Test "import-yaml" ====
  File "/cvmfs/cms-ib.cern.ch/jenkins-env/python/shared/yaml/__init__.py", line 2, in <module>
    from error import *
ModuleNotFoundError: No module named 'error'

---> test import-yaml had ERRORS
 
^^^^ End Test import-yaml ^^^^
```

unrelated ?

mrodozov:
unrelated. for some weird reason on Arm and PPC this test is failing.
error is a file inside yaml - if you start python3 and import yaml it's ok
but if you run python3 -c 'import yaml' it's failing.
and that's after we changed the test to be python3

makortel:
Out of curiosity, how much does the PTX add to the (shared) object file size? (even on qualitative order-of-magnitude scale)

fwyzard:
Here is a comparison of the libraries from a working area I had around.

##### before (no PTX)
```
1.4M    pluginEventFilterEcalRawToDigiPlugins.so
724K    pluginEventFilterHcalRawToDigiGPUPlugins.so
600K    pluginHeterogeneousCoreCUDATestPlugins.so
8.2M    pluginRecoLocalCaloEcalRecProducersPlugins.so
1.8M    pluginRecoLocalCaloHGCalRecProducersPlugins.so
6.7M    pluginRecoLocalCaloHcalRecProducers.so
1.5M    pluginRecoLocalTrackerSiPixelClusterizerPlugins.so
1.9M    pluginRecoLocalTrackerSiPixelRecHitsPlugins.so
13M     pluginRecoPixelVertexingPixelTripletsPlugins.so
2.1M    pluginRecoPixelVertexingPixelVertexFindingPlugins.so
38M     total
```

##### after (with PTX)
```
1.4M    pluginEventFilterEcalRawToDigiPlugins.so
748K    pluginEventFilterHcalRawToDigiGPUPlugins.so
616K    pluginHeterogeneousCoreCUDATestPlugins.so
8.6M    pluginRecoLocalCaloEcalRecProducersPlugins.so
1.9M    pluginRecoLocalCaloHGCalRecProducersPlugins.so
7.0M    pluginRecoLocalCaloHcalRecProducers.so
1.6M    pluginRecoLocalTrackerSiPixelClusterizerPlugins.so
2.0M    pluginRecoLocalTrackerSiPixelRecHitsPlugins.so
15M     pluginRecoPixelVertexingPixelTripletsPlugins.so
2.3M    pluginRecoPixelVertexingPixelVertexFindingPlugins.so
41M     total
```

So, around 10% of the size for the CUDA-enabled plugins, on average ?

makortel:
Thanks. So noticeable but not large. Do you think the JIT'ting would be intended only for cases where new architecture appears after the build, or should we think of mainly JIT'ting the code?

fwyzard:
For the moment I would consider it as a fallback option for new architectures (e.g. Ampere), and to possibly improve the performance on less used architectures (e.g. JIT'ting for `sm_61` *may* have some benefit for a GeForce GTX 1080 instead of using the binary for `sm_60`).

Somewhere in the NVIDIA documentation I've see a suggestion that we should build  with `-gencode arch=compute_60,code=sm_60 -gencode arch=compute_61,code=sm_61 -gencode arch=compute_62,code=sm_62 -gencode arch=compute_70,code=sm_70 -gencode arch=compute_72,code=sm_72 -gencode arch=compute_75,code=sm_75 -gencode arch=compute_80,code=sm_80 -gencode arch=compute_86,code=[sm_86,compute_86]` ... which would increase both compilation times and library sizes significantly.

I consider this PR a reasonable compromise.

makortel:
> I consider this PR a reasonable compromise.

Sure, thanks. I was thinking more the future, e.g. should we add `sm_80` etc there at some point?

smuzaffar:
+externals

fwyzard:
> Sure, thanks. I was thinking more the future, e.g. should we add `sm_80` etc there at some point?

Sure - well, maybe once we actually have some Ampere card to test :-)

smuzaffar:
test parameters:
- addpkg = HeterogeneousCore

smuzaffar:
please test

smuzaffar:
abort

we need to wait for next IB with SCRAMV3 for external PR testing

smuzaffar:
please test

PR #6864, opened 2021-05-03
[SCRAM] Fixes for recursive scram build

smuzaffar:


smuzaffar:
test parameters:
- addpkg = FWCore

smuzaffar:
please test

smuzaffar:
please test

mrodozov:
@smuzaffar there was one unit test failing for arm and ppc (import-yaml) that is not failing anymore (which is fine of course)
could it have been because of scram update ? now it's not failing but it just miraculously disappeared

last it was showing that it's searching for yaml from the jenkins-env:
```

===== Test "import-yaml" ====
  File "/cvmfs/cms-ib.cern.ch/jenkins-env/python/shared/yaml/__init__.py", line 2, in <module>
    from error import *
ModuleNotFoundError: No module named 'error'


```

smuzaffar:
I think it was failing because it was picking up yaml from /cvmfs/cms-ib.cern.ch/jenkins-env/python/shared which was wrong and we should start cleaning up /cvmfs/cms-ib.cern.ch/jenkins-env . 

and yes it is fixed due to new SCRAM as new SCRAM remove PYTHONPATH before running any scram command . This means for "scram build runtests" the PYTHONPATH is unset and that is why /cvmfs/cms-ib.cern.ch/jenkins-env/python/shared is not seen by python

PR #6874, opened 2021-05-05
[SCRAM] Fix scram build -j without N option bug

smuzaffar:
This PR fixes two issues
- `scram b -j` should run `gmake -j $(nproc)` . It was broken with SCRAM V3. This should be fixed now.
- `scram b disable-biglib` is broken with SCRAMV3. Disabling the biglib was commenting the biglib paths using `#` in the `config/Self.xml` file. `#` was a valid comment character for SCRSM V2 but as SCRAM V3 uses proper xml parser, so this was not a valid comment. BuildRules are fixed to use proper `<!--  -->` comment

### DIFF
--- a/SCRAMV1.spec
+++ b/SCRAMV1.spec
@@ -1,4 +1,4 @@
-### RPM lcg SCRAMV1 V3_00_21
+### RPM lcg SCRAMV1 V3_00_22
 ## NOCOMPILER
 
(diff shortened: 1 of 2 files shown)


smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
+externals
failures are not related to scram changes

PR #6877, opened 2021-05-05
[SCRAM] Protect against failure when run in container

smuzaffar:


smuzaffar:
please test

smuzaffar:
+externals
coral built and `scam project` command worked fine. That is enough to test the change

PR #6901, opened 2021-05-12
Update ROOT master and remove root_numpy

mrodozov:
remove redundant symbol that might have cause trouble


mrodozov:
please test with cms-sw/cmssw#33707

PR #6910, opened 2021-05-16
Enable libunwind-toolfile and rebuild the external

mrodozov:
after some failures I found out libunwind is not built on PPC
https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_12_0_X/master/cmssw-tool-conf.spec#L180
and I'll enable it for this night IB to see what wrong with it
gperf requires libunwind


mrodozov:
please test

PR #6929, opened 2021-05-20
Triton: enable GPU support

kpedro88:
The Triton client must be compiled with GPU support enabled in order to use GPU shared memory (via the `cudaIpcMemHandle_t` object).


mrodozov:
please test

kpedro88:
please test

kpedro88:
please test

fwyzard:
@smuzaffar @mrodozov is there a way to make these changes conditional on whether CUDA is available or not ?

mrodozov:
can you elaborate a little more what you have in mind

fwyzard:
Sure...

We assume that some architectures may not support CUDA (e.g. if we make builds for GCC 11, we would have no CUDA support).
In that case, what happens to the externals that depend on CUDA ?

For CMSSW packages we can use
```xml
<iftool name="cuda">
  ...
</iftool>
```
in the buildfiles.

For some externals, the dependency on CUDA is a hard one (e.g. it makes no sense to build cuDNN if we don't have CUDA).

For other externals (like LLVM, Alpaka, Triton, etc.) we can still build the external with only some features disabled.

Do we have something like this already in place ?
If not, is there a mechanism we could use ?

mrodozov:
In such case where we know from before there will be no cuda like for gcc11
we'll wrap the option GPU=ON/OFF in a condition, I think we already have examples for compiler version.
But we have to know "from before" what is the condition - in this case it will be if gcc_version > 10
then it comes to use of rpm macroses, like in here:

https://github.com/cms-sw/cmsdist/blob/4f5e67a29b506c0312331b3c6a1ce4496e616f6e/herwig7.spec#L31

I think we have shorter example, but in the case of gcc > 10 
we 1. wrap` Requires: cuda` in conditions 2. wrap the GPU=ON/OFF in condition. 
We are already using such conditions for ifarch == something

https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_12_0_X/master/mkfit.spec#L22

fwyzard:
OK, so we don't have a centralised (or not) system in place yet.

Let's keep it in mind and add it the first time we need it, instead of
peppering individual checks in each tool.

mrodozov:
Unfortunately we'll have to explicitely describe what is happening in this case wrap everything that depends on cuda in conditions, or device something clever like define rpm macro (if requires: cuda)
Or maybe I cant think of anything of the sort right now

fwyzard:
I was thinking that we could add a new file (e.g. `cuda_is_available.file`) that defines a new macro (e.g. `%cuda_is_available`) so the packages that optionally depend on CUDA could do something like
```spec
## INCLUDE cuda_is_available

%if %cuda_is_available
  # configure with CUDA
%else
  # configure without CUDA
%endif
```

kpedro88:
@fwyzard thanks for raising this point - it would indeed be useful to make the CUDA dependence optional. Once we figure out the right approach for cmsdist here, the few blocks of CUDA-dependent code added in https://github.com/cms-sw/cmssw/pull/33801 can be wrapped in `ifdef`s (and thereby compiled out when CUDA is not available).

smuzaffar:
@fwyzard , you can use `%{cuda_gcc_support}` macro ( defined here https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_12_0_X/master/rpm-preamble.file#L298 ) in the `%prep, %build, %install` sections to contidtionally enable/disable cuda related changes. We already use this for `cupla` ( https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_12_0_X/master/cupla.spec#L43 ) . 

Is this what you were looking for?

kpedro88:
@smuzaffar thanks, I've added a commit that uses this feature. Do we have a CUDA-disabled build to test it? (It works fine on my machine with CUDA enabled.)

smuzaffar:
@kpedro88 , no currently all our gcc versions are supported by cuda. Soon we will setup GCC 11 IBs and that might have cuda disabled

fwyzard:
@smuzaffar this approach leaves the
```spec
Requires: cuda
```
dependency, though ?

Isn't that a problem for an installation where CUDA is not available ?

kpedro88:
please test

mrodozov:
please test
openssl removal created conflicts

mrodozov:
@kpedro88 @fwyzard we merge this as it is and continue on the flags topic additionally, correct ?

mrodozov:
+externals

kpedro88:
@mrodozov yes, from my side this is ready to go.

smuzaffar:
> @smuzaffar this approach leaves the
> 
> ```python
> Requires: cuda
> ```
> 
> dependency, though ?
> 
> Isn't that a problem for an installation where CUDA is not available ?

@fwyzard , we can not avoid this. For archs where cuda is not available, we can just provide a dummy cuda.spec

fwyzard:
That is why I was suggesting to implement it as a specfile macro that could be used with `%if` .

Anyway, this is a more general comment, which should not block this PR.

PR #6930, opened 2021-05-20
DXR needs py2-parsimonious to run

mrodozov:
dxr-cmssw-index jobs is failing because of missing py2 package DXR was using without requiring it


mrodozov:
please test

PR #6936, opened 2021-05-25

mrodozov:
get latest commits

### DIFF
--- a/root.spec
+++ b/root.spec
@@ -3,8 +3,8 @@
 ## INITENV +PATH PYTHON3PATH %{i}/lib
 ## INITENV SET ROOTSYS %{i}
 
-%define tag 4e1d2ff8e7b16adfe93e1e0e69962df180ccb046
-%define branch cms/master/fdb7d15
+%define tag cb7356c64e7180054a1ddc1ee17c63012c9dffcd
+%define branch cms/master/db3d424
 %define github_user cms-sw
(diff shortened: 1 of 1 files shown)


mrodozov:
please test

makortel:
In `oneTBB` v2021.2.0
```
$ git grep -i captured_exception
test/tbb/test_task_group.cpp:#if TBB_USE_CAPTURED_EXCEPTION
test/tbb/test_task_group.cpp:    typedef tbb::captured_exception TestException;
```
which leaves me wondering how exactly this error could be triggered?

mrodozov:
I'm reading that commit 
https://github.com/root-project/root/commit/a0b5e3dad6ed778feb1855e24d0bf077d88c9ef8
and I wonder if this logic is correct

makortel:
From the discussion in https://github.com/root-project/root/pull/8169 the PR was motivated by TBB 2018, and indeed there `git grep` shows many hits. So something that has been removed since.

makortel:
The logic is most certainly not correct, TBB 2021.2 does not even have `tbb/tbb_config.h`.

mrodozov:
hahjo figured whats wrong and is pushing it for a fix

mrodozov:
please test

mrodozov:
please test with https://github.com/cms-sw/cmssw/pull/33875

PR #6869, opened 2021-05-03
Removal of many more python2 tools

davidlange6:
Extends the work of #6846

PR #6955, opened 2021-06-01
dropping more python2 tools 

davidlange6:
This is a piece of #6869 which has problems that I've not yet understood..


davidlange6:
please test

davidlange6:
please test

smuzaffar:
@davidlange6 , do you know who needs `pyOpenSSL` ? If it is not needed then we can drop it and `cryptography` (which does not support OpenSSL 1.0.2). There is security issue https://github.com/cms-sw/cmsdist/security/dependabot/pip/requirements.txt/cryptography/open to update `cryptography` but it does not build for CC7 as OpenSSL version on CC7 is 1.0.2k. 

Also can you please update these packages to avoid the security issues https://github.com/cms-sw/cmsdist/security/dependabot
```
-lxml==4.6.2; python_version>'3.0'
+lxml==4.6.3; python_version>'3.0'
-pydantic==1.8 ; python_version>'3.0'
+pydantic==1.8.2 ; python_version>'3.0'
-urllib3==1.26.3
+urllib3==1.26.5
```

davidlange6:
> On Jun 1, 2021, at 12:43 PM, Malik Shahzad Muzaffar ***@***.***> wrote:
> 
> 
> @davidlange6 , do you know who needs pyOpenSSL ? If it is not needed then we can drop it and cryptography (which does not support OpenSSL 1.0.2). There is security issue https://github.com/cms-sw/cmsdist/security/dependabot/pip/requirements.txt/cryptography/open to update cryptography but it does not build for CC7 as OpenSSL version on CC7 is 1.0.2k.

Sounds like we should drop it "The Python Cryptographic Authority strongly suggests the use of pyca/cryptography where possible. If you are using pyOpenSSL for anything other than making a TLS connection you should move to cryptography and drop your pyOpenSSL dependency.”

I can update things later after I see how the tests go on the current PR.

davidlange6:
test parameters:
 - full_cmssw = true

davidlange6:
please test

davidlange6:
this error is in the IB already - so not related. Perhaps this update is ok to go now.

mrodozov:
+externals
unless @smuzaffar has anything to add. not anticipated failures will appear if any, anyway 😃 I hope there will be none.
after yesterdays discussion on going for py3 I can think one thing that we should test for py3 - the dxr. we are already building it for py3 but it has to be ran with the py3 executable (say from the py3 IB)

smuzaffar:
looks good

PR #6962, opened 2021-06-02
Another set of python2 tools to be removed

davidlange6:


davidlange6:
test parameters:
  - full_cmssw = true

davidlange6:
please test

smuzaffar:
@davidlange6 , looks like this contains the changes of #6955 too ..right?
Can we drop `cryptography` too as no packages requires it any more (only pyOpenSSL was using it)

davidlange6:
Yes, this contains #6955.  I don’t remember why we added cryptography - I’ll try to remind myself. Perhaps indeed its not needed.

davidlange6:
I don't find anything that suggests that the cryptography package is important for some use case - I'll drop cryptography when I next update this (or in the next PR in case this one doesn't need updating)

PR #6989, opened 2021-06-05
[DXR] Use get dxr sources from cms-externals

smuzaffar:


smuzaffar:
please test with cms-sw/cmssw#33984

smuzaffar:
+externals

PR #7007, opened 2021-06-10
[DXR] Added deps on py3-Pygments which is required at runtime by dxr

smuzaffar:


smuzaffar:
please test

smuzaffar:
+externals
externals build and unit tests passed. no nee to wait for rest of the tests as they do not use dxr

PR #7019, opened 2021-06-11
cleaning up unused py2 tools ; restoring tensorflow as part of python_tools.spec

davidlange6:
(For DEVEL)


davidlange6:
test parameters:
  - full_cmssw = true

davidlange6:
please test

davidlange6:
Seems that I forgot to commit a file worth of changes. Will fix this afternoon

On Jun 12, 2021 9:42 AM, cmsbuild ***@***.***> wrote:

-1

Summary: https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-3ab08a/15916/summary.html
COMMIT: 2d7cd22<https://github.com/cms-sw/cmsdist/commit/2d7cd22e38c635eacb09be9c64e8145c5c0eaa56>
CMSSW: CMSSW_12_0_DEVEL_X_2021-06-11-2300/slc7_amd64_gcc900
User test area: For local testing, you can use /cvmfs/cms-ci.cern.ch/week0/cms-sw/cmsdist/7019/15916/install.sh to create a dev area with all the needed externals and cmssw changes.

External Build

I found compilation warning when building: See details on the summary page.

davidlange6:
please test

davidlange6:
please test

davidlange6:
There are a few unit tests in PhysicsTools/PythonAnalysis that need to be cleaned up tomorrow.

davidlange6:
please test with cms-sw/cmssw#34106

davidlange6:
please test with cms-sw/cmssw#34106

smuzaffar:
please test with cms-sw/cmssw#34106 for CMSSW_12_0_DEVEL_X

davidlange6:
I believe these failures are all in the mainstream IB or expected in DEVEL - @smuzaffar ?

smuzaffar:
that is correct @davidlange6. What about comparison differences, do you think those are due to python3 used by cmsRun?

davidlange6:
The 7 dqm histogram failures are all message logger - so “normal”.

I’m puzzled how to have reco comparison differences without dqm histogram comparison differences. So, ok, something we have no DQM monitoring of - slimmedJetsPuppi..

The test done with the mainstream build did not have the comparison differences. How can I find what the baseline was for the reco comparisons for the build starting from the DEVEL branch? Was it not a recent DEVEL IB?

smuzaffar:
> How can I find what the baseline was for the reco comparisons for the build starting from the DEVEL branch

you can click on the "Details" link for `cms/7019/DEVEL/slc7_amd64_gcc900/comparison` status which should take you to comparison jenkins job ( in this case https://cmssdt.cern.ch/jenkins/job/compare-root-files-short-matrix/43455/ ) . As we ony generate baseline for 12.0.X so comparison is always done with normal 12.0.X IB.

davidlange6:
Ok, thanks - so that is presumably a python3 difference (feature or bug - to be understood..). I will investigate that. But thats independent of this PR.

davidlange6:
the puppi differences are due to different order of things the python config. 

Eg, 

245924c245933
<     discriminatorSources: VInputTag tracked  = {pfDeepCSVJetTagsAK8PFPuppiSoftDropSubjets:probbb,pfDeepCSVJetTagsAK8PFPuppiSoftDropSubjets:probb,pfCombinedMVAV2BJetTagsAK8PFPuppiSoftDropSubjets,pfCombinedInclusiveSecondaryVertexV2BJetTagsAK8PFPuppiSoftDropSubjets}
---
>     discriminatorSources: VInputTag tracked  = {pfCombinedInclusiveSecondaryVertexV2BJetTagsAK8PFPuppiSoftDropSubjets,pfCombinedMVAV2BJetTagsAK8PFPuppiSoftDropSubjets,pfDeepCSVJetTagsAK8PFPuppiSoftDropSubjets:probbb,pfDeepCSVJetTagsAK8PFPuppiSoftDropSubjets:probb}

presumably due to https://github.com/cms-sw/cmssw/blob/074f0c69af4f4669623b0ef33c8c73f73d8d287f/PhysicsTools/PatAlgos/python/recoLayer0/bTagging_cff.py#L62

So if we are depending on the order of things coming out of a dictionary, that is a bug (perhaps just a bug in the reco comparisons.

smuzaffar:
lets get this in for DEVEL

smuzaffar:
@davidlange6 , 3 of DEVEL IBs unit tests can be fixed if they use `python3` e.g.
- https://github.com/cms-sw/cmssw/blob/master/FWCore/PythonFramework/test/runPythonTests.sh#L10-L15
- https://github.com/cms-sw/cmssw/blob/master/Validation/Geometry/test/genHGCalPlots.sh#L12
- https://github.com/cms-sw/cmssw/blob/master/Validation/Geometry/test/genTrackerPlots.sh#L11-L16

should we just fix this in CMSSW_DEVEL branch or define soem env e.g CMSRUN_PYTHON=`python|python3` and use this in these scripts?

davidlange6:
I think the two in Validation can be changed also in mainstream. The one in FWCore needs to be consistent with the fwk python version- I can make a PR for the two in Validation.

davidlange6:
Sorry, I’m wrong. The two in Validation use the FWCore/PythonFramework so have the same issue that the python version needs to make the python version of the fwk. (Which I don’t think we have an easy way to discover)

smuzaffar:
you can try but I think the tests in Validation should also use same python as used by FWK

smuzaffar:
can we do something like the following to find the FWK python? 
```
ldd `which cmsRun` | grep libpython
```

or add a helper script cms-python which just runs the above and return either `python2 or python3`

davidlange6:
Yes, I think so - now is the limited lifetime of this script worthwhile to introduce? (Or maybe there is a longer term use case?)

But, eg, this works in a bash script - we could add that to the three unit tests

```
#!/bin/bash
{
python3 -c "from FWCore.PythonFramework.CmsRun import CmsRun" 2>/dev/null &&echo "using python3"
}||
{
python -c "from FWCore.PythonFramework.CmsRun import CmsRun" 2>/dev/null && echo "using python2"
}

```

smuzaffar:
Ido not see any long term use case, so fixing these 3 sciprts should be enough. By the way, you want to use `python2` instead of `python` asin PY3 IBs, python is python3 :-)

davidlange6:
Ok, lets see if I can fix all three tests

PR #7035, opened 2021-06-16
[TBB] Commit in master suggested by TBB devs

mrodozov:
See https://github.com/cms-sw/cmssw/issues/33636#issuecomment-862394219
this is building. it's not a tag but we need this change (hopefully it fixes our issue)

### DIFF
--- a/tbb.spec
+++ b/tbb.spec
@@ -1,7 +1,7 @@
 ### RPM external tbb v2021.2.0
 
-%define tag %{realversion}
-%define branch onetbb_2021
+%define tag 478de5b1887c928e52f029d706af6ea640a877be
+%define branch master
 %define github_user oneapi-src
 %define github_repo oneTBB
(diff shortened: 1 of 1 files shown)


mrodozov:
please test

makortel:
Would it make sense to test this first in non-production IBs? I suppose ARM and PPC would make most sense.

I'm a bit concerned of updating the production IB to a non-release version. At minimum, I think, this should not be merged for production IB before pre3 is cut.

mrodozov:
after the test we can
1. merge it 
2. build 1 devel 1 ppc and 1 arm IB
3. revert it until we are sure it's ok for master (and TBB people make a new tag maybe)

makortel:
The crashes in PPC seem to be frequent-enough that a single IB could be sufficient to check if the crashes get fixed.

mrodozov:
right, I noticed the IB failed and then went to see what wrong:
https://github.com/oneapi-src/oneTBB/blob/8b6f831c68170231ad2748962d48e7d279d35084/src/tbbmalloc/frontend.cpp#L806-L819
I'm asking the tbb devs what to do because improvising can make things worse 😅
In short there is no ppc block and the compiler skips that part failing to compile when meets `getSmallObjectIndex`

PR #7042, opened 2021-06-17
[G4][geant4_only] Update geant4 to 10.7.2

mrodozov:
Get the latest G4 tag in the G4 IB

### DIFF
--- a/geant4.spec
+++ b/geant4.spec
@@ -1,4 +1,4 @@
-### RPM external geant4 10.7.1
+### RPM external geant4 10.7.2
 ## INCLUDE compilation_flags
 %define tag %{realversion}
 %define branch geant4-10.7-release


mrodozov:
please test

civanch:
@mrodozov , it is not clear why there is build error in RECO sub-library while only Geant4 version is changed.

mrodozov:
@civanch we don't have a fresh IB that includes this change in mkFit . This evening we'll have it and restart the test
the IB used here is CMSSW_12_0_GEANT4_X_2021-06-15-2300 which doesn't include the change needed.
We don't build that IB everyday as we do for other releases

mrodozov:
please test
now we have geant4 IB (since this evening)

civanch:
@mrodozov , I think the test is now positive, the fact that simulations histories are different and we have no regression means, that we have to do private validation and PPD validation campaign. This PR may be merged to special branches GEANT4 and G4VECGEOM.

PR #7054, opened 2021-06-22
restore pyyaml on python2 as needed by conddb 

davidlange6:
some conditions tools need to be consistent with the python used by boost-python (which is still python2). These rely on pyyaml.


davidlange6:
please test

smuzaffar:
+externals
looks good

PR #7071, opened 2021-06-28
update crabclient DEV

belforte:
first CRABClient which more or less works in python3


smuzaffar:
please test with cms-sw/pkgtools#176

smuzaffar:
+externals
the external build process is finished successfully, no need to wait for the rest of the tests

PR #7074, opened 2021-06-29
Update tag  number for root 6.24

mrodozov:
the current tag is changed to 6.24.03


smuzaffar:
trivial update, so no need to test it. IB should be enough to test it

PR #7081, opened 2021-06-30
remove pyminuit and its dependencies

davidlange6:
to quote the pyminuit repo "Update in 2015: Guys, PyMinuit is ancient." It won't support python3, we already have professor2 rather than professor. None of these appear to be used in cmssw.


davidlange6:
please test

smuzaffar:
assign generators

smuzaffar:
@mkirsano,@SiewYan,@alberto-sanchez,@agrohsje,@GurpreetSinghChahal  any objections on dropping `professor` from the cmssw distribution? As @davidlange6 mentioned, we already have `professor2` so do we need to keep `professor` ?

agrohsje:
Agreed but double-checking with the tuning people. Already pinged them.

sensrcn:
We (GEN-17-002) use `professor` only in CMSSW_9_x and 10_x releases. We don't need `professor` in CMSSW_12_x.

mseidel42:
Professor 1.4 had an experimental feature for taking into account covariance matrices.
But to be honest, it was rarely used, and this can also be hacked into Professor 2.

We should probably get the successor tool, [apprentice](https://github.com/HEPonHPC/apprentice), into CMSSW and make sure that it has all the features that we need.

smuzaffar:
+externals
Thanks for the feedback @cms-sw/generators-l2 . So we will go ahead and drop `professor` for 12.0.X
For `apprentice` , it is available Pypi so we should be able to include it when needed. Please open an issue in cms-sw/cmssw repo when you need this.

agrohsje:
+generators
thanks @smuzaffar

PR #7082, opened 2021-06-30
updated xrootd and blackhat to use py3 only; drop distcc/ccache

smuzaffar:
- Updated `xrootd` to only build for Python3
- Updated `blackhat` to use python3 though it looks like blackhat internally is not building `pythoninterface`  even with `--enable-pythoninterface` set). So basically we do not need python dependency here but as `configure` script looks for it so that is why we have to include it.
- Dropped unused packages `distcc and ccache`
- Cleanup `rpm-preamble`


smuzaffar:
please test

smuzaffar:

smuzaffar:
please test for CMSSW_12_0_X/slc7_ppc64le_gcc9

smuzaffar:
please test for CMSSW_12_0_X/slc7_ppc64le_gcc9

smuzaffar:
+externals

PR #7084, opened 2021-06-30
Py3 drop unused pkgs

smuzaffar:
- apache-ant, apache-tomcat : Nothing in cmssw uses it and it was never part of cmssw distribution
- cvs2git: This is not used any more, we built and shipped it when we did the cvs to git migration
- geneva: It was dropped a year ago as it failed to build with new pythia8 and also it does not support python3


smuzaffar:
please test

smuzaffar:
+externals
external build process is successful

PR #7096, opened 2021-07-03
[DEVEL] Disable python for lhapdf: need lhapdf 6.3.0 for py3.9

smuzaffar:
@davidlange6 , looks like lhapdf 6.2.3 does not build with python 3.9 that is why DEVEL IBs are not building
For now I have disabled the python for lhapdf, we need lhapdf 6.3.0 which works with python 3.9


smuzaffar:
please test for CMSSW_12_0_DEVEL_X

smuzaffar:
+externals
tests looks good, lets get this in DEVEL IBs

PR #7112, opened 2021-07-07
remove python2 pip support and remaining python2 packages except for future

davidlange6:


davidlange6:
please test

smuzaffar:
@davidlange6 , we also need to update https://github.com/cms-sw/cmssw/blob/5244916401e27d97ce2dace3341db78c5c08098f/PhysicsTools/PythonAnalysis/test/BuildFile.xml#L29-L31 . I think just
```
<test name="testNumPy" command="python3 -c 'import numpy'"/>
```
should be enough now ( i.e drop the `<iftool ...>` condition)

davidlange6:
Yes, looks like the best way to go is to make that change after this is merged

smuzaffar:
please test
lets use a newer IB

davidlange6:
my guess is that none of these will be fixed. I need to finish my PR for replacing /usr/bin/env python with /usr/bin/env python3..

davidlange6:
well, but maybe I'm wrong. Keras/tensorflow etc have been python3 only for a while..

smuzaffar:
Tensorflow for sure is py3 only since long

davidlange6:
please abort

davidlange6:
please test

davidlange6:
sorry - my mistake..

davidlange6:
at least some (all?) of the other errors will be fixed by a PR merged (to move Configuration scripts to python3).

smuzaffar:
@davidlange6 , any idea about reco comparison diffs?

davidlange6:
That is interesting… smells like a python dict sorting issue. But its definitely not obvious. 

Is there a log file from the reco comparison someplace?

smuzaffar:
yes, from the summary page you can go to `Comparison with the baseline` ( e.g. https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_12_0_X_2021-07-07-2300+508a55/43907/ ) then all logs should be there

smuzaffar:
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_12_0_X_2021-07-07-2300+508a55/43907/validateJR/all_OldVSNew_RunPhoton2012Bwf4p53/all_OldVSNew_RunPhoton2012Bwf4p53.log is log for one of the failed reco comparison

davidlange6:
Thanks - not sure how I missed these..

So the reco comparison script really needs to match labels in Jet::pairDiscriVector_ rather than using the index of elements. 

But that said, I have no idea why the order of that vector would have changed unless the comparison baseline was with cmsRun (or perhaps cmsDriver) with python2.  But using CMSSW_12_0_X_2021-07-07-2300 as the baseline  should have avoided that.

makortel:
> So the reco comparison script really needs to match labels in Jet::pairDiscriVector_ rather than using the index of elements.

FYI @cms-sw/reconstruction-l2

slava77:
> > So the reco comparison script really needs to match labels in Jet::pairDiscriVector_ rather than using the index of elements.
> 
> FYI @cms-sw/reconstruction-l2

it's a known limitation of the script

slava77:
> > > So the reco comparison script really needs to match labels in Jet::pairDiscriVector_ rather than using the index of elements.
> > 
> > 
> > FYI @cms-sw/reconstruction-l2
> 
> it's a known limitation of the script

I think that we can "manually" deal with cases where either the python version changes or the jet tagging configuration itself changes.

is there expectation that the order is not going to be repeatable now? In this case, we'll need to reconsider what to do with the comparison script

davidlange6:
no, the order is not expected to change regularly (but I don't think I understand why in changed in this PR)

davidlange6:
ah - but you see this in a second PR. - :( - I can try to reproduce

davidlange6:
Update - I can reproduce just by dumping the python of step3 of workflow 25. There are differences, clearly due to dictionary sorting differences. I can also reproduce just by dumping the same step3 workflow twice. Eg, there is some randomness which then affects these reco plots.

A trivial reproducer of iterating over a dictionary does not show the same effect. (which makes sense as the order of iterators over dictionaries is reproducible in py3.8).  Anyway, looking for the underlying cause...

slava77:
> Update - I can reproduce just by dumping

same here; I'm curious if the order can be enforced in various extensions of the `discriminatorSources`

the following may be the gymnastics that leads to reordering
https://github.com/cms-sw/cmssw/blob/1cd8d24459220fe8dd876941666453d7051b17f0/PhysicsTools/PatAlgos/python/tools/jetTools.py#L243-L255

makortel:
> > Update - I can reproduce just by dumping
> 
> same here; I'm curious if the order can be enforced in various extensions of the `discriminatorSources`
> 
> the following may be the gymnastics that leads to reordering
> https://github.com/cms-sw/cmssw/blob/1cd8d24459220fe8dd876941666453d7051b17f0/PhysicsTools/PatAlgos/python/tools/jetTools.py#L243-L255

It uses `set`, found 
> `set`s are not ordered in Python 3.6, not even as a CPython implementation detail.

https://stackoverflow.com/a/45589769

makortel:
Also [documentation of 3.9](https://docs.python.org/3/tutorial/datastructures.html#sets) says "A set is an **unordered** collection with no duplicate elements."

slava77:
> Also [documentation of 3.9](https://docs.python.org/3/tutorial/datastructures.html#sets) says "A set is an **unordered** collection with no duplicate elements."

I suppose a sort  of what starts as `    acceptedBtagDiscriminators = list()` before https://github.com/cms-sw/cmssw/blob/1cd8d24459220fe8dd876941666453d7051b17f0/PhysicsTools/PatAlgos/python/tools/jetTools.py#L819-L823
can solve the problem here.

makortel:
> > Also [documentation of 3.9](https://docs.python.org/3/tutorial/datastructures.html#sets) says "A set is an **unordered** collection with no duplicate elements."
> 
> I suppose a sort of what starts as ` acceptedBtagDiscriminators = list()` before https://github.com/cms-sw/cmssw/blob/1cd8d24459220fe8dd876941666453d7051b17f0/PhysicsTools/PatAlgos/python/tools/jetTools.py#L819-L823
> can solve the problem here.

Or `btagDiscriminators = sorted(discriminators)` in https://github.com/cms-sw/cmssw/blob/1cd8d24459220fe8dd876941666453d7051b17f0/PhysicsTools/PatAlgos/python/tools/jetTools.py#L250
but I didn't study the code at all to say which one would make more sense.

slava77:
> Or `btagDiscriminators = sorted(discriminators)`

testing ...

slava77:
> > Or `btagDiscriminators = sorted(discriminators)`
> 
> testing ...

cms-sw/cmssw#34409

davidlange6:
Nice catch! (your tests won't actually test if the problem is fixed - we need to merge that change and then test a new PR to see that...

(anyway, this PR is now understood I think)

smuzaffar:
yes @davidlange6 this PR is ready to go in but as we are building 12.0.0.pre4 right now so I will merge it after pre4.

davidlange6:
Great! (No rush from my side!) — I’ll come back to the unit tests mentioned above likely on Monday.

smuzaffar:
test parameters:
- full_cmssw = true

smuzaffar:
please test

smuzaffar:
please test
@davidlange6 many unit tests failed , like like they need py3 updates. Anyway, I am re-starting the test to get latest IB

smuzaffar:
please test

smuzaffar:
@davidlange6 , most of these are failing due to missing `six`, I think I just need to update the scripts to use `python3`. If you are not working in these then I can open a PR in next 30 mins

smuzaffar:
test parameters:
- full_cmssw = true

smuzaffar:
please test

smuzaffar:
@davidlange6 , lets see if cms-sw/cmssw#34471,cms-sw/cmssw#34472,cms-sw/cmssw#34473,cms-sw/cmssw#34474,cms-sw/cmssw#34475 set of PRs fixes the rest of the issues

smuzaffar:
ok one unit test fails now ( src/FWCore/ParameterSet/test/TestFWCoreParameterSetDriver ) this should be fixed by https://github.com/cms-sw/cmssw/pull/34477

smuzaffar:
please test

smuzaffar:
@davidlange6 , now `TestFWCoreParameterSetDriver` is failing in the same way as it is failing in PY3 IBs. Can you please check this unit tests in PY3 IBs? No idea why `testhep_ml` failed now, may be a glitch

smuzaffar:
the framework test is being followed here https://github.com/cms-sw/cmssw/pull/34477#issuecomment-879534593

davidlange6:
I can have a look in ~2 hours - no problem

smuzaffar:
please test

smuzaffar:
looks good, lets get this in

PR #7115, opened 2021-07-08
avoid using hard-coded python version

smuzaffar:
Avoid hard coded pythn verson. This should fix the DEVEL IB build errors where boost_python is looking for version `38` where as it should have been `39`


smuzaffar:
+externals

PR #7140, opened 2021-07-15
[DEVEL] enable build with python

smuzaffar:
Python build was disabled as lhapdf 6.2.3 did not work with python 3.9. As now we have newer lhapdf version ( 6.3.0) so lets re-enable the python


smuzaffar:
please test

PR #7145, opened 2021-07-17
Fix for DXR, use py3-dxr instead of py2-dxr

smuzaffar:


smuzaffar:
please test

smuzaffar:
+externals

PR #7147, opened 2021-07-19
scram-tools repo to provide scram toolfile definitions

smuzaffar:

PR #7142, opened 2021-07-15
Update Triton to v2.11

kpedro88:
This PR updates Triton to its latest release. The Triton repository has been refactored into separate components, so the external is renamed to `triton-inference-client`.

In the process of refactoring, access to some helper functions was lost, so these are installed by hand in the spec file. This should be fixed in an upcoming version of Triton. (Hopefully the various CMake issues currently solved in the spec file will also be addressed.)

The Triton updates require a newer version of CMake, so that is done here as well.

This PR requires an associated CMSSW PR (to be submitted simultaneously) to compile: https://github.com/cms-sw/cmssw/pull/34508

PR #7175, opened 2021-07-27
remove unneeded triton toolfile

kpedro88:
#7142 replaced triton-inference-server with triton-inference-client, but #7147 was submitted around the same time, so it still had triton-inference-server-toolfile.xml. This file is not needed, so it can be removed.


kpedro88:
please test

smuzaffar:
+externals

PR #7201, opened 2021-08-04
fix EvtGen data path

mkirsano:


mkirsano:
please test

smuzaffar:
CMSSW 11.3 and 12.0 are using evtgen 2.0 , so I guess this fix should be backported to those release cycles too .... right @mkirsano

mkirsano:
In principle yes, but this is not very critical as in many cases the CMS specific data files are used

smuzaffar:
+externals

PR #7209, opened 2021-08-11
keep cmssw tools to drop in a common file

smuzaffar:
The PR makes sure that the tools dropped for full cmssw release are also dropped for patch release. Thisis done via  common cmssw-drop-tool.file which is included by both cmssw and cmssw-patch.

This should fix the issue  reported here cms-sw/cmssw#34841


smuzaffar:
please test

smuzaffar:
+externals
cms external and tools file look good. merging it and start an IB now, So that at 23h00 we can build a patch IB to see the effect

PR #7231, opened 2021-08-23
added suport to build indiviual toolfiles

smuzaffar:
This PR allows to optionally build `<package>-toolfile`. This is helpfull when you only want to build a single package and its toolfile instead of full cmssw-tool-conf. 

Cleanup of some of the existing `*-tool-conf` specs


smuzaffar:
please test

PR #7260, opened 2021-09-01
ThePEG - remove `-g` flag from the compilation

mrodozov:
This removes the default CXXFLAGS="-g -O2" (same for CFLAGS) to -g0 -O2 -DNDEBUG to keep the optimization, set debug info to 0 and remove assertions if any
The original lib is 123 mb, the resulting new one 6512 bytes without further check whats in it (if any debug symbols remain)
The debug info for the failing job is reduced


mrodozov:
please test

smuzaffar:
+externals

PR #7279, opened 2021-09-08
Add the cuda-compatible-runtime test as a new external

fwyzard:


fwyzard:
please test

cuda-compatible-runtime.spec
```diff
@@ -0,0 +1,27 @@
...
+gcc -std=c99 -O2 -Wall %{SOURCE0} -I $CUDA_ROOT/include -L $CUDA_ROOT/lib64 -L $CUDA_ROOT/lib64/stubs -l cudart_static -l cuda -ldl -lrt -pthread -static-libgcc -o %_builddir/build/cuda-compatible-runtime || true
+
+%install
+
+mkdir %{i}/test
+if [ -f %_builddir/build/cuda-compatible-runtime ]; then
+  cp %_builddir/build/cuda-compatible-runtime %{i}/test/cuda-compatible-runtime
+else
+  ln -s /usr/bin/false %{i}/test/cuda-compatible-runtime
+fi
```
smuzaffar, line 24: @fwyzard , I do not understand this logic here. This is run on the build machine, so if it fails to build then what does that mean?

fwyzard: It could be a system for which we don't have a CUDA version working (e.g. GCC 12 when it comes out, and before CUDA is updated for it).
If that happens, I thought it safer not to make the build fail, and just replace the script used for the check with something that will always fail (e.g. `/usr/bin/false`).

fwyzard: Good news: the link to /usr/bin/false' works as intended.
Bad news: building the `cuda-compatible-runtime` binary failed on what should be a supported architecture :-(

Is there a way to access the build logs for this PR ?

fwyzard: OK, found them at https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-9a8f2f/18380/external/cuda-compatible-runtime/1.0-60cce5540ed7db5c5397f8d4cf6b0744/log

fwyzard: Ah... the sources on gitlab are redirecting to the login page :-(
How do I make cmsbuild log in ?

smuzaffar: ah this is coming from a private repo. Can we move the `test.c` to come public repo? Though the `cmsbuild` user has valid krb5 token but as you are downloading a single file using `https` protocol and that requires username/password. I do not know if there is a gitlab url which can use krb5 token to download a single file

fwyzard: I think I found a way to make it work, now I'm working through a couple other issues...

fwyzard:
please test

cuda-compatible-runtime.spec
```diff
@@ -0,0 +1,26 @@
+### RPM external cuda-compatible-runtime 1.0
+
+%define branch master
+%define commit 18d5a51bfb32fbb7b765dd2eecd14e193cce8126
+
+Source: git+https://:@gitlab.cern.ch:8443/cms-patatrack/%{n}.git?obj=%{branch}/%{commit}&export=%{n}&filter=test.c&output=/%{n}-%{realversion}.tgz
```
smuzaffar: this still means that only users with read access to this repo can build this external ... right ? This will stop other users to build externals

fwyzard:
All CMS members should have access.

cuda-compatible-runtime.spec
smuzaffar: please change the `filter=test.c` to `filter=./test.c` . cmsBuild uses https://github.com/cms-sw/pkgtools/blob/V00-32-XX/cmsBuild#L531 to filter the files which means they will have `./` infront

fwyzard:
please test

fwyzard:
If this works, the next step would be to replace `$CMSSW_BASE/config/SCRAM/hooks/runtime/00-nvidia-drivers` according to https://github.com/cms-sw/cmssw-config/pull/85 .

smuzaffar:
@fwyzard ,  to update `00-nvidia-drivers` , you need to open pr for https://github.com/cms-sw/cmssw-config e.g. https://github.com/cms-sw/cmssw-config/blob/scramv3/SCRAM/hooks/runtime/00-nvidia-drivers

fwyzard:
Done.

fwyzard:
please test

smuzaffar:
+externals

PR #7284, opened 2021-09-08
Drop cgal

iarspider:
Notice: CGAL is now (>= 5.0) a header-only library.


iarspider:
please test

smuzaffar:
@iarspider , looks like this external is also not needed any more. It was only used ion one cmssw package but has been dropped ( https://github.com/cms-sw/cmssw/pull/18684#issuecomment-302248100 ) few years ago. I would suggest to remove this external too.

smuzaffar:
test parameters:
- full_cmssw = true

smuzaffar:
please test

smuzaffar:
+externals

PR #7293, opened 2021-09-09
[BuildRules] Use CUDA runtime compatibility to set cuda env

smuzaffar:
this has been tested via https://github.com/cms-sw/cmssw-config/pull/85


smuzaffar:
+externals

PR #7294, opened 2021-09-10
Remove doxygen from cmssw dist

smuzaffar:
Doxygen is only used to generate cmssw doxygen doc and there is no need to distribute it via cmssw. It can be used from the system ( cmssw/*:amd64 containers do contain doxygen )


smuzaffar:
please test

smuzaffar:
+externals

PR #7305, opened 2021-09-14
use default aiohttp version 3.7.4

smuzaffar:
This should avoid the security warning https://github.com/cms-sw/cmsdist/security/dependabot/pip/requirements.txt/aiohttp/open as github system treats 3.7.4 as newer than 3.7.4.post0

### DIFF
--- a/pip/requirements.txt
+++ b/pip/requirements.txt
@@ -12,7 +12,7 @@
 #flags or patching, please use package_name.file
 #############################################################################
 absl-py==0.13.0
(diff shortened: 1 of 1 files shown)


smuzaffar:
please test

smuzaffar:
+externals
`aiohttp` is only used by `dqmgui` and nothing in cmssw use it. As externals build process is successful so no need to wait for the rest of the tests. I am merging it for 11h00 IB

PR #7310, opened 2021-09-15
Update form and clean up recipe

iarspider:


iarspider:
please test

form.spec
```diff
@@ -8,12 +8,12 @@ Requires: zlib
 %endif
 
 %prep
-%setup -q -n form-4.1
+%setup -q -n form-4.2.1
```
smuzaffar: @iarspider , please change this to `form-%{realversion}` so that next time we can just update the version in one place

PR #7318, opened 2021-09-17
Drop graphviz

iarspider:


iarspider:
please test

smuzaffar:
@iarspider , I think we can drop graphviz, it was used by doxygen but we have dropped that. So this can be remove too. Just move the spec in the archive dir

iarspider:
Will do that, thanks!

iarspider:
please test

smuzaffar (approved):

PR #7341, opened 2021-09-29
Buildrule tag update to V07-00-00

smuzaffar:
tag is same as V06-03-06. This is to allow old release to still use V06 with out private header file usage check


smuzaffar:
please test

smuzaffar:
+externals
new tag `V07-00-00` is same as `V06-03-06`

PR #7344, opened 2021-09-29
[12.0.X] Build rules updated to V06-03-07

smuzaffar:
This is same as  V06-03-06 but without private header file check enable. 12.0.X is not clean fo private header usage.


smuzaffar:
test parameters:
- full_cmssw = true

smuzaffar:
please test

PR #7348, opened 2021-09-30
simplify bootstrap-driver

smuzaffar:
drop unused archs and simplify the boot strap driver generation.


smuzaffar:
+externals
PR tests do not make use of it. It is only used when we build bootstrap for new archs.

PR #7353, opened 2021-10-01
update conan and fix dip build

smuzaffar:
Looks like conan server certificate is expire and it resulted in failure to build `dip`
```
ERROR: Failed requirement 'log4cplus/2.0.5' from 'dip/6.0.0'
ERROR: HTTPSConnectionPool(host='center.conan.io', port=443): Max retries exceeded with url: /v1/ping (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: certificate has expired (_ssl.c:1129)')))
```

new conan version seems to fix this issue.

PR #7354, opened 2021-10-02
build dip with cmake instead of conan

smuzaffar:
Changed dip build to use `cmake` to build it. `conan` download `log4cplus` sources during the build which mean re-build of dip might fail due to conan server issues ( as we have seen https://github.com/cms-sw/cmsdist/pull/7353 )
- added `log4cplus` as separate package and build with cmake
- build dip and platform-dependent using cmake


smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
+externals

PR #7355, opened 2021-10-04
Test if yoda_pyroot.patch is still necessary (after dropping Py2)

iarspider:


iarspider:
please test

davidlange6:
this patch has nothing to do with python2 vs 3..   I'm not sure that the Yoda python interface is tested so its not obvious that a positive test result is sufficient to conclude that this patch is not needed...

smuzaffar:
To support py2/3 in root, here was what was suggested by root team

```
As a follow up of our discussion a couple of weeks ago, we've been
working with Massimiliano on this PR (not merged yet):

https://github.com/root-project/root/pull/5083

The highlights are:
- We will place all the libs (included the PyROOT ones) in the
ROOTSYS/lib directory, instead of using subdirectories for each Python
version. The PyROOT libraries will have the version number in the
library name, e.g. libcppyy2_7.so, to allow to build for both Python2
and Python3 at the same time. This solves the problem of having
duplicate library names when symlinking everything into the lib of CMSSW.
- We will only generate one libTPython.so. The reason is the PCMs: we
can't have libTPython2_7.so and libTPython3_6.so, plus their respective
PCMs, because all the PCMs are loaded by ROOT and we would have
duplicated symbols.
- To solve the Yoda case, we propose that Yoda does not link anymore
against libTPython, but against libcppyy. libcppyy contains the
functionality that Yoda needs, it is just a matter of including a
different header and slightly changing the function names that are being
invoked (we can give you more details about this). libcppyy will have
two versions (e.g. libcppyy2_7 and libcppyy3_6), so Yoda will be able to
be used from both Pythons. 
```

I think without this patch yoda was even failing to build

smuzaffar:
please test for CMSSW_12_1_ROOT6_X

davidlange6:
interesting... thanks for the correction. It seems however that this change might be desirable in any case, but indeed better for the Yoda team to maintain that.

smuzaffar:
please test

smuzaffar:
+externals

PR #7411, opened 2021-10-26
[llvm] use same host triple as set by gcc

smuzaffar:
For ppc64le gcc11 builds, gcc properly use host triple to `powerpc64le-redhat-linux-gnu` while llvm uses `powerpc64le-unknown-linux-gnu` which causes cmssw build to fail as llvm could not find gcc headers. This change should make sure that same host tripel is used by llvm and gcc


smuzaffar:
please test

smuzaffar:
test parameters:
- addpkg = CondFormats,CondTools,CondCore

smuzaffar:
please test

PR #7416, opened 2021-10-28
cleanup scram provides

smuzaffar:
- Drop extra provides from SCRAM: There is no PERL modules in SCRAM any more
- Added python3 provide in fakesystem package


smuzaffar:
please test

smuzaffar:
+externals
Build of externals went fine.

PR #7417, opened 2021-10-28
[git] Added extra perl provides to avoid system package deps

smuzaffar:
We do not make much use of git perl interface. These extra provides are to avoid perl system modules dependencies. If one needs to use git perl interface then they need to install extra perl modules on system


smuzaffar:
please test

PR #7419, opened 2021-10-29
Update cuda-compatible-runtime to add kernel checks

fwyzard:


fwyzard:
enable gpu

fwyzard:
please test

fwyzard:

fwyzard:
please test for cc8_amd64_gcc9

fwyzard:
please test for slc7_amd64_gcc10

fwyzard:

fwyzard:
please abort

fwyzard:
please test

fwyzard:
@gartung this should be the first part of the fix for the automatic checks, to support the environment at NERSC

fwyzard:
please test

fwyzard:
@gartung, https://github.com/cms-sw/cmssw-config/pull/87 should be the second part of the fix for the automatic checks, to support the environment at NERSC

fwyzard:

fwyzard:
please test for cc8_amd64_gcc9

fwyzard:

fwyzard:
please test for slc7_amd64_gcc10

gartung:
I can confirm that the changes made to the CMSSW config and this executable set the correct LD_LIBRARY_PATH on Perlmutter at NERSC.

fwyzard:
Thanks for the check!

cuda-compatible-runtime.spec
```diff
@@ -1,18 +1,22 @@
...
 %prep
 %setup -n %{n}
 
 %build
+## INCLUDE cuda-flags
+# defines nvcc_stdcxx and cuda_flags_4
+
 rm -rf %{_builddir}/build && mkdir %{_builddir}/build
-gcc -std=c99 -O2 -Wall test.c -I $CUDA_ROOT/include -L $CUDA_ROOT/lib64 -L $CUDA_ROOT/lib64/stubs -l cudart_static -l cuda -ldl -lrt -pthread -static-libgcc -o %{_builddir}/build/cuda-compatible-runtime # || true
+$CUDA_ROOT/bin/nvcc %{nvcc_stdcxx} -O2 -g %{cuda_flags_4} test.cu -I $CUDA_ROOT/include -L $CUDA_ROOT/lib64 -L $CUDA_ROOT/lib64/stubs --cudart static -ldl -lrt --compiler-options '-Wall -pthread' -o %{_builddir}/build/cuda-compatible-runtime # || true
```
smuzaffar: @fwyzard , isn't it going to fail if gcc version not compatible with cuda?

smuzaffar: @fwyzard ping

fwyzard: > @fwyzard , isn't it going to fail if gcc version not compatible with cuda?

Yes, good point.
If that happens, there really is no version of the runtime we want to use...

smuzaffar: @fwyzard , should we fix this so that it does not fail the build process? I mean in that case we should still create cuda-compatible-runtime which exit with non-zero code e.g.

```
if [ $(%{cuda_gcc_support}) = true ] ; then
  $CUDA_ROOT/bin/nvcc %{nvcc_stdcxx} -O2 -g %{cuda_flags_4} test.cu -I $CUDA_ROOT/include -L $CUDA_ROOT/lib64 -L $CUDA_ROOT/lib64/stubs --cudart static -ldl -lrt --compiler-options '-Wall -pthread' -o %{_builddir}/build/cuda-compatible-runtime
else
  echo "CUDA ${CUDA_VERSION} is not compatiable with GCC ${GCC_VERSION}" > %{_builddir}/build/cuda-compatible-runtime
  echo "false" >> %{_builddir}/build/cuda-compatible-runtime
  chmod +x %{_builddir}/build/cuda-compatible-runtime
fi
```

smuzaffar: @fwyzard , what do you think about https://github.com/cms-sw/cmsdist/pull/7419/files#r741837428 ?

fwyzard: sorry, I forgot :-(
let me see...

fwyzard: If it works, sounds like a good idea.
Can we actually use `%{cuda_gcc_support}` before CUDA has been set up ?

I guess I can make the changes and try...

fwyzard: In fact, the `%install` part was already setting up a symlink to `/usr/bin/false` is the build failed; but I like the idea of a more verbose message.

fwyzard: as an alternative - is there a way to use the system compiler for building this, instead of the one bundled with CMSSW ?

fwyzard:
please test

fwyzard:
do we have an architecture that does not support CUDA ?
maybe something with gcc12 ?

smuzaffar:
no, currently we do not have any arch without cuda/gcc support. I have tested your change by forcing cuda command to fail and generate the script. All look good

```
>./test/cuda-compatible-runtime
>echo $?
1
>./test/cuda-compatible-runtime -h
Usage: ./test/cuda-compatible-runtime [-h|-v]

Options:
  -h        Print a help message and exits.
  -v        Be more verbose.
>./test/cuda-compatible-runtime -v
CUDA 11.4.2-0939a3504c82d9c20346029080003d72 is not compatible with GCC 9.3.0
```

smuzaffar:
please test

smuzaffar:
@fwyzard , when we see a message like 
https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-49f516/20771/runTheMatrixGPU-results/11634.506_TTbar_14TeV+2021_Patatrack_PixelOnlyTripletsGPU+TTbar_14TeV_TuneCP5_GenSim+Digi+Reco+HARVEST/step2_TTbar_14TeV+2021_Patatrack_PixelOnlyTripletsGPU+TTbar_14TeV_TuneCP5_GenSim+Digi+Reco+HARVEST.log
```
CUDA runtime version 11.4, driver version 11.5, NVIDIA driver version 495.29.05
CUDA device 0: Tesla T4 (sm_75)
```
i.e. `CUDA device 0` does this mean we did not find any usable device?

smuzaffar:
+externals

fwyzard:
> @fwyzard <https://github.com/fwyzard> , when we see a message like
>
> CUDA runtime version 11.4, driver version 11.5, NVIDIA driver version 495.29.05
> CUDA device 0: Tesla T4 (sm_75)
>
> i.e. CUDA device 0 does this mean we did not find any usable device?
>

No, it's fine, it means that it found one device.
CUDA devices count from 0.

>

PR #7422, opened 2021-11-01
update bootstrap files

smuzaffar:
This only updates the bootstrap driver files and it should not re-build any external


smuzaffar:
please test

smuzaffar:
+externals
looks good as externals build without any rebuild.

PR #7424, opened 2021-11-03
allow version suffix options

smuzaffar:
- update crab suffix due to change in crab-build
- explicitly update versions to avoid version suffix


smuzaffar:
please test

smuzaffar:
please test

PR #7426, opened 2021-11-03
Py3 cleanup: drop py3-rep

smuzaffar:
Dropped `py3-rep` and extra py3 packages which are only used by `py3-rep`


smuzaffar:
please test with cms-sw/cmssw#35878

smuzaffar:
+externals
dropped few unused python packages

PR #7438, opened 2021-11-09
New specfile for py3 crabserver REST interface

mapellidario:
While we migrate the dmwm/CRABServer code related to crabserver REST interface, we convened that it is convinient to have two separate specfiles, one for a py2 environment for crabserver and one for py3.

After the migration is complete, we plan to replace `crabserver.spec` with `crabserver-py3.spec`.

PR #7452, opened 2021-11-16
crabserver - single specfile for both py2 and py3 env

mapellidario:
Supersedes #7438.

While we migrate the dmwm/CRABServer code related to crabserver REST interface, we convened that it is convinient to have a single specfile that can build both py2 and py3 environments, so that we do not need to adapt the name of the application (`crabserver`) down the road, for example in in dmwm/CMSKubernetes, dmwm/deployment, in our own code, ...

After the migration is complete, we plan to deprecate all the `%if` statements and to keep the py3 version only.

### DIFF
--- a/crabserver.spec
+++ b/crabserver.spec
@@ -1,22 +1,36 @@
 ### RPM cms crabserver v3.210701
 
+# This specfile accepts both following types of tags:
+# - v3.210701 -> builds py2 environment
+# - py3.211104patch1 -> build py3 environment
+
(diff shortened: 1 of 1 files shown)


muhammadimranfarooqi:
+1 to efed0b3ed8fff7f0c26b6f9eb503f88615234b1a on slc7_amd64_gcc630.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/7452/test_efed0b3-20211116-181504.log" target="_blank">here</a>.

PR #7462, opened 2021-11-23
move data packages to github

smuzaffar:
http://cmsdoc.cern.ch is behind SSO and data sources can not be downloaded. This PR proposes to move these data to either github or cmsrep


smuzaffar:
please test

smuzaffar:
+externals
No data file was changed, these are just moved to either github or cached on cmsrep ( for non-sso access)

PR #7463, opened 2021-11-24
always create realverson directory in share area

smuzaffar:
This change avoids duplicate data directories in `RPM_PREFIX/share/cms/data-*` we now always create `realversion` directory


smuzaffar:
please test

PR #7464, opened 2021-11-24
Use original config version instead of wrapper

ddaina:
solves https://github.com/dmwm/CRABClient/issues/5119

FYI @belforte


ddaina:
please test

smuzaffar:
+externals

PR #7466, opened 2021-11-25
Added cs8 support, remove harded coded usage of amd64

smuzaffar:
- use `uname -m` instead of hardcoded `amd64`


smuzaffar:
please test

PR #7473, opened 2021-11-25
Update 00-nvidia-drivers to check if the CUDA devices can run a kernel

smuzaffar:
see https://github.com/cms-sw/cmssw-config/pull/87


smuzaffar:
please test

PR #7484, opened 2021-12-01
CRAB: update crab-prod to v3.211130

ddaina:
fyi @belforte

### DIFF
--- a/crab-build.file
+++ b/crab-build.file
@@ -40,7 +40,7 @@ echo dbs >> %{i}/etc/crab_py_pkgs.txt
 echo RestClient >> %{i}/etc/crab_py_pkgs.txt
 
 #Copy WMCore
-if [ "%{crab_type}" != "dev" ] ; then
+if [ "%{crab_type}" == "pre" ] ; then
   for pkg in %{wmcore_packages} ; do
     if [ -d %{_builddir}/WMCore/src/python/${pkg} ] ; then
       rsync -a %{_builddir}/WMCore/src/python/${pkg} %{i}/lib/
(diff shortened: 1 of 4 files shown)


ddaina:
please test

crab-build.file
```diff
@@ -40,7 +40,7 @@ echo dbs >> %{i}/etc/crab_py_pkgs.txt
 echo RestClient >> %{i}/etc/crab_py_pkgs.txt
 
 #Copy WMCore
-if [ "%{crab_type}" != "dev" ] ; then
+if [ "%{crab_type}" == "pre" ] ; then
```
smuzaffar, line 43: @ddaina , is this what you want? `pre` will use `WMCore` and `prod` and `dev` will use   `CRABClient/WMCoreConfiguration.py` ?

belforte: yes, that's what we want. We use `pre` as `previous` so users can fall back to it in case of problems when we move from `dev` to `prod`. We do not have means to do a menaingful validation in "preprod" after we do our best in `dev` which in this case includes having CMSSW_12 users use it !

smuzaffar:
+externals

crab-build.file
smuzaffar: my bad, I thought `pre` stands for for pre-production/pre-release :-)

belforte: no reason to feel bad. It was a dirty expedient I dreamed off. And not standard in any way. Yes, it is a devious use of the `pre` prefix , sorry about that. Did not want to ask you to introduce a `crab-old`

PR #7493, opened 2021-12-09
Upgrade dbs3-client to v4.0.1

todor-ivanov:
!!! DO NOT MERGE YET !!!


muhammadimranfarooqi:
+1 to 4a64e6d03199abdebd49ce7c6e2b9006a7a012cb on slc7_amd64_gcc630.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/7493/test_4a64e6d-20211209-141504.log" target="_blank">here</a>.

todor-ivanov:
FYI @belforte 
Hi, Stefano, this PR is in order to start  testing of the new dbs3-client version in testbed and since  CRABServer and TaskWorker also depend on this I wanted you to give green light before merging this. Here is the related issue in WMCore where we keep track of this upgrade.

https://github.com/dmwm/WMCore/issues/10884

belforte:
thanks for headup Todor. Let me add @mapellidario as well.
I am confused as to how we can test this. Seems that simply from some point onward all builds will use new version. Wouldn't it be better to have a differently name spec file to give us flexibility ? A less monolithic approach works better in current CRAB development cycle, which is heavily based on continuous integration.

todor-ivanov:
Actually looking twice this is the same also for T0, 
So we would need to warn also T0 team: @drkovalskyi  ,  @germanfgv , @jhonatanamado ,

amaltaro:
It's also important to mention that, since cmsweb-testbed DBSReader was switched to the Go server implementation **yesterday**, this new client is likely already needed by any service that needs to talk to DBS server in cmsweb-testbed.

Stefano, I guess by "less monolithic approach" you meant to have a handle to keep different versions between testbed and production? If so, then I think the best is to deal with that once we move away of RPMs.

belforte:
yeah.. you see things from WMAgent/WMCore POV only. I need to be able to build prod and dev rpm's with different version of the dependencies, since I can't influence the timing of you changing the dependencies.
Maybe I misunderstood the message from Todor though. Is this a request for us to test the new DBS client, or did you test already and we will simply use it are very unlikely to find problems ? I think that common spec files used by many should not be changed w/o testing first.

amaltaro:
> yeah.. you see things from WMAgent/WMCore POV only. I need to be able to build prod and dev rpm's with different version of the dependencies, since I can't influence the timing of you changing the dependencies.

Okay, thanks for confirming this! Are you using cmsdist in a different way to accomplish it? Because I fail to see how we can have a spec version for dev/testbed/production (of course, some control can be done through deployments and tags deployed in the cluster).
We could accomplish it with a few diff ways though, like:
1) have a test and prod spec for every single spec (this won't fly!!! maintaining multiple specs for the same package is a nightmare and people keep using the wrong spec)
2) have different branches in cmsdist
3) or wait for the deprecation of cmsdist and control everything within each project (through a requirements text file, or something similar)

>  Maybe I misunderstood the message from Todor though. Is this a request for us to test the new DBS client, or did you test already and we will simply use it are very unlikely to find problems ? I think that common spec files used by many should not be changed w/o testing first.

We wanted to decouple the package update from the DBSReader tests. But I agree, this touches many services and we better have some basic tests performed before pushing this change in.

todor-ivanov:
hi @belforte we haven't tested it yet. This  PR was supposed to give us the chance to have a fully deployed version with all dependencies (both first and n-th order) of WMCore central services and WMAgent linked to the latest dbs3-client version, so we can start proper validation.

belforte:
I agree with all the rationals, I simply thought that a spec file with a new name and a corresponding edit in the Requires list of the packages used for the test would have achieved the needed "isolation".  But a new branch in cmsdist is maybe cleaner. All in all those comp_gcc*** names are ... well.. not so important to stick to! Simply that our automated build picks things from gcc_630 so that's our "production branch".  Let's not make a storm in a teacup anyhow, we have all reasons to expect dbs client v4 to work. Maybe we can just go for it in CRAB as well. Is this only for python3 builds ?

belforte:
answering myself, yes, this only affects py3 builds which for CRAB are currently experimantal anyhow. So go ahead :-)
Sorry if I got scared too soon !

Answering earlier Alan's question, yeah.. having different dependencies for different CRAB versions is a bit ugly atm https://github.com/cms-sw/cmsdist/blob/cf99360cd1bedb7ff2f735a53929c4fda0ccbe87/crabserver.spec#L14-L30
but if we want to look at `deprecation of cmsdist ` let's have an ad hoc discussion

muhammadimranfarooqi:
+1 to 22e8afebceca8d921e42beaeec4b2019e8b89ce6 on slc7_amd64_gcc630.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/7493/test_22e8afe-20211210-082004.log" target="_blank">here</a>.

amaltaro:
Thanks Todor for getting this started! After resolving an issue with dbs3-pycurl, I created this PR updating both dbs3-client and dbs3-pycurl: https://github.com/cms-sw/cmsdist/pull/7508

Can you please close this one? @todor-ivanov
Thanks

todor-ivanov:
Thanks @amaltaro ! Closing this one.

PR #7503, opened 2021-12-10
Fixes needed for GCC11 IBs

smuzaffar:
GCC 11 IBs rebuilds these packages and fails as cmsBuild adds the hash suffix. I force updated the versions so that cmsBuild do not add hash suffixes


smuzaffar:
please test

smuzaffar:
+externals

PR #7504, opened 2021-12-11
fix git to not fail is build direction is not reachable

smuzaffar:
This fixes the errors like [a] when the directory used for building git package not reachable. This happens as `/data/cmsbuild` directory exists but is not readable by the user who is now trying to use git.

[a]
```
fatal: unable to access '/data/cmsbuild/muz/w/tmp/BUILDROOT/41d8deb8bae89161b7233b9737d9ca24/opt/cmssw/slc7_aarch64_gcc9/external/git/2.23.0-41d8deb8bae89161b7233b9737d9ca24/etc/gitconfig': Permission denied
```


smuzaffar:
please test

smuzaffar:
+externals

PR #7509, opened 2021-12-14
New dasgoclient version

vkuznet:
It provides proper URLs to `config dataset=/z/v/x` queries pointing to ReqMgr config cache. Switched to new GO compiler and updated code dependencies.


smuzaffar:
please test

PR #7522, opened 2021-12-17
New dasgoclient version

vkuznet:
This PR fixes issue reported in  https://github.com/cms-sw/cmssw/issues/36527

Summary: DAS must support multiple services URLs due to different frontends, like cmsweb.cern.ch, cmsweb-testbed.cern.ch, cmsweb-k8s-prod.cern.ch. In addition, some services like DBS uses different instances, like prod vs int, etc. To accommodate all possible use-cases DAS library first check for service URL, e.g. defined by DBS_URL, or RUCIO_URL, etc, then uses das maps, and finally construct proper URL. Depending on a client, e.g. dasgoclient CLI or das2go web client, the behavior may change, since CLI should use full URL, while web may use just API end-point and point to the same host, etc.


smuzaffar:
please test

Issue #7534, opened 2022-01-07
Use case-insensitive comparision for checking dependencies of python packages


iarspider:
There is a problem with the way cmsdist checks python packages for missing dependencies.

The check is done by first getting package name from output of pip show ( https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_12_3_X/master/build-with-pip.file#L105 ), then looking for that name in output of pip check ( https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_12_3_X/master/build-with-pip.file#L107 ). The problem is in capitalization of PKG_NAME. E.g., for `py3-cachecontrol`, the value of `PKG_NAME` is `CacheControl`, but in the output of `pip3 check` the package is named `cachecontrol` (all lowercase), and the following grep ^${PKG_NAME} will not return anything
```
$ pip3 check --disable-pip-version-check cachecontrol
cachecontrol 0.12.6 requires msgpack, which is not installed.
```

We should do a case-insensitive grep for `PKG_NAME`.

smuzaffar:
> Ivan Razumov
>The check is done by first getting package name from output of pip show ( https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_12_3_X/master/build-with-pip.file#L105 ), then looking for that name in output of pip check ( https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_12_3_X/master/build-with-pip.file#L107 ). The problem is in capitalization of PKG_NAME. For py3-cachecontrol, the value of PKG_NAME is CacheControl, but in the output of pip3 check :
$ pip3 check --disable-pip-version-check cachecontrol
cachecontrol 0.12.6 requires msgpack, which is not installed.
> the package is named cachecontrol (all lowercase), and the following grep ^${PKG_NAME} will not return anything.
> I think making a case-insensitive grep (grep -i) should work. I will create a PR with that fix + any missing dependencies that are exposed after this fix.
> One of the bigger missing dependencies is rust (for py3-setuptools-rust -> py3-cryptography). 
> to avoid building and distributing rust, we can try using pre-built wheel for py3-cryptography
> same issue is with py3-pygithub: it's PKG_NAME name is PyGithub, but pip check shows it as pygithub.
> a different issue with py3-keyring - it doesn't determine it's version correctly: the log file contains line Successfully installed keyring-0.0.0. Thus grepping for keyring 23.1.0 will not return anything

[...]

PR #7537, opened 2022-01-11
 Fix check for requirements (#7534), add missing dependencies

iarspider:
@smuzaffar I can move c823a9b7a60bd11c72c76816852bda5cf9d8f3d8 (proposed fix for #7534) to a separate PR to avoid rebuilding unchanged python packages if necessary.


iarspider:
please test

iarspider:
There are some dependencies (build ones) that are not caught by `pip check`. This is a side-effect of using `--no-build-isolation`. For example, `py3-correctionlib` actually requires `py3-make` and `py3-cmake` (although it seems that it can be installed even if these packages are missing and only `make` and `cmake` commands are available).

smuzaffar:
@iarspider , yes please remove the change for `build-with-pip-file`. I think all you need is to add the following  in
- cmsdist/pip/cachecontrol.file
```
%define patchsrc0 sed -i -e 's|name="CacheControl"|name="cachecontrol"|' setup.py
```

- cmsdist/pip/keyring.file
```
%define patchsrc0 sed -i -e 's|\\[metadata\\]|[metadata]\\nversion = %{realversion}|' setup.cfg
```

note that `keyring` version is set to `0.0.0.0` and check at https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_12_3_X/master/build-with-pip.file#L107  is still going to fail as we also search for `%realversion`. 

I think we can improve `build-with-with.file` but I would suggest to change it only when we have updated `gcc, python3, pip3 or setuptools`

smuzaffar:
I have a `rust.spec` which allows to use or llvm, I will open a PR which you can use for this PR test

iarspider:
@smuzaffar `py3-keyring` should be fixed now (let me check once more) - the version is set with using `py3-setuptools-scm`, which was one of the missing deps. For `py3-cachecontrol`, this won't be necessary once we fix `build-with-pip.file`. For `rust` - this would be useful, thanks!

smuzaffar:
see https://github.com/cms-sw/cmsdist/pull/7538 for rust.
I would suggest to fix `cachecontrol.file` and undo the `build-with-pip.file` changes. I have a better fix for `build-with-pip.file` which we will only include when we will updateany base poackage. For now, it is better to include these changes in 12.3.X without massive rebuilt

iarspider:
> I would suggest to fix `cachecontrol.file` and undo the `build-with-pip.file` changes. I have a better fix for `build-with-pip.file` which we will only include when we will updateany base poackage. For now, it is better to include these changes in 12.3.X without massive rebuilt

Sure. I have a few more fixes (for build-time dependencies), will push them soon.

iarspider:
please test

smuzaffar:
why not use the rust from #7538? The one you have is not going to work for non-x86_64 archs

iarspider:
please abort

iarspider:
@smuzaffar I thought it was not yet ready. Merged.

iarspider:
please test

cmssw-tool-conf.spec
```diff
@@ -139,6 +139,7 @@ Requires: python_tools
 Requires: dasgoclient
 Requires: mxnet-predict
 Requires: dablooms
+BuildRequires: rust
```
smuzaffar: as you already have `rust` dep in `setuptools-rust` so please remove this line. I added it to make sure that PR tests build `rust`.

pip/setuptools-rust.file
```diff
@@ -0,0 +1 @@
+Requires: rust py3-setuptools-scm py3-semantic-version py3-toml py3-typing-extensions
```
smuzaffar: remove rust from `Requires` and add `BuildRequires: rust` so that we do not deploy it with cmssw. We only need it at build time

smuzaffar: by the way, is `rust` a deps on this or `crytography`? I think it is needed by  `crytography`  so better to add `BuildRequires: rust` in  `crytography.file`

iarspider: crytography needs setuptools-rust, which needs rust.

smuzaffar: ok. I just wanted to see if we can avoid distributing rust along with cmssw. May be then just add
```
BuildRequires: rust
```
for both `setuptools-rust.file` and `crytography.file`. This way rust will not be deployed along with cmssw

iarspider: Sure.

pip/cryptography.file
```diff
@@ -0,0 +1 @@
+Requires: py3-setuptools-rust py3-cffi
```
smuzaffar, line 3: please also add
```
%define PipPreBuild export CARGO_HOME=%{_tmppath}/cargo_home
```
this will allow cargo to use our buidl area instead of `~/.cargo` directory

build-with-pip.file
```diff
@@ -104,7 +104,7 @@ if [ $(cat files.list | wc -l) -eq 1 ] ; then
    pip3 install --no-clean --no-deps --no-index --no-build-isolation --no-cache-dir --disable-pip-version-check --user -v %{PipInstallOptions} %{PipBuildOptions} $PIPFILE
    PKG_NAME=$(pip3 show %{pip_name} --disable-pip-version-check | grep '^Name:' | sed 's|^Name: *||;s| ||g')
    [ "${PKG_NAME}" = "" ] && exit 1
-   DEPS=$(pip3 check --disable-pip-version-check | grep "^${PKG_NAME}  *%{realversion}  *requires " | sed 's|,.*||;s|.* |py3-|' %{PipDepsPost} | tr '\n' ' ')
+   DEPS=$(pip3 check --disable-pip-version-check | grep -i "^${PKG_NAME}  *%{realversion}  *requires " | sed 's|,.*||;s|.* |py3-|' %{PipDepsPost} | tr '\n' ' ')
```
smuzaffar: do you still need this change?

iarspider: I will remove it before final test

smuzaffar: if you remove it before the PR tests then it will take less time to run tests (otherwise it will rebuild every thing for PR tests) :-)

iarspider: Yes, but it will also catch anything I might've missed

smuzaffar: in that case you might want to remove matching of `%{realversion}`  ( as we have seen some packages returns version `0.0.0.0` and that is not caught by this check.

iarspider:
please test

pip/correctionlib.file
```diff
@@ -1,3 +1,3 @@
 Requires: py3-pybind11 py3-numpy py3-pydantic py3-python-rapidjson py3-rich
-BuildRequires: py3-setuptools-scm py3-scikit-build py3-cmake py3-make
+BuildRequires: py3-setuptools-scm py3-scikit-build cmake py3-make
```
smuzaffar, line 2: do we need py3-make ?

iarspider: Technically, yes. Actually - not sure.

iarspider:
please test for slc7_aarch64_gcc11

iarspider:
please test for slc7_ppc64le_gcc11

smuzaffar:
please test for slc7_aarch64_gcc11

smuzaffar:
please test for cs8_amd64_gcc11

iarspider:
@smuzaffar I think it's ok to merge. The error doesn't seem to be related to PR

smuzaffar:
+externals

PR #7546, opened 2022-01-14
Updated cmsos to support rhel OS

smuzaffar:
For Red Hat Enterprise version 8 and above , cmsos should now return rhelX instead of slcX


smuzaffar:
please test

smuzaffar:
+externals

PR #7582, opened 2022-01-24
[AArch64] build cmssw aarch64 with -mno-outline-atomics file

smuzaffar:
see https://github.com/cms-sw/cmssw/issues/36788

### DIFF
--- a/OpenBLAS.spec
+++ b/OpenBLAS.spec
@@ -22,7 +22,7 @@ make %{build_opts} TARGET=CORE2
 make %{build_opts} TARGET=ARMV8
 %endif
 %ifarch ppc64le
-make %{build_opts} CFLAGS="%{ppc64le_build_flags}"
+make %{build_opts} CFLAGS="%{arch_build_flags}"
 %endif
 
 %install
--- a/OpenBLAS.spec
+++ b/OpenBLAS.spec
@@ -19,7 +19,7 @@ Patch1: OpenBLAS-disable-tests
 make %{build_opts} TARGET=CORE2
 %endif
 %ifarch aarch64
-make %{build_opts} TARGET=ARMV8
+make %{build_opts} TARGET=ARMV8 CFLAGS="%{arch_build_flags}"
 %endif
 %ifarch ppc64le
 make %{build_opts} CFLAGS="%{arch_build_flags}"
--- a/boost.spec
+++ b/boost.spec
@@ -1,5 +1,5 @@
 ### RPM external boost 1.75.0
-
+## INCLUDE compilation_flags
 %define tag 3defebd61ecb0970c0046c85384bb34ec9572ac3
 %define branch cms/v%realversion
 %define github_user cms-externals
@@ -17,6 +17,12 @@ case %cmsos in
   *) TOOLSET=gcc ;;
 esac
 
+# enable boost::mpi
+%if "%{?arch_build_flags:set}" == "set"
+echo 'using gcc : : : <cxxflags>"%{arch_build_flags}" ;' > user-config.jam
+echo 'using gcc : : : <cflags>"%{arch_build_flags}" ;' >> user-config.jam
+%endif
+
 pushd tools/build
   sh bootstrap.sh ${TOOLSET}
   mkdir ./tmp-boost-build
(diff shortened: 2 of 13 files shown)


smuzaffar:
test parameters:
- full_cmssw = true
- enable_test = threading
- workflow_threading = 4.44,13.0,11634.24,140.56

smuzaffar:
please test for slc7_aarch64_gcc11

fwyzard:
Building with `-mno-outline-atomics` will require hardware support for the LSE instructions from the ARMv8.1 ISA.
If our build machines support only the ARMv8.0 ISA, the build is expected to fail at run time :-/

dan131riley:
> Building with `-mno-outline-atomics` will require hardware support for the LSE instructions from the ARMv8.1 ISA.

I thought the point of the outline atomics was to enable use of LSE at runtime, so I expected `-mno-outline-atomics` to revert to the pre-10.1 behavior that didn't use LSE instructions.

> If our build machines support only the ARMv8.0 ISA, the build is expected to fail at run time :-/

Our build machines (at least the ones I'm aware of) have been upgraded to ThunderX2 that support the ARMv8.1 ISA, and since gcc 10.1 have been using the LSE instruction set at runtime.  I was wondering if the use of LSE was causing the mysterious failures we've been seeing, so I was hoping that flag would disable LSE.

dan131riley:
> /bin/sh: line 1: 2956474 Illegal instruction     (core dumped) /usr/bin/touch /data/cmsbuild/jenkins_b/workspace/ib-run-pr-tests/CMSSW_12_3_X_2022-01-23-2300/tmp/slc7_aarch64_gcc11/cache/edm_edmPluginRefresh

This is really weird--/usr/bin/touch crashed?  I have no idea how adding that flag could cause that to crash.

fwyzard:
> Our build machines (at least the ones I'm aware of) have been upgraded to ThunderX2 that support the ARMv8.1 ISA

Ah, OK.
Do we have interactive access to any of those nodes ?

smuzaffar:
@fwyzard , you can login to those node. I am going to send you the instructions via email

fwyzard:
> > Building with `-mno-outline-atomics` will require hardware support for the LSE instructions from the ARMv8.1 ISA.
> 
> I thought the point of the outline atomics was to enable use of LSE at runtime, so I expected `-mno-outline-atomics` to revert to the pre-10.1 behavior that didn't use LSE instructions.

OK, I guess I misread the GCC's documentation:

> `-moutline-atomics`
> `-mno-outline-atomics`
>
>    Enable or disable calls to out-of-line helpers to implement atomic operations. These helpers will, at runtime, determine if the LSE instructions from ARMv8.1-A can be used; if not, they will use the load/store-exclusive instructions that are present in the base ARMv8.0 ISA.
>
>    This option is only applicable when compiling for the base ARMv8.0 instruction set. If using a later revision, e.g. `-march=armv8.1-a` or `-march=armv8-a+lse`, the ARMv8.1-Atomics instructions will be used directly. The same applies when using `-mcpu=` when the selected cpu supports the `lse` feature. This option is on by default.

So
  - with `-march=armv8.1-a` or later, `-moutline-atomics`/`-mno-outline-atomics` has no effect, and LSE are always used
  - with `-march=armv8-a`, `-moutline-atomics` uses helpers to check if LSE are available and use them, or fall back to the old instructions otherwise
  - with `-march=armv8-a`, `-mno-outline-atomics` always uses the  old instructions
  
?

smuzaffar:
please test for slc7_aarch64_gcc11

lets re-run the tests

dan131riley:
> * with `-march=armv8.1-a` or later, `-moutline-atomics`/`-mno-outline-atomics` has no effect, and LSE are always used
> * with `-march=armv8-a`, `-moutline-atomics` uses helpers to check if LSE are available and use them, or fall back to the old instructions otherwise
> * with `-march=armv8-a`, `-mno-outline-atomics` always uses the  old instructions

I think that's correct.  We aren't specifying `-march` or `-mcpu` so I'd expect it to use the base ARMv8 instructions with `-mno-outline-atomics` (in fact, I got started on this line of investigation due to a handful crashes in the outline LSE routines).

smuzaffar:
correct, we are not setting `-mcpu/arch` for aarch64. We have two arm64 nodes 
- thunderx2: olarm-202
- thunderx: olarm-102

but we are currently using olarm-202 only. Let me build all externals and cmssw with `-mcpu=armv8-a` on  olarm-202 and see if that helps.

dan131riley:
> Refitted track not found in list
> ----- End Fatal Exception -------------------------------------------------

Well, that kills that theory.  Not only does the error reproduce with LSE disabled, but it also shows that the error reproduces in a single-threaded test (which I had not noticed before), which rules out synchronization primitives.

smuzaffar:
@dan131riley , yes looks like outline-atomics is not the issue. I have built gcc with outline atomics disabled ( https://github.com/cms-sw/cmsdist/pull/7593 ) and still the same error ( https://github.com/cms-sw/cmsdist/pull/7593#issuecomment-1025225126 )

PR #7605, opened 2022-02-04
Update crab client

belforte:


smuzaffar:
please test

belforte:
@smuzaffar  I hope there is a better solution than detect the 27a60cc501cf00215a28461ccdc2a04b suffix by this error and update the spec with it !
Anyhow this is not urgent since Andrea Sciaba' just found a subtle bug in CRABClient (which only affects HammerCloud of course) and at this point I'd rather fix that as well.  Still, kindly let me know how to solve this version naming.

smuzaffar:
@belforte , as `crab-build.file` file is updated so basically all `crab-*.spec` need to change the suffix. In this case suffix in `crab-dev.spec` also needs changes

belforte:
got it, thanks. Will try again with a new crab-dev tag and the fix for HammerCloud

smuzaffar:
please test

smuzaffar:
+externals

PR #7674, opened 2022-03-08
[TBB] Update to v2021.5.0 which contasins fixes for glibc-2.34

smuzaffar:
https://github.com/oneapi-src/oneTBB/pull/394
https://github.com/oneapi-src/oneTBB/pull/519

### DIFF
--- a/tbb.spec
+++ b/tbb.spec
@@ -1,4 +1,4 @@
-### RPM external tbb v2021.4.0
+### RPM external tbb v2021.5.0
 
 %define tag %{realversion}
 %define branch onetbb_2021


smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
@makortel @Dr15Jones any objection on updating TTB to latest v2021.5.0 version?

Dr15Jones:
I'm good with switching to the new version. As part of that, I'd like the `task_handle` API to be available. That seems to be controlled by 

https://github.com/oneapi-src/oneTBB/blob/690aaf497a78a75ff72cddb084579427ab0a8ffc/include/oneapi/tbb/detail/_config.h#L523-L525

and it is unclear to me if we get it for free from `__TBB_BUILD` or if we need to enable `TBB_PREVIEW_TASK_GROUP_EXTENSIONS`.

The API is now part of the official TBB documentation so it is odd that they still treat it as a _preview_ in the code itself.

makortel:
No objection, but at this stage perhaps after 12_3_0_pre6 has been cut?

smuzaffar:
please test

Dr15Jones:
@smuzaffar what about TBB_PREVIEW_TASK_GROUP_EXTENSIONS?

smuzaffar:
@Dr15Jones , `TBB_BUILD` is by default ON for this version (https://github.com/oneapi-src/oneTBB/blob/v2021.5.0/CMakeLists.txt#L101 ) so `task_handle` should be available with this verison

smuzaffar:
+externals

Dr15Jones:
@smuzaffar I tested using the new feature of TBB in CMSSW_12_4_0_pre2  (the function `tbb::task_group::defer` which is needed to create a `tbb::task_handle` ) and it failed (compiler says the function does not exist).

Therefore we do need TBB_PREVIEW_TASK_GROUP_EXTENSIONS to be set.

makortel:
> Therefore we do need TBB_PREVIEW_TASK_GROUP_EXTENSIONS to be set.

@smuzaffar Could we add this flag to TBB build configuraton?

smuzaffar:
@Dr15Jones , can you please try `USER_CXXFLAGS=-DTBB_PREVIEW_TASK_GROUP_EXTENSIONS=1 scram build` . If this works then we just need to update tbb.xml ( scram toolfile definition )

Dr15Jones:
It works when I do.
```
USER_CXXFLAGS=-DTBB_PREVIEW_TASK_GROUP_EXTENSIONS scram build
```
[I was able to fix silly errors in my code so everything now builds using the new TBB API]

PR #7677, opened 2022-03-09
configure pythia8 to use hepmc3 and hepmc2

mkirsano:
Configure pythia8 to be able to use both conversions: to HepMC2 and HepMC3

### DIFF
--- a/pythia8.spec
+++ b/pythia8.spec
@@ -2,12 +2,12 @@
 
 Source: https://pythia.org/download/pythia83/%{n}%{realversion}.tgz
 
-Requires: hepmc lhapdf
+Requires: hepmc hepmc3 lhapdf
 
 %prep
 %setup -q -n %{n}%{realversion}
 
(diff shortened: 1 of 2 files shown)


mkirsano:
please test

smuzaffar:
please test

smuzaffar:
assign generators
any objections on this change?

Saptaparna:
looks fine, +1 from me.

smuzaffar:
@mkirsano , I would suggest to lets include it in DEVEL IBs to get full tests and if every thing goes fine in DEVEL IBs then I will back port it for master, any objections?

mkirsano:
If this is a normal (and hopefully quick) procedure developed for such cases, no objections

smuzaffar:
as this is bit tricky ( mean pythia8 build with two different versions of hepmc) so I would like to test all relvals and unit tests before including it for next pre-release. I am merging it now and hopefully tomorrow we will have the results of the IB

PR #7692, opened 2022-03-15
Dasgoclient v02.04.45

vkuznet:
The new version brings the following:
- adds gzip encoding to HTTP requests
- adds keep-alive HTTP headers
- adds `-noKeepAlive` option to turn off keep-alive header
- adds `GODEBUG=asyncpreemptoff=1` option on Linux to prevent race condition on AFS
- change details DBS query on `file dataset=...` DAS query when user does not provide `-json` flag and when no DAS filters are used. This will significantly reduce DBS output for large datasets
- adds `dnsCache` option to avoid extensive DNS queries of cmsweb.
- adds binary build on macOS using Apple Silicon chip

### DIFF
--- a/dasgoclient.spec
+++ b/dasgoclient.spec
@@ -1,4 +1,4 @@
-### RPM cms dasgoclient v02.04.42
+### RPM cms dasgoclient v02.04.45
 ## NOCOMPILER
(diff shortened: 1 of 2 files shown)


smuzaffar:
please test

smuzaffar:
+externals

PR #7698, opened 2022-03-17
[BuildRule] Update rules to remove scram build packages generated during unit tests

smuzaffar:


smuzaffar:
please test with cms-sw/cmssw#37260

smuzaffar:
+externals
unit tests passed and scram has properly updated the recursing scram message.

PR #7704, opened 2022-03-21
Build sherpa with `std=c++17`

iarspider:
Try building Sherpa with `std=c++17` (same as ROOT, which is a Sherpa dependency). It is currently built with `std=c++0x` (which is an archaic form of `std=c++11`)


iarspider:
please test

smuzaffar:
+externals

PR #7708, opened 2022-03-22
Revert "Build sherpa with `std=c++17`"

smuzaffar:
Reverts cms-sw/cmsdist#7704
Lets test workflow 534.0 with old c++ std


smuzaffar:
please test workflow 534.0

smuzaffar:
+externals
@iarspider 534.0 workflow passed with this revert. So looks like sherpa with `std=c++17` does not work ( may be because of grid-packs)

PR #7709, opened 2022-03-23
[xrootd] drop python-python3 deps

smuzaffar:
this is not needed for xrootd version 5.4.2


smuzaffar:
please test for rhel8_amd64_gcc10

PR #7710, opened 2022-03-23
[Data] Remove Geometry XML data package which are not used any more

smuzaffar:
@iarspider noticed that we are not building these packages


smuzaffar:
please test

smuzaffar:
+externals

PR #7793, opened 2022-04-15
[G4] Fixed Geant4 compound materials construction

smuzaffar:
see https://github.com/cms-externals/geant4/pull/59


smuzaffar:
please test

PR #7797, opened 2022-04-19
[SCRAM] treat rhel and its colnes as same

smuzaffar:
SCRAM should not complain like [a], it should treat al rhel and its clones as same

[a]
```
WARNING: Developer's area is created for architecture cs9_amd64_gcc11 while your current OS is el9_amd64.
```


smuzaffar:
please test

smuzaffar:
+externals

PR #7763, opened 2022-04-07
Drop legacy MC generators

iarspider:
@cms-sw/generators-l2 We would like to drop the following MC generators:

* cascade
* charybdis
* herwig 6.520
* jimmy
* photos 215.5
* ~~pythia6~~
* ~~hydjet 1.9.1~~
* tauola 27.121.5
* toprex

These are ancient and not developed anymore

PR #7806, opened 2022-04-20
Remove herwig 6.520, jimmy

iarspider:
Split from https://github.com/cms-sw/cmsdist/pull/7763


qliphy:
merge
see https://github.com/cms-sw/cmssw/pull/37612

PR #7812, opened 2022-04-21
Force rebuild toolfile is scram-tools.file is update

smuzaffar:
`cmsBuild` does not rebuild `<tool>-toolfile` even if one has explicitly update a toolfile under `scram-tools.file`. This change forces every toolfile package to depend on the contents of `scram-tools.file` which allows `cmsBuild` to rebuild the toolfile


smuzaffar:
please test

PR #7815, opened 2022-04-22
updated ROCm tools

smuzaffar:
- No need to define `LIBDIR` for `rocm-rocrand.xml` as it depends on `rocm`
- enable symlink creation for `rocm` libs. Previously this was disabled to avoid getting `rocm/llvm` libs which conflicits with llvm libs from cmssw externals
- Fix `ROOT_INCLUDE_PATH` to have all `INCLUDE` directories. If a tool have multiple include paths then `join="1"` attribute is needed

FYI @fwyzard

### DIFF
--- a/scram-tools.file/tools/rocm/rocm-rocrand.xml
+++ b/scram-tools.file/tools/rocm/rocm-rocrand.xml
@@ -5,7 +5,6 @@
   <lib name="rocrand"/>
   <client>
     <environment name="ROCM_ROCRAND_BASE" default="@TOOL_ROOT@"/>
-    <environment name="LIBDIR"  default="$ROCM_ROCRAND_BASE/lib"/>
     <environment name="INCLUDE" default="$ROCM_ROCRAND_BASE/include/hiprand"/>
     <environment name="INCLUDE" default="$ROCM_ROCRAND_BASE/include/rocrand"/>
   </client>
--- a/scram-tools.file/tools/rocm/rocm.xml
+++ b/scram-tools.file/tools/rocm/rocm.xml
@@ -13,7 +13,6 @@
     <environment name="INCLUDE"   default="$ROCM_BASE/hip/include"/>
     <environment name="INCLUDE"   default="$ROCM_BASE/hsa/include"/>
   </client>
-  <flags SKIP_TOOL_SYMLINKS="1"/>
   <flags ROCM_FLAGS="-fno-gpu-rdc --amdgpu-target=gfx900 --gcc-toolchain=$COMPILER_PATH -D__HIP_PLATFORM_HCC__ -D__HIP_PLATFORM_AMD__"/>
   <!-- REM_CXXFLAGS from llvm/llvm-cxxcompiler.xml -->
   <flags ROCM_HOST_REM_CXXFLAGS="-Wno-non-template-friend"/>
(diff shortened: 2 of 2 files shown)


smuzaffar:
please test with https://github.com/cms-sw/cms-bot/pull/1751

fwyzard:
> * If a tool have multiple include paths then `join="1"` attribute is needed

Is this a new feature ?
At least, I didn't know about it.

smuzaffar:
yes, this is recently add. I was also not aware of this issue :-) for the following like, scram uses the last value of `$INCLUDE` and set it to `ROOT_INCLUDE_PATH`
```
```
I added `join` attribute to add all the include paths of the tool

fwyzard:
what about the case where `ROOT_INCLUDE_PATH` is given more than once, like [scram-tools.file/tools/vecgeom/vecgeom_interface.xml](https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_12_4_X/master/scram-tools.file/tools/vecgeom/vecgeom_interface.xml) ?

```xml

<tool name="vecgeom_interface" version="@TOOL_VERSION@">
  <info url="https://gitlab.cern.ch/VecGeom/VecGeom"/>
  <client>
    <environment name="VECGEOM_INTERFACE_BASE" default="@TOOL_ROOT@"/>
    <environment name="INCLUDE" default="$VECGEOM_INTERFACE_BASE/include"/>
    <environment name="INCLUDE" default="$VECGEOM_INTERFACE_BASE/include/VecGeom"/>
  </client>
  <runtime name="ROOT_INCLUDE_PATH" value="$VECGEOM_INTERFACE_BASE/include" type="path"/>
  <runtime name="ROOT_INCLUDE_PATH" value="$VECGEOM_INTERFACE_BASE/include/VecGeom" type="path"/>
  <use name="root_cxxdefaults"/>
</tool>
```

fwyzard:
> I added `join` attribute to add all the include paths of the tool

By the way, why did you decide to add a new attribute, instead of directly implementing the new behaviour ?
Is it to preserve the old behaviour in the other files ?

From a quick `grep`, it looks like th eonly other tool file with multiple `INCLUDE` lines feeding into a `ROOT_INCLUDE_PATH` is [scram-tools.file/tools/xrootd/xrootd.xml](https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_12_4_X/master/scram-tools.file/tools/xrootd/xrootd.xml) .

smuzaffar:
> > I added `join` attribute to add all the include paths of the tool
> 
> By the way, why did you decide to add a new attribute, instead of directly implementing the new behaviour ? Is it to preserve the old behaviour in the other files ?
> 
> From a quick `grep`, it looks like th eonly other tool file with multiple `INCLUDE` lines feeding into a `ROOT_INCLUDE_PATH` is [scram-tools.file/tools/xrootd/xrootd.xml](https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_12_4_X/master/scram-tools.file/tools/xrootd/xrootd.xml) .

I added the new attribute so that it does not change any existing tool definition. Now one can use the new `join` attribute 

```
<runtime name="ROOT_INCLUDE_PATH" value="$INCLUDE" type="path" join="1"/>
```
OR explicit multiple entries ( just like the one in vecgeom_interface.xml)
```
  <runtime name="ROOT_INCLUDE_PATH" value="$VECGEOM_INTERFACE_BASE/include" type="path"/>
  <runtime name="ROOT_INCLUDE_PATH" value="$VECGEOM_INTERFACE_BASE/include/VecGeom" type="path"/>
```

both are same.

PR #7735, opened 2022-03-31
Update vecgeom to 1.1.20

iarspider:
Closes #7734

PR #7821, opened 2022-04-26
Attempt to fix compilation warnings and optimize usage of VecGeom

civanch:
Currently in any local compilation there is a warning related to VecGeom and Geant4:

warning: "G4GEOM_USE_USOLIDS" redefined
   34 | #define G4GEOM_USE_USOLIDS
<command-line>: note: this is the location of the previous definition

There is no run time problems but this warning means some inconsistency of compilation options.

This PR is an attempt to remove it and also to remove G4CutTubs problem within the same version of VecGeom 1.1.18


civanch:
@smuzaffar , if this PR will be OK it may substitute #7735

smuzaffar:
test parameters:
- full_cmssw = true

smuzaffar:
please test

civanch:
@smuzaffar , I see that SimG4CMS/Calo/plugins/BuildFile.xml does not have explicit
geant4core. It may get it indirectly.... but build fails in this plugin. Should it be added?

smuzaffar:
@civanch , I see that https://github.com/cms-sw/cmssw/blob/master/SimG4CMS/Calo/plugins/CaloSteppingAction.cc#L266-L267 we try to access `G4Trap::GetZHalfLength()` but in `libG4geometry.so` there is no G4Trap but there is `G4UTrap::GetZHalfLength() const`. This is the reason we have these link issues.

civanch:
@smuzaffar , I step back from geant4.spec change.

smuzaffar:
please test

smuzaffar:
+externals
this is for special G4VECGEOM IBs

PR #7863, opened 2022-05-12
[ROOT 6.24] Attempt at Geom optimization

smuzaffar:


smuzaffar:
this has been tested via https://github.com/cms-sw/root/pull/168 , so lets get this in for DEVEL IB

PR #7865, opened 2022-05-13
[BuildRules] Do not create python/package symlink to src/package/python

smuzaffar:
There was a long pending request to not generate __init__.py under src tree (https://hypernews.cern.ch/HyperNews/CMS/get/sw-develtools/1894.html ). New build rules implement this and now it not only generate any __init__.py file but also still allows developers to create/modify and use src/Package/python/*.py files without re-running `scram build`. New build rules has been tested in special `PY_X` IBs

### DIFF
--- a/cmssw-patch-build.file
+++ b/cmssw-patch-build.file
@@ -34,7 +34,7 @@
 %endif
 
 %define PatchReleaseLink \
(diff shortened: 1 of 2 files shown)


smuzaffar:
please test

smuzaffar:
enable profiling

PR #7866, opened 2022-05-13
[SCRAM] Allow conditional dependency

smuzaffar:
SCRAM needs to support conditional dependencies e.g. add dependency on one set of cmssw packages when alpaka backend is cuda. Now scram allows to add `<use name="Foo/Bar" when="alpaka-cuda"/>`


smuzaffar:
please test

PR #7871, opened 2022-05-16
Define TBB_PREVIEW_TASK_GROUP_EXTENSIONS

makortel:
In order to use tbb::task_group::defer and tbb::task_handle.

Following https://github.com/cms-sw/cmsdist/pull/7674#issuecomment-1096121522.


makortel:
@smuzaffar I just took a blind try based on your comment in https://github.com/cms-sw/cmsdist/pull/7674#issuecomment-1096121522. Is this the right way to define that macro?

@Dr15Jones FYI

Dr15Jones:
This 
https://oneapi-src.github.io/oneTBB/main/reference/task_group_extensions.html

says we should set that macro to 1.

smuzaffar:
thanks @makortel , yes this is right place to set the flag. as @Dr15Jones suggested, please set it to 1 ( just add `=1`)

makortel:
Thanks, done.

makortel:
@cmsbuild, please test

smuzaffar:
+externals

PR #7896, opened 2022-05-29
[cms-common] cmsrel/cmsenv changed to functions for bash

smuzaffar:
- see https://github.com/cms-sw/cms-common/pull/4 
- update default gcc version to gcc10 for slc7 and above


smuzaffar:
please test

PR #7899, opened 2022-05-30
[cmssw-env] Added `--ignore-mount dir1,dir2` option to ignore mount some path

### LINKED ISSUE #2: Mounting of /eos
These scripts automatically mount /eos. This might not be desirable in general:
* at FNAL, we discourage use of the FUSE mount because of frequently-observed instabilities: https://uscms.org/uscms_at_work/computing/LPC/usingEOSAtLPC.shtml#doNot
* this also caused a problem for some tests of https://github.com/cms-sw/cmssw/pull/32576, which uses a non-CMS Singularity image whose entrypoint performs `find /`; on lxplus with /eos mounted, this takes forever and spits out many permission denied messages

It would be nice to have some toggle to turn this off or on. Personally, I would prefer if the default were "off", but I am open to rebuttals on that point before making a PR.

smuzaffar:
this fixes https://github.com/cms-sw/cmssw-osenv/issues/2

### DIFF
--- a/cmssw-osenv.spec
+++ b/cmssw-osenv.spec
@@ -1,10 +1,10 @@
-### RPM cms cmssw-osenv 220530.0
+### RPM cms cmssw-osenv 220601.0
 ## NOCOMPILER
 ## NO_VERSION_SUFFIX
 
 # ***Do not change minor number of the above version. ***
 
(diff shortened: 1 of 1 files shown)


smuzaffar:
please test

PR #7910, opened 2022-06-04
Update to HEAD of libunwind master. Patch is no longer needed.

gartung:
The latest patches for the crash in Dwarf64 are on the master branch of libunwind.


gartung:
enable profiling

gartung:
please test

PR #7927, opened 2022-06-13
[SCRAM] Allow conditional dependency for different flavors

smuzaffar:
Allow 
```
< use name="foo/bar" for="alpaka/cuda"/>
```


smuzaffar:
please test

PR #7932, opened 2022-06-14
[BuildRules] Fix findDependency whcih was broken due to python compile rules changes

smuzaffar:
Build rules in 12.5.X were updated to not create `cmssw/python/Package` symlink point to `cmssw/src/Package/Python`. This broke the https://github.com/cms-sw/cmssw-config/blob/scramv3/SCRAM/findDependencies.py for python dependency checking. New build rules tag `V07-01-06` should fix this issue


smuzaffar:
please test

PR #7933, opened 2022-06-15
Fix genreflex cppflags

smuzaffar:
Currentl BuildRules do not allow to include cppflags from tools for building root dictionaries. This change makes `GENREFLEX_CPPFLAGS` an overridable flag so that tools can set 
```
<flags GENREFLEX_CPPFLAGS="-Dflags"/>
```


smuzaffar:
please test
FYI @fwyzard , this change should allow tools to set `GENREFLEX_CPPFLAGS` which can be then passed to `genreflex`

PR #7947, opened 2022-06-20
[security] Update notebook for Token bruteforcing

smuzaffar:
Fixes https://github.com/cms-sw/cmsdist/security/dependabot/129


smuzaffar:
please test

smuzaffar:
+externals
notebok external is built and unittests ran fine

PR #7963, opened 2022-06-27
[SCRAM] Prefer user env over project env if explicilt ignored via .scramrc

### LINKED ISSUE #30: Scram messes with UTF-8 configuration
Running `scramv1 runtime -sh`, I see than one of the variables exported is `LANG=C`. However that messes up my locale configuration, which I've set to `en_GB.UTF8`. An annoying consequence is that the magic tab completion in `zsh` picks up all kinds of strange behaviours...

It seems `C.UTF-8` is fine, so the issue is really related to the encoding.

Is it really necessary to mess with the locale/encoding when configuring the environment?

smuzaffar:
CMSSW env sets some env variables e.g. `LANG=C`  which can break user tool e.g. see https://github.com/cms-sw/SCRAM/issues/30 . New SCRAMV3 now allows user to instruct scram to not set env if explicitly ignored via `~/.scramrc/runtime` file e.g.

```
ignore: LANG
ignore: LC_ALL LC_CTYPE
```
entries in `~/.scramrc/runtime` file will instruct scram to not set `LANG, LC_ALL and LC_CTYPE` env 

Fixes https://github.com/cms-sw/SCRAM/issues/30

### DIFF
--- a/SCRAMV1.spec
+++ b/SCRAMV1.spec
@@ -1,8 +1,8 @@
-### RPM lcg SCRAMV1 V3_00_45
+### RPM lcg SCRAMV1 V3_00_46
 ## NOCOMPILER
 ## NO_VERSION_SUFFIX
 
-%define tag 65fcc4ad6429620c2be617f1d81893bd2bf37525
+%define tag fbfceadf49a3c3b9a765e9bd66b330f5041b257a
 %define branch SCRAMV3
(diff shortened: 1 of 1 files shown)


smuzaffar:
please test

smuzaffar:
+externals

PR #7987, opened 2022-07-09
Disable aggressive compiler optimizations for VecGeom

hahnjo:
It seems that gcc10 miscompiles (some of) the VecGeom code at `-O3`. In my tests, switching back to `-O2` fixes the crashes. More details and discussion if that is indeed a compiler problem or an issue in the code in https://sft.its.cern.ch/jira/projects/VECGEOM/issues/VECGEOM-600

### DIFF
--- a/vecgeom.spec
+++ b/vecgeom.spec
@@ -23,6 +23,7 @@ cmake ../%{n}-%{realversion} \
   -DCMAKE_AR=$(which gcc-ar) \
   -DCMAKE_RANLIB=$(which gcc-ranlib) \
   -DCMAKE_BUILD_TYPE=Release \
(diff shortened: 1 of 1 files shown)


iarspider:
please test

hahnjo:
A quick update on some further investigations: The crashes are exposed by a change that landed for VecGeom 1.1.19, but the miscompilation can also be reproduced with earlier versions when using gcc10. So this change is probably also a candidate for application to production releases...

iarspider:
please test for CMSSW_12_5_G4VECGEOM_X/el8_amd64_gcc10

smuzaffar:
thanks @hahnjo , https://github.com/cms-sw/cmsdist/pull/7992 include this change for productions releases

PR #8062, opened 2022-09-02
[BuildRules] prefix multi-arch lib directories with scram_

smuzaffar:
see https://github.com/cms-sw/cmssw/issues/39286 issue


smuzaffar:
please test

PR #8064, opened 2022-09-04
Don't use haswell as a target

gartung:
A feature of LD with glibc 2.6+ is that is searches directories in LD_LIBRARY_PATH for subdirectories with the string haswell in their name regardless of position. This causes a subdir name haswell or even scram-haswell to be added to LD_LIBRARY_PATH.

### DIFF
--- a/vectorization/cmsdist_packages.py
+++ b/vectorization/cmsdist_packages.py
@@ -20,10 +20,10 @@ if machine() == "x86_64":
     "rivet",
   ]
   VALID_TARGETS = {
-    "nehalem":     "-march=nehalem",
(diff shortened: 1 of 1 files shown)


smuzaffar:
-externals
There was a bug in python ( i think) due to which scram was not resetting the env at built time. So if `cmsenv` has set env for `avx2 or avx512` then without https://github.com/cms-sw/SCRAM/commit/ca205b622e7376a9734df96452568d5fdd72ea40  scram is still not going to reset the env for `scram build` and release area `avxN` libs are still going to pick up `avxN libs` + `default libs`. A simple scram update is enough to avoid the issue

PR #8065, opened 2022-09-04
[BuildRules] Do not set multi-arch env if not enabled

smuzaffar:
Setting CMSSW env for multi-target first and then setting env for non-multi-target cmssw release causes scram to set wrong LD_LIBRARY_PATH. New cmssw-config tag makes sure to ignore `SCRAM_TARGET` env ( which could be set by multi-trget release) for a non-multi-target release


smuzaffar:
please test

smuzaffar:
please test

PR #8084, opened 2022-09-13
[Boost] Update boost 1.80.0

smuzaffar:


smuzaffar:
please test

smuzaffar:
@ggovi @cms-sw/db-l2 any objections on updating boost to version 1.80?

francescobrivio:
@smuzaffar is the target of the boost update CMSSW_12_6_X?

smuzaffar:
yes @francescobrivio  12.6.X

ggovi:
@smuzaffar need to run some test, will let you know asap

smuzaffar:
test parameters:
- full_cmssw = true

smuzaffar:
please test for CMSSW_12_6_CLANG_X

tvami:
assign db

tvami:
@smuzaffar is there a reason why this couldnt go into the first 13_X_Y release? I think it would be more natural to keep the 12_X_Y with the same boost. Thanks.

smuzaffar:
> @smuzaffar is there a reason why this couldnt go into the first 13_X_Y release? I think it would be more natural to keep the 12_X_Y with the same boost. Thanks.

no there is no specific reason. We can include it either in 13_X_Y or include it in 12.6.DEVEL to give you more time to test it. Do you have any specific reason to keep same boost for 12.X (we are going to update `root` for 12.X which might be a bigger jump as compare to boost update :-) )?

francescobrivio:
> > @smuzaffar is there a reason why this couldnt go into the first 13_X_Y release? I think it would be more natural to keep the 12_X_Y with the same boost. Thanks.
> 
> no there is no specific reason. We can include it either in 13_X_Y and include it in 12.6.DEVEL to give you more time to test it. Do you have any specific reason to keep same boost for 12.X (we are going to update `root` for 12.X which might be a bigger jump as compare to boost update :-) )?

`boost` updates always require special handling, mainly because of the O2Os which might disrupt (read kill) data-taking.
Now, as far as I know, 12_6_X isn't meant for data-taking, but things can change depending on the LHC schedule, so I agree with @tvami that postponing this to 13_X_X would make our (alcadb) life much easier.

fwyzard:
12.6.x aims to be out in December... I don't think there are any chances it
can be used for data taking this year.

tvami:
for AlCaDB the `boost` change is much more relevant than the `root` change. If the conditions are produced with a new `boost` that cannot be read by the older releases, so that's why I'd prefer to have a clear boundary at the 13_X_Y

ggovi:
From my first investigation it looks like the transition 1.78 -> 1.80 does not break the serialisation. I mean, it seems that 1.78 CAN read payloads written in 1.80.  However, I would prefer to make more tests with 'heavier' payloads. 
BTW, running the condTestRegression.py on singularity with arch=el8_amd64_gcc10, I get the failure:
===== Test "condTestRegression" ====
FATAL:   container creation failed: mount /etc/tnsnames.ora->/etc/tnsnames.ora error: while mounting /etc/tnsnames.ora: destination /etc/tnsnames.ora doesn't exist in container

__init__() took 0.650 ms

================================================================================

fwyzard:
Makes sense... in fact, there are no differences in boost serialisation between 1.78.0 and 1.80.0: https://github.com/boostorg/serialization/compare/boost-1.78.0...boost-1.80.0 .

smuzaffar:
@ggovi , Looks like your system is missing `/etc/tnsnames.ora` , I would suggest to copy it from lxplus. OR use cmsdev nodes

ggovi:
@smuzaffar I'm running on cmsdev23

smuzaffar:
@ggovi , how do you start singularity? For me `cmssw-el8` works
```
[muzaffar@cmsdev32 ~]$ cmssw-el8
Singularity> ls -l /etc/tnsnames.ora
-rw-r--r--. 1 root root 213978 Sep 16 06:11 /etc/tnsnames.ora
```

smuzaffar:
please test

smuzaffar:
please test for CMSSW_13_0_CLANG_X

smuzaffar:
@tvami @ggovi , CMSSW_13_0_X is open now, any objections on integrating boost 1.80 in to it?

francescobrivio:
> @tvami @ggovi , CMSSW_13_0_X is open now, any objections on integrating boost 1.80 in to it?

@smuzaffar given the update has been moved to 13_X and some preliminary checks from Giacomo are successful (https://github.com/cms-sw/cmsdist/pull/8084#issuecomment-1249047571) I think we don't have any objections.
But I do see that the tests report some unitTest failures for:
`cms/8084/CLANG/el8_amd64_gcc10/unittest — Some unit tests were failed. `
Unfortunately I can't find them in the logs...maybe they come from an older test trial?

smuzaffar:
thanks @francescobrivio for quick reply.

> Unfortunately I can't find them in the logs...maybe they come from an older test trial?

Yes those were old test, latest tests https://github.com/cms-sw/cmsdist/pull/8084#issuecomment-1328020956  and https://github.com/cms-sw/cmsdist/pull/8084#issuecomment-1328029556 (with full cmssw build) show that all unit tests passed.

francescobrivio:
+db

- differences observed with full release build (https://github.com/cms-sw/cmsdist/pull/8084#issuecomment-1328029556) are not observed in "standard tests" (https://github.com/cms-sw/cmsdist/pull/8084#issuecomment-1328020956)
- @ggovi since this will be merged in 13_0_X only this gives you a bit more time to perform more extensive tests on the new boost version if needed

smuzaffar:
+externals

PR #8087, opened 2022-09-14
Do not set COMPILER_HOST env variables

smuzaffar:
Move the `COMPILERHOST` definition in `rocm` tool ( where it is needed)


smuzaffar:
please test

scram-tools.file/tools/rocm/env.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash -e
+export COMPILER_HOST=$(gcc -dumpmachine)
+export COMPILER_PATH=${GCC_ROOT}
```
fwyzard: Does this set the path to what it is at runtime (after the installation and `cmsrel ...`) or what it was when the external itself was built ?

smuzaffar: yes, the path will be relocated at install time (just it it is done for other toolfiles)

smuzaffar: this file is only used at build time and the `@COMPILER_PATH@` in [scram-tools.file/tools/rocm/rocm.xml](https://github.com/cms-sw/cmsdist/pull/8087/files#diff-b43b9ccc937203603b2df4858063a3278b4c872b00e2c30a3c3f2e899e068c88) is replaced with the value of $COMPILER_PATH. At install time , we relocate all toolfiles

smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
+externals

PR #8090, opened 2022-09-16
fix geant4 cmake files relocation

smuzaffar:
This should fix the build errors in Geant4 IBs which are failing to build `dd4hep` as the geant4 cmake files were not properly relocated. This PR fixes the relocation for current G4 and new G4 version


smuzaffar:
please test

smuzaffar:
+externals

PR #8099, opened 2022-09-21
Build pacparser python module

### LINKED ISSUE #8098: Install pacparser python module
The pacparser.spec file here currently does `make install` which installs the C-based pacparser library.  It would be helpful if it would also do `make install-pymod` as suported by its [Makefile](https://github.com/manugarg/pacparser/blob/master/src/Makefile).  We would like it to be able to then be imported into python from the cms-sw environment with `import pacparser` so we can do things like is shown in the [pacparser python example](https://github.com/manugarg/pacparser/blob/master/examples/pactest.py).

@stlammel was asking me yesterday how to parse a proxy auto config file for a SAM test, and I think this would be the easiest way for him to do it.

If it's not too hard to teach someone how to test this, @edytuk could implement it.  I am assuming the PR would be made against the current default branch of this repository.  I see that @aandvalenzuela (hi, Andrea!) has submitted PRs here, maybe she could teach @edytuk how to do it.

smuzaffar:
Enabled pymod for pacparser.
Fixes #8098

### DIFF
--- a/pacparser-python-fix.patch
+++ b/pacparser-python-fix.patch
@@ -7,7 +7,7 @@ index 9760300..fbb436d 100644
  
  def runtests(pacfile, testdata, tests_dir):
 -  py_ver = '.'.join([str(x) for x in sys.version_info[0:2]])
-+  py_ver = ''.join([str(x) for x in sys.version_info[0:2]])
++  py_ver = "%s*" % sys.version_info[0]
    try:
      pacparser_module_path = glob.glob(os.path.join(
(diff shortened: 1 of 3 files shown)


smuzaffar:
please test

smuzaffar:
please test for el8_aarch64_gcc10

smuzaffar:
please test for el8_ppc64le_gcc10

smuzaffar:
please test

smuzaffar:
+externals

PR #8158, opened 2022-10-31
Add patch to disable builting Rivet documentation

iarspider:
Building `analyses.dat` and `analyses.dat` cause `libRivet.so` to be loaded - which could've been built with vector instructions not supported by build machine. While the failure of command used to generate `doc/analists/analyses.{dat,json}` does not cause built to fail, lack of generated files does fail the build.


iarspider:
please test for CMSSW_12_6_SKYLAKEAVX512_X

smuzaffar:
please test
lets start test for default build

smuzaffar:
please test

smuzaffar:
+externals
DQM differences are only in LogMessage

PR #8172, opened 2022-11-08
[SCRAM] map cs/alma/rocky/rhel to el

smuzaffar:


smuzaffar:
please test workflow 512.0

PR #8181, opened 2022-11-15
set jemalloc hugepagesize on aarch64 to avoid huge address space allocations

dan131riley:
The OpenLab ARM64 systems have Hugepagesize set to 0.5GB.  jemalloc uses the huge page size for its internal arrays and bitmaps, so having a very large value allocates large amounts of address space that often never get used.  For details, see:

https://github.com/cms-sw/cmssw/issues/31306

This PR sets the jemalloc huge page size to the same 16MB value used by the ppc builds.  Even that probably wastes some address space--the jemalloc default if it can't find the pagesize is 2MB.

Completely untested, it has been a long time since I built anything in cmsdist by hand.

### DIFF
--- a/jemalloc-debug.spec
+++ b/jemalloc-debug.spec
@@ -3,10 +3,10 @@
 
 %build
 XOPTS=""
-case %{cmsplatf} in
-  # set the huge page size to 16M on ARMv8
-  *_aarch64_*) XOPTS="--with-lg-hugepage=24" ;;
-esac
+%ifarch aarch64
+# set the huge page size to 16M ( log2(16M)=24 ) on ARMv8
+XOPTS="--with-lg-hugepage=24"
+%endif
 
 ./autogen.sh ${XOPTS} \
   --enable-shared \
--- a/jemalloc-debug.spec
+++ b/jemalloc-debug.spec
@@ -2,7 +2,13 @@
 ### RPM external jemalloc-debug %{jemalloc_version}
 
 %build
-./autogen.sh \
+XOPTS=""
+case %{cmsplatf} in
+  # set the huge page size to 16M on ARMv8
+  *_aarch64_*) XOPTS="--with-lg-hugepage=24" ;;
+esac
+
+./autogen.sh ${XOPTS} \
   --enable-shared \
   --disable-static \
   --disable-doc \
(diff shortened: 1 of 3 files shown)


jemalloc-debug.spec
```diff
@@ -2,7 +2,13 @@
 ### RPM external jemalloc-debug %{jemalloc_version}
 
 %build
-./autogen.sh \
+XOPTS=""
+case %{cmsplatf} in
+  # set the huge page size to 16M on ARMv8
+  *_aarch64_*) XOPTS="--with-lg-hugepage=24" ;;
```
makortel: What is the connection between 16M (in comment) and 24 (argument)?

dan131riley: In the jemalloc config, anything with "lg" is log base2, so log2(16MB) = 24.

makortel:
@cmsbuild, please test for el8_aarch64_gcc10

dan131riley:
I created a dev area for local testing, added FWCore/Services, and verified that the previously failing test succeeds with this change.

jemalloc-debug.spec
```diff
@@ -2,7 +2,13 @@
...
+esac
```
smuzaffar: please use rpm buildin macros e.g. `ifarch` instead of `cmsplatf`
```suggestion
%ifarch aarch64
# set the huge page size to 16M ( log2(16M)=24 ) on ARMv8
XOPTS="--with-lg-hugepage=24"
%endif
```

smuzaffar: and same in other 2 specs

smuzaffar:
test parameters:
- addpkg = FWCore/Services

jemalloc-debug.spec
dan131riley: The suggestion from @smuzaffar included adding that information into the comment.

jemalloc-debug.spec
dan131riley: Done, thanks.  That's much cleaner, was following an old example.

smuzaffar:
please test for el8_aarch64_gcc10

smuzaffar:
please test
start tests for production arch too

smuzaffar:
+externals

PR #8200, opened 2022-11-25
Update dasgoclient to use revXX as version suffix, fix for finding correct exec

smuzaffar:
- Use `dasgoclient_tag.revXX` as dasgoclient package directory to make sure we get the latest build of it.
- Drop extra hash from the install path
- Use `CMSDIST_FILE_REVISION` feature of cmsdist to copy the latest script back to cms_path/common

### DIFF
--- a/dasgoclient.spec
+++ b/dasgoclient.spec
@@ -1,8 +1,10 @@
-### RPM cms dasgoclient v02.04.49
+### RPM cms dasgoclient v20221125.0
 ## NOCOMPILER
-Source0: https://github.com/dmwm/dasgoclient/releases/download/%{realversion}/dasgoclient_amd64
-Source1: https://github.com/dmwm/dasgoclient/releases/download/%{realversion}/dasgoclient_aarch64
-Source2: https://github.com/dmwm/dasgoclient/releases/download/%{realversion}/dasgoclient_ppc64le
+## NO_VERSION_SUFFIX
+%define dasgoclient_tag v02.04.49
+Source0: https://github.com/dmwm/dasgoclient/releases/download/%{dasgoclient_tag}/dasgoclient_amd64
+Source1: https://github.com/dmwm/dasgoclient/releases/download/%{dasgoclient_tag}/dasgoclient_aarch64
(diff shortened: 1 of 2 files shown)


smuzaffar:
FYI @vkuznet

smuzaffar:
please test

vkuznet:
@smuzaffar , could you please elaborate an use-case why do we need such changes now? My understanding was that once new tag version of dasgoclient appears we only change it and make one build. Why do we need now a date?

smuzaffar:
@vkuznet , currently we are using dasgoclient tag as version so we get `cms/dasgoclient/<dasgoclient_tag>-<hash1>`  [a] . Now if we have to change the contents on package itself (e.g. in this case I had to update the `dasgoclient` wrapper ) then the newly build package will be `<dasgoclient_tag>-<hash2>` and there is no simple way to find which one is newer. 

[a]
```
> ls /cvmfs/cms.cern.ch/el8_amd64_gcc11/cms/dasgoclient
v02.04.48-f313cb6ef87ad3c001647c10c2b6837e 
v02.04.49-ccf3472cbe7cb8eb7e8ab0f565628681
```

smuzaffar:
if you do not like the date a version then we can add a suffix e.g. `<dasgoclient_tag>.r00` ( just like we do for crab https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_13_0_X/master/crab-dev.spec )

vkuznet:
@smuzaffar , thanks for explanation. In this case I rather prefer CRAB approach but I suggest to improve it further. Since the suffix is CMS specific part, it should be in a form, `-cms-01` or `-cmsbuild01`. Doing this way we can increment suffix, and it will be distinguished from naming convention used by tag of the package. Do you mind to switch to suffix approach instead of a date?

smuzaffar:
@vkuznet , sure I can switch to suffix approach. I can not use `-cmsXX` or anything which starts with `-c` as this will not help with sorting the old version which can also have `-[0-9a-f]`. I suggest to use `-revXX`

vkuznet:
ok, let's settle on `-revXX` suffix then.

smuzaffar:
please test

smuzaffar:
please test

PR #8203, opened 2022-11-28

iarspider:


iarspider:
please test

smuzaffar:
this looks good but I do not understand why root comparison job takes 15h for root master changes. For now it is good to go in but need to study which part of comparison is taking more time

PR #8225, opened 2022-12-10
[12.6.X] Update tag for RecoEgamma-ElectronIdentification to V01-12-00

smuzaffar:


smuzaffar:
this is needed by https://github.com/cms-sw/cmssw/pull/40250 which is already merged

PR #8233, opened 2022-12-15
Enable -std=c++20 for CMSSW build

smuzaffar:
FYI @makortel

### DIFF
--- a/cuda-flags.file
+++ b/cuda-flags.file
@@ -16,7 +16,7 @@
 %define llvm_cuda_arch %(echo $(for ARCH in %cuda_arch; do echo "$ARCH"; done) | sed -e"s/ /,/g")
 
 # C++ standard to use for building host and device code with nvcc
-%define nvcc_stdcxx -std=c++17
+%define nvcc_stdcxx -std=c++20
 
 # generate optimised code
 %define cuda_flags_0 -O3
--- a/cuda.spec
+++ b/cuda.spec
@@ -1,8 +1,8 @@
-### RPM external cuda 11.5.2
+### RPM external cuda 12.0.1
 ## INITENV +PATH LD_LIBRARY_PATH %i/lib64
 
 %define runpath_opts -m compute-sanitizer -m drivers -m nvvm
-%define driversversion 495.29.05
+%define driversversion 525.85.12
 
 %ifarch x86_64
 Source0: https://developer.download.nvidia.com/compute/cuda/%{realversion}/local_installers/%{n}_%{realversion}_%{driversversion}_linux.run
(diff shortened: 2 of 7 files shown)


smuzaffar:
test parameters:
- full_cmssw = true

smuzaffar:
please test

fwyzard:
@smuzaffar, CUDA 12.0 was released last week, with support for GCC 12 and c++20.
I'm taking the opportunity to redesign the CUDA spec files, I will make a PR for it in the coming days.

makortel:
I also just came across with https://github.com/root-project/root/pull/11311#issuecomment-1252819197 so I'm less optimistic now that "partially switching" to C++20 would work...

fwyzard:
Can we build ROOT in c++20 mode ?

smuzaffar:
we can try

smuzaffar:
please test

makortel:
> Can we build ROOT in c++20 mode ?

As far as I can tell, no. The LLVM 13 was integrated in ROOT master ~last week (earlier use LLVM 9), and here is their open PR to support for building with C++20 https://github.com/root-project/root/pull/11874 .

smuzaffar:
please test for CMSSW_13_2_ROOT628_X

fwyzard:
@smuzaffar are we already using CUDA 12 for this ?

I just reproduced and submitted another bug report...
  - https://forums.developer.nvidia.com/t/compiling-a-catch2-application-with-nvcc-std-c-20-leads-a-crash-in-cudafe/255088
  - https://developer.nvidia.com/nvidia_bug/4139863 (protected)
  - https://github.com/fwyzard/nvidia_bug_4139863

smuzaffar:
no @fwyzard , this is still based on cuda 11.5

fwyzard:
ah... so we would use c++20 with gcc and g++17 with nvcc ?

smuzaffar:
> ah... so we would use c++20 with gcc and g++17 with nvcc ?

ah right, we also have https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_13_2_X/master/cuda-flags.file#L19 
```
%define nvcc_stdcxx -std=c++17
```

I thought we were getting it form gcc xml. let me try to move the c++std flag to its own dedicated flag file.

fwyzard:
Mhm, OK, but keep in mind that CUDA 11.5 (up to CUDA 11,8, included) do not support C++20.
The support was added only with CUDA 12.

smuzaffar:
ok, so may be I need to setup a new IB for testing c++20 with
- gcc 12 or gcc 11
- cuda 12
- root 6.28

smuzaffar:
please test

smuzaffar:
please test

makortel:
> ok, so may be I need to setup a new IB for testing c++20 with
> 
> * gcc 12 or gcc 11
> * cuda 12
> * root 6.28

Yeah, given the complexity a new IB sounds like a good choice (sorry).

makortel:
```
/data/cmsbld/jenkins/workspace/ib-run-pr-tests/CMSSW_13_2_X_2023-05-31-2300/src/FWCore/MessageLogger/interface/ErrorObj.icc:61:27: error: 'fmt' is not a constant expression
   61 |     auto str = fmt::format(fmt, args...);
      |                ~~~~~~~~~~~^~~~~~~~~~~~~~
```
cppreference has for [`std::format()`](https://en.cppreference.com/w/cpp/utility/format/format)
> As of [P2216R3](https://wg21.link/P2216R3), it is an error if the format string is not a constant expression. [std::vformat](https://en.cppreference.com/w/cpp/utility/format/vformat) can be used in this case.

So the `ErrorObj::format()` etc need to be fixed (possibly adding `ErrorObj::vformat()` for the case of runtime-known formatting strings)

makortel:
https://github.com/cms-sw/cmssw/pull/41888 makes an attempt to fix the MessageLogger's interface. At least the MessageService unit tests compile in the C++20-enabled build of https://github.com/cms-sw/cmsdist/pull/8233#issuecomment-1572210519

smuzaffar:
We now have 13.2.CPP20 IBs setup which use
- GCC 12
- ROOT 6.28 (built with c+=20)
- CUDA 12
- TF with disabled GPU

closing this PR now

PR #8237, opened 2022-12-19
Minor updates to the TBB build configuration

fwyzard:
Build TBB in c++17 mode.
Build TBB with support for hwloc 2.5 and later.


fwyzard:
please test

aandvalenzuela:
Hi @fwyzard,
Just to confirm, do these changes also apply for `gcc12` (https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_13_0_X/g12/tbb.spec)? Thanks!

fwyzard:
Yes, they should be fine for any recent version of GCC.

PR #8246, opened 2023-01-12
use INCLUDE instead of IMPORT: to fix recursive parsing issue

smuzaffar:
cmsBuild does not recursively parse the `IMPORT` file , so use `INCLUDE` instead


smuzaffar:

PR #8249, opened 2023-01-17
Update to CUDA 12.0.0 and NVIDIA drivers 525.60.13

fwyzard:
The main changes since CUDA 11.5.x are
  - support for Lovelace (sm_87) and Hopper (sm_90) architectures (since CUDA 11.8)
  - support for GCC 12
  - support for C++20 (with GCC 10 and later)

See https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html for the full CUDA 12.0.x release notes and change log.

### DIFF
--- a/cuda.spec
+++ b/cuda.spec
@@ -1,8 +1,8 @@
-### RPM external cuda 11.5.2
+### RPM external cuda 12.0.0
 ## INITENV +PATH LD_LIBRARY_PATH %i/lib64
 
 %define runpath_opts -m compute-sanitizer -m drivers -m nvvm
-%define driversversion 495.29.05
+%define driversversion 525.60.13
 
 %ifarch x86_64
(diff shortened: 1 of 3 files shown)


fwyzard:
please test

fwyzard:
please test

fwyzard:
OK, so basically all the "cool" stuff (cuDNN, pycuda, and anything that depends on those) is in the way of updating to CUDA 12.0, supporting GCC 12, and C++20.

smuzaffar:
please test
pycuda 2022.2.2 has some fixes for cuda 12, so lets see if it works

fwyzard:
The linker errors and [support matrix](https://docs.nvidia.com/deeplearning/cudnn/support-matrix/index.html) confirm that CUDA 12 is currently not supported by cuDNN.

PR #8266, opened 2023-01-26
Set the CXXFLAGS for the host compiler when using ROCm libraries

fwyzard:


fwyzard:
please test

scram-tools.file/tools/rocm/rocm.xml
```diff
@@ -13,7 +13,8 @@
     <environment name="INCLUDE"   default="$ROCM_BASE/hip/include"/>
     <environment name="INCLUDE"   default="$ROCM_BASE/hsa/include"/>
   </client>
-  <flags ROCM_FLAGS="-fno-gpu-rdc --amdgpu-target=gfx900 --target=@COMPILER_HOST@ --gcc-toolchain=$(COMPILER_PATH) -D__HIP_PLATFORM_HCC__ -D__HIP_PLATFORM_AMD__"/>
+  <flags CXXFLAGS="-D__HIP_PLATFORM_HCC__ -D__HIP_PLATFORM_AMD__"/>
```
fwyzard: @smuzaffar is it OK to use `CXXFLAGS` for preprocessor macros, or should I use a different "flags" ?

fwyzard:
please test

fwyzard:
I think it should be fine to merge this, in order to have the updates flags in the IBs for testing the first ROCm packages.

scram-tools.file/tools/rocm/rocm.xml
smuzaffar: @fwyzard , its better to use `<flags CPPDEFINES=".."/>` like https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_13_0_X/master/scram-tools.file/tools/boost/boost_header.xml#L11-L13

fwyzard: Done - hopefully with the right syntax.
By the way, should we use `CPPDEFINES` instead of `CXXFLAGS` also in the various alpaka tools ?

fwyzard:
please test

scram-tools.file/tools/rocm/rocm.xml
smuzaffar: > By the way, should we use CPPDEFINES instead of CXXFLAGS also in the various alpaka tools ?

yes that will also help. This way scram puts all the cppdefines close to each other in the generated compilation command and then these will not be passed to the link command

smuzaffar:
+externals

PR #8271, opened 2023-01-29
[Buildrules] Treat ROCM_CXXFLAGS/LDFLAGS as valid compiler flags

smuzaffar:


smuzaffar:
please test with https://github.com/cms-sw/cmssw/pull/40619

smuzaffar:
test parameters:
- full_cmssw = true

smuzaffar:
please test

smuzaffar:
@fwyzard , any idea why we get `isRocmDeviceSupported.hip.cc.o:(.hipFatBinSegment+0x8): undefined reference to '__hip_fatbin' collect2: error: ld returned 1 exit status` ? I tried adding `--hip-link` to `hipcc` but still not success

fwyzard:
I think it's because the final link is being done with `g++` instead of `hipcc`.

smuzaffar:
> I think it's because the final link is being done with `g++` instead of `hipcc`.

@fwyzard that is true and using hipcc forlinking needs a careful thinking and I will do not in next PR . By the way, I was able to link and run hip application if I explicitly link with `rccl` library  (i.e I added `<lib name="rccl"/>` in rocm.xml). So can we still use `g++` and link `rccl`?

fwyzard:
No, that is not enough to make the program work.
```
$ g++ -O2 ... tmp/el8_amd64_gcc11/src/HeterogeneousCore/ROCmServices/bin/rocmComputeCapabilities/rocmComputeCapabilities.cpp.o tmp/el8_amd64_gcc11/src/HeterogeneousCore/ROCmServices/bin/rocmComputeCapabilities/isRocmDeviceSupported.hip.cc.o ... -lamdhip64 -lrccl -o tmp/el8_amd64_gcc11/src/HeterogeneousCore/ROCmServices/bin/rocmComputeCapabilities/rocmComputeCapabilities
```
does link, but the resulting binary does not run:
```
$ ./tmp/el8_amd64_gcc11/src/HeterogeneousCore/ROCmServices/bin/rocmComputeCapabilities/rocmComputeCapabilities
"Cannot find Symbol"
Aborted (core dumped)
```

I think because it is still missing the generation and linking of the device-side code and kernel launches.

PR #8284, opened 2023-02-02
Update Alpaka and add support for ROCm

fwyzard:
Update the version of alpaka to the HEAD of the develop branch as of 2023.02.01, corresponding to the commit alpaka-group/alpaka@a68c866cc6c .

Major changes:
  - remove the functions to pin/unpin an existing buffer; host memory buffers pinned/mapped for a particular accelerator platform can be obtained with `allocMappedBuf(...)`;
  - implement accelerator tags, to list all available accelerators;
  - drop the Boost.Fiber back-end;
  - add support for newer CUDA and ROCm back-ends;
  - add optional support for mdspan

Other changes:
  - add deduction guide for `Vec`;
  - add element-wise_min and max functions;
  - make CUDA/ROCm mangled kernel names as short as possible;
  - add math hyperbolic functions;
  - refactor `ConcurrentExecPool` and `QueueGenericThreadsNonBlocking`.
  - implement trait constants.

Add support for the HIP/ROCm alpaka back-end.

### DIFF
--- a/scram-tools.file/tools/alpaka/alpaka-rocm.xml
+++ b/scram-tools.file/tools/alpaka/alpaka-rocm.xml
@@ -1,0 +1,9 @@
+<tool name="alpaka-rocm" version="@TOOL_VERSION@">
+  <use name="alpaka"/>
+  <use name="rocm"/>
+  <!-- host comiplation should run with ALPAKA_HOST_ONLY defined -->
+  <flags CXXFLAGS="-DALPAKA_ACC_GPU_HIP_ENABLED -DALPAKA_HOST_ONLY"/>
+  <flags GENREFLEX_CPPFLAGS="-DALPAKA_ACC_GPU_HIP_ENABLED -DALPAKA_HOST_ONLY"/>
+  <!-- device comiplation should run without ALPAKA_HOST_ONLY defined -->
+  <flags ROCM_FLAGS="-DALPAKA_ACC_GPU_HIP_ENABLED -UALPAKA_HOST_ONLY"/>
+</tool>
--- a/alpaka.spec
+++ b/alpaka.spec
@@ -1,7 +1,7 @@
-### RPM external alpaka develop-20220902
+### RPM external alpaka develop-20230201
 ## NOCOMPILER
 
-%define git_commit b518e8c943a816eba06c3e12c0a7e1b58c8faedc
+%define git_commit a68c866cc6c3019748cf127b4eac61be38e7f687
 
 Source: https://github.com/alpaka-group/%{n}/archive/%{git_commit}.tar.gz
 Requires: boost


fwyzard:
please test with cms-sw/cmssw-config#91, cms-sw/cmssw#40678

scram-tools.file/tools/alpaka/alpaka-rocm.xml
```diff
@@ -0,0 +1,9 @@
+<tool name="alpaka-rocm" version="@TOOL_VERSION@">
+  <use name="alpaka"/>
+  <use name="rocm"/>
+  <!-- host comiplation should run with ALPAKA_HOST_ONLY defined -->
+  <flags CXXFLAGS="-DALPAKA_ACC_GPU_HIP_ENABLED -DALPAKA_HOST_ONLY"/>
+  <flags GENREFLEX_CPPFLAGS="-DALPAKA_ACC_GPU_HIP_ENABLED -DALPAKA_HOST_ONLY"/>
+  <!-- device comiplation should run without ALPAKA_HOST_ONLY defined -->
+  <flags ROCM_FLAGS="-DALPAKA_ACC_GPU_HIP_ENABLED -UALPAKA_HOST_ONLY"/>
```
smuzaffar, line 8: As `CXXFLAGS` already has `-DALPAKA_ACC_GPU_HIP_ENABLED` so it is not needed here

fwyzard: Yes, you're right.

However, I've copied this from the CUDA case, and I would prefer to clean them all up together in a separate PR.

smuzaffar: `-UALPAKA_HOST_ONLY` should work too but if you prefer then you can use
```
<flags REM_ROCM_HOST_CXXFLAGS="-DALPAKA_HOST_ONLY"/>
```

fwyzard: Following what you told me the other day, I was going to move the flags to `CPPDEFINES` - but I guess we do not have a `REM_CPPDEFINES` or `REM_ROCM_HOST_CPPDEFINES` functionality ?

fwyzard: So I think we can stick to `CXXFLAGS` and use `REM_CUDA_HOST_CXXFLAGS`/`REM_ROCM_HOST_CXXFLAGS` .

smuzaffar: yes that is correct, currently there is no `REM_ROCM_HOST_CPPDEFINES` , so lets stick to `CXXFLAGS`

smuzaffar:
@fwyzard , input test failure are not related to this change. Do you want to add more changes here otherwise I am happy to include it for next IB.

fwyzard:
@smuzaffar sorry, I missed the message earlier.

This PR needs to go in together with cms-sw/cmssw#40678; it would be great if both could enter in time for tonight's IB.

@perrotta @rappoccio FYI.

perrotta:
+1
- It has to go with cms-sw/cmssw#40678

perrotta:
merge

PR #8291, opened 2023-02-06
[BuildRules] Disable LTO flags for ROCM products

smuzaffar:
I think problem is mixing LTO objects build with llvm (hipcc) and gcc. For now lets disable LTO for rocm products. I will check if for rocm products ( where .hip.cc files exists) we can compile default `.cc` files with llvm instead of `gcc`

### DIFF
--- a/scram-project-build.file
+++ b/scram-project-build.file
@@ -54,7 +54,7 @@ BuildRequires: dwz
 %endif
 
 %if "%{?configtag:set}" != "set"
-%define configtag       V07-05-06
+%define configtag       V07-05-07
 %endif
 
 %if "%{?cvssrc:set}" != "set"


smuzaffar:
test parameters:
- addpkg = HeterogeneousCore

smuzaffar:
please test

smuzaffar:
FYI @fwyzard @makortel

fwyzard:
> I will check if for rocm products ( where .hip.cc files exists) we can compile default `.cc` files with llvm instead of `gcc`

Most likely we can - but I'm not sure if we should.

smuzaffar:
please test for CMSSW_13_0_LTO_X

smuzaffar:
+externals
this shoudl fix the LTO build errors for rocm binaries

PR #8304, opened 2023-02-09
Add (dummy) cmssw-pr-package for tests

iarspider:


smuzaffar:
@iarspider , This package is needed by all cmssw release cycles. So may be better to keep it in cms-bot/pr_tests area ? otherwise you have to fetch it explicitly for older release cycles

PR #8314, opened 2023-02-14
add RecoTracker-DisplacedRegionalTracking V00-01-00 data

slava77:
needed in context of testing cms-sw/cmssw#40643 

I'm guessing that the external was not propagated to the cmsdist because that part was still manual 2 years ago; but perhaps also because the underlying PR cms-sw/cmssw#32806 was not merged back then

### DIFF
--- a/data/cmsswdata.txt
+++ b/data/cmsswdata.txt
@@ -3,6 +3,7 @@
 #Once a non-default section is empty then cleanup that section and remove its cmsdist/${PACKAGE_TYPE}.file
 #If there is no customization for the packae then remove its .spec and .file
 [default]
+RecoTracker-DisplacedRegionalTracking=V00-01-00
 RecoTracker-MkFit=V00-12-00
 RecoTauTag-TrainingFiles=V00-07-00
 PhysicsTools-NanoAOD=V01-03-00


data/cmsswdata.txt
```diff
@@ -87,4 +88,4 @@ RecoBTag-SecondaryVertex=V02-00-04
 RecoBTag-SoftLepton=V01-00-01
 RecoHI-HiJetAlgos=V01-00-01
 RecoParticleFlow-PFTracking=V13-01-00
-SimTransport-HectorProducer=V01-00-01
\ No newline at end of file
+SimTransport-HectorProducer=V01-00-01
```
slava77, line 91: no idea why this diff showed up. I was editing online and modified only the first lines (above)

smuzaffar:
@slava77 , the PR https://github.com/cms-sw/cmsdist/pull/6907 was closed as the cmssw PE https://github.com/cms-sw/cmssw/pull/32806 which needed it was closed too.

data/cmsswdata.txt
smuzaffar: this happens when previously bot had edited the file and might have removed a new line or line feed char. So do not worry about it

mmusich:
is there anything preventing this to be merged? It seems it's needed in order to have meaningful tests at https://github.com/cms-sw/cmssw/pull/40643

perrotta:
+1
- As far as I can see it does not interact with any other workflow

perrotta:
please test

PR #8318, opened 2023-02-15
added RecoTracker-DisplacedRegionalTracking in cmssw deps

smuzaffar:


mmusich:
seems that technically https://github.com/cms-sw/cmssw/pull/40643 run with this addition. Should it be just merged?

smuzaffar:
@mmusich ,  this should go in along with https://github.com/cms-sw/cmssw/pull/40643 . I would suggest to wait for https://github.com/cms-sw/cmssw/pull/40643 to fully signed and then we can include it.

mmusich:
> I would suggest to wait for https://github.com/cms-sw/cmssw/pull/40643 to fully signed and then we can include it.

is there a strong reason to not merge now? 
I think it happened several times in the past that a PR depending on some external update was merged before the corresponding external was available leading to failures in IBs.
At any rate the actual input files in `cms-data` are already there.

smuzaffar:
@mmusich , this is a new data external and it is still not part of cmssw distribution. I would like to not include it in distribution unless the cmssw changes which require it converge. https://github.com/cms-sw/cmsdist/pull/6907 was opened 20months ago to include this data external but the cmssw PR https://github.com/cms-sw/cmssw/pull/32806 did not converge. So I would suggest to wait for https://github.com/cms-sw/cmssw/pull/40643 to converge first

slava77:
recall that a requirement for an external triggers the PR tests to use the head of the IB and could introduce unwanted differences (and repeated tests until by luck there are no relevant new PRs).

So, unless merging the external breaks the baseline, perhaps it's worth to consider changing the policy and merge?

mmusich:
> I would like to not include it in distribution unless the cmssw changes which require it converge. https://github.com/cms-sw/cmsdist/pull/6907 was opened 20months ago to include this data external but the cmssw PR https://github.com/cms-sw/cmssw/pull/32806 did not converge.

@smuzaffar, while I see from where you are coming from, I really think that this time the PR is set to converge on a reasonable timescale (as there is push from physics to have it).

smuzaffar:
itis merged now

PR #8319, opened 2023-02-16
CepGen 1.1.0, CMake 3.25.2

forthommel:
This PR bumps the version of CepGen to 1.1.0. Changelog in https://cepgen.hepforge.org/changelog.html.

As a collateral damage, the CMake version is bumped to 3.25.2.

### DIFF
--- a/cepgen.spec
+++ b/cepgen.spec
@@ -1,4 +1,4 @@
-### RPM external cepgen 1.0.2patch1
+### RPM external cepgen 1.1.0
 
 Source: https://github.com/cepgen/cepgen/archive/refs/tags/%{realversion}.tar.gz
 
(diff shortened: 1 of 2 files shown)


smuzaffar:
please test

forthommel:
Hi @smuzaffar,
I see that `cmake_path` was only introduced in CMake v>=3.20, while the default one used appears to be 3.18. Is there a way to bump the version? Or should I issue a patch removing the usage of this command?

smuzaffar:
@forthommel , I would suggest to bump `cmake` version to `3.25.2` in your branch ( the version we have in cmsdist is 2 years old)

forthommel:
Thanks for the quick reply, @smuzaffar! Just pushed!

smuzaffar:
please test

smuzaffar:
+externals

forthommel:
Many thanks for the very fast and fruitful interaction, @smuzaffar !

PR #8325, opened 2023-02-20
Disable gpu usage for tensorflow in gcc12

aandvalenzuela:
`13_1_X` IBs are broken for `el8_amd64_gcc12`.


aandvalenzuela:
please test for el8_amd64_gcc12

smuzaffar:
as there is no gcc12 IB, so we can not test it. I would suggest to merge it and trigger a gcc12 Ib at 11h00

PR #8326, opened 2023-02-20
[CXXMODULE] Remove root_12193 patch for rootmodule

aandvalenzuela:
`CXXMODULE` iB fails when applying this patch. After checking it, I think it might not be needed anymore (See https://github.com/cms-sw/root/blob/cms/master/94b3064bb5/interpreter/cling/include/cling/std.modulemap#L376-L380)


aandvalenzuela:
test parameters:

- full_cmssw = true

aandvalenzuela:
please test

davidlange6:
Right - that PR got merged into master. Thanks!

smuzaffar:
as externals build so this looks good to go in ( module Ibs are broken anyway)

PR #8327, opened 2023-02-21
[ROOT6] Root update master

aandvalenzuela:


aandvalenzuela:
please test

aandvalenzuela:
I think unit test errors are related to the failures in IBs, maybe we should test it again once the errors are gone.

smuzaffar:
+externals
agree that unit tests failures are not related to this and there is no need to rerun tests.

PR #8335, opened 2023-02-24
Disable LTO for CLANG IBs

aandvalenzuela:
We are disabling LTO for CLANG IBs since we cannot link against external libraries, e.g., dd4hep, built with gcc.


aandvalenzuela:
test parameters:
- full_cmssw = true
- addpkg = HeterogeneousCore/AlpakaTest, SimG4Core/DD4hepGeometry, SimG4Core/Geometry

aandvalenzuela:
please test

Issue #8343, opened 2023-02-28
Update UTM to 0.11.0


epalencia:
An updated utm library (version 0.11.0) is expected to be released soon. It will contain the following new features:

- new MUS2 object (for HTM triggers)
- new MU-INDEX cut for MU object (for monitoring seeds)
- new ZDCP, ZDCM objects (HCAL ZDC+/ZDC-)
- new ADT object and ADT-ASCORE cut (anomaly detection trigger)

Documentation for integration in emulator/tools will be provided with the release.

n.b. an updated TME will follow up after the release (~ mid March).

(Information from Bernhard Arnold)

smuzaffar:
thanks @epalencia for the information. please ping here again once the new tag is ready or feel free to open a PR with the required changes for https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_13_1_X/master/utm.spec

epalencia:
Just for completeness, since this update is needed for the data taking, a backport to 13_0 will be made as well.

epalencia:
The updated library is available: https://gitlab.cern.ch/cms-l1t-utm/utm/tree/utm_0.11.0

perrotta:
@elfontan

elfontan:
For reference, related PRs:
- Master: https://github.com/cms-sw/cmsdist/pull/8352
- Backport to CMSSW_13_0_Xhttps://github.com/cms-sw/cmsdist/pull/8354 

Relevant L1 updates relying on this update:
- Fix of the uGT L1_SingleMuX_EMTF, L1_SingleMuX_OMTF and L1_SingleMuX_BMTF algos ([CMSLITDPG-1075](https://its.cern.ch/jira/browse/CMSLITDPG-1075))
- New L1 seed for two loose HMT showers ([CMSLITDPG-1087](https://its.cern.ch/jira/browse/CMSLITDPG-1087))

smuzaffar:
UTM 0.11.2 has been integrated in 13.1.X and 13.0.X

PR #8352, opened 2023-03-03
Update UTM to version 0.11.1

epalencia:
UTM version 0.11.0 is available: https://gitlab.cern.ch/cms-l1t-utm/utm/tree/utm_0.11.0. It  contains the following new features:
- new MUS2 object (for HTM triggers)
- new MU-INDEX cut for MU object (for monitoring seeds)
- new ZDCP, ZDCM objects (HCAL ZDC+/ZDC-)
- new ADT object and ADT-ASCORE cut (anomaly detection trigger)

This PR addresses: https://github.com/cms-sw/cmsdist/issues/8343

A backport to 13_0 is needed.

### DIFF
--- a/utm.spec
+++ b/utm.spec
@@ -1,4 +1,4 @@
-### RPM external utm utm_0.10.0
+### RPM external utm utm_0.11.0
 Source: git+https://gitlab.cern.ch/cms-l1t-utm/utm.git?obj=master/%{realversion}&export=%{n}-%{realversion}&output=/%{n}-%{realversion}.tgz
 BuildRequires: gmake
 Requires: xerces-c boost
--- a/utm.spec
+++ b/utm.spec
@@ -1,4 +1,4 @@
-### RPM external utm utm_0.11.0
+### RPM external utm utm_0.11.1
 Source: git+https://gitlab.cern.ch/cms-l1t-utm/utm.git?obj=master/%{realversion}&export=%{n}-%{realversion}&output=/%{n}-%{realversion}.tgz
 BuildRequires: gmake
 Requires: xerces-c boost


smuzaffar:
please test

smuzaffar:
@epalencia , is there any cmssw PR to go with this update?

epalencia:
>

epalencia:
@smuzaffar, not yet, they will come

smuzaffar:
please test

smuzaffar:
@epalencia , new utm breaks couple of unit tests https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-2cf760/31075/unitTests/failed.html 
```
std::bad_alloc exception
----- Begin Fatal Exception 05-Mar-2023 02:34:38 CET-----------------------
An exception of category 'BadAlloc' occurred while
   [0] Processing global begin Run run: 1
   [1] Prefetching for module L1TGlobalSummary/'l1tGlobalSummary'
   [2] Calling method for EventSetup module L1TUtmTriggerMenuESProducer/'L1TriggerMenu'
Exception Message:
A std::bad_alloc exception was thrown.
The job has probably exhausted the virtual memory available to the process.
----- End Fatal Exception -------------------------------------------------
```

smuzaffar:
please test

epalencia:
@smuzaffar, is there a way to reproduce these errors locally?

smuzaffar:
yes @epalencia , either connect to a CentOS8 node or login to cmsdev20 using singularity e.g.
```
# login to cmsdev20
ssh cmsdev20
# create user work dir
mkdir /build/`whoami`
cd  /build/`whoami`
# start container
cmssw-el8
# setup build area using PR artifacts
/cvmfs/cms-ci.cern.ch/week1/cms-sw/cmsdist/8352/31079/install.sh
cd CMSSW_13_1_X_2023-03-05-2300
# setup env
cmsenv
# Run tests
runL1-O2O-scramTests.sh -k uGT,uGTrs -d uGMT,EMTF,OMTF,BMTF,CALO 400000 l1_trg_cosmics2022/v43 l1_trg_rs_cosmics2020/v23
```

aloeliger:
@smuzaffar I'm taking a look at this today. I don't get exactly those errors, however, I do get something slightly similar at the beginning:

```
Singularity> runL1-O2O-scramTests.sh -k uGT,uGTrs -d uGMT,EMTF,OMTF,BMTF,CALO 400000 l1_trg_cosmics2022/v43 l1_trg_rs_cosmics2020/v23
CMSSW_BASE = /build/aloelige/CMSSW_13_1_X_2023-03-05-2300
PWD = /build/aloelige/CMSSW_13_1_X_2023-03-05-2300
.	FrameworkJobReport.xml	cfipython  external	    iov_before	      lib   poison  static
..	biglib			config	   ibeos_cache.txt  l1config.db       logs  python  test
.SCRAM	bin			doc	   include	    lastLogForFM.txt  objs  src     used-ibeos-sort
INFO: ADDITIONAL CMS OPTIONS:   subsystemLabels=uGT,uGTrs
mkdir: cannot create directory '/build/aloelige/CMSSW_13_1_X_2023-03-05-2300/': File exists
Writing to sqlite_file:/build/aloelige/CMSSW_13_1_X_2023-03-05-2300//l1config.db instead of ORCON.
----- Begin Fatal Exception 08-Mar-2023 11:29:20 CET-----------------------
An exception of category 'ConditionDatabase' occurred while
   [0] Constructing the EventProcessor
   [1] Constructing ESSource: class=PoolDBESSource label='l1conddb'
Exception Message:
Tag "L1TriggerKeyExt_Stage2v0_hlt" has not been found in the database. from IOVProxy::load 
----- End Fatal Exception -------------------------------------------------
TSC payloads absent; writing  subsystemLabels=uGT,uGTrs now
Singletons:
  copyDBAuth        : .
                    - Authentication path for copy DB
  copyDBConnect     : sqlite_file:/build/aloelige/CMSSW_13_1_X_2023-03-05-2300//l1config.db
                    - Connection string for copy DB
  copyNonO2OPayloads: 1
                    - Copy DTTF TSC payloads from ORCON
  dropFromJob       : uGMT,EMTF,OMTF,BMTF,CALO
                    - Comma-separated list of systems for which we must not create WriterProxy
  logTransactions   : 0
                    - Record transactions in log DB
  onlineDBAuth      : /afs/cern.ch/user/signed char/aloelige/
                    - Authentication path for online DB
  onlineDBConnect   : oracle://cms_omds_adg/CMS_TRG_R
                    - Connection string for online DB
  outputDBAuth      : /afs/cern.ch/user/signed char/aloelige/
                    - Authentication path for output DB
  outputDBConnect   : sqlite_file:/build/aloelige/CMSSW_13_1_X_2023-03-05-2300//l1config.db
                    - Connection string for output DB
  overwriteKeys     : 0
                    - Overwrite existing keys
  protoDBAuth       : /afs/cern.ch/user/signed char/aloelige/
                    - Authentication path for prototypes' DB
  protoDBConnect    : oracle://cms_orcon_adg/CMS_CONDITIONS
                    - Connection string for prototypes' DB
  rsKey             : l1_trg_rs_cosmics2020/v23
                    - RS key
  subsystemLabels   : uGT,uGTrs
                    - Comma-separated list of specific payloads to be processed
  tagUpdate         : 
                    - Comma-separated list of column-separated pairs relating type to signed char new tagBase
  tscKey            : l1_trg_cosmics2022/v43
                    - TSC key
  unsafe            : 
                    - Comma-separated list of systems that we do not care about anymore
Lists:

Lifted transaction safe for: ['']
Will create only the following writers: cms.vstring(
    'L1TriggerKeyExtRcd@L1TriggerKeyExt',
    'L1TGlobalPrescalesVetosFractO2ORcd@L1TGlobalPrescalesVetosFract',
    'L1TUtmTriggerMenuO2ORcd@L1TUtmTriggerMenu'
)
----- Begin Fatal Exception 08-Mar-2023 11:29:33 CET-----------------------
An exception of category 'ConditionDatabase' occurred while
   [0] Constructing the EventProcessor
   [1] Constructing ESProducer: class=L1SubsystemKeysOnlineProdExt label='L1SubsystemKeysOnlineExt'
Exception Message:
Required Key File "/afs/cern.ch/user/a/aloelige/.cms_cond/db.key" is missing or unreadable. from DecodingKey::init 
----- End Fatal Exception -------------------------------------------------
----- Begin Fatal Exception 08-Mar-2023 11:29:33 CET-----------------------
An exception of category 'ConditionDatabase' occurred while
   [0] Constructing the EventProcessor
   [1] Constructing ESProducer: class=L1SubsystemKeysOnlineProdExt label='L1SubsystemKeysOnlineExt'
Exception Message:
Required Key File "/afs/cern.ch/user/signed char/aloelige/.cms_cond/db.key" is missing or unreadable. from DecodingKey::init 
----- End Fatal Exception -------------------------------------------------
L1-O2O-ERROR: could not write TSC payloads
L1-O2O-ERROR: could not write TSC payloads
```

Is the initial fatal error with the conddb/`Tag "L1TriggerKeyExt_Stage2v0_hlt" has not been found in the database` expected/handled?

The tag itself I find in the conddb [here](https://cms-conddb.cern.ch/cmsDbBrowser/list/Prod/tags/L1TriggerKeyExt_Stage2v0_hlt), having been inserted in 2016, so that seems a bit strange to me...

smuzaffar:
> mkdir: cannot create directory '/build/aloelige/CMSSW_13_1_X_2023-03-05-2300/': File exists

@aloeliger , you can ignore this error or avoid it by setting `export SCRAM_TEST_NAME=FooBar`  

> Required Key File "/afs/cern.ch/user/signed char/aloelige/.cms_cond/db.key" is missing or unreadable. from DecodingKey::init 

this is real error for you. You need to have ~/.cms_cond/db.key in order to connect to condition DB. I guess you need to check iwth alca/db ppl to get access to this file

aloeliger:
I dropped a line to the L1 alca/db people about getting this file. However, is the initial fatal error with the conddb/`Tag "L1TriggerKeyExt_Stage2v0_hlt" has not been found in the database` expected/handled?

Even when I just try to run the test in the `CMSSW_13_1_X_2023-03-05-2300` IB with nothing else added that error occurs.

smuzaffar:
@aloeliger , the `"L1TriggerKeyExt_Stage2v0_hlt" has not been found in the database` error we also get in the IBs but overall it runs fine ( see log https://cmssdt.cern.ch/SDT/cgi-bin/logreader/el8_amd64_gcc11/CMSSW_13_1_X_2023-03-08-1100/unitTestLogs/L1TriggerConfig/L1TConfigProducers#/1474 )

aloeliger:
@smuzaffar Thank you, then the _truly_ fatal error is the memory one? Okay, once the L1 ALCA/DB contact gets back to me about that file I'll see what I can find out about these memory errors.

panoskatsoulis:
Hi @aloeliger ,
as @smuzaffar says the error with the message `"L1TriggerKeyExt_Stage2v0_hlt" has not been found in the database` is not the real issue behind this unit test failure.

As explained also in the AlCa channel these jobs consist of 3 parts (check&produce Rcd, update IOV, check again - link[1]). this message is the result of the first check that is required to be failed so the producers to be run and tested. This is the main target of these unit tests.

Looking into the error from the failed unit tests below I'm afraid the issue is not the unit test itself, but the GT ESProducers (menu and prescales) that are called to create a potentially new L1Menu Rcd that keys are describing.
Seems like the new UTM is not compatible with the ESProducers or something, so the new Rcd fails to be created.

```diff
%MSG-double DBOutputService:   L1TGlobalPrescalesVetosOnlineProd:L1TGlobalPrescalesVetosOnlineProd@callESModule  03-Mar-2023 15:33:30 CET Run: 4294967295 Event: 1 Logger.h:60
Fetching tag info for L1TriggerKeyListExt_Stage2v0_hlt
%MSG
%MSG-int L1-O2O: L1TGlobalPrescalesVetosOnlineProd:   L1TGlobalPrescalesVetosOnlineProd:L1TGlobalPrescalesVetosOnlineProd@callESModule  03-Mar-2023 15:33:30 CET Run: 4294967295 Event: 1
Producing L1TGlobalPrescalesVetos with TSC:RS key = ugt_pp/v1:ugt_rs_cosmics2020/v4
%MSG
%MSG-int L1-O2O: L1TGlobalPrescalesVetosOnlineProd:   L1TGlobalPrescalesVetosOnlineProd:L1TGlobalPrescalesVetosOnlineProd@callESModule  03-Mar-2023 15:33:30 CET Run: 4294967295 Event: 1
+Producing L1TUtmTriggerMenu with key =ugt_l1_l1menu_collisions2020/v8
%MSG

-std::bad_alloc exception
-%MSG-double DBOutputService:   L1TUtmTriggerMenuOnlineProd:L1TUtmTriggerMenuOnlineProd@callESModule  03-Mar-2023 15:33:30 CET Run: 4294967295 Event: 1 Logger.h:60
+Fetching tag info for L1TriggerKeyListExt_Stage2v0_hlt
%MSG
%MSG-int L1-O2O: L1TUtmTriggerMenuOnlineProd:   +L1TUtmTriggerMenuOnlineProd:L1TUtmTriggerMenuOnlineProd@callESModule  03-Mar-2023 15:33:30 CET Run: 4294967295 Event: 1
+Producing L1TUtmTriggerMenu with key =ugt_l1_l1menu_collisions2020/v8
%MSG

-std::bad_alloc exception
----- Begin Fatal Exception 03-Mar-2023 15:33:31 CET-----------------------
An exception of category 'BadAlloc' occurred while
   [0] Processing  Event run: 4294967295 lumi: 1 event: 1 stream: 0
   [1] Running path 'p'
+   [2] Prefetching for module L1CondDBPayloadWriterExt/'L1CondDBPayloadWriterExt'
-   [3] Calling method for EventSetup module L1TGlobalPrescalesVetosOnlineProd/''
Exception Message:
-A std::bad_alloc exception was thrown.
-The job has probably exhausted the virtual memory available to the process.
----- End Fatal Exception -------------------------------------------------
----- Begin Fatal Exception 03-Mar-2023 15:33:31 CET-----------------------
An exception of category 'BadAlloc' occurred while
   [0] Processing  Event run: 4294967295 lumi: 1 event: 1 stream: 0
   [1] Running path 'p'
+   [2] Prefetching for module L1CondDBPayloadWriterExt/'L1CondDBPayloadWriterExt'
-   [3] Calling method for EventSetup module L1TGlobalPrescalesVetosOnlineProd/''
Exception Message:
-A std::bad_alloc exception was thrown.
The job has probably exhausted the virtual memory available to the process.
----- End Fatal Exception -------------------------------------------------
L1-O2O-ERROR: could not write TSC payloads
L1-O2O-ERROR: could not write TSC payloads

---> test L1O2O_L1TMenu had ERRORS
```

while local testing I would start from these ESProducers

[1] https://github.com/cms-sw/cmssw/blob/e7caf242a3dce3197eb979a42de55eb853c621e9/L1TriggerConfig/L1TConfigProducers/scripts/runL1-O2O-scramTests.sh#L79-L82

makortel:
Looking at the `L1TUtmTriggerMenuESProducer` (that in https://github.com/cms-sw/cmsdist/pull/8352#issuecomment-1455035201 is the one throwing the exception), it does
```cpp
L1TUtmTriggerMenuESProducer::ReturnType L1TUtmTriggerMenuESProducer::produce(const L1TUtmTriggerMenuRcd& iRecord) {
  //const L1TUtmTriggerMenu * cmenu = reinterpret_cast<const L1TUtmTriggerMenu *>(tmeventsetup::getTriggerMenu("/afs/cern.ch/user/t/tmatsush/public/tmGui/test-menu.xml"));
  const L1TUtmTriggerMenu* cmenu =
      reinterpret_cast<const L1TUtmTriggerMenu*>(tmeventsetup::getTriggerMenu(m_L1TriggerMenuFile));
  return ReturnType(cmenu);
}
```
https://github.com/cms-sw/cmssw/blob/e7caf242a3dce3197eb979a42de55eb853c621e9/L1Trigger/L1TGlobal/plugins/L1TUtmTriggerMenuESProducer.cc#L91-L96

The `reinterpret_cast` is potentially dangerous, and indeed the `tmeventsetup::getTriggerMenu()` returns pointer to `esTriggerMenuHandle` that itself inherits from `esTriggerMenu`
https://gitlab.cern.ch/cms-l1t-utm/utm/-/blob/utm_0.11.0/tmEventSetup/tmEventSetup.cc#L27-45
https://gitlab.cern.ch/cms-l1t-utm/utm/-/blob/utm_0.11.0/tmEventSetup/include/utm/tmEventSetup/esTriggerMenuHandle.hh#L20
https://gitlab.cern.ch/cms-l1t-utm/utm/-/blob/utm_0.11.0/tmEventSetup/include/utm/tmEventSetup/esTriggerMenu.hh#L23

This type cast is undefined behavior (i.e. can break at any time in any way), that has been reported in https://github.com/cms-sw/cmssw/issues/30819 (and also touched in https://github.com/cms-sw/cmssw/issues/35010).

panoskatsoulis:
yes exactly, this cast is widely used for bit-level operations and when one knows exactly what would get by using it.
@epalencia @aloeliger this should be returned as feedback to L1T-GT experts, understand exactly why is written this way, and if/why it crashes by using the new UTM

aloeliger:
@elfontan I believe you are responsible for some of the global trigger. Could you also tag in any other GT experts I may be missing?

> Looking at the `L1TUtmTriggerMenuESProducer` (that in [#8352 (comment)](https://github.com/cms-sw/cmsdist/pull/8352#issuecomment-1455035201) is the one throwing the exception), it does
> 
> ```c++
> L1TUtmTriggerMenuESProducer::ReturnType L1TUtmTriggerMenuESProducer::produce(const L1TUtmTriggerMenuRcd& iRecord) {
>   //const L1TUtmTriggerMenu * cmenu = reinterpret_cast<const L1TUtmTriggerMenu *>(tmeventsetup::getTriggerMenu("/afs/cern.ch/user/t/tmatsush/public/tmGui/test-menu.xml"));
>   const L1TUtmTriggerMenu* cmenu =
>       reinterpret_cast<const L1TUtmTriggerMenu*>(tmeventsetup::getTriggerMenu(m_L1TriggerMenuFile));
>   return ReturnType(cmenu);
> }
> ```
> 
> https://github.com/cms-sw/cmssw/blob/e7caf242a3dce3197eb979a42de55eb853c621e9/L1Trigger/L1TGlobal/plugins/L1TUtmTriggerMenuESProducer.cc#L91-L96
> 
> The `reinterpret_cast` is potentially dangerous, and indeed the `tmeventsetup::getTriggerMenu()` returns pointer to `esTriggerMenuHandle` that itself inherits from `esTriggerMenu` https://gitlab.cern.ch/cms-l1t-utm/utm/-/blob/utm_0.11.0/tmEventSetup/tmEventSetup.cc#L27-45 https://gitlab.cern.ch/cms-l1t-utm/utm/-/blob/utm_0.11.0/tmEventSetup/include/utm/tmEventSetup/esTriggerMenuHandle.hh#L20 https://gitlab.cern.ch/cms-l1t-utm/utm/-/blob/utm_0.11.0/tmEventSetup/include/utm/tmEventSetup/esTriggerMenu.hh#L23
> 
> This type cast is undefined behavior (i.e. can break at any time in any way), that has been reported in [cms-sw/cmssw#30819](https://github.com/cms-sw/cmssw/issues/30819) (and also touched in [cms-sw/cmssw#35010](https://github.com/cms-sw/cmssw/issues/35010)).

Do you know why this test is written in this way?

makortel:
> > This type cast is undefined behavior (i.e. can break at any time in any way), that has been reported in [cms-sw/cmssw#30819](https://github.com/cms-sw/cmssw/issues/30819) (and also touched in [cms-sw/cmssw#35010](https://github.com/cms-sw/cmssw/issues/35010)).
> 
> Do you know why this test is written in this way?

It seems to me, and was also pointed out it https://github.com/cms-sw/cmssw/issues/30819, the underlying problem is deeper in the interaction of the UTM and related CMSSW code than just this test.

aloeliger:
Well, looking at the [`utm/tmEventSetup`](https://gitlab.cern.ch/cms-l1t-utm/utm) files, the objects seem designed to be almost perfect mirrors of the `CondFormats/L1TObjects` objects. I could try to write some functions that would translate one into the other without the need for `reinterpret_cast` tomorrow when I wake up. Before I do that though, I haven't worked with `reinterpret_cast` before, do failures of this kind of pointer reinterpretation produce the `std::bad_alloc` seen or a more traditional style segfault?

makortel:
> the objects seem designed to be almost perfect mirrors of the `CondFormats/L1TObjects` objects

Unfortunately "almost perfect mirror" is not sufficient for bit-wise reinterpretation to work. The exact rules are complicated https://en.cppreference.com/w/cpp/language/reinterpret_cast . E.g. virtual functions (as noted in https://github.com/cms-sw/cmssw/issues/30819) make the behavior of the reinterpreted object undefined.

> Before I do that though, I haven't worked with `reinterpret_cast` before, do failures of this kind of pointer reinterpretation produce the `std::bad_alloc` seen or a more traditional style segfault? 

As I wrote earlier, under undefined behavior literally anything can happen (including "it seems to work today"). I could easily imagine an incorrect interpretation of some bits as std containers could lead to `std::bad_alloc` being thrown.

elfontan:
Good morning all, and thank you for this work and checks.
I believe that people who dealt with this kind of updates in the past and might know something more are Vladimir Rekovic and Richard Cavanaugh, while on the utm side the experts are Bernhard Arnold and Manfred Jeitler. 

P.S. 
I don't understand why I cannot tag people in cms-dist (and I know that there were issues in tagging me before), any idea? I can open a dedicated email thread with the uGT people in the meanwhile.

arnobaer:
Hi @aloeliger @makortel I completely agree and I'm aware of this `reinterpret_cast` issue for many years now and it is quite a pain to extend/update utm without breaking bit-wise compatibility with the emulator.

It would be a significant improvement to replace the `reinterpret_cast` into constructing a new L1UtmTriggerMenu object from data provided by the returned esTriggerMenu object.

elfontan:
Hi @arnobaer, thank you!
For my understanding, could you please elaborate a bit more on what you mean with _"breaking bit-wise compatibility with the emulator"_? Does it mean that in principle at the current stage we should update the emulator according to the new utm before trying to incorporate them in CMSSW?

arnobaer:
Hi @elfontan instead of reusing the allocated memory from `esTriggerMenu` by `L1UtmTriggerMenu` it would be safer to copy only the required data into a new `L1UtmTriggerMenu` object, more like...
```cpp
L1UtmTriggerMenu* getL1UtmTriggerMenu()
{
  const auto* tm = tmeventsetup::getTriggerMenu(filename);
  auto* l1tm = new L1UtmTriggerMenu();
  l1tm->setName(tm->getName());
  l1tm->setScaleSetName(tm->getScaleSetName());
  ...
  delete tm;
  return l1tm;
}
```
That would also allow for free extension of utm classes without breaking the emulator (and freeing utm from being an emulator hostage). The open question is who will implement this?

arnobaer:
I released a patch utm_0.11.1 that fixes a minor bug for logging and a minor fix that broke building python bindings on a manylinux2014 docker image, but this might be unrelated to the above issue.
https://gitlab.cern.ch/cms-l1t-utm/utm/tree/utm_0.11.1

arnobaer:
@aloeliger in case you are interested here is a boilerplate snippet traversing the required data.
https://gist.github.com/arnobaer/8e2004a76645310a98fb9af82cab4406

n.b. it appears that some setter methods are missing in some L1Utm* classes, but it should be easy to add them.

aloeliger:
> @aloeliger in case you are interested here is a boilerplate snippet traversing the required data. https://gist.github.com/arnobaer/8e2004a76645310a98fb9af82cab4406
> 
> n.b. it appears that some setter methods are missing in some L1Utm* classes, but it should be easy to add them.

@arnobaer Thanks. I've been working on similar stuff today, but have been distracted by numerous things. I will try to have a set of one-to-the-other style functions available soon, and I as soon as I can get a key to recreate the error, I will see if this is the root cause. From what I understand though, it may be good to get rid of the cast anyways even if it isn't.

aloeliger:
@arnobaer In the utm firmware, it doesn't seem like in the `TriggerMenu` that the [`external_map_`](https://gitlab.cern.ch/arnold/utm/-/blob/master/tmEventSetup/esTriggerMenu.hh#L183) is retrievable in any way, from the base class, and is not used anywhere outside of a `TriggerMenuHandle`. Is that the case? Is this class member vital to function? 

Similarly for the [`token_to_condition_`](https://gitlab.cern.ch/arnold/utm/-/blob/master/tmEventSetup/esTriggerMenu.hh#L184). Is this necessary?

Can these safely be neglected in the translation to an `L1Utm` object in CMSSW? The CMSSW versions of these objects have no way of retrieving the equivalent variable either, and are not set anywhere (outside of the reinterpret cast from existing es objects I would guess).

arnobaer:
@aloeliger yes, I assume `token_to_condition_` is not relevant for CMSSW. `external_map_` contains a map of available external signals connected to uGT, but also seems not relevant here. I didn't dare to add a getter method as I was a bit afraid to break something in CMSSW when updating the headers affected by the `reinterpret_cast`.

aloeliger:
All that being said, I think I have a(n admittedly quite verbose) translation between the `tm/es` objects and `L1TUtm` objects.

It would change this:
```c++
L1TUtmTriggerMenuESProducer::ReturnType L1TUtmTriggerMenuESProducer::produce(const L1TUtmTriggerMenuRcd& iRecord) {
  //const L1TUtmTriggerMenu * cmenu = reinterpret_cast<const L1TUtmTriggerMenu *>(tmeventsetup::getTriggerMenu("/afs/cern.ch/user/t/tmatsush/public/tmGui/test-menu.xml"));
  const L1TUtmTriggerMenu* cmenu =
      reinterpret_cast<const L1TUtmTriggerMenu*>(tmeventsetup::getTriggerMenu(m_L1TriggerMenuFile));
  return ReturnType(cmenu);
}
```

Into something a little more like this:
```c++
L1TUtmTriggerMenuESProducer::ReturnType L1TUtmTriggerMenuESProducer::produce(const L1TUtmTriggerMenuRcd& iRecord) {
  const tmeventsetup::esTriggerMenu* theEsMenu = tmeventsetup::getTriggerMenu(m_L1TriggerMenuFile);
  auto l1Menu = UtmTmesHelpers::makeEquivalentL1TUtmTriggerMenu(*theEsMenu);
  delete theEsMenu;
  return make_unique<const L1TUtmTriggerMenu>(l1Menu);
}
```

By way of a function (in hindsight, I probably should have just written constructors for the `L1TUtm` objects that used this code instead) that looks something to the tune of this:
```c++
    L1TUtmTriggerMenu makeEquivalentL1TUtmTriggerMenu(esTriggerMenu esMenu)
    {
        //This is trickier
        //We have mapes of algorithms, conditions, and scales that we need to convert
        //Start with the algorithm mapping
        std::map<std::string, L1TUtmAlgorithm> l1AlgMap;
        auto esAlgMap = esMenu.getAlgorithmMap();
        for(auto it = esAlgMap.begin(); it != esAlgMap.end(); ++it)
            l1AlgMap[it->first] = makeEquivalentL1TUtmAlgorithm(it->second);

        std::map<std::string, L1TUtmCondition> l1CondMap;
        auto esCondMap = esMenu.getConditionMap();
        for(auto it = esCondMap.begin(); it != esCondMap.end(); ++it)
            l1CondMap[it->first] = makeEquivalentL1TUtmCondition(it->second);
        
        std::map<std::string, L1TUtmScale> l1ScaleMap;
        auto esScaleMap = esMenu.getScaleMap();
        for(auto it = esScaleMap.begin(); it != esScaleMap.end(); ++it)
            l1ScaleMap[it->first] = makeEquivalentL1TUtmScale(it->second);
        
        L1TUtmTriggerMenu l1Menu = L1TUtmTriggerMenu(
            l1AlgMap,
            l1CondMap,
            l1ScaleMap,
            //Needed? Function to get external map
            //Needed? Function to get token condition
            esMenu.getName(),
            esMenu.getVersion(),
            esMenu.getComment(),
            esMenu.getDatetime(),
            esMenu.getFirmwareUuid(),
            esMenu.getScaleSetName(),
            esMenu.getNmodules(),
            0// version is hard set to 0.
        );
        return l1Menu;

    }
```

This introduces a header into `CondFormats/L1TriggerObjects` that I am calling `utmTotmEventSetupHelpers.h`, that contains these functions in a namespace `UtmTmesHelpers`. A number of the `L1TUtm` objects have to have had constructors added that set their individual values (again, probably should have just written `L1TUtm` constructors  for the `es` objects, but I was tired and not thinking straight). 

I have also only written the functions for `es` -> `L1TUtm` translation, which is a bit easier given the control I have over `L1TUtm` objects. I have less ability to manipulate what is settable for an `es` object. @arnobaer The handle derived classes should largely have setter functions for all the base clase object members that do not otherwise have setters correct? I haven't checked them all, but I thought that was generally happening. I know there is code in CMSSW that actually uses `es` objects in preference over the `L1TUtm` objects, so having the ability to at least create a handle for these objects with all members settable may be necessary to remove all reinterpreting casts

I have tested that my version of all this compiles, and will start going through whatever tests I can. 

@makortel I'll defer to your expertise in this: Is the `reinterpret_cast` egregious enough that I should open a PR for this (or something like this that is less of a draft) regardless of whether or not it would fix the unit test causing the UTM version upgrade issue?

arnobaer:
@aloeliger the handle classes are used as (internal) _builders_ to create the actual objects from tmTable maps and scales. I would propose to extend the L1Utm classes with overloaded constructors, either taking all the member data or taking an es objects reference and populate the members inside the constructors (making use of initializer list for most of the values), like:
```cpp
class L1TUtmCutValue {
  ...
  L1TUtmCutValue(const tmeventsetup::esCutValue cutValue)
  : value(cutValue.getValue()),
    index(cutValue.getIndex()),
    version(0)
  {}
  ...
};

class L1TUtmCut {
  ...
  L1TUtmCut(const tmeventsetup::esCut& cut)
  : name_(cut.getName()),
    object_type_(cut.getObjectType()),
    cut_type_(cut.getCutType()),
    minimum_(L1UtmCutValue(cut.getMinimum())),
    maximum_(L1UtmCutValue(cut.getMaximum())),
    data_(cut.getData()),
    key_(cut.getKey()),
    version(0)
  {}
  ...
};

class L1TUtmObject {
  ...
  L1TUtmObject(const tmeventsetup::esObject& object)
  : name_(object.getName()),
    type_(object.getType()),
    comparison_operator_(object.getComparisonOperator()),
    bx_offset_(object.getBxOffset()),
    threshold_(object.getThreshold()),
    ext_signal_name_(object.getExternalSignalName()),
    ext_channel_id_(object.getExternalChannelId()),
    version(0)
  {
    cuts_.reserve(object.getCuts().size());
    for (const auto& cut: object.getCuts()) {
      cuts_.emplace_back(L1UtmCut(cut));
    }
  }
  ...
};

class L1TUtmCondition {
  ...
  L1TUtmCondition(const tmeventsetup::esCondition& condition)
  : name_(condition.getName()),
    type_(condition.getType()),
    version(0)
  {
    objects_.reserve(condition.getObjects().size());
    for (const auto& object: condition.getObjects()) {
      objects_.emplace_back(L1UtmObject(object));
    }
    cuts_.reserve(condition.getCuts().size());
    for (const auto& cut: condition.getCuts()) {
      cuts_.emplace_back(L1UtmCut(cut));
    }
  }
  ...
};
```

aloeliger:
@arnobaer Okay. However, This code in the global trigger:

https://github.com/cms-sw/cmssw/blob/166eab1458287d877b7b06a931153f21b74e2093/L1Trigger/L1TGlobal/plugins/TriggerMenuParser.cc#L175

Contains a reinterpret from an `L1TUtm` object back to an `tm/es` one so I need a way to get from `L1TUtm` to `tm/es`. 

I don't have control over the constructors of `tm/es` objects, and performing either my quick draft's method, or the constructor method is out of my reach. From what I can see in the available code for basic objects (taking algorithm as an example),  these objects are designed with a blank/default constructor, and without any available method for setting any of it's members: https://gitlab.cern.ch/arnold/utm/-/blob/master/tmEventSetup/esAlgorithm.hh. How can I access the members for the creation of equivalent `es` objects for creating `es` objects from `L1TUtm` ones. For that matter, I didn't think my c++ was that out of date and rusty, but these members are protected, if they aren't set by derived classes, how are these members ever set and used?

aloeliger:
Does that code need to be done in `es` objects?

aloeliger:
Unless I'm mistaken, It seems like the only reason this code was written in this way, was to have access to the enums in https://gitlab.cern.ch/arnold/utm/-/blob/master/tmEventSetup/esTypes.hh ? And even then, it should work with just the L1Utm object, shouldn't it? Are these casts at all necessary? If necessary, could the enums simply be ported to CMSSW?

arnobaer:
@aloeliger I don't think so. I see no reason in casting back to es* types as all the L1TUtm* types provide the necessary getter methods anyway. A quick fix for `L1TriggerMenuParser` would be removing the `reinterpret_cast`, using `auto` instead explicit type declarations for variables and replace the method argument types to L1TUtm* types.

I see no problem in comparing L1TUtm* types with enums form `esTypes.hh`. Althougth it would be nice to decouple the enums form utm and CMSSW using a mapping, but that might be something for future refactoring.

aloeliger:
Okay, Let me try to do this all in constructors and get rid of all the reinterprets and see if that compiles and will run tests. 

@makortel I'll make sure this doesn't get lost in the thread: Is the `reinterpret_cast` egregious enough that I should open a PR for this (or something like this that is less of a draft) regardless of whether or not it would fix the unit test causing the UTM version upgrade issue?

makortel:
@aloeliger 

> I'll make sure this doesn't get lost in the thread: Is the `reinterpret_cast` egregious enough that I should open a PR for this (or something like this that is less of a draft) regardless of whether or not it would fix the unit test causing the UTM version upgrade issue?

The `reinterpret_cast` should indeed be replaced with a translation logic in any case. 

Just to note, the ESProducer `produce()` could be made more concise along
```cpp
L1TUtmTriggerMenuESProducer::ReturnType L1TUtmTriggerMenuESProducer::produce(const L1TUtmTriggerMenuRcd& iRecord) {
  std::unique_ptr<const tmeventsetup::esTriggerMenu> theEsMenu{tmeventsetup::getTriggerMenu(m_L1TriggerMenuFile)};
  return make_unique<const L1TUtmTriggerMenu>(UtmTmesHelpers::makeEquivalentL1TUtmTriggerMenu(*theEsMenu));
}
```
or
```cpp
  using ReturnType = std::optional<L1TUtmTriggerMenu>;
  ...
L1TUtmTriggerMenuESProducer::ReturnType L1TUtmTriggerMenuESProducer::produce(const L1TUtmTriggerMenuRcd& iRecord) {
  std::unique_ptr<const tmeventsetup::esTriggerMenu> theEsMenu{tmeventsetup::getTriggerMenu(m_L1TriggerMenuFile)};
  return UtmTmesHelpers::makeEquivalentL1TUtmTriggerMenu(*theEsMenu);
}
```

aloeliger:
@smuzaffar I have opened a PR to CMSSW that should fix the `reinterpret_cast` issue found by @makortel: https://github.com/cms-sw/cmssw/pull/41036. If that PR passes testing, is it possible to test this PR using those changes? I seem to recall that being one of buildbot's neat features.

smuzaffar:
please test with cms-sw/cmssw#41036

aloeliger:
Sorry, there seem to be some confusing build failures that are not present in my set-up. I need some time to debug this PR further.

aloeliger:
Well, that's disappointing. It's failing the same tests. I have a key from ALCA/DB now, so hoepfully I should be able to recreate these in the new setup.  I'll investigate a little and see what I can figure out.

aloeliger:
So, just at a glance, we have two failures, `L1TUtmTriggerMenuESProducer` and `L1TGlobalPrescalesVetosOnlineProd`, even after the PR for `reinterpret_cast` removed it from `L1TUtmTriggerMenuESProducer` (worth nothing is that `L1TGlobalPrescalesVetosOnlineProd` _does_ contain a `reinterpret_cast`, but that can't be the primary issue). 

The only remaining common change for these producer files between `utm` v0.10.x and v0.11.x is the firmware code itself. Looking at the files, they both contain only one call to firmware functions and objects, and it is common between themselves, and that is a call to `tmeventsetup::getTriggerMenu()` which is my remaining suspect on this error.

Looking at the utm code, this seems to have undergone quite a bit of change between v0.10.x and v0.11.x

in v0.11.x the version of the function we call is this: https://gitlab.cern.ch/cms-l1t-utm/utm/-/blob/master/tmEventSetup/tmEventSetup.cc#L27-45

and it's v0.10.x counterpart is this: https://gitlab.cern.ch/cms-l1t-utm/utm/-/blob/utm_0.10.1/tmEventSetup/tmEventSetup.cc#L94-113

These are relatively similar, both versions have further calls to `tmtable::xml2menu` , but this function seems identical between versions and is likely not a culprit.

The difference between these two then is that v0.11.x directly returns a new `esTriggermenuHandle` (a derived class of `esTriggerMenu`) where v0.10.x returns a call to `_getTriggerMenu()` (both with similar arguments of menu, scale, and extSignal)

the bodies of `_getTriggerMenu` and the constructor of `esTriggerMenuHandle` are quite similar looking, largely going through and concerning themeselves with the setting of `esTriggerMenuHandle` values. `_getTriggerMenu` does this via calls to `getValue` which is designed to return a string pulled out of a map provided to it https://gitlab.cern.ch/cms-l1t-utm/utm/-/blob/utm_0.10.1/tmEventSetup/tmEventSetup.cc#L138-152. The v0.11.x version of this calls on `tmtable::get()` which is, by the looks of it, some code over tmtable "rows" which are std::maps of strings, and is defined here: https://gitlab.cern.ch/cms-l1t-utm/utm/-/blob/master/tmTable/tmTable.cc#L39-47. This code also looks basically the same taking into account that a row is map of strings.

After doing some reading, it seems `std::bad_alloc` is the choice of any std datatype with memory that can be allocated, throwing this error whenever memory cannot be allocated (typically because it runs out of memory to do so). One possibility is that somewhere one of these setters may be failing to allocate something when members are set or is trying to do so infinitely or something. I also see this difference: https://gitlab.cern.ch/cms-l1t-utm/utm/-/blob/master/tmEventSetup/esTriggerMenuHandle.cc#L288 between v 0.11.x and v0.10.x, which, without much understanding, just looking at it, doesn't seem like it would be causing bad allocations, but I figure I should ask.

@arnobaer Have there been any major changes to how datatypes with memory allocation are handled in the new utm library? Especially anything connected to `esTriggerMenu`? What is this `if (Algorithm::isGate(token)) continue;` line that is different between v0.11.x and v0.10.x?

Thank you for your help with this.

aloeliger:
```c++
std::cout<<"getting trigger menu!"<<std::endl;
  std::shared_ptr<L1TUtmTriggerMenu> pMenu(
      const_cast<L1TUtmTriggerMenu *>(reinterpret_cast<const L1TUtmTriggerMenu *>(tmeventsetup::getTriggerMenu(iss))));
  std::cout<<"got trigger menu!"<<std::endl;
  ```
  
  ->
  
  ```
 getting trigger menu!

std::bad_alloc exception
%MSG-double DBOutputService:   L1TUtmTriggerMenuOnlineProd:L1TUtmTriggerMenuOnlineProd@callESModule  12-Mar-2023 09:54:12 CET Run: 4294967295 Event: 1 Logger.h:60
Fetching tag info for L1TriggerKeyListExt_Stage2v0_hlt
%MSG
%MSG-int L1-O2O: L1TUtmTriggerMenuOnlineProd:   L1TUtmTriggerMenuOnlineProd:L1TUtmTriggerMenuOnlineProd@callESModule  12-Mar-2023 09:54:12 CET Run: 4294967295 Event: 1
Producing L1TUtmTriggerMenu with key =ugt_l1_l1menu_collisions2020/v8
%MSG

std::bad_alloc exception
----- Begin Fatal Exception 12-Mar-2023 09:54:13 CET-----------------------
An exception of category 'BadAlloc' occurred while
   [0] Processing  Event run: 4294967295 lumi: 1 event: 1 stream: 0
   [1] Running path 'p'
   [2] Prefetching for module L1CondDBPayloadWriterExt/'L1CondDBPayloadWriterExt'
   [3] Calling method for EventSetup module L1TGlobalPrescalesVetosOnlineProd/''
Exception Message:
A std::bad_alloc exception was thrown.
The job has probably exhausted the virtual memory available to the process.
----- End Fatal Exception -------------------------------------------------
----- Begin Fatal Exception 12-Mar-2023 09:54:13 CET-----------------------
An exception of category 'BadAlloc' occurred while
   [0] Processing  Event run: 4294967295 lumi: 1 event: 1 stream: 0
   [1] Running path 'p'
   [2] Prefetching for module L1CondDBPayloadWriterExt/'L1CondDBPayloadWriterExt'
   [3] Calling method for EventSetup module L1TGlobalPrescalesVetosOnlineProd/''
Exception Message:
A std::bad_alloc exception was thrown.
The job has probably exhausted the virtual memory available to the process.
----- End Fatal Exception -------------------------------------------------
L1-O2O-ERROR: could not write TSC payloads
L1-O2O-ERROR: could not write TSC payloads 
  ```

**Comment by arnobaer** [comment]
> @arnobaer Have there been any major changes to how datatypes with memory allocation are handled in the new utm library? Especially anything connected to `esTriggerMenu`? What is this `if (Algorithm::isGate(token)) continue;` line that is different between v0.11.x and v0.10.x?

There is only a single memory allocation using `new` to create and return the `tmeventsetup::esTriggerMenu` in function `tmeventsetup::getTriggerMenu`. In the new version I refactored the local function `tmeventsetup::_getTriggerMenu` into the constructor of `tmeventsetup::esTriggerMenuHandle`.

Function ``tmgrammar::Algorithm::isGate`` reads a bit cryptic but is used to test if a token returend from the algorithm parser is either a unary or binary logical operator (NOT, AND, OR, XOR) or a condition expression (e.g. `MU42`). Again, as this function is used in CMSSW emulator code I was not able to refactor this into e.g. `isOperator` to make its use more verbose.

@aloeliger have you tried to build against [utm_0.11.1](https://gitlab.cern.ch/cms-l1t-utm/utm/-/tree/utm_0.11.1/) patch?

Meanwhile I checked with Valgrind, without errors, but approx. 220 MiB of memory are allocated by loading a recent collisions menu.

**Comment by aloeliger** [comment]
I've sent @epalencia a message to see if he can update the PR. @smuzaffar do you know how much virtual memory is associated with the unit testing jobs?

**Comment by smuzaffar** [comment]
@aloeliger , these jobs run on a 16core VM with 112GB memory. We do not limit memory and tests should have access to all
```
stack size                  (kbytes, -s) unlimited
cpu time                    (seconds, -t) unlimited
max user processes          (-u) 471289
virtual memory              (kbytes, -v) unlimited
file locks                  (-x) unlimited
```

aloeliger:
So, unlikely to be upset by a 220 Mb menu load. Or even by slight memory errors by the sounds of it.

epalencia:
> I've sent @epalencia a message to see if he can update the PR. @smuzaffar do you know how much virtual memory is associated with the unit testing jobs?

I have created a new PR using the updated library: https://github.com/cms-sw/cmsdist/pull/8375

PR #8357, opened 2023-03-04
Replace c++1z with c++17 [13.0.x]

fwyzard:
Replace the legacy flag `-std=c++1z` with the official flag `-std=c++17`.

According to the `gcc` man page `c++1z` is deprecated:
> **-std=**
>     Determine the language standard.   This option is currently only supported when compiling C or C++.
>   - **c++17**
>   - **c++1z**
>       The 2017 ISO C++ standard plus amendments.  The name **c++1z** is deprecated.

`nvcc` does not support `c++1z`:
> **--std** {**c++03**|**c++11**|**c++14**|**c++17**|**c++20**}           (**-std**)                          
>         Select a particular C++ dialect.  Note that this flag also turns on the corresponding
>         dialect flag for the host compiler.
>         Allowed values for this option:  'c++03','c++11','c++14','c++17','c++20'.

`clang` does documents only `c++17` and does not mention `c++1z`, but it accepts it for compatibility.

As `-std=c++1z` is now an alias for `-std=c++17`, no changes are expected.

### DIFF
--- a/scram-tools.file/tools/gcc/env.sh
+++ b/scram-tools.file/tools/gcc/env.sh
@@ -34,7 +34,7 @@ source ${SCRAM_TOOLS_BIN_DIR}/os_libdir.sh
 # optimizations as they become available in gcc.
 
 GCC_CXXFLAGS=""
-GCC_CXXFLAGS="$GCC_CXXFLAGS -std=c++1z -ftree-vectorize"
+GCC_CXXFLAGS="$GCC_CXXFLAGS -std=c++17 -ftree-vectorize"
 GCC_CXXFLAGS="$GCC_CXXFLAGS -Werror=array-bounds -Werror=format-contains-nul -Werror=type-limits"
 GCC_CXXFLAGS="$GCC_CXXFLAGS -fvisibility-inlines-hidden"
 GCC_CXXFLAGS="$GCC_CXXFLAGS -fno-math-errno --param vect-max-version-for-alias-checks=50"


fwyzard:
backport #8356

fwyzard:
please test

fwyzard:
urgent

@perrotta, @rappoccio, IMHO this makes sense only if it enters 13.0.0.
Since it touches the gcc specs, it will likely trigger a complete rebuild of all the externals, so it's probably not worth backporting it to 13.0.x after 13.0.0 is out.

smuzaffar:
+externals

@fwyzard , this only touches the gcc toolfile which are only used by cmssw and coral packages so it does not rebuilt everything.

fwyzard:
> @fwyzard , this only touches the gcc toolfile which are only used by cmssw and coral packages so it does not rebuilt everything.

Wait... so how do we pass the `-std` option to the other externals ? Like ROOT ?

smuzaffar:
@fwyzard , we do not have common flags for all externals. Each externals can use its own build flags . For root or most cmake based project we run cmake to enable C++17 standards https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_13_1_X/master/root.spec#L92

fwyzard:
Ah, I see.

But isn't it dangerous to potentially use different C++ standards when building an external (`-std` flag comes from the spec files) vs when using its header files in CMSSW (`-std` flag comes from the gcc configuration) ?

smuzaffar:
it is only dangerous if different standards generate different symbols (ABI) which one can not link. This happened when we moved to c++11 where ABI was changed and we could not link external built with old ABI. But c++11 and above has the same ABI so it is not dangerous. We update the the standard for an external when it needs it e.g. an external which only uses features of c++14 can be build with c++17 flags  but if it requires c++17 then we can not build it with c++14.

fwyzard:
Well, no, it's also dangerous if any of the headers uses something like
```c++
#if __cplusplus >= 201703L
// only available with c++17
...
#endif
```
to define different functionality.

If user code is compiled for c++17 and _uses_ that functionality, it may cause link time errors.

smuzaffar:
sure, such cases should cause link error and we had seen such issues in past with tensorflow when it was compiled with c++14 causing it to use absl::string_view as std::string_view was only available in c++17. Mostly we catch these issues during PR tests time.

fwyzard:
OK, with 13.0.0 out, I don't think we need to backport this.

PR #8355, opened 2023-03-03
[TBB] Fix concurrent_[bounded_]queue correctness on weak memory models

smuzaffar:
Applied https://github.com/oneapi-src/oneTBB/pull/782 patch on top of v2021.8.0 .
This should fix the [testFWCoreUtilities failure in ARM IBs](https://github.com/cms-sw/cmssw/issues/34647)

PR #8358, opened 2023-03-04
[TBB] Fix concurrent_[bounded_]queue correctness on weak memory models [13.0.x]

fwyzard:
Applied [oneapi-src/oneTBB#782](https://github.com/oneapi-src/oneTBB/pull/782) patch on top of v2021.8.0 .
This should fix the [testFWCoreUtilities failure in ARM IBs](https://github.com/cms-sw/cmssw/issues/34647)

### DIFF
--- a/tbb-782.patch
+++ b/tbb-782.patch
@@ -1,0 +1,320 @@
+From 2f1b2a2055da106b98859b60a785c0c45829dc79 Mon Sep 17 00:00:00 2001
+From: Alexei Katranov <alexei.katranov@intel.com>
+Date: Mon, 21 Feb 2022 15:03:10 +0300
+Subject: [PATCH 1/6] Fix cq & cbq correctness on ARM
+
+Signed-off-by: Alexei Katranov <alexei.katranov@intel.com>
+---
+ include/oneapi/tbb/concurrent_queue.h         |  6 +++--
+ .../tbb/detail/_concurrent_queue_base.h       | 14 +++++------
+ test/tbb/test_concurrent_queue_whitebox.cpp   | 23 +++++++++----------
+ 3 files changed, 22 insertions(+), 21 deletions(-)
+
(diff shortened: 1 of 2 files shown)


fwyzard:
type bugfix

fwyzard:
backport #8355

fwyzard:
please test

fwyzard:

fwyzard:

fwyzard:
urgent

@perrotta, @rappoccio, as this is a bug fix, can we have it in 13.0.0 ?

fwyzard:

smuzaffar:
please test

smuzaffar:

perrotta:
@fwyzard I started this morning the build of CMSSW_13_0_1 and I forgot to (test and, if successful) include this one
I could stop the build, if needed. However, as far as I understand this is not strictly necessary, and it could even enter a forthcoming new 13_0_X release (expected soon because also PPS needs to include some update for the data taking): if so, I would not delay the ongoing build of 13_0_1, requested by HLT and L1T.
Please let us know,

fwyzard:
Hi @perrotta,
you can go ahead with the ongoing build.

fwyzard:
The long version is: this PR fixes a problem in TBB queues on ARM and Power - but we've lived with it for years now, so empirically it should not break anything.

Given how TBB queues are used in the framework, it may lead to a non optimal reuse of resources, but it should introduce any incorrect behaviour.

While there may be other code that benefits from the optimal behaviour, a `concurrent_queue` cannot really guarantee that it is empty (another thread could push to it right after the check was made), so nothing should actually rely on it.

So, the fix is good to have, but not critical.

smuzaffar:
+externals

smuzaffar:
merging it for next 13.0.X IB so that it can be part of next 13.0.2 release

PR #8361, opened 2023-03-06
[BuildRules] Changes for running unit tests from separate test dirs

smuzaffar:
- properly cleanup up top level unit test directory
- add test directories of release/full release in path
- Run test scripts using full path e.g. for `<test name="FOO" command="bar.sh"/>`  test will be run as `$CMSSW_BASE/src/Package/Name/test/bar.sh`

These changes will allow to run unit tests from their separate test paths. This will avoid the issue if two or more tests writing in to same file in $CMSSW_BASE

### DIFF
--- a/scram-project-build.file
+++ b/scram-project-build.file
@@ -54,7 +54,7 @@ BuildRequires: dwz
 %endif
 
 %if "%{?configtag:set}" != "set"
-%define configtag       V07-07-03
+%define configtag       V07-07-04
 %endif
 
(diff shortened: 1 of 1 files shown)


smuzaffar:
test parameters:
- full_cmssw = true
- skip_tests = static

smuzaffar:
please test

smuzaffar:
+externals

lets get this in, we still need to fix many unit tests so that they use full paths instead of `src/Package`

PR #8362, opened 2023-03-06
Add more libamdhip64.so.5 symbol versions

fwyzard:
The installation of CMSSW 13.0.0/13.1.0-pre1 from RPMs fails complaining about missing symbols, all the way up to hip_5.3:
```
error: Failed dependencies:
        libamdhip64.so.5(hip_5.3)(64bit) is needed by cms+cmssw+CMSSW_13_1_0_pre1-1-1.x86_64
```

This PR includes all symbol versions in the spec file's Provides.

The list of symbol versions supported by the library can be extracted e.g. with
```
nm -D lib/libamdhip64.so --defined-only | cut -s -d @ -f3 | sort -u
```

PR #8363, opened 2023-03-06
Add more libamdhip64.so.5 symbol versions [13.0.x]

fwyzard:
The installation of CMSSW 13.0.0/13.1.0-pre1 from RPMs fails complaining about missing symbols, all the way up to hip_5.3:
```
error: Failed dependencies:
        libamdhip64.so.5(hip_5.3)(64bit) is needed by cms+cmssw+CMSSW_13_1_0_pre1-1-1.x86_64
```

This PR includes all symbol versions in the spec file's Provides.

The list of symbol versions supported by the library can be extracted e.g. with
```
nm -D lib/libamdhip64.so --defined-only | cut -s -d @ -f3 | sort -u
```

### DIFF
--- a/rocm.spec
+++ b/rocm.spec
@@ -7,6 +7,13 @@ Provides: librocm-core.so.1()(64bit)
 Provides: librocm_smi64.so.5()(64bit)
 Provides: libamdhip64.so.5()(64bit)
 Provides: libamdhip64.so.5(hip_4.2)(64bit)
+Provides: libamdhip64.so.5(hip_4.3)(64bit)
+Provides: libamdhip64.so.5(hip_4.4)(64bit)
+Provides: libamdhip64.so.5(hip_4.5)(64bit)
+Provides: libamdhip64.so.5(hip_5.0)(64bit)
+Provides: libamdhip64.so.5(hip_5.1)(64bit)
+Provides: libamdhip64.so.5(hip_5.2)(64bit)
+Provides: libamdhip64.so.5(hip_5.3)(64bit)
 
 # This rpm packages only symlinks to an installation that is already on CVMFS.
 # Configure pkgtools to keep the static libraries, to avoid actually trying to


fwyzard:
type bugfix

fwyzard:
backport #8362

fwyzard:
please test

fwyzard:
I'm afraid this may prevent the installation of CMSSW_13_0_0 on the HLT farm.

smuzaffar:
for HLT form, they just need to update https://github.com/cms-sw/cmspkg/blob/master/puppet-module/lib/puppet/provider/package/cmspkg.rb which should allow to upgrade cms common packages to latest revision.

fwyzard:
> ... upgrade cms common packages to latest revision.

Ehm, could you explain how that helps with the missing requirements ?

smuzaffar:
`fakesystem` is a package which is used to provide extra provides. The new cmspkg puppet provider makes sure that it upgrade the cms common packages for system (i.e. it runs `cmspkg -y --upgrade-packages upgrade` before install `cmssw`). Latest revision of fakesystem provide the missing requirement ( https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_13_1_X/master/fakesystem.spec#L109 ) .

Note that we treat rocm as system package as we do not bundle it along with cmssw itself. so instead if updating rocm spec, it should be better to update the fakesystem for missing provides

fwyzard:
> Note that we treat rocm as system package as we do not bundle it along with cmssw itself. so instead if updating rocm spec, it should be better to update the fakesystem for missing provides

OK - shall we move all the `Provides` there then ?

By the way, is the fakesystem package architecture dependent ?

smuzaffar:
>  OK - shall we move all the Provides there then ?

We can move all the rocm provides there but I was thinking to start bundling runtime libs (if we can distribute) along with cmssw software stack and then we can get rid of these dummy provides

> By the way, is the fakesystem package architecture dependent ?

currently it is union of all provides of all archs so it is kind of arch independenct but now I am thinking that some of these ( rocm part) can be conditional (e..g only for x86_64)

PR #8372, opened 2023-03-12
Build ONNXRuntime for same cuda archs as the rest of cms externals

smuzaffar:
This should fix ONNXRuntime for Cuda 12 which does not support cuda arch 37 ( enabled by default by ONNXRuntime)


smuzaffar:
please test

smuzaffar:
+externals

but this will not fix cuda 12 build issues as onnxruntime does not support cuda 12 yet ( https://github.com/microsoft/onnxruntime/issues/14495 )

PR #8381, opened 2023-03-14
[do not merge] Disable the use of GNU unique symbols

fwyzard:
From `man gcc`:
> `-fno-gnu-unique`
> On systems with recent GNU assembler and C library, the C++ compiler uses the "STB_GNU_UNIQUE" binding to make sure that definitions of template static data members and static local variables in inline functions are unique even in the presence of "RTLD_LOCAL"; this is necessary to avoid problems with a library used by two different "RTLD_LOCAL" plugins depending on a definition in one of them and therefore disagreeing with the other one about the binding of the symbol.  But this causes "dlclose" to be ignored for affected DSOs; if your program relies on reinitialization of a DSO via "dlclose" and "dlopen", you can use `-fno-gnu-unique`.

Since CMSSW loads plugins using `RTLD_GLOBAL`, not `RTLD_LOCAL`, we could check the impact of disabling the GNU unique extension, that does not seem to be supported by clang.


smuzaffar:
test parameters:
- full_cmssw = true

fwyzard:
please test

fwyzard:
please test

fwyzard:
image

spurious failure ?

fwyzard:
please test

PR #8382, opened 2023-03-14
Explicitly use the GNU binutils ld.bfd linker

fwyzard:
As we configure them, gcc and clang already use the GNU bfd linker by default (`ld` or `ld.bfd`, they are the same in our externals).

Instead `hipcc` uses its version of `ld.lld` by default.

This can cause inker problems in some corner cases, for example when linking objects with weak `V` symbols with objects with unique `u` symbols: `ld.bfd` and `ld.gold` seem to handle that, while `ld.lld` fails with a duplicate symbol error.

Rather than changing how gcc and clang generate symbols, it looks like the problem can be fixed instructing `hipcc` to use the GNU bfd linker.
This improves consistency with the rest of our build system, possibly avoiding other corner cases.

Setting the flag for gcc should propagate automatically to clang and hipcc, making sure all compilers consistently use the same linker.

### DIFF
--- a/scram-tools.file/tools/gcc/env.sh
+++ b/scram-tools.file/tools/gcc/env.sh
@@ -40,6 +40,9 @@ GCC_CXXFLAGS="$GCC_CXXFLAGS -fvisibility-inlines-hidden"
 GCC_CXXFLAGS="$GCC_CXXFLAGS -fno-math-errno --param vect-max-version-for-alias-checks=50"
 GCC_CXXFLAGS="$GCC_CXXFLAGS -Xassembler --compress-debug-sections"
 
+# Explicitly use the GNU binutils ld.bfd linker
+GCC_CXXFLAGS="$GCC_CXXFLAGS -fuse-ld=bfd"
+
 case $(uname -m) in
   aarch64 ) GCC_CXXFLAGS="$GCC_CXXFLAGS -fsigned-char -fsigned-bitfields" ;;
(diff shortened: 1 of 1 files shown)


fwyzard:
test parameters:
  - full_cmssw = true

fwyzard:
please test

fwyzard:
> ---> test DRNTest had ERRORS

The error is
```
%MSG-e TritonService:  AfterSourceConstruction 14-Mar-2023 16:17:33 CET  pre-events
CMS_TRITON_GRPC_PORT: 8001
WARNING: While bind mounting '/data:/data': destination is already in the mount point list
INFO:    underlay of /etc/grid-security required more than 50 (105) bind mounts
ERROR:   container cleanup failed: no instance found with name triton_server_instance_13a406f3-14ac-481f-b210-3ad414afec99
FATAL:   container creation failed: mount hook function failure: mount /etc/tnsnames.ora->/etc/tnsnames.ora error: while mounting /etc/tnsnames.ora: destination /etc/tnsnames.ora doesn't exist in container

FATAL:   failed to start instance: while running /usr/libexec/apptainer/bin/starter: exit status 255
Error for command "stop": no instance found
```

so I assume it is unrelated to this PR.

smuzaffar:
+externals
we have seen this unit test randomly fails or random nodes.

smuzaffar:

smuzaffar:

lets tests this change for non-x86_64 archs too

smuzaffar:
this looks good

PR #8389, opened 2023-03-16
Rename jemalloc libraries with debug and profiling options enabled.

gartung:


gartung:
please test with https://github.com/cms-sw/cmssw/pull/41071

gartung:
please test with https://github.com/cms-sw/cmssw/pull/41071

smuzaffar:
@gartung , looks like you also need to update https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_13_1_X/master/cmssw-drop-tools.file and remove `jemalloc-debug and jemalloc-prof` from there otherwise these tools will not be available in cmssw env.

smuzaffar:
please test with https://github.com/cms-sw/cmssw/pull/41071

I have updated `cmssw-drop-tool`

smuzaffar:
@gartung , this did not work. patchelf shows that lib name is still `libjemalloc.so.2`. So although scram uses `-ljemalloc-prof` at link time but the actual library linked is `libjemalloc.so.2` ( due to soname registered in the library) [b]. I think if you run the following after renaming the libs then compiler will be able to properly link the jemalloc-prof

```
patchelf --set-soname  libjemalloc-prof.so.2  libjemalloc-prof.so.2
```

I have tried it locally and it works

[a]
```
Singularity>  patchelf --print-soname  libjemalloc-prof.so
libjemalloc.so.2
Singularity>  patchelf --print-soname  libjemalloc-prof.so.2
libjemalloc.so.2
```

[b]
```
Singularity> ldd tmp/el8_amd64_gcc11/src/FWCore/Framework/bin/cmsRunJE/cmsRunJE | grep jemalloc
        libjemalloc.so.2 => /build/muz/xx/CMSSW_13_1_X_2023-03-16-1100/external/el8_amd64_gcc11/lib/libjemalloc.so.2 (0x00007ff33dcf1000)
```

gartung:
please test with https://github.com/cms-sw/cmssw/pull/41071

gartung:
please test with https://github.com/cms-sw/cmssw/pull/41071

gartung:
please test with https://github.com/cms-sw/cmssw/pull/41071

smuzaffar:
@gartung , this look good now. `cmsRunJE` is now properly linked to `libjemalloc-prof`. While we are at this, may be you can clean these tools up and move the common section from jemalloc*.spec in to `jemalloc-common.file`. I am thinking that `%install` and `%post` sections can be moved to `jemalloc-common.file` e.g. in jemalloc-common we can have 

```
%install
make install
%if "%{n}" != "jemalloc"
  # Rename jemalloc libs
  mv %{i}/lib/libjemalloc.so.2 %{i}/lib/lib%{n}.so.2
  rm %{i}/lib/libjemalloc.so
  ln -sf lib%{n}.so.2 %{i}/lib/lib%{n}.so
  patchelf --set-soname  lib%{n}.so.2  %{i}/lib/lib%{n}.so.2
  # Make sure there are no other libs. If there are then we should fail and updade the recipe
  if [ $(ls %{i}/lib/lib* | grep -v lib%{n}. | wc -l) -gt 0 ] ; then exit 1; fi
%endif

%post
%{relocateConfig}bin/jemalloc.sh
%{relocateConfig}bin/jemalloc-config
```

and remove `%install` and %post` sections from the jemalloc*.spec files

gartung:
please test with https://github.com/cms-sw/cmssw/pull/41071

smuzaffar:
+externals
thanks @gartung , looks good now

PR #8396, opened 2023-03-20
Update ROOT in master

aandvalenzuela:
Align master with root626. New commit on 17 of March (https://github.com/root-project/root/commit/5858905487bba8769a8d612d1e3aff20a3acb63d).

PR #8400, opened 2023-03-22
Update ROOT in nonLTO branch

aandvalenzuela:
Testing to see if comparison differences in https://github.com/cms-sw/cmsdist/pull/8396 are due to LTO


aandvalenzuela:
please test

smuzaffar:
test parameters:
- baseline = self

smuzaffar:
please test
 
lets use the `baseline = self` so that the comparison is also run against nonlto IB

smuzaffar:
so nonlto does not show any reco comparison differences

PR #8403, opened 2023-03-22
[SCRAM] disply BuildFile name in case of parsing errors

smuzaffar:


smuzaffar:
test parameters:
- addpkg = FWCore,DataFormats

smuzaffar:
please test

smuzaffar:
+externals
this is change in scram which is only used at build time. Build passed so this is good to go in

PR #8407, opened 2023-03-24
[FWLite] Added utm dependency

smuzaffar:
https://github.com/cms-sw/cmssw/pull/41036 added utm dependency in CondFormats/L1TObjects (which is part of fwlite distribution). This PR proposes to add utm as fwlite deps which should fix the fwlite build errors in IBs.


smuzaffar:
for now we do not have any PR tests to check this. So I am merging it and restarting the FWLite for last night. @iarspider @aandvalenzuela we should add a PR test to build fwlite package in case fwlite* files are changes in cmsdist

PR #8429, opened 2023-04-05
Mass-update python packages

iarspider:


iarspider:
please test

iarspider:
please abort

iarspider:
please test

iarspider:
please test

iarspider:
@smuzaffar we can either downgrade `py3-scons` or drop it completely (since it's only used by `openloops` and it comes with it's own copy).

smuzaffar:
@iarspider , drop it if it is only needed by openloops (which as you mentioned comes with its own copy)

iarspider:
please test

iarspider:
please test

iarspider:
please test

iarspider:

iarspider:

iarspider:

iarspider:

iarspider:
Unit test failure: also happens in master. RelVal failures: complain that `ROOT PCM /scratch/cmsbuild/jenkins_a/workspace/ib-run-pr-tests/testBuildDir/el8_ppc64le_gcc11/lcg/root/6.26.11-5398835674af07ddeac0bd6be0987f84/lib/libROOTVecOps_rdict.pcm file does not exist`

iarspider:

smuzaffar:
please test

lets re-run tests for production arch too

build-with-pip.file
```diff
@@ -24,7 +24,7 @@
 %endif
 
 %if "%{?pip_name:set}" != "set"
-%define pip_name %(echo %n | cut -f2-10 -d-)
+%define pip_name %(echo %n | cut -f2-10 -d- | tr '-' '_')
```
smuzaffar: @iarspider , why this change was needed? Ideally we should have the correct package name in requirement.txt file i.e. `%n` should already be the correct.

Note that this change means all the `py3-*` packages will be rebuilt. If this change is required then we should also see if we can update pip and setuptools.

iarspider: @smuzaffar because this is how tarballs are named: the packages' names have `-`, but tarballs have `_`.

iarspider:
To check and update:
```
pip==23.1.1
setuptools==67.7.2 
```

build-with-pip.file
smuzaffar: thanks @iarspider , if it is just the tar file name then we should fix the https://github.com/cms-sw/pkgtools/blob/V00-34-XX/cmsBuild#L623-L627 code and check for both

- '%s-%s.tar.gz' % (pack, pkg[1].strip())
- '%s-%s.tar.gz' % (pack.repalce("-","_"), pkg[1].strip())

`pip_name` is also used for searching/installing package so it should match what we have in requirements.txt.

Change in  pkgtools/cmsBuild will allow us to find the correct tar file without rebuilding all the py3 packages

iarspider:
please test
Updated pip, setuptools, reverted changes to build-with-pip

iarspider:
please test

iarspider:
please test

iarspider:
@smuzaffar I would suggest installing pip, wheel and setuptools from wheels, since (a) they are pure-python packages, (b) we don't patch them, and (c) they can be a pain to build, since they need each other when building from source:

* setuptools: https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/py-setuptools/package.py#L201-L206
* wheel: https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/py-wheel/package.py#L82-L84
* pip: https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/py-pip/package.py#L98-L101

smuzaffar:
@iarspider , we never had issue with building these tools. It is always the other packages which failed due these updates. What failed this time?

iarspider:
> @iarspider , we never had issue with building these tools. It is always the other packages which failed due these updates. What failed this time?

[py3-wheel](https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-26fbda/32192/externals/py3-wheel/0.37.1-8e8cef1adc2331996c1fbe95d371d3e9/log) with new setuptools and pip

smuzaffar:
try wheel `0.40.0`, I think version `0.38.3` fixed this recursive dep issue

iarspider:
please abort

iarspider:
please test

smuzaffar:
ah too bad, `0.40.0 , Wheel now uses flit-core as a build backend ` (https://wheel.readthedocs.io/en/stable/news.html)  and `flit-core` requires `wheel` :-(

smuzaffar:
please test

hopefully this will fix the wheel/flit-core issue

pip/requirements.txt
```diff
@@ -214,8 +214,9 @@ notebook==6.5.3
 notebook-shim==0.2.2
 #numba==0.56.4
 numexpr==2.8.4
+# setuptools version <64 is needed by numpy: https://github.com/pypa/setuptools/issues/3549
```
iarspider, line 217: @smuzaffar I think this warning should be in py3-setuptools.spec

build-with-pip.file
```diff
@@ -60,8 +61,10 @@
 %define pip pip3&package_dependency=py3-pip
 Requires: python3 py3-pip py3-setuptools
 %if "%{n}" != "py3-wheel"
+%if "%{?PipDropWheelDep:set}" != "set"
```
iarspider: `PipNoWheelDep`?

pip/requirements.txt
smuzaffar: I agree, can you please update it. Also see if you can squash the commit.  there are too many commits for trying something and then reverting it :-(

build-with-pip.file
smuzaffar: either `PipNoWheelDep` and then check for `== "set"` or `PipDropWheelDep` and check for `!= "set"` are same :-)

smuzaffar: I really have no preference here. feel free to change it :-)

smuzaffar:
@iarspider update  `pip/flit-core.file` for `PipNoWheelDep` too

iarspider:
please test

smuzaffar:
please test

oops, forgot the `pip/wheel.file` .

smuzaffar:
please test

smuzaffar:
please test

smuzaffar:

smuzaffar:

smuzaffar:

iarspider:

iarspider:

iarspider:

iarspider:
please test

iarspider:

smuzaffar:
+externals
this looks good, last commit just fixes some port relocation issues. Lets get this in tonight IB

iarspider:
timed out while downloading thepeg from hepforge, let's try again

iarspider:
please abort

PR #8442, opened 2023-04-17
Update Millepede, GBL versions, addition of LAPACK  solution method

consuegs:


tvami:
test parameters:
 - addpkg  = Alignment/MillePedeAlignmentAlgorithm

tvami:
@cmsbuild , please test

consuegs:
In anticipation to this PR the pede test case was tested on lxplus and the following worked:

```
make pede \
SUPPORT_LAPACK64=yes \
LAPACK64=MKL \
LAPACK64_LIBS_DIR="/cvmfs/projects.cern.ch/intelsw/oneAPI/linux/x86_64/2022/mkl/2022.1.0/lib/intel64" \
LAPACK64_LIB=mkl_rt   

export OMP_STACKSIZE=60M; MKL_THREADING_LAYER=GNU; export OMP_NUM_THREADS=10; export MKL_NUM_THREADS=10; \
export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/cvmfs/projects.cern.ch/intelsw/oneAPI/linux/x86_64/2022/mkl/2022.1.0/lib/intel64;

./pede -t
```

millepede.spec
```diff
@@ -8,7 +8,11 @@ Requires: zlib
 %build
 make \
   ZLIB_INCLUDES_DIR="${ZLIB_ROOT}/include" \
-  ZLIB_LIBS_DIR="${ZLIB_ROOT}/lib"
+  ZLIB_LIBS_DIR="${ZLIB_ROOT}/lib" \
+  SUPPORT_LAPACK64=yes \
+  LAPACK64=MKL \
+  LAPACK64_LIBS_DIR="/cvmfs/projects.cern.ch/intelsw/oneAPI/linux/x86_64/2022/mkl/2022.1.0/lib/intel64" \
+  LAPACK64_LIB=mkl_rt
```
smuzaffar, line 15: We should not pick up libraries from cvmfs. Required dependencies should be expilcitly added.
Also this will break non-x86_64 builds (ppc64le, aarch64)

consuegs: I see, thanks. What should be done in case the library can not be explicitly added because not available with CMSSW?

```
scram tool info mkl     
SCRAM error: >>>> Tool mkl is not setup for this project area. <<<<
```

i.e. @smuzaffar is there a way to use mkl?

If not, I would add a commit and change mkl to openblas, explicitly adding openblas.

```
scram tool info openblas

Tool info as configured in location /afs/cern.ch/cms/CAF/CMSALCA/ALCA_TRACKERALIGN/data/commonValidation/CMSSW_13_0_2
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Name : openblas
Version : 0.3.15-26c67b8b638762cfd2e2bcfc936e3ec7
++++++++++++++++++++
```

iarspider: @smuzaffar maybe we can add a "recipe" for Intel MKL based on recipe for V-tune?

smuzaffar: @consuegs , yes please go ahead with `openblas` which we use every where in cmssw.
@iarspider , I do not think we have `v-tune` recipe in cmssw. We have `intel-vtune` but that also points to cvmfs and only available at CERN and for `x86_64`

smuzaffar:
closing it, please open a new one with openblas deps

PR #8466, opened 2023-04-25
Search for unpack containers under /cvmfs/singularity.opensciencegrid.org too

smuzaffar:
to allow running `cmssw-env --cmsos cmssw/cms:rhel8`


smuzaffar:
please test

PR #8469, opened 2023-04-26
llvm does not support -fno-fat-lto-objects

smuzaffar:
optimization flag '-fno-fat-lto-objects' is not supported


smuzaffar:
please test with cms-sw/cmssw#41403

smuzaffar:
+externals

PR #8471, opened 2023-04-26
[rust] Delete .git files which are breaking slc7 builds

smuzaffar:
This should fix slc7 builds

### DIFF
--- a/rust.spec
+++ b/rust.spec
@@ -14,6 +14,8 @@ Requires: llvm
 
 %prep
 %setup -n %{n}-%{realversion}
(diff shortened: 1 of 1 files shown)


smuzaffar:
please test

smuzaffar:

fwyzard:
Why do we even have Rust in cmsdist ?

smuzaffar:
`rust` is needed by `py3-cryptography` which is needed by `py3-poetry` ( a python packaging and deps managment tool required by many python packages) :-(

smuzaffar:
+externals

rust builds for both el8 and slc7 now

PR #8473, opened 2023-04-27
drop numba and llvmlite

smuzaffar:
llvmlite support of newer llvm version is very slow. It is only needed by `numba` which is not used any where in cmssw production jobs. So This PR proposes to drop `numba` and `llvmlite`


smuzaffar:
test parameters:
- full_cmssw =true
- pull_request = cms-sw/cmssw#41435
- skip_tests = static

smuzaffar:
please test

smuzaffar:
+externals

PR #8485, opened 2023-05-02
fastjet 3.4.1 and fastjet-contrib 1.051

smuzaffar:
Update [fastjet 3.4.1](https://fastjet.fr/all-releases.html) with bug fixes
```
- Resolved MAJOR BUG that arose with full thread-safety enabled (reported by Ludo Scyboz),
where PseudoJet::reset_momentum and PseudoJet::operator= failed to update internally cached
values of phi() and rap(). In some instances related issues could also lead to race conditions.
- eliminated NaN from square-root of negative mean areas in background estimation (now returns zero) 
```

### DIFF
--- a/fastjet-contrib.spec
+++ b/fastjet-contrib.spec
@@ -1,8 +1,8 @@
-### RPM external fastjet-contrib 1.044
-%define tag 03f2fb3c7e26248f5cab3d6c52fab3e112342113
+### RPM external fastjet-contrib 1.051
+%define tag aa1972cd2b3e8aa2c76764f14122e8f728c23712
 %define branch cms/v%{realversion}
 %define github_user cms-externals
-Source: git+https://github.com/%github_user/%{n}.git?obj=%{branch}/%{tag}&export=%{n}-%{realversion}&foo=1&output=/%{n}-%{realversion}.tgz
+Source: git+https://github.com/%github_user/%{n}.git?obj=%{branch}/%{tag}&export=%{n}-%{realversion}&output=/%{n}-%{realversion}.tgz
 Requires: fastjet
 %define keep_archives true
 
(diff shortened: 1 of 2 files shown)


smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
@perrotta @rappoccio is it too late to get this in for 13.1.X ? 
@makortel , this version of fastjet has a bug fix for thread safety but I am not sure if we are affected by this

perrotta:
@smuzaffar I intended to build pre4 ~now
Since I did not start it yet, it is not (yet) too late
If the inclusion is safe, I think we can merge it and start building pre4 right after: just decide quickly, feel free to consider "+orp" here

smuzaffar:
+externals
PR tests look good, DQM/Reco comparison show differences in MessageLogger, so let get this in then.

smuzaffar:
Thanks @perrotta , I have merged it now. If it helps then I can start an IB too

perrotta:
> If it helps then I can start an IB too

Do you think it is needed?
I would start the build o pre4 now anyhow, but we can still have a look at the results of the IB tests before the uploads

smuzaffar:
11h00 IB does not contain this change, so you have to wait for 23h00 IB results. If this is fine then no need to build an extra IB

makortel:
> this version of fastjet has a bug fix for thread safety but I am not sure if we are affected by this

Good question. I don't recall seeing fastjet-specific thread safety problems in a long time, but I don't expect our testing to be able catch all possible problems :) Since the comparisons were clean, I'd be tempted to backport these fixes to 13_0_X (but maybe after 13_1_0_pre4 has been validated?

PR #8487, opened 2023-05-02
Fix for llvm-static checks

smuzaffar:
Set `CCC_OVERRIDE_OPTIONS="^--gcc-toolchain=$(GCC_CXXCOMPILER_BASE)"` for scan-build so that `c++-analyzer` command can find gcc headers


smuzaffar:
please test with https://github.com/cms-sw/cmssw/pull/41480

Just to check is llvm-static checks work

PR #8511, opened 2023-05-23
Update CMSSW 13.3.x to CUDA 12.0.1

fwyzard:
CUDA-related updates:
  - update to CUDA 12.0.1 and NVIDIA drivers 525.85.12;
  - update to cuDNN version 8.8.0.121 for CUDA 12;
  - update UCX to version 1.14.1 for compatibility with CUDA 12.

### DIFF
--- a/cuda.spec
+++ b/cuda.spec
@@ -1,8 +1,8 @@
-### RPM external cuda 11.5.2
+### RPM external cuda 12.0.1
 ## INITENV +PATH LD_LIBRARY_PATH %i/lib64
 
 %define runpath_opts -m compute-sanitizer -m drivers -m nvvm
-%define driversversion 495.29.05
+%define driversversion 525.85.12
 
 %ifarch x86_64
 Source0: https://developer.download.nvidia.com/compute/cuda/%{realversion}/local_installers/%{n}_%{realversion}_%{driversversion}_linux.run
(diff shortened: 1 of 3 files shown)


fwyzard:
please test

fwyzard:
See https://github.com/cms-sw/cmssw/issues/41748 .

fwyzard:
enable gpu

fwyzard:
please test

fwyzard:
So, it looks like we cannot update CUDA because TensorFlow does not support
the recent versions ?

smuzaffar:
@fwyzard , yes unfortunately this is true. Though we have cuda 12.1.1 in gcc12 IBs but we have disabled tensorflow gpu there.

fwyzard:
Can we roll back to CUDA 12.0.1 in the GCC 12 builds ?
CUDA 12.1.1 fails to compile some of our code...

fwyzard:
(I can make the PRs for that)

smuzaffar:
> (I can make the PRs for that)

thanks, yes please open a PR.

smuzaffar:
> Can we roll back to CUDA 12.0.1 in the GCC 12 builds ? CUDA 12.1.1 fails to compile some of our code...

yesterday I reported it [here](https://forums.developer.nvidia.com/t/cicc-compilation-error-and-debug-flag/27910/23?u=shahzad.muzaffar) but do not know how to provide a simple re-producer

fwyzard:
> thanks, yes please open a PR.

https://github.com/cms-sw/cmsdist/pull/8515

fwyzard:
> So, it looks like we cannot update CUDA because TensorFlow does not support the recent versions ?

See https://github.com/tensorflow/tensorflow/issues/60691:

> The latest tensorflow stable version 2.12 is compatible with CUDA 11.8 and cuDNN 8.6. Please take a look at the official document for the tested build [configurations](https://www.tensorflow.org/install/source).
> 
> Going forward for the upcoming releases, tensorflow might move to CUDA 12.0 or higher version. We will announce that information in the mentioned official doc once the next release happens.

smuzaffar:
not needed as moving to gcc12 already brings in cuda 12.2

PR #8512, opened 2023-05-23
[ROOT6] Root update master

aandvalenzuela:


aandvalenzuela:
please test

smuzaffar:
+externals

perrotta:
@smuzaffar this update went in almost in coincidence with the start of build of CMSSW_13_2_0_pre1, see https://github.com/cms-sw/cmssw/issues/41758
Please confirm that there are no issue with it: alternatively we can stop and restart that build.

iarspider:
As far as I understand, this (and the other ROOT PR) do not affect
releases, since they target their own cmsdist branches that are only used
for ROOT6_X and ROOT628_X IBs

smuzaffar:
Right, it is for root6 ibs and is independent if 13.2.x ibs and releases

PR #8513, opened 2023-05-23
[ROOT628] Updated root to tip of branch v6-28-00-patches

aandvalenzuela:


aandvalenzuela:
please test

smuzaffar:
+externals

perrotta:
@smuzaffar this update went in almost in coincidence with the start of build of CMSSW_13_2_0_pre1, see https://github.com/cms-sw/cmssw/issues/41758
Please confirm that there are no issue with it: alternatively we can stop and restart that build.

smuzaffar:
@perrotta this is for root628 ibs only and it is independent of 13.2.x ibs and releases. So no need to stop or rebuild 13.2.0.pre1

PR #8521, opened 2023-05-25
[13_0_X] Update (add) tag for Configuration-Generator to V01-06-00

perrotta:
Move Configuration-Generator data to new tag, see
cms-data/Configuration-Generator#13


perrotta:
please test

perrotta:
@smuzaffar the build error says that the tag for Configuration-Generator is duplicated, which is true.
But when I open https://github.com/cms-sw/cmsdist/edit/IB/CMSSW_13_0_X/master/data/cmsswdata.txt I don't see the previous tag any more: that's why I only added it without removing the old one.
What I am doing wrong?
(The same issue shows up in the proposed 12_6_X backport)

PR #8515, opened 2023-05-23
Roll back to CUDA 12.0.1 [gcc 12]

fwyzard:
Roll back to CUDA 12.0.1, as CUDA 12.1.1 fails to build the `RecoPixelVertexing/PixelTriplets package` (the `cicc` CUDA compiler crashes).

PR #8527, opened 2023-05-30
Roll back to CUDA 12.1.1 [gcc 12]

smuzaffar:
Reverts cms-sw/cmsdist#8515

### DIFF
--- a/cuda-flags.file
+++ b/cuda-flags.file
@@ -22,7 +22,7 @@
 %define cuda_flags_0 -O3
 
 # generate debugging information for device code
-%define cuda_flags_1 --generate-line-info --source-in-ptx --display-error-number
+%define cuda_flags_1 --generate-line-info --display-error-number
 
 # imply __host__, __device__ attributes in constexpr functions
 %define cuda_flags_2 --expt-relaxed-constexpr
--- a/cuda.spec
+++ b/cuda.spec
@@ -1,8 +1,8 @@
-### RPM external cuda 12.0.1
+### RPM external cuda 12.1.1
 ## INITENV +PATH LD_LIBRARY_PATH %i/lib64
 
 %define runpath_opts -m compute-sanitizer -m drivers -m nvvm
-%define driversversion 525.85.12
+%define driversversion 530.30.02
 
 %ifarch x86_64
 Source0: https://developer.download.nvidia.com/compute/cuda/%{realversion}/local_installers/%{n}_%{realversion}_%{driversversion}_linux.run
--- a/ucx.spec
+++ b/ucx.spec
@@ -1,5 +1,7 @@
-### RPM external ucx 1.14.1
-Source: https://github.com/openucx/%{n}/archive/refs/tags/v%{realversion}.tar.gz
+### RPM external ucx 1.14.x
+%define branch v1.14.x
+%define tag bb39346ac1aebc67cb88c4419429de78a1260bda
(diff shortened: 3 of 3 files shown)


smuzaffar:
test parameters:
- full_cmssw = true
- enable = gpu

smuzaffar:
please test

fwyzard:
Do we gain anything from CUDA 12.1 over 12.0 ?
If not, I would prefer to stay with 12.0 and keep the `--src-in-ptx` option.

smuzaffar:
@fwyzard , I am just testing that if cuda `12.1.1` and no `--src-in-ptx` flag allow us to move forward with gcc12/cuda12. Note that cuda `12.0.1` with gcc12  has build errors 

https://cmssdt.cern.ch/SDT/cgi-bin/buildlogs/el8_amd64_gcc12/CMSSW_13_2_X_2023-05-29-1100/RecoLocalTracker/SiStripClusterizer
```
gcc/x86_64-redhat-linux-gnu/12.2.1/include/avx512fp16intrin.h(38): error: vector_size attribute requires an arithmetic or enum type
```

smuzaffar:
even removing `--source-in-ptx` does not  work for cuda12/gcc12 due to error [a] (which is there for both cuda 12.0.1 and 12.1.1)


[a]
```
>> Compiling  RecoLocalTracker/SiStripClusterizer/plugins/SiStripRawToClusterGPUKernel.cu
gcc/12.2.1/lib/gcc/x86_64-redhat-linux-gnu/12.2.1/include/avx512fp16intrin.h(38): error: vector_size attribute requires an arithmetic or enum type
                  __v8hf __attribute__ ((__vector_size__ (16)));
```

fwyzard:
hi @smuzaffar , as a workaround you can add
```c++
#undef _Float16
```
right after
```
#include <cub/cub.cuh>
```
in [`RecoLocalTracker/SiStripClusterizer/plugins/SiStripRawToClusterGPUKernel.cu`](https://github.com/cms-sw/cmssw/blob/master/RecoLocalTracker/SiStripClusterizer/plugins/SiStripRawToClusterGPUKernel.cu).

While this _should_ be safe for us (we don't use _Float16 in `SiStripRawToClusterGPUKernel.cu`) I do not know if this can have any consequences in general.

I'll open a bug report with NVIDIA.

fwyzard:
See
  - https://forums.developer.nvidia.com/t/including-cub-header-breakes-compilation-with-gcc-12-and-sse2-or-better/255018
  - https://developer.nvidia.com/nvidia_bug/4139266 (protected)
  - https://github.com/fwyzard/nvidia_bug_4139266

fwyzard:
After feedback from NVIDIA, the workaround should be safe, see https://github.com/cms-sw/cmssw/pull/41819 .

In the meantime, a fix is being prepared upstream, see https://github.com/NVIDIA/libcudacxx/pull/476 .

smuzaffar:
thanks a lot @fwyzard for following this up.  As you like to stay with cuda 12.0.1 with `--source-in-ptx`  so I am closing this PR

jrhemstad:
@fwyzard I wanted to go out of my way to thank you for providing such an excellent bug triage and reproducer in https://github.com/fwyzard/nvidia_bug_4139266. 

We had been trying to track this down from other bug reports and it would have taken me forever to figure out the `_Float16` macro conflict. 

If only every bug report and reproducer was this good :wink:.

PR #8532, opened 2023-05-31
sqlite: use zlib from cms externals

smuzaffar:
Make sure to build sqlite with zlib from CMS externals
**NOTE**: As sqlite is used by `python3`, so this change rebuilds nearly all the externals. so Do not merge it yet. May be when we have to update the GCC 11 compiler or python3 version then this change can go it. 

For GCC12 and GCC13, we can already include this change

### DIFF
--- a/sqlite.spec
+++ b/sqlite.spec
@@ -1,10 +1,12 @@
 ### RPM external sqlite 3.36.0
+Requires: zlib
 Source: https://www.sqlite.org/2021/sqlite-autoconf-3360000.tar.gz
 
 %prep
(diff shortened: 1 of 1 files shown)


makortel:
> have to update the GCC 11 compiler

In the core sw meeting yesterday my GCC update question should have been on GCC 11 (instead of 12), they just released 11.4, should we look into updating?

smuzaffar:
ah right, gcc 11.4 was out on 29th May. yes we can already start testing it

smuzaffar:
closing in favor of #8545

PR #8568, opened 2023-06-26
stop distributing old DBS Client with CRAB. Fix #8566

### LINKED ISSUE #8566: setup of DBS_CLIENT via CRAB broken in CMSSW 12 +
this line
https://github.com/cms-sw/cmsdist/blob/d34e50e14f8a0cb4bd911942ecb1f14586451923/crab/crab-setup.sh.file#L15
fails becasue `python` can not be found resulting in an annoying message.

Rather than fixing for python3, I propose to remove this, since old [DBS Client](http://github.com/dmwm/DBS) is not compatible with current DBS server. 
We should also stop pulling in old DBS Client in https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_13_2_X/master/crab-build.file

@smuzaffar if you agree I will make a PR for both files and the crab-prod/pre/dev spec files

belforte:


belforte:
Let's give this a try !

aandvalenzuela:
please test

belforte:
@aandvalenzuela @smuzaffar I am curious why this was merged but still
`/cvmfs/cms-ib.cern.ch/latest/common/crab-setup.sh` (date Jul 2nd) has the old version.
Same for the `csh` variant.

Those scripts still try to define `$DBS3_CLIENT_ROOT` even if, correctly, there is no DBS anymore in
`/cvmfs/cms-ib.cern.ch/sw/x86_64/nweek-02792/share/cms/crab-prod/v3.230417.01` 

Nothing is broken, simply the annoying message `-bash: python: command not found` is still present.

Is this simply a matter of waiting ? Or should I have done some other changed in the PR ? Or .... ?

thanks

aandvalenzuela:
Hello @belforte, I am not 100% sure but I think we should update also the `FILE_REVISION` in the `.file` files. I am opening a PR to test this. Thanks for noticing!

smuzaffar:
> Hello @belforte, I am not 100% sure but I think we should update also the `FILE_REVISION` in the `.file` files. I am opening a PR to test this. Thanks for noticing!

correct, `CMSDIST_FILE_REVISION` should be incremented

belforte:
Thanks. About "next time", is this something CRAB people should learn ? Or something that you usually take care of ?

If the former, what are the condistions under which CMSDIST_FILE_REVISION number in it must also be incremented ? I see that not all changed down in the past came with a change in that variable.

Can you add a comment about that, in case ?

Issue #8571, opened 2023-06-27
CUDA RT IB fails to build `onnxruntime` pkg


aandvalenzuela:
Hello,

The dedicated IB for cuda runtime fails to build `onnxruntime` pkg. From the latest CUDART IB log:
```
FAILED: CMakeFiles/onnxruntime_providers_cuda.dir/data/cmsbld/jenkins/workspace/build-any-ib/w/BUILD/el8_amd64_gcc11/external/onnxruntime/1.14.1-e4f32e7ee87ac6022c8ef3aa4cfda0b3/onnxruntime-1.14.1/onnxruntime/core/providers/cuda/math/binary_elementwise_ops_impl.cu.o 
[...]

PR #8545, opened 2023-06-14
GCC 11.4; CUDA 11.8; cleanup for cmsplatf/cmsos use

smuzaffar:
- Update GCC version to 11.4.1 and its prerequisites and build with zstd support ( changes backported from gcc 12 branch)
- CUDA 11.8 which has GCC 11.4 support
- Fixes to resolve https://github.com/cms-sw/cmsdist/issues/7571 i.e. clean up for the usage of `cmsplatf` and `cmsos`
- Added missing dependencies for sqlite, swig, py3-pillow, gnuplot

PR #8572, opened 2023-06-27
[CUDART] CUDA update to 11.8.0

aandvalenzuela:
It seems that the automatic forward of https://github.com/cms-sw/cmsdist/pull/8545 did not work for cuda runtime (See https://github.com/cms-sw/cmsdist/issues/8571).


aandvalenzuela:
please test for CMSSW_13_2_CUDART_X

PR #8580, opened 2023-07-06
Update file revision in crab `.file` files

aandvalenzuela:
I think updating the file revisions may fix https://github.com/cms-sw/cmsdist/pull/8568#issuecomment-1623163421.

FYI, @smuzaffar @belforte


aandvalenzuela:
please test

- I think it has no effect on the testing, but let's make sure.

PR #8583, opened 2023-07-07
Update XRootD to version 5.6.1

amadio:
@smuzaffar Please review. Thanks!

### DIFF
--- a/xrootd.spec
+++ b/xrootd.spec
@@ -8,7 +8,7 @@
 %define github_user xrootd
 Source: git+https://github.com/%github_user/xrootd.git?obj=%{branch}/v%{tag}&export=%{n}-%{realversion}&output=/%{n}-%{realversion}.tgz
 
-BuildRequires: cmake gmake autotools
+BuildRequires: cmake gmake autotools py3-pip
 Requires: zlib libuuid curl davix
 Requires: python3 py3-setuptools
 Requires: libxml2
--- a/xrootd.spec
+++ b/xrootd.spec
@@ -1,4 +1,4 @@
-### RPM external xrootd 5.5.4
+### RPM external xrootd 5.6.1
 ## INITENV +PATH LD_LIBRARY_PATH %i/lib64
 ## INITENV +PATH PYTHON3PATH %{i}/${PYTHON3_LIB_SITE_PACKAGES}
 
@@ -23,13 +23,13 @@ Requires: libxml2
 sed -i -e 's|UUID REQUIRED|UUID |' cmake/XRootDFindLibs.cmake
 
 %build
-# By default xrootd has perl, fuse, krb5, readline, and crypto enabled. 
-# libfuse and libperl are not produced by CMSDIST.
+# By default xrootd has fuse, krb5, readline, and crypto enabled.
+# libfuse is not produced by CMSDIST.
 
 rm -rf ../build; mkdir ../build; cd ../build
 cmake ../%n-%{realversion} \
   -DCMAKE_INSTALL_PREFIX=%{i} \
-  -DUSER_VERSION=%{realversion} \
+  -DXRootD_VERSION_STRING=%{realversion} \
   -DCMAKE_BUILD_TYPE=Release \
   -DFORCE_ENABLED=ON \
   -DENABLE_FUSE=FALSE \
@@ -37,11 +37,11 @@ cmake ../%n-%{realversion} \
   -DXRDCL_ONLY=TRUE \
   -DENABLE_KRB5=TRUE \
   -DENABLE_READLINE=TRUE \
-  -DENABLE_CRYPTO=TRUE \
   -DCMAKE_SKIP_RPATH=TRUE \
   -DENABLE_PYTHON=TRUE \
   -DENABLE_HTTP=TRUE \
   -DXRD_PYTHON_REQ_VERSION=3 \
+  -DPIP_OPTIONS="--verbose" \
   -DCMAKE_CXX_FLAGS="-I${LIBUUID_ROOT}/include -I${DAVIX_ROOT}/include" \
   -DUUID_INCLUDE_DIR="${LIBUUID_ROOT}/include" \
   -DUUID_LIBRARY="${LIBUUID_ROOT}/lib64/libuuid.%{soext}" \
--- a/xrootd.spec
+++ b/xrootd.spec
@@ -6,7 +6,7 @@
 %define tag %{realversion}
 %define branch master
 %define github_user xrootd
-Source: https://xrootd.slac.stanford.edu/download/v%{realversion}/%{name}-%{realversion}.tar.gz
+Source: https://xrootd.slac.stanford.edu/download/v%{realversion}/%{n}-%{realversion}.tar.gz
 
 BuildRequires: cmake gmake autotools py3-pip
 Requires: zlib libuuid curl davix
--- a/xrootd.spec
+++ b/xrootd.spec
@@ -6,7 +6,7 @@
 %define tag %{realversion}
 %define branch master
 %define github_user xrootd
-Source: git+https://github.com/%github_user/xrootd.git?obj=%{branch}/v%{tag}&export=%{n}-%{realversion}&output=/%{n}-%{realversion}.tgz
+Source: https://xrootd.slac.stanford.edu/download/v%{realversion}/%{name}-%{realversion}.tar.gz
 
 BuildRequires: cmake gmake autotools py3-pip
 Requires: zlib libuuid curl davix
@@ -19,7 +19,7 @@ Requires: libxml2
 %endif
 
 %prep
-%setup -n %n-%{realversion}
+%setup -n %{n}-%{realversion}
 sed -i -e 's|UUID REQUIRED|UUID |' cmake/XRootDFindLibs.cmake
 
 %build
@@ -29,7 +29,6 @@ sed -i -e 's|UUID REQUIRED|UUID |' cmake/XRootDFindLibs.cmake
 rm -rf ../build; mkdir ../build; cd ../build
 cmake ../%n-%{realversion} \
   -DCMAKE_INSTALL_PREFIX=%{i} \
-  -DXRootD_VERSION_STRING=%{realversion} \
   -DCMAKE_BUILD_TYPE=Release \
   -DFORCE_ENABLED=ON \
   -DENABLE_FUSE=FALSE \


xrootd.spec
```diff
@@ -23,25 +23,25 @@ Requires: libxml2
...
-  -DUSER_VERSION=%{realversion} \
+  -DXRootD_VERSION_STRING=%{realversion} \
   -DCMAKE_BUILD_TYPE=Release \
   -DFORCE_ENABLED=ON \
   -DENABLE_FUSE=FALSE \
   -DENABLE_VOMS=FALSE \
   -DXRDCL_ONLY=TRUE \
   -DENABLE_KRB5=TRUE \
   -DENABLE_READLINE=TRUE \
-  -DENABLE_CRYPTO=TRUE \
```
amadio, line 40: This option has been removed. It is always required and enabled now.

xrootd.spec
```diff
@@ -23,25 +23,25 @@ Requires: libxml2
...
   -DCMAKE_SKIP_RPATH=TRUE \
   -DENABLE_PYTHON=TRUE \
   -DENABLE_HTTP=TRUE \
   -DXRD_PYTHON_REQ_VERSION=3 \
+  -DPIP_OPTIONS="--verbose" \
```
amadio, line 43: This is to make it easier to debug build failures of the Python bindings.

aandvalenzuela:
please test

iarspider:
@amadio if I'm reading the error and the code correctly, xrootd wants CMake modules for `libuuid`, which are not present in our libuuid (taken from [`util-linux` package](http://www.kernel.org/pub/linux/utils/util-linux) ).

amadio:
> @amadio if I'm reading the error and the code correctly, xrootd wants CMake modules for `libuuid`, which are not present in our libuuid (taken from [`util-linux` package](http://www.kernel.org/pub/linux/utils/util-linux) ).

Not quite, XRootD has its own `Findlibuuid.cmake` module, which uses pkg-config, but indeed there were changes in this area on how we find libuuid, so there could be bugs needing correction. Can you please place the full log somewhere I can read it? I cannot see the logs by clicking the link. Thank you!

iarspider:
@amadio https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-d7fa3c/33598/externals/xrootd/5.6.0-e003158de589d277e8676c5bf226d77a/log

smuzaffar:
@amadio , looks like the xrootd change https://github.com/xrootd/xrootd/blob/master/cmake/Findlibuuid.cmake#L37-L39 is overriding the UUID variables set by our spec https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_13_2_X/master/xrootd.spec#L46-L47

amadio:
Hi Shahzad, I have a fix for it and will make a patch release soon with it. I will update this pr once it's out.

amadio:
@smuzaffar I've updated XRootD to fix the uuid issue. I think everything should work now. Let's wait for the CI to see for sure.

smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
@amadio looks like root's xrootd cmake module https://github.com/root-project/root/blob/master/cmake/modules/FindXROOTD.cmake is not happy with new version scheme. Previously it was not able to find the correct xrootd version [a] and forcing it to `1000000000` :-) and now `xrdversnum` is set to `5.6.1` [b] which breaks various comparisons statements like https://github.com/root-project/root/blob/master/cmake/modules/FindXROOTD.cmake#L75 


[a]
```
root-6.26.11/cmake/modules/FindXROOTD.cmake(33):  file(STRINGS ${XROOTD_INCLUDE_DIR}/XrdVersion.hh xrdvers REGEX ^#define XrdVERSION  )
root-6.26.11/cmake/modules/FindXROOTD.cmake(34):  string(REGEX REPLACE #define[ ]+XrdVERSION[ ]+  xrdvers ${xrdvers} )
root-6.26.11/cmake/modules/FindXROOTD.cmake(35):  if(${xrdvers} STREQUAL \"unknown\" )
root-6.26.11/cmake/modules/FindXROOTD.cmake(36):  math(EXPR xrdversnum 1000000000 )
root-6.26.11/cmake/modules/FindXROOTD.cmake(52):  if(${xrdversnum} EQUAL 300030000 )
root-6.26.11/cmake/modules/FindXROOTD.cmake(55):  else()
root-6.26.11/cmake/modules/FindXROOTD.cmake(56):  SET(XROOTD_FOUND TRUE )
```

[b]
```
oot-6.26.11/cmake/modules/FindXROOTD.cmake(33):  file(STRINGS ${XROOTD_INCLUDE_DIR}/XrdVersion.hh xrdvers REGEX ^#define XrdVERSION  )
root-6.26.11/cmake/modules/FindXROOTD.cmake(34):  string(REGEX REPLACE #define[ ]+XrdVERSION[ ]+  xrdvers ${xrdvers} )
root-6.26.11/cmake/modules/FindXROOTD.cmake(35):  if(${xrdvers} STREQUAL \"unknown\" )
root-6.26.11/cmake/modules/FindXROOTD.cmake(37):  else()
root-6.26.11/cmake/modules/FindXROOTD.cmake(38):  string(REGEX REPLACE [^v\\.]+  xrdversdots ${xrdvers} )
root-6.26.11/cmake/modules/FindXROOTD.cmake(39):  if(${xrdversdots} STREQUAL v.. )
root-6.26.11/cmake/modules/FindXROOTD.cmake(46):  else()
root-6.26.11/cmake/modules/FindXROOTD.cmake(48):  string(REGEX REPLACE [v\"]  xrdvers ${xrdvers} )
root-6.26.11/cmake/modules/FindXROOTD.cmake(49):  string(SUBSTRING ${xrdvers} 0 8 xrdversnum )
root-6.26.11/cmake/modules/FindXROOTD.cmake(52):  if(${xrdversnum} EQUAL 300030000 )
root-6.26.11/cmake/modules/FindXROOTD.cmake(55):  else()
root-6.26.11/cmake/modules/FindXROOTD.cmake(56):  SET(XROOTD_FOUND TRUE )
root-6.26.11/cmake/modules/FindXROOTD.cmake(60):  if(XROOTD_FOUND )
root-6.26.11/cmake/modules/FindXROOTD.cmake(63):  if(NOT XROOTD_FIND_QUIETLY )
root-6.26.11/cmake/modules/FindXROOTD.cmake(64):  if(${xrdvers} STREQUAL \"unknown\" )
root-6.26.11/cmake/modules/FindXROOTD.cmake(66):  else()
root-6.26.11/cmake/modules/FindXROOTD.cmake(67):  message(STATUS Found Xrootd version num: ${xrdvers} (setting -DROOTXRDVERS=${xrdversnum}) )
-- Found Xrootd version num: 5.6.1 (setting -DROOTXRDVERS=5.6.1)
```

amadio:
Setting the version is something that was fixed in 5.6. Maybe it's time to retire ROOT's FindXROOTD.cmake and use the one provided by XRootD itself. Let's see what they say, I will file a bug when I'm back from vacations.

smuzaffar:
thanks @amadio. I am not sure how far back root can update this. xrootd is one of the package which we do back port to old cmssw releases and I am not sure if root is going to fix it in root 5 release cycle. I would suggest that xrootd should still support old versioning somehow  so that we can easily backport it to old cmssw release cycles

amadio:
Hi @smuzaffar, I'm back from vacations.

> I would suggest that xrootd should still support old versioning somehow so that we can easily backport it to old cmssw release cycles

I'm happy to create a fix that will work for CMS-SW, but I don't know what you mean with support old versioning. We moved from shell scripts to CMake to generate a version, but otherwise things are mostly the same. The problem seems to be that ROOT needed XRootD to not set a version (i.e. use "unknown" as fallback), while now it's properly set. I'd like to avoid reverting this fix in XRootD, as the problem is in ROOT. I think it should not be hard to backport a fix to `FindXROOTD.cmake` to all versions of ROOT that need the change.

smuzaffar:
@amadio , xrootd is special external which we often backport to old cmssw releases. Those old cmssw release cycles use old versions of root e.g CMSSW_10_6_X release, where we recently have backported xrootd 5.5.4, uses root 6.14. So if we have to backport xrootd 5.6.1 to CMSSW_10_6_X then I just want to make sure root 6.14 is able to build against new xrootd. This can be done by either a configure flag to xrootd to support old version so that root build can use it or fix all old versions of root to be able to build with both type of xrootd version schemes.

amadio:
@smuzaffar Ok, I will make sure ROOT 6.14 can build against XRootD 5.6.2, which I plan to release soon. I will close this pull request and open a new one for 5.6.2 when it's out.

smuzaffar:
thanks @amadio . Note that it is not only root 6.14 but all root versions between 6.14 and root master :-) but I hope your fix  for root 6.14 can be applied to other root versions

amadio:
Yes, I think that not much change happened since 6.14, so the fix should be applicable to all versions.

xrootd.spec
```diff
@@ -8,7 +8,7 @@
 %define github_user xrootd
 Source: git+https://github.com/%github_user/xrootd.git?obj=%{branch}/v%{tag}&export=%{n}-%{realversion}&output=/%{n}-%{realversion}.tgz
```
amadio: Please use `Source: https://xrootd.slac.stanford.edu/download/v%{version}/%{name}-%{version}.tar.gz`, which corresponds to the official tarball. I think you may have problems using a dynamically generated tarball from git.

smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
please test for CMSSW_13_3_ROOT6_X

smuzaffar:
please test

smuzaffar:
+externals

looks good

smuzaffar:
@amadio , looks like it fails to find uuid on rhel9 [a]. It worked for rhel8 as the libuuid.so contains the `uuid_generate_random` symbol but for rhel9 the symbol is [b]. Previously things worked as we were setting [UUID_LIBRARY](https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_13_3_X/master/xrootd.spec#L45-L46) but now those values are [overridden](https://github.com/xrootd/xrootd/blob/master/cmake/Findlibuuid.cmake#L37-L39). We need a fix as `5.6.1` version has been merged and rhel9 builds are now failing

[a]
```
-- Looking for uuid_generate_random
-- Looking for uuid_generate_random - not found
-- Looking for uuid_generate_random in uuid
-- Looking for uuid_generate_random in uuid - not found
-- Found PkgConfig: /build/muz/xrd/w/el9_amd64_gcc11/external/autotools/1.5-cb0339419cfc82288971bbef0a7bc503/bin/pkg-config (found version "0.29.2") 
-- Checking for module 'uuid'
--   No package 'uuid' found
CMake Error at /build/muz/xrd/w/el9_amd64_gcc11/external/cmake/3.25.2-ad4f35b5a88cc925c835889f6a6fd837/share/cmake-3.25/Modules/FindPkgConfig.cmake:607 (message):
```

[b]
```
0000000000002aa0 T __uuid_generate_random@@UUIDD_PRIVATE
0000000000002b40 T uuid_generate_random@@UUID_1.0
```

smuzaffar:
for now I can just patch `xrootd` to drop the https://github.com/xrootd/xrootd/blob/master/cmake/Findlibuuid.cmake#L37-L39 section

amadio:
@smuzaffar Sorry about that. Those lines were added in ROOT to always force a search, and I ported the find module into XRootD. I can see that unfortunately that unsets the variables you pass in the command line. I will drop that section in 5.6.2. I didn't notice this problem before because I usually set CMAKE_PREFIX_PATH rather than individual variables. You may be able to do the same in your build to avoid having to patch XRootD.

amadio:
I see that in https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_13_3_X/master/xrootd.spec#L47 you do add `${LIBUUID_ROOT}` to `CMAKE_PREFIX_PATH`. Maybe if you drop the two lines above setting individual variables, it will work. In any case, I will remove the problematic section from the find module in the next release.

smuzaffar:
@amadio , I have dropped the two lines but then it fails with error [a] i.e. `pkgconfig` fails to find our `libuuid`.


[a]
```
-- Looking for uuid_generate_random
-- Looking for uuid_generate_random - not found
-- Looking for uuid_generate_random in uuid
-- Looking for uuid_generate_random in uuid - not found
-- Found PkgConfig: /build/muz/xrd/w/el9_amd64_gcc11/external/autotools/1.5-cb0339419cfc82288971bbef0a7bc503/bin/pkg-config (found version "0.29.2") 
-- Checking for module 'uuid'
--   No package 'uuid' found
CMake Error at /build/muz/xrd/w/el9_amd64_gcc11/external/cmake/3.25.2-ad4f35b5a88cc925c835889f6a6fd837/share/cmake-3.25/Modules/FindPkgConfig.cmake:607 (message):
  A required package was not found
```

smuzaffar:
For now, https://github.com/cms-sw/cmsdist/pull/8661/files works for rhel9 i.e.
- drop https://github.com/xrootd/xrootd/blob/master/cmake/Findlibuuid.cmake#L58 and explicitly set `_have_libuuid` to avoid going in to pkgconfig section
- explicitly set `-DCMAKE_CXX_FLAGS="-I${LIBUUID_ROOT}/include"` and `-DCMAKE_SHARED_LINKER_FLAGS="-L${LIBUUID_ROOT}/lib64"`

amadio:
@smuzaffar I will try to reproduce the problem and make a better solution for it. But it would be good to have access to logs where the problem happens, which I cannot see right now.

smuzaffar:
@amadio , build log is available at https://cmssdt.cern.ch/SDT/jenkins-artifacts/build-any-ib/CMSSW_13_3_X_2023-08-21-2300/el9_amd64_gcc11/160691/external/xrootd/5.6.1-9590c1bd68c0a374fefb81d2f5f8d24d/log but you can also re-produce it on lxplus by running the following commands
```
ssh lxplus
cd /tmp/`whoami`
git clone -b IB/CMSSW_13_3_X/cs9 https://github.com/cms-sw/cmsdist
git clone https://github.com/cms-sw/pkgtools
#start el9 singularity container
/cvmfs/cms.cern.ch/common/cmssw-el9
./pkgtools/cmsBuild --weekly -a el9_amd64_gcc11 -i work -j 10 build xrootd
```
this should fail and points you the logs file e.g. 
```
* The action "build-external+xrootd+5.6.1-864cb0bbe0cafa9507eb8a6bfe0a01bf" was not completed successfully because Failed to build xrootd. Log file in /tmp/muzaffar/work/BUILD/el9_amd64_gcc11/external/xrootd/5.6.1-864cb0bbe0cafa9507eb8a6bfe0a01bf/log. Final lines of the log file:

```

amadio:
I cannot access that build log, I'm no longer in the CMS groups I used to be in. I will try in the container.

amadio:
BTW, if the problem is pkg-config not finding the library, you may need to set `PKG_CONFIG_PATH` if it's in a non-standard location (i.e. so it finds the .pc file needed).

smuzaffar:
the libuuid build in cms does not contain pkgconfig data ( no .pc file) so setting `PKG_CONFIG_PATH` will not help ( you can try setting it in xrootd.spec and see if it works)

amadio:
Hi @smuzaffar,
```
Starting to process package external+xrootd+5.6.1-ea84880c9e20786a9e2337d3897f58c7
Checking if xrootd is cached.
xrootd is available from remote repository "cms.week1". Not building. Downloading and installing...
Package external+xrootd+5.6.1-ea84880c9e20786a9e2337d3897f58c7 found in repository
Nothing else to be done, exiting.
```
How do I force a re-build? I cannot find the option in the help of cmsBuild.

smuzaffar:
@amadio , just add an empty line in cmsdist/xrootd.spec file. This will force cmsBuild to rebuild the package.

iarspider:
> Hi @smuzaffar,
> 
> ```
> Starting to process package external+xrootd+5.6.1-ea84880c9e20786a9e2337d3897f58c7
> Checking if xrootd is cached.
> xrootd is available from remote repository "cms.week1". Not building. Downloading and installing...
> Package external+xrootd+5.6.1-ea84880c9e20786a9e2337d3897f58c7 found in repository
> Nothing else to be done, exiting.
> ```
> 
> How do I force a re-build? I cannot find the option in the help of cmsBuild.

Modify the recipe for xrootd (e.g. by adding a blank line or a comment somewhere).

amadio:
@smuzaffar @iarspider Ah, of course! I should have thought of that! Thank you!

smuzaffar:
I would suggest to remove the work around I added  to make it build for rhel9. So please remove  the following lines from the spec and then try to build
https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_13_3_X/cs9/xrootd.spec#L23
https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_13_3_X/cs9/xrootd.spec#L44-L45

amadio:
@smuzaffar I found out the problem, the [fix](https://github.com/xrootd/xrootd/commit/e53d4c8dd48b497d1f2fcba7238077bde41331b2) was not so easy :-). On macOS, I had to use `set_property` directly with `INTERFACE_SYSTEM_INCLUDE_DIRECTORIES` to avoid messing up the ordering of header includes and break compilation, but that made it such that in your case, the include directory was not added for compilation. Now I fixed this in a different way, by forcing CMake to find framework headers last and using `target_include_directories` in the `uuid::uuid` interface library target, which fixes your case, so when I release 5.6.2, we can apply the patch below and things should just work:
```diff
--- a/xrootd.spec
+++ b/xrootd.spec
@@ -1,4 +1,4 @@
-### RPM external xrootd 5.6.1
+### RPM external xrootd 5.6.2
 ## INITENV +PATH LD_LIBRARY_PATH %i/lib64
 ## INITENV +PATH PYTHON3PATH %{i}/${PYTHON3_LIB_SITE_PACKAGES}
 
@@ -20,7 +20,6 @@ Requires: libxml2
 
 %prep
 %setup -n %{n}-%{realversion}
-sed -i -e 's|^ *check_library_exists("uuid" "uuid_generate_random".*$|set(_have_libuuid True)|' cmake/Findlibuuid.cmake
 
 %build
 # By default xrootd has fuse, krb5, readline, and crypto enabled.
@@ -41,8 +40,6 @@ cmake ../%n-%{realversion} \
   -DENABLE_HTTP=TRUE \
   -DXRD_PYTHON_REQ_VERSION=3 \
   -DPIP_OPTIONS="--verbose" \
-  -DCMAKE_CXX_FLAGS="-I${LIBUUID_ROOT}/include" \
-  -DCMAKE_SHARED_LINKER_FLAGS="-L${LIBUUID_ROOT}/lib64" \
   -DCMAKE_PREFIX_PATH="${ZLIB_ROOT};${PYTHON3_ROOT};${LIBXML2_ROOT};${LIBUUID_ROOT};${CURL_ROOT};${DAVIX_ROOT}"
 
 make %makeprocesses VERBOSE=1
```

PR #8601, opened 2023-07-18
Enable debug mode option for root, geant4, dd4hep and vecgeom

aandvalenzuela:


smuzaffar:
@aandvalenzuela , thinking about it again I think instead of having separate `<package_name>_debug` macros for each package how about doing something like this
- `cmsBuild` can be run with `--define cms_debug_packages=root,geant4,any-other-pkg`
- In cmsdist/rpm-preamble.file we define a helper macro e.g `is_debug_build` which can be called from the package spec file e.g.
```
%if %{is_debug_build}
-DCMAKE_BUILD_TYPE=Debug \
%else
-DCMAKE_BUILD_TYPE=Release \
%endif
```
The helper macro `is_debug_build` should  `echo 1` is the package name  `%n` is in `%{cms_debug_packages}` otherwise it should `echo 0`.

Once this is done then we can make use of `DEBUG_EXTERNALS` cms-bot/config.map parameter and add `DEBUG_EXTERNALS=root` for the IBs for which we want to build root in debug mode. we need to update https://github.com/cms-sw/cms-bot/blob/master/build-cmssw-ib-with-patch#L178-L183  so that it can then pass `--define cms_debug_packages=${DEBUG_EXTERNALS}` to `cmsBuild` options

smuzaffar:
please test

by default ( without `--define cms_debug_packages=root` ) root should be build in `Release` mode

aandvalenzuela:
please test

dd4hep.spec
```diff
@@ -24,7 +24,11 @@ Requires: root boost clhep xerces-c geant4
   -DCMAKE_AR=$(which gcc-ar) \\\
   -DCMAKE_RANLIB=$(which gcc-ranlib) \\\
   -DCMAKE_CXX_STANDARD=17 \\\
+%if %{is_debug_build}
+  -DCMAKE_BUILD_TYPE=Debug \\\
+%else
```
smuzaffar: and this is not working. I would suggest to add [a] above the https://github.com/cms-sw/cmsdist/pull/8601/files#diff-71d575042aa0c5063c7a3b3456d09216a283b3805936d7ee23702fac27a752e9R16 and then just use
```
-DCMAKE_BUILD_TYPE=%{pkg_build_type} \\\
```


[a]
```
%if %{is_debug_build}
%define pkg_build_type Debug
%else
%define pkg_build_type Release
%endif
```

smuzaffar: or may be you can also add the following in rpm-preamble.file and then just use `-DCMAKE_BUILD_TYPE=%{cmake_build_type}` in spec files :-)
```
%if %{is_debug_build}
%define cmake_build_type Debug
%else
%define cmake_build_type Release
%endif
```

aandvalenzuela:
please test

root.spec
```diff
@@ -40,13 +40,8 @@ export CXXFLAGS="${CXXFLAGS} %{arch_build_flags}"
...
   -G Ninja \
-%if %{is_debug_build}
-  -DCMAKE_BUILD_TYPE=Debug \
-  -DLLVM_BUILD_TYPE=Debug \
-%else
-  -DCMAKE_BUILD_TYPE=Release \
-  -DLLVM_BUILD_TYPE=Release \
-%endif
+  -DCMAKE_BUILD_TYPE=%{pkg_build_type} \
+  -DLLVM_BUILD_TYPE=%{pkg_build_type} \
```
smuzaffar: please use `cmake_build_type` every where

aandvalenzuela:
please test

smuzaffar:
+externals
this looks good. As expacted, by default these packages were built in `Release` mode

PR #8622, opened 2023-07-31
SCRAMV2: Do not create CMS_PAtH/etc/scramrc files which are taken care by SCRAmv1

smuzaffar:
Bug fix: if SCRAMV2 is installed before SCRAMV1 (which the default scram used by cmssw) then it creates CMS_PATH/etc/scramrc directory without creating the CMS_PATH/etc/scramrc/links.db file which can avoid scram looking in to extra installations. This change makes sure that CMS_PATH/etc/scramrc  is only creating by SCRAMV1


smuzaffar:
please test

PR #8637, opened 2023-08-04
drop py3-py package which is now part of pytest

smuzaffar:
this should fix the [ReDoS in py library when used with subversion](https://github.com/cms-sw/cmsdist/security/dependabot/200) issue


smuzaffar:
please test

smuzaffar:
+externals

PR #8639, opened 2023-08-07
TF: keep read cpp-standard first

smuzaffar:
If TF cuda is disabled (e.g. in GCC12 branch due to cuda 12) then cpp-standard read via cuda-flag does not set cms_cxx_standard. This change makes sure that  cpp-standard is loaded first


smuzaffar:
please test

PR #8642, opened 2023-08-07
Force check dependencies for provides and requires

aandvalenzuela:
Testing with `oracle` pkg. `AutoReq` to be removed also in [‎`ucx`](https://github.com/aandvalenzuela/cmsdist/blob/a99b9b044d3ca31f8e5c3e3977c6c0c5414b3a53/ucx.spec#L11), [`llvm`](https://github.com/aandvalenzuela/cmsdist/blob/a99b9b044d3ca31f8e5c3e3977c6c0c5414b3a53/llvm.spec#L8), [‎`openmpi`](https://github.com/aandvalenzuela/cmsdist/blob/a99b9b044d3ca31f8e5c3e3977c6c0c5414b3a53/openmpi.spec#L11), [`cuda-compatible-runtime`](https://github.com/aandvalenzuela/cmsdist/blob/a99b9b044d3ca31f8e5c3e3977c6c0c5414b3a53/cuda-compatible-runtime.spec#L8) and [`cuda`](https://github.com/aandvalenzuela/cmsdist/blob/a99b9b044d3ca31f8e5c3e3977c6c0c5414b3a53/cuda.spec#L17).

### DIFF
--- a/cuda.spec
+++ b/cuda.spec
@@ -14,7 +14,6 @@ Source0: https://developer.download.nvidia.com/compute/cuda/%{realversion}/local
 Source0: https://developer.download.nvidia.com/compute/cuda/%{realversion}/local_installers/%{n}_%{realversion}_%{driversversion}_linux_sbsa.run
 %endif
 Requires: python3
-AutoReq: no
 
 %prep
 
--- a/cuda.spec
+++ b/cuda.spec
@@ -14,13 +14,7 @@ Source0: https://developer.download.nvidia.com/compute/cuda/%{realversion}/local
 Source0: https://developer.download.nvidia.com/compute/cuda/%{realversion}/local_installers/%{n}_%{realversion}_%{driversversion}_linux_sbsa.run
 %endif
 Requires: python3
-
-# FIXME: Needed by cuda when using default AutoReq. Provided by glibc, but not by glibc in our image.
-Provides: ld-linux.so.2
-Provides: ld-linux.so.2(GLIBC_2.3)
-Provides: libc.so.6(GLIBC_2.0)
-Provides: libc.so.6(GLIBC_2.1)
-Provides: libc.so.6(GLIBC_2.1.3)
+AutoReq: no
 
 %prep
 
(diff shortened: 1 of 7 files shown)


aandvalenzuela:
test parameters:

- full_cmssw = true

aandvalenzuela:
please test

- Let's test first with our regular image and then with the buildtime image

smuzaffar:
please test

no need to run full_cmssw (change is only to check if oracle package deps are available in the platform seed or not)

aandvalenzuela:
please test

aandvalenzuela:
please test

aandvalenzuela:
please test

aandvalenzuela:
@smuzaffar those libraries are provided by `glibc` that is present in the driver and the container image (runtime).  In this case we should move it to the buildtime image, right? Even if we move it to the buildtime image, do we need to add those `*.so` libraries as provides? I thought they should be already there by `glibc`.

aandvalenzuela:
please test

cuda.spec
```diff
@@ -14,7 +14,13 @@ Source0: https://developer.download.nvidia.com/compute/cuda/%{realversion}/local
...
 %endif
 Requires: python3
-AutoReq: no
+
+# FIXME: Needed by cuda when using default AutoReq. Provided by glibc, but not by glibc in our image.
+Provides: ld-linux.so.2
+Provides: ld-linux.so.2(GLIBC_2.3)
+Provides: libc.so.6(GLIBC_2.0)
+Provides: libc.so.6(GLIBC_2.1)
+Provides: libc.so.6(GLIBC_2.1.3)
```
smuzaffar: looks like there are many system deps for cuda. I would suggest to keep `AutoReq: no` for cuda

aandvalenzuela:
please test

smuzaffar:

smuzaffar:

aandvalenzuela:

smuzaffar:
please test

smuzaffar:

smuzaffar:

smuzaffar:

smuzaffar:
please test for el9_amd64_gcc11

PR #8644, opened 2023-08-08
Fix relocation for py-tensorflow

iarspider:


iarspider:
please test

smuzaffar:
@iarspider , as discussed, this is not going to fix it. Problem is that the files are actually generated via `tensorflow-sources` packages and are packed in a wheel bundle. Relocation works if both  `tensorflow-sources`  and `py3-tensorflow` packages are build in same path. I think then proper fix will be
- In tensorflow-sources package: before creating the wheel,  use `sed` to replace the `%{cmsroot}` with fixed `%{installroot}` for these files.
- In py3-tensorflow: use `sed` again to replace `%{installroot}` with `$CMS_INSTALL_PREFIX` for these files.

Note that make sure that these files exists (e.g. if TF is built without cuda support then some of these files will be missing)

iarspider:
> @iarspider , as discussed, this is not going to fix it. Problem is that the files are actually generated via `tensorflow-sources` packages and are packed in a wheel bundle. Relocation works if both `tensorflow-sources` and `py3-tensorflow` packages are build in same path. I think then proper fix will be
> 
> * In tensorflow-sources package: before creating the wheel,  use `sed` to replace the `%{cmsroot}` with fixed `%{installroot}` for these files.
> 
> * In py3-tensorflow: use `sed` again to replace `%{installroot}` with `$CMS_INSTALL_PREFIX` for these files.
> 
> 
> Note that make sure that these files exists (e.g. if TF is built without cuda support then some of these files will be missing)

Thanks, working on it.

iarspider:
please test

smuzaffar:
this looks very complicated solution. why not just just do

- Define list of files to be checked for relocation
```
%define RelocateFile  \
    lib/python%{cms_python3_major_minor_version}/site-packages/tensorflow/include/external/local_config_cuda/cuda/cuda/cuda_config.h \
         lib/python%{cms_python3_major_minor_version}/site-packages/tensorflow/include/third_party/gpus/cuda/cuda_config.h \
         lib/python%{cms_python3_major_minor_version}/site-packages/tensorflow/python/platform/build_info.py
```

- After installation, relocate files (if exists) to replace all build path to `%{cmsroot}`
```
%define PipPostInstall \
  for f in %{RelocateFile} ; do \
    [ ! -f %{i}/$f ] || sed -i -e 's|/[^ ]/%{cmsplatf}/external/|%{cmsroot}/%{cmsplatf}/external/|' %{i}/$f; \
 done
```

- Relocate at rpm install time
```
%define PipPostPost \
for f in %{RelocateFile} ; do \
  [ ! -f $RPM_INSTALL_PREFIX/$f ] || %{relocateConfig}/$f
done
```

iarspider:
Thanks, I will open a new PR with this approach.

PR #8667, opened 2023-08-25
Add missing rocm deps on zstd

smuzaffar:
fwlite IBs failed due to missing deps. things worked for cmssw as zstd was explicitly installed via `cmssw-tool-conf` and that is why rpm did not complain
```
Singularity> ./common/cmspkg -a el8_amd64_gcc11 env -- rpm -q --requires external+rocm+5.6.0-b3e8e46972b3112a27027ca2f3ea0e1b | grep zstd
libzstd.so.1()(64bit)
```

### DIFF
--- a/rocm.spec
+++ b/rocm.spec
@@ -1,5 +1,4 @@
 ### RPM external rocm 5.6.0
-## NOCOMPILER
 
 %if "%{rhel}" == "7"
 # allow rpm2cpio dependency on the bootstrap bundle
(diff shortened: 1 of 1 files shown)


smuzaffar:
please test

fwyzard:
@smuzaffar should we add also the dependency on zlib and elfutils (from the gcc package) ?

smuzaffar:
@fwyzard , no need to explicitly add those as both of these are indirectly in deps list due to `python3` dependency. I will also remove the `## NOCOMPILER` which is not needed as `gcc` is anyway in dep due to `python3`

smuzaffar:
please test

PR #8676, opened 2023-09-01
Add OpenMP support for ROCm

fwyzard:


fwyzard:
please test

makortel:
Could you add something about the motivation for adding OpenMP support?

fwyzard:
> Could you add something about the motivation for adding OpenMP support?

@VinInn is using it for local tests, I though we could add it in CMSSW.

makortel:
So what does it do exactly? Add an ability to use OpenMP offloading to AMD GPUs? Or use OpenMP as "HIP backend"? Or something else?

fwyzard:
> So what does it do exactly? Add an ability to use OpenMP offloading to AMD GPUs? Or use OpenMP as "HIP backend"? Or something else?

According to https://rocm.docs.amd.com/en/latest/about/compatibility/openmp.html , it should support OpenMP offloading to AMD GPUs.

fwyzard:
Update from @VinInn: the OpenMP packages are required to compile any source that uses `#pragma omp ...` with `hipcc`.

fwyzard:
Superseded by #9050 .

PR #8682, opened 2023-09-05
update monit tools version

nikodemas:
Hi,

thank you for the guidelines, @smuzaffar . If I understand correctly, the version on the first line is then assigned to the `realversion` variable, right?

Changes here:
- update monit tools version
- delete decommissioned service (`ggus_parser`)

FYI @leggerf @brij01 @vkuznet

### DIFF
--- a/cmsmon-tools.spec
+++ b/cmsmon-tools.spec
@@ -1,4 +1,4 @@
-### RPM cms cmsmon-tools 0.5.35
+### RPM cms cmsmon-tools go-0.6.3
 ## NOCOMPILER
 
 %define arch amd64
@@ -11,7 +11,7 @@
 %define heyver 0.0.2
 %define k8s_info_ver 0.0.1
 %define gocurlver 0.0.4
-%define monit_commands monit ggus_parser alert annotationManager nats-sub nats-pub dbs_vm
+%define monit_commands monit alert annotationManager nats-sub nats-pub dbs_vm
 %define common_commands promtool amtool prometheus hey stern trivy k8s_info gocurl
 %define flags -ldflags="-s -w -extldflags -static" -p %{compiling_processes}
 Source0: https://github.com/dmwm/CMSMonitoring/releases/download/%{realversion}/cmsmon-tools.tar.gz
--- a/cmsmon-tools.spec
+++ b/cmsmon-tools.spec
@@ -1,4 +1,4 @@
-### RPM cms cmsmon-tools go-0.6.3
+### RPM cms cmsmon-tools 0.6.3
 ## NOCOMPILER
 
 %define arch amd64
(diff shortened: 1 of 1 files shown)


smuzaffar:
@nikodemas , yes the version on the first line becomes `%{realversion}`. 
By the way, why changing the versioning scheme? Note that there is a script https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_13_3_X/master/cmsmon-tools.spec#L99-L108 which search for the latest version installed and run the cmsmon commands from that versions.

smuzaffar:
by the way, if you really want to change the version scheme then I would suggest to use `0.6.3` in the first line and update https://github.com/cms-sw/cmsdist/pull/8682/files#diff-15398a788af5b8c0e165e8cca9a44fb7637daebb4850bf66571da9512729a925R17 to contain the extra `go-` i.e. it should be something like
```
https://github.com/dmwm/CMSMonitoring/releases/download/go-%{realversion}/cmsmon-tools.tar.gz
```

smuzaffar:
I also noticed that [finding latest version logic](https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_13_3_X/master/cmsmon-tools.spec#L105) will not work when you reach version `0.6.10`

cmsmon-tools.spec
```diff
@@ -1,4 +1,4 @@
-### RPM cms cmsmon-tools 0.5.35
+### RPM cms cmsmon-tools go-0.6.3
```
iarspider: I would suggest putting `go-` part into the URL (line 17), and keep the version purely numeric, but will let @smuzaffar have the final say.

nikodemas:
@smuzaffar @iarspider thank you for your comments.

I think the versioning scheme was changed some time ago when new GitHub pipelines were set up in our repository to trigger the creation of different images based on different tags and together with this release tag scheme got changed.

smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
@nikodemas @vkuznet  , `cmsmon` new versionhas been installed under `/cvmfs/cms-ib.cern.ch/sw/x86_64/week1/cmsmon/` but looks like the `0.6.X` release requires `libc` with `GLIBC_2.34' which is not available on lxplus/lxplus8 [a]. Currently latest cmsmon version is only available in our cms-ib area, if it gets deployed on production cvmfs  (/cvmfs/cms.cern.ch) then it will break all these cmsmon tools for lxplus/lxplus8. Looks like it release was built for el9 only as it works on lxplus9 [b].

We need to fix this as soon as possible if these tools are suppose to work on el7/el8

[a]
```
[muzaffar@lxplus801 ~]$  /cvmfs/cms-ib.cern.ch/sw/x86_64/week1/cmsmon/monit
/cvmfs/cms-ib.cern.ch/sw/x86_64/week1/cmsmon/../el8_amd64_gcc11/cms/cmsmon-tools/0.6.3-6a37f2e77e6891b96b25facb687ba0f9/monit: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by /cvmfs/cms-ib.cern.ch/sw/x86_64/week1/cmsmon/../el8_amd64_gcc11/cms/cmsmon-tools/0.6.3-6a37f2e77e6891b96b25facb687ba0f9/monit)
/cvmfs/cms-ib.cern.ch/sw/x86_64/week1/cmsmon/../el8_amd64_gcc11/cms/cmsmon-tools/0.6.3-6a37f2e77e6891b96b25facb687ba0f9/monit: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by /cvmfs/cms-ib.cern.ch/sw/x86_64/week1/cmsmon/../el8_amd64_gcc11/cms/cmsmon-tools/0.6.3-6a37f2e77e6891b96b25facb687ba0f9/monit)
```

[b]
```
[muzaffar@lxplus901 ~]$ /cvmfs/cms-ib.cern.ch/sw/x86_64/week1/cmsmon/monit
2023/09/08 16:04:24 Please provide valid token
```

PR #8697, opened 2023-09-12
Update monit tools version

nikodemas:
Fix the issue with static executable (see https://github.com/cms-sw/cmsdist/pull/8682#issuecomment-1711731431, https://github.com/dmwm/CMSMonitoring/pull/239 and https://github.com/dmwm/CMSMonitoring/pull/238).

FYI @vkuznet @leggerf @brij01

### DIFF
--- a/cmsmon-tools.spec
+++ b/cmsmon-tools.spec
@@ -2,7 +2,7 @@
 # Version policy: [0-9]+.[0-9]+.[0-9]+ e.g 0.5.9, 0.5.10 etc.    #
 # Finding the latest version depend on this version policy       #
 ##################################################################
-### RPM cms cmsmon-tools 0.6.3
+### RPM cms cmsmon-tools 0.6.5
 ## NOCOMPILER
 
 %define arch amd64
--- a/cmsmon-tools.spec
+++ b/cmsmon-tools.spec
@@ -2,7 +2,7 @@
 # Version policy: [0-9]+.[0-9]+.[0-9]+ e.g 0.5.9, 0.5.10 etc.    #
 # Finding the latest version depend on this version policy       #
 ##################################################################
-### RPM cms cmsmon-tools 0.6.5
+### RPM cms cmsmon-tools 0.6.6
 ## NOCOMPILER
 ## NO_VERSION_SUFFIX
 
--- a/cmsmon-tools.spec
+++ b/cmsmon-tools.spec
@@ -4,6 +4,7 @@
 ##################################################################
 ### RPM cms cmsmon-tools 0.6.5
 ## NOCOMPILER
+## NO_VERSION_SUFFIX
 
 %define arch amd64
 %define linuxarch linux-amd64
@@ -28,7 +29,6 @@ Source6: https://github.com/vkuznet/k8s_info/releases/download/%k8s_info_ver/k8s
 Source7: https://github.com/aquasecurity/trivy/releases/download/v%trivyver/trivy_%{trivyver}_Linux-64bit.tar.gz
 Source8: https://github.com/vkuznet/gocurl/releases/download/%gocurlver/gocurl-tools.tar.gz
 AutoReq: no
-BuildRequires: go
 
 # RPM macros documentation
 # http://www.rpm.org/max-rpm/s1-rpm-inside-macros.html
@@ -45,8 +45,6 @@ BuildRequires: go
 %setup -D -T -b 8 -n gocurl-tools
 
 %build
-export CGO_ENABLED=0
-export GOCACHE=%{_builddir}/gocache
 
 %install
 cd %{_builddir}/cmsmon-tools


smuzaffar:
please test

smuzaffar:
@nikodemas , it still fails on lxplus/lxplus8 [a]
 
[a]
```
[muzaffar@lxplus800 ~]$ /cvmfs/cms-ci.cern.ch/week0/PR_3774b2b4/cmsmon/monit --help
/cvmfs/cms-ci.cern.ch/week0/PR_3774b2b4/cmsmon/../el8_amd64_gcc11/cms/cmsmon-tools/0.6.5/monit: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by /cvmfs/cms-ci.cern.ch/week0/PR_3774b2b4/cmsmon/../el8_amd64_gcc11/cms/cmsmon-tools/0.6.5/monit)
/cvmfs/cms-ci.cern.ch/week0/PR_3774b2b4/cmsmon/../el8_amd64_gcc11/cms/cmsmon-tools/0.6.5/monit: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by /cvmfs/cms-ci.cern.ch/week0/PR_3774b2b4/cmsmon/../el8_amd64_gcc11/cms/cmsmon-tools/0.6.5/monit)
```

smuzaffar:
@vkuznet , the old version `0.5.35` also links against `libc` but that does not fail [a]. I think previously the release was built on CentOS7 node while new version `0.6.X` is built on may be rhel9/alma9?
```
[muzaffar@lxplus800 ~]$ ldd /cvmfs/cms.cern.ch/el8_amd64_gcc11/cms/cmsmon-tools/0.5.35-9b563cabfb89b1b9a4f4444339006648/monit
        linux-vdso.so.1 (0x00007ff91d877000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007ff91d42b000)
        libc.so.6 => /lib64/libc.so.6 (0x00007ff91d066000)
        /lib64/ld-linux-x86-64.so.2 (0x00007ff91d64b000)
```

vkuznet:
@smuzaffar I did not follow how PR was made but I gave instruction to @nikodemas to built static executables. If you still see dependency on GLIC it means that the built still did not produced static executables and @nikodemas should fix that. I think spec file should properly updated to pick up latest release tag which should be in place after CI/CD pipeline in CMSMonitoring. @nikodemas I suggest you manually download release tar ball and test `monit` tool on lxplus. You should be able to run `ldd /path/monit` command and it should says that it is statically linked executable. Once it is the case, you must change release tag and hash(es) (I think) to pick up in a spec which should provide only statically linked executables.

smuzaffar:
@vkuznet , as I wrote earlier, the `libc` dependency was always there even the old version of these tools links against `libc`. So I think problem is not the static build. Looks like it was built on a newer OS (with newer glibc)
[a]
```
[muzaffar@lxplus7125 cmsmon-tools]$ ldd /cvmfs/cms.cern.ch/slc7_amd64_gcc10/cms/cmsmon-tools/0.5.32/monit
        linux-vdso.so.1 =>  (0x00007ffced16c000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f68e608a000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f68e5cbc000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f68e62a6000)
```

vkuznet:
yes, I understand, and my point is that we should build fully static executable. The one you referring is not a static executable as it depends on system libraries, so it is dynamically linked executable (even though it is small subset of system libs). I pointed @nikodemas that there is a way to built fully statically linked executables and he should just find proper set of options to go compiler. For instance, auth-proxy-server release (https://github.com/dmwm/auth-proxy-server/releases/download/0.2.47/auth-proxy-tools_amd64.tar.gz) provides such executables and I built them using this [Makefile](https://github.com/dmwm/auth-proxy-server/blob/master/Makefile). Here is a proof:
```
# on lxplus
curl -ksLO https://github.com/dmwm/auth-proxy-server/releases/download/0.2.47/auth-proxy-tools_amd64.tar.gz
tar xfz auth-proxy-tools_amd64.tar.gz
ldd auth-proxy-tools_amd64/auth-proxy-server
        not a dynamic executable
```
So, it is a question of flags for specific compiler version. Since I do not know now which OS and GoLang compiler is used in CMS Monitoring I can't say much, but it is possible to build pure statically linked executable and I provided all examples.

nikodemas:
Hi,

thank you for your help @smuzaffar @vkuznet. I have managed to build a static executable by adding `CGO_ENABLED=0` before the `go build` command (as was done in Valentin's `make` files), have tested it out by downloading the `tar` files to the all versions of `lxplus` and recreated the release. Is it ok if I leave the same version number (`0.6.5`)?

smuzaffar:
@nikodemas , I would suggest to update the version.

smuzaffar:
please test

smuzaffar:
+externals

PR #8712, opened 2023-09-21
[cms-common] Added scram site runtime hook to add gsl/bin in to PATH

smuzaffar:
For old releases, CMS distributed `external/gsl/bin` is not added in the PATH which means `gsl-config` is picked up from system while using `libgsl.so` from cms externals. Grid packs makes use of `gsl-config` and fails ( see https://github.com/cms-sw/cmssw/issues/42716 for datails). This PR should fix the old release cycles and already built releases as it adds a `SCRAM` site based runtime hook to add CMS distributed `gsl/bin` in to PATH.

### DIFF
--- a/cms-common.spec
+++ b/cms-common.spec
@@ -1,8 +1,8 @@
 ### RPM cms cms-common 1.0
-## REVISION 1229
+## REVISION 1230
 ## NOCOMPILER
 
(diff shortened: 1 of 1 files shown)


smuzaffar:
please test

smuzaffar:
+externals

PR #8755, opened 2023-10-13
Disable LTO for ASAN builds

aandvalenzuela:
To test if new ASAN errors are related to LTO, as requested in https://github.com/cms-sw/cmssw/issues/43003.

### DIFF
--- a/compilation_flags_lto.file
+++ b/compilation_flags_lto.file
@@ -1,11 +1,19 @@
 #Enable/disable LTO builds
 %define enable_lto 1
 
-#LTO related flags
+#Disable LTO for ppc64le
 %ifarch ppc64le
 %undefine enable_lto
 %define enable_lto 0
 %endif
+
+#Disable LTO for ASAN
+%if "%(case %realversion in (*_ASAN_X*) echo true ;; (*) echo false ;; esac)" == "true"
+%undefine enable_lto
+%define enable_lto 0
+%endif
+
+#LTO related flags
 %if "%{enable_lto}" == "1"
 %define lto_build_flags -flto -fipa-icf -flto-odr-type-merging -fno-fat-lto-objects -Wodr
 %else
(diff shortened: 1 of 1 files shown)


compilation_flags_lto.file
```diff
@@ -1,11 +1,19 @@
...
 %undefine enable_lto
 %define enable_lto 0
 %endif
+
+#Disable LTO for ASAN
+%if "%(case %realversion in (*_ASAN_X*) echo true ;; (*) echo false ;; esac)" == "true"
+%undefine enable_lto
+%define enable_lto 0
+%endif
+
```
smuzaffar: @aandvalenzuela , this is not going to work. this file is not directly used by cmssw so `realversion` will not container `ASAN`. I would suggest to just change `%define enable_lto 1` to `%define enable_lto 0` on the 2nd line of this file and try to build full cmssw for ASAN

aandvalenzuela: Ah, nice! In fact, I was a bit unsure of how to filter based on flavor in this file. Do I make the change, trigger the IB, and revert the change?

smuzaffar: No, no need to build the IB. Just the PR tests should be enough to see if LTO is causing any issues for ASAN

aandvalenzuela:
please test for CMSSW_13_3_ASAN_X
- full_cmssw = true

smuzaffar:
test parameters:
- full_cmssw = true

smuzaffar:
> please test for CMSSW_13_3_ASAN_X
> 
>     * full_cmssw = true

this is not going to work. `full_cmssw = true` should be under `test parameters`

smuzaffar:
please test for CMSSW_13_3_ASAN_X

aandvalenzuela:
> > please test for CMSSW_13_3_ASAN_X
> > ```
> > * full_cmssw = true
> > ```
> 
> this is not going to work. `full_cmssw = true` should be under `test parameters`

My bad, sorry!

PR #8765, opened 2023-10-17
Update GitPython to version 3.1.37

smuzaffar:
New version contains the `Blind local file inclusion` issue
```
In order to resolve some git references, GitPython reads files from the .git directory, 
in some places the name of the file being read is provided by the user, GitPython 
doesn't check if this file is located outside the .git directory. This allows an attacker 
to make GitPython read any file from the system.
```


smuzaffar:
please test

smuzaffar:
+externals

PR #8780, opened 2023-10-23
Add conifer external

thesps:
This PR adds a new external `conifer` that is to be used for emulation of BDTs in the L1T. We just need to pull in a single header file ([this one](https://github.com/thesps/conifer/blob/master/conifer/backends/cpp/include/conifer.h)).

This is my first time adding an external, so to check: the header file itself does `#include "nlohmann/json.hpp"`, which is already an external. Do I need to add a requirement on `json`?

### DIFF
--- a/scram-tools.file/tools/conifer/conifer.xml
+++ b/scram-tools.file/tools/conifer/conifer.xml
@@ -1,7 +1,8 @@
 <tool name="conifer" version="@TOOL_VERSION@">
   <client>
     <environment name="CONIFER_BASE" default="@TOOL_ROOT@"/>
-    <environment name="INCLUDE" default="$CONIFER_BASE/conifer/backends/cpp/include"/>
+    <environment name="INCLUDE" default="$CONIFER_BASE/include"/>
   </client>
+  <use name="json"/>
 </tool>
 
--- a/scram-tools.file/tools/conifer/conifer.xml
+++ b/scram-tools.file/tools/conifer/conifer.xml
@@ -1,0 +1,7 @@
+<tool name="conifer" version="@TOOL_VERSION@">
+  <client>
+    <environment name="CONIFER_BASE" default="@TOOL_ROOT@"/>
+    <environment name="INCLUDE" default="$CONIFER_BASE/conifer/backends/cpp/include"/>
+  </client>
+</tool>
+
--- a/cmssw-tool-conf.spec
+++ b/cmssw-tool-conf.spec
@@ -17,6 +17,7 @@ Requires: bz2lib
 Requires: cepgen
 Requires: classlib
 Requires: clhep
+Requires: conifer
 Requires: coral
 Requires: cppunit
 Requires: cpu_features
--- a/conifer.spec
+++ b/conifer.spec
@@ -1,0 +1,16 @@
+### RPM external conifer 1.2
+## NOCOMPILER
+
+Source: https://github.com/thesps/%{n}/archive/v%{realversion}.tar.gz
+
+%prep
+%setup -n %{n}-%{realversion}
+
+%build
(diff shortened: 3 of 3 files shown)


iarspider:
> Do I need to add a requirement on `json`?

Yes

thesps:
Thanks, done.

iarspider:
@cmsbuild please test

scram-tools.file/tools/conifer/conifer.xml
```diff
@@ -0,0 +1,7 @@
+<tool name="conifer" version="@TOOL_VERSION@">
+  <client>
+    <environment name="CONIFER_BASE" default="@TOOL_ROOT@"/>
+    <environment name="INCLUDE" default="$CONIFER_BASE/conifer/backends/cpp/include"/>
```
smuzaffar: please change `$CONIFER_BASE/conifer/backends/cpp/include`  to `$CONIFER_BASE/include` ( as you copied the header in `%{i}/include` ).

Also please add `<use name="json"/>` after `</client>`

tomalin:
Shouldn't this PR also delete /L1Trigger/Phase2L1ParticleFlow/interface/conifer.h and migrate the small amount of PF code that refers to this? Or will that be done in a separate PR?

smuzaffar:
@thesps , please also include `conifer` as dependency of [cmssw-tool-conf](https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_13_3_X/master/cmssw-tool-conf.spec)  otherwise it will not be part of cmssw.

smuzaffar:
> Shouldn't this PR also delete /L1Trigger/Phase2L1ParticleFlow/interface/conifer.h and migrate the small amount of PF code that refers to this? Or will that be done in a separate PR?

@tomalin , not this PR but yes there should be a cmssw PR to delete L1Trigger/Phase2L1ParticleFlow/interface/conifer.h and start using `conifer` external

smuzaffar:
please test

(just to make sure external is built properly)

smuzaffar:
+externals

This looks good to go in. @thesps , is cmssw PR ready to make use of this change?

smuzaffar:
I see https://github.com/cms-sw/cmssw/pull/42663 waiting for new external. 
Let get this in for next IB

thesps:
Okay now having tried this from the IB last night `CMSSW_13_3_X_2023-10-25-2300`, changing an include in CMSSW to `#include "conifer.h"` there's a compile error from the conifer header on `error: comparison of integer expressions of different signedness` and `-Werror=sign-compare`.
What's the best fix:
- change this code in the conifer library itself, make a release, update the spec here to point to new relase
- suppress the error (not sure where)
?

At least we picked up the external okay...

smuzaffar:
>  change this code in the conifer library itself, make a release, update the spec here to point to new relase 

yes, it should be fixed upstream in `conifer`

thesps:
I'll do this ASAP (but probably in about 2 hours from now due to other commitments). Then, after fixing and updating here, I would like to use the `User test area` setup from the CI (as described [here](https://github.com/cms-sw/cmsdist/pull/8780#issuecomment-1775688181)). Is there a way to use the externals from CI but a 'normal'/master/recent-release CMSSW area with sparse checkout? I would like to compile and run from just one package/sub-package to see if the new external is working okay.

smuzaffar:
In your cmssw dev area (based on one of the latest IBs), you just need to update `CONIFER_BASE` in  `config/toolbox/$SCRAM_ARCH/tools/selected/conifer.xml` to point to your local installation of `conifer` and run `scram setup conifer`. This will allow scram to pick the `conifer.h` from your local directory.

PR #8789, opened 2023-10-27
[Geant4] G4TransportationWithMsc: Always update momentum direction

### LINKED ISSUE #8783: Backport bug fix for Geant4 11.1
We need to backport a fix to Geant4 11.1.2:
```patch
diff --git a/source/processes/electromagnetic/utils/src/G4TransportationWithMsc.cc b/source/processes/electromagnetic/utils/src/G4TransportationWithMsc.cc
index bf828c8b7ec..601121913b9 100644
--- a/source/processes/electromagnetic/utils/src/G4TransportationWithMsc.cc
+++ b/source/processes/electromagnetic/utils/src/G4TransportationWithMsc.cc
@@ -372,8 +372,7 @@ G4double G4TransportationWithMsc::AlongStepGetPhysicalInteractionLength(
             // 2) Call SampleScattering(), which *may* change it.
             const G4ThreeVector displacement =
               mscModel->SampleScattering(fTransportEndMomentumDir, minSafety);
-            // 3) Get the changed direction and inform G4Transportation.
-            fMomentumChanged         = true;
+            // 3) Get the changed direction.
             fTransportEndMomentumDir = *fParticleChangeForMSC->GetProposedMomentumDirection();
 
             const G4double r2 = displacement.mag2();
@@ -469,6 +468,12 @@ G4double G4TransportationWithMsc::AlongStepGetPhysicalInteractionLength(
 
         fParticleChange.ProposeTrueStepLength(totalTruePathLength);
 
+        // Inform G4Transportation that the momentum might have changed due
+        // to scattering. We do this unconditionally to avoid the situation
+        // where the last step is done without MSC and G4Transportation reset
+        // the flag, for example when running without field.
+        fMomentumChanged = true;
+
         return totalGeometryStepLength;
       }
     }
```

Since commit 8d156a75988a99af0fdeab074851c4bace99b580 cmsdist is using the sources from https://github.com/Geant4/geant4. Looking into past discussions, we cannot include patches in such setup. Can the `geant4.spec` be changed to build from https://github.com/cms-externals/geant4/tree/cms/v11.1.2? Then I will open a PR in that repo to apply the above change.

cc @smuzaffar @civanch

smuzaffar:
Fixes https://github.com/cms-sw/cmsdist/issues/8783
see https://github.com/cms-externals/geant4/pull/76

### DIFF
--- a/geant4.spec
+++ b/geant4.spec
@@ -2,9 +2,9 @@
 ## INCLUDE compilation_flags
 ## INCLUDE compilation_flags_lto
 ## INCLUDE cpp-standard
-%define tag cf4c6712c60cdf2b4d8799e34e176917c2de358d
-%define branch master
-%define github_user Geant4
+%define tag 2d5992376773d172321f01710a82943ead5bb435
+%define branch cms/v%{realversion}
+%define github_user cms-externals
 Source: git+https://github.com/%github_user/%{n}.git?obj=%{branch}/%{tag}&export=%{n}.%{realversion}&output=/%{n}.%{realversion}-%{tag}.tgz
 
 BuildRequires: cmake gmake


smuzaffar:
please test

smuzaffar:
please test

lets re-run based on latest IB

smuzaffar:
+externals

PR #8794, opened 2023-10-30
Add fix for segfault in Min Bias

Dominic-Stafford:
A segfault was reported for generating Min Bias events with Herwig. The root cause seems to be that both the "read" and "run" steps are run in a single job (whereas in standalone Herwig they are run in two different jobs). This leads to issues as two static pointers are made to an object which is deleted and recreated between the steps, removing the static keyword seems to resolve the issue (tested in CMSSW_10_6). Herwig authors agree this seems a reasonable fix, and they plan to include it in future versions of Herwig, this patch resolves the issue in CMSSW in the meantime

### DIFF
--- a/herwig7.spec
+++ b/herwig7.spec
@@ -18,12 +18,14 @@ BuildRequires: autotools
 Patch0: herwig_Matchbox_mg_py3
 Patch1: herwig7-fxfx-fix
 Patch2: LHEEventNumFxFx
+Patch3: herwig_MB
 
 %prep
 %setup -q -n Herwig-%{realversion}
 %patch0 -p1
 %patch1 -p1
 %patch2 -p1
+%patch3 -p1
 
 # Regenerate build scripts
 autoreconf -fiv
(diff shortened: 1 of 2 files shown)


smuzaffar:
please test

smuzaffar:
+externals

smuzaffar:
assign generators

@cms-sw/generators-l2 can you please review and approve this change?

menglu21:
+1

smuzaffar:
@cms-sw/orp-l2 , this is ready to go in 13.3.X, feel free to sign it

smuzaffar:
@cms-sw/generators-l2 @cms-sw/orp-l2 , what about this fix? should this go in 13.3.X?

antoniovilela:
> @cms-sw/generators-l2 @cms-sw/orp-l2 , what about this fix? should this go in 13.3.X?

Yes, sorry for the delay.

antoniovilela:
+1

PR #8849, opened 2023-12-01
Update crab-dev.spec

belforte:
patch CRABClient/bin/crab to use python2 for CMSSW_10_1

### DIFF
--- a/crab-dev.spec
+++ b/crab-dev.spec
@@ -3,7 +3,7 @@
 #For any other change, increment version_suffix
 ##########################################
 %define version_suffix 00
-%define crabclient_version v3.231030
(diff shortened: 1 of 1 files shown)


belforte:
@smuzaffar sorry to bug. Pinging for a "please test"

smuzaffar:
@belforte , do you really want to break the version convention? Although `v3.231030.p1` will work and https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_14_0_X/master/crab.spec#L54 should properly prefer  `v3.231030.p1.00` over `v3.231030.00` but if possible can we stick to original version scheme?

belforte:
I see. Let me use a standard tag name then !

PR #8850, opened 2023-12-01
crab-setup - support zsh login shell

mapellidario:
### problem

Current `crab-setup.sh` does not support zsh on login shells.

### description

When opening a shell on lxplus, the shell is of type "login", so 

```plaintext
> [15:24] lxplus909 ~
> ps -p$$ -ocmd=
-zsh
```

which makes the check ` [ "$(ps -p$$ -ocmd=)" = "zsh" ]` in `crab-setup.sh` to return false, and

```plaintext
autoload -U +X compinit && compinit
autoload -U +X bashcompinit && bashcompinit
```

are not executed, so zsh does not know about the function `complete`, and we ultimately can not import bash completions into zsh.

### solution

change the condition to match also zsh login shell, something like the following should do the job.

```diff
- [ "$(ps -p$$ -ocmd=)" = "zsh" ]
+ [[ "$(ps -p$$ -ocmd=)" = *"zsh" ]]
```

I tested it and it works.

fyi @belforte @novicecpp , I know you use bash, but maybe a user asks for it

### DIFF
--- a/crab/crab-setup.sh.file
+++ b/crab/crab-setup.sh.file
@@ -12,7 +12,7 @@ esac
 export CRABCLIENT_TYPE
 crab_shared_dir="@CMS_PATH@/share/cms/crab/@CRAB_COMMON_VERSION@"
 export PYTHONPATH="${crab_shared_dir}/lib/${CRABCLIENT_TYPE}${PYTHONPATH:+:$PYTHONPATH}"
-if [ "$(ps -p$$ -ocmd=)" = "zsh" ] ; then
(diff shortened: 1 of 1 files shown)


belforte:
I am not against this. Looks harmless to users of other shells !

smuzaffar:
@mapellidario , please also increment the `CMSDIST_FILE_REVISION` at line number 2. That will allow system to deploy this new file

mapellidario:
Ooops sorry i forgot, thanks for the reminder!

smuzaffar:
please test

smuzaffar:
+externals
looks good

PR #8856, opened 2023-12-04
save crab-prod as crab-pre, then bump to v3.231101

belforte:
after v3.231101 client was validated on all CMSSW release series in our list https://github.com/dmwm/CRABClient/issues/5271


smuzaffar:
please test

smuzaffar:
+externals

PR #8888, opened 2023-12-18
Ul pythi8310

menglu21:
bump pythia to 8310 to be able to use correct FxFx for top samples, especially for ttW process


stahlleiton:
@menglu21 : Did you managed to increase the version of PYTHIA to 8.310?
There are new functionalities for heavy ions that are only present in PYTHIA 8.308 or beyond, which I would be interested in using in CMSSW_13_0_X.

PR #8892, opened 2023-12-18
Update ROOT 6.14 to include forward-compatibility patch

nothingface0:
Update ROOT 6.14 ref to include forward-compatibility patch for root files created by versions >=6.30.

Discussion [here](https://github.com/cms-sw/cmssw/issues/43590).

Initial issue [here](https://github.com/cms-DQM/dqmgui_prod/issues/18)

@smuzaffar Are there any other changes that need to be included?

### DIFF
--- a/root.spec
+++ b/root.spec
@@ -2,7 +2,7 @@
 ## INITENV +PATH PYTHONPATH %{i}/lib
 ## INITENV SET ROOTSYS %{i}
 %define tag c99f334c8d6bd505f4f0e453dcc654a938020e10
-%define branch cms/v6-14-00-patches/94610d9
+%define branch cms/v6-14-00-patches/65ed49a
 %define github_user cms-sw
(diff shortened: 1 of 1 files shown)


root.spec
```diff
@@ -2,7 +2,7 @@
 ## INITENV +PATH PYTHONPATH %{i}/lib
 ## INITENV SET ROOTSYS %{i}
 %define tag c99f334c8d6bd505f4f0e453dcc654a938020e10
-%define branch cms/v6-14-00-patches/94610d9
+%define branch cms/v6-14-00-patches/65ed49a
```
smuzaffar, line 5: this will not work. We do not have  cms/v6-14-00-patches/65ed49a branch in https://github.com/cms-sw/root/. Let me open a PR with the proper fix

dpapagiannis: Thanks, closing this.

PR #8905, opened 2023-12-20
Update Intel oneAPI compiler to version 2024.0

fwyzard:


fwyzard:
please test

fwyzard:
@smuzaffar do we have anything in CMSDIST or CMSSW that actually tests the Intel compiler ?

smuzaffar:
> @smuzaffar do we have anything in CMSDIST or CMSSW that actually tests the Intel compiler ?

No, I don't think any thing in cmssw depend on it

fwyzard:
OK, so the update cannot break anything... on the other hand, we won't know if it works until we try to use it.
Maybe next year we can try to add some tests.

smuzaffar:
+externals

PR #8920, opened 2024-01-08
celeritas: Disable OpenMP and fix the lib link order

smuzaffar:


smuzaffar:
please test with cms-sw/cmssw#43676

smuzaffar:
+externals

`celeritas` external was properly built. Using it in cmssw (via https://github.com/cms-sw/cmssw/pull/43676 ) did not show any compilation or linking errors

PR #8935, opened 2024-01-10
Triton model versioning

kpedro88:
This PR is kept as a draft until the cms-data updates are integrated (see https://github.com/cms-sw/cmssw/issues/43695).

This PR provides a tool to automatically check and update model checksums in config.pbtxt files. This tool is installed in CMSSW for general use. A small patch is applied to protobuf to make the tool work more nicely. The tool also uses the Triton python client, which is added as a Python package external.

### DIFF
--- a/data/cmsTritonChecksumTool
+++ b/data/cmsTritonChecksumTool
@@ -6,35 +6,6 @@ from argparse import ArgumentParser, ArgumentDefaultsHelpFormatter
 from google.protobuf import text_format, message, descriptor
 from tritonclient import grpc
 
-# patched version for more succinct printing of repeated messages
-def PrintMessage(self, message):
-    if self.message_formatter and self._TryCustomFormatMessage(message):
-        return
-    if (message.DESCRIPTOR.full_name == text_format._ANY_FULL_TYPE_NAME and
-            self._TryPrintAsAnyMessage(message)):
-        return
-    fields = message.ListFields()
-    if self.use_index_order:
-        fields.sort(key=lambda x: x[0].number if x[0].is_extension else x[0].index)
-    for field, value in fields:
-        if text_format._IsMapEntry(field):
-            for key in sorted(value):
-                entry_submsg = value.GetEntryClass()(key=key, value=value[key])
-                self.PrintField(field, entry_submsg)
-        elif field.label == descriptor.FieldDescriptor.LABEL_REPEATED:
-            if self.use_short_repeated_primitives: # no other conditions here
-                self._PrintShortRepeatedPrimitivesValue(field, value)
-            else:
-                for element in value:
-                    self.PrintField(field, element)
-        else:
-            self.PrintField(field, value)
-
-    if self.print_unknown_fields:
-        self._PrintUnknownFields(unknown_fields.UnknownFieldSet(message))
-
-text_format._Printer.PrintMessage = PrintMessage
-
 def get_checksum(filename, chunksize=4096):
     import hashlib
     with open(filename, 'rb') as f:
--- a/data/cmsTritonChecksumTool
+++ b/data/cmsTritonChecksumTool
@@ -1,0 +1,133 @@
+#!/usr/bin/env python3
+
+import os, sys
+from collections import OrderedDict
+from argparse import ArgumentParser, ArgumentDefaultsHelpFormatter
+from google.protobuf import text_format, message, descriptor
+from tritonclient import grpc
+
+# patched version for more succinct printing of repeated messages
+def PrintMessage(self, message):
+    if self.message_formatter and self._TryCustomFormatMessage(message):
+        return
+    if (message.DESCRIPTOR.full_name == text_format._ANY_FULL_TYPE_NAME and
+            self._TryPrintAsAnyMessage(message)):
+        return
+    fields = message.ListFields()
+    if self.use_index_order:
+        fields.sort(key=lambda x: x[0].number if x[0].is_extension else x[0].index)
+    for field, value in fields:
+        if text_format._IsMapEntry(field):
+            for key in sorted(value):
+                entry_submsg = value.GetEntryClass()(key=key, value=value[key])
+                self.PrintField(field, entry_submsg)
+        elif field.label == descriptor.FieldDescriptor.LABEL_REPEATED:
+            if self.use_short_repeated_primitives: # no other conditions here
+                self._PrintShortRepeatedPrimitivesValue(field, value)
+            else:
+                for element in value:
+                    self.PrintField(field, element)
+        else:
+            self.PrintField(field, value)
+
+    if self.print_unknown_fields:
+        self._PrintUnknownFields(unknown_fields.UnknownFieldSet(message))
+
+text_format._Printer.PrintMessage = PrintMessage
+
+def get_checksum(filename, chunksize=4096):
+    import hashlib
+    with open(filename, 'rb') as f:
+        file_hash = hashlib.md5()
+        while chunk := f.read(chunksize):
+            file_hash.update(chunk)
+    return file_hash.hexdigest()
+
+def update_config(args):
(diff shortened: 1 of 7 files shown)


scram-project-build.file
```diff
@@ -312,6 +312,9 @@ find external/%cmsplatf -type l | xargs ls -l
 %{?PatchReleaseSymlinkRelocate:%PatchReleaseSymlinkRelocate}
 echo "%{cmsroot}" > %{i}/config/scram_basedir
 
+# install checksum tool for local use (via scram)
+cp data/cmsTritonChecksumTool %i/bin/
```
kpedro88, line 316: @smuzaffar is this the right way to install the tool into CMSSW? Running a full build to test this is a bit beyond my current resources...

smuzaffar: @kpedro88 , `scram-project-build` is used to build scram based project e.g. coral/cmssw etc. If this tool need to go in to cmssw/bin directory then why not add it under some package in cms-sw/cmssw repo?

data/cmsTritonChecksumTool
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env python3
```
smuzaffar, line 1: @kpedro88 , I think it is better to add this in to cmssw repo

smuzaffar: ah OK, this is need for cms-data externals. So yes it sould not go in to cmssw

scram-project-build.file
smuzaffar: ah ok @kpedro88 , after reading https://github.com/cms-sw/cmssw/issues/43695 I remember where this tool is used. It should be copied here in scram-project-build.file. 

This tool is needed by cms-data externals. Can you please remind me how to run this for a cms-data repo? All cms-data external use https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_14_0_X/master/cmsswdata-github.file . So for selected cms-data externals (where we want to run it), we need to create a file e.g. `cmsdist/data/data-HeterogeneousCore-SonicTriton.file` with contents like
```
%define PostBuild %{cmsdist_directory}/data/cmsTritonChecksumTool <args>
```

so that ` %{cmsdist_directory}/data/cmsTritonChecksumTool ` can be run [here](https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_14_0_X/master/cmsswdata-github.file#L30)

kpedro88: Yes, it needs to be here so it can run on cms-data packages during build time, and will then be copied into CMSSW. If there's a better place to execute that copy operation, I'm happy to move it.

My untested attempt to execute the tool during build time is here: (not in this PR, because I want to have everything in place before we start enforcing it)
https://github.com/kpedro88/cmsdist/commit/44f13f27f211d3a0e737e321a4696e8a26e75f32

This will be a no-op for most cms-data packages that don't have any config.pbtxt files, but using `find` avoids having to maintain a manual list (which is just one more thing to synchronize, i.e. one more potential source of failures when people forget to synchronize it).

smuzaffar: thanks @kpedro88 , about https://github.com/kpedro88/cmsdist/commit/44f13f27f211d3a0e737e321a4696e8a26e75f32, we should not modify the content of cmsswdata-github.file file otherwise it will rebuild all the data externals. We can add
```
%define PostBuild for CONFIG in $(find . -name config.pbtxt); do %{cmsdist_directory}/data/cmsTritonChecksumTool -c $CONFIG; done
```

in the `cmsdist/data/data-<Package-Name>.file` for only those data externals which contain `config.pbtxt`

kpedro88: My concern, though, is that this is an additional manual step that has to be taken any time a Triton model is added to a (new or existing) cms-data package. This build-time check should be automatic. Rebuilding all the data externals would only happen once, right? Is that such a big problem?

Should I move the line from the "Prep" step to the "Build" step either way?

smuzaffar: yes rebuilding all data means re-distributing 20GB of data again. Although this is not a concern for CVMFS distribution but for online where cmssw is install on local disk this can cause issues.

Adding new data external is automated, bot know how to add a new data external in cmsdist. We can teach bot to create this cmsdist/data/data-<Package-Name>.file for new data external if it contains config.pbtxt

smuzaffar: @kpedro88 , see https://github.com/cms-sw/cms-bot/pull/2143

pip/requirements.txt
```diff
@@ -359,6 +359,8 @@ toolz==0.7.1
 tornado==6.3.3
 tqdm==4.65.0
 traitlets==5.9.0
+# always sync version number with triton-inference-client.spec (C++ version)
+tritonclient==2.25.0
```
smuzaffar, line 363: @kpedro88 , this alone is not going to include python tritonclient in cmssw. Please add `py3-tritonclient` dependency in https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_14_0_X/master/python_tools.spec .

scram-project-build.file
kpedro88: @smuzaffar thanks! This looks like a good approach that satisfies both of our concerns.

A few questions:
1. the call to `cmsTritonChecksumTool -c $CONFIG` will go in the file `data/cmsTritonPostBuild.file`, once we're ready to enable it?
2. will this automatically propagate to the existing cms-data repos that contain config.pbtxt files, or do we have to add the `data-X-Y.file` files for those manually in cmsdist?

smuzaffar: > 1. the call to cmsTritonChecksumTool -c $CONFIG will go in the file data/cmsTritonPostBuild.file, once we're ready to enable it?

Yes

> will this automatically propagate to the existing cms-data repos that contain config.pbtxt files, or do we have to add the data-X-Y.file files for those manually in cmsdist?

No , not by default. `data-X-Y.file` will be created only when those data repos opens a new PR.

smuzaffar:
@kpedro88 , I have tried to build locally and I think this is not the right approch.

1. Which this approch the data packages start depending on many externals (e.g py3-tritonclient py3-protobuf which brings in other externals) and will get rebuild any time one of these externals gets updated.
2.  py3-tritonclient from upstream is only available in wheel form ( no sources) and has cuda dependency. We should enable the PYTHON_GRPC flag in our triton client spec to get the python triton client instead of using prebuild wheel

Why not we just add a unit test in cmssw which runs `cmsTritonChecksumTool` by searching all `config.pbtxt` files under `CMSSW_SEARCHPATH` . We can make sure that unit tests is run during PR tests whenever any cmsswdata is changed. This way we do not rebuild data externals due to change in tools needed by `cmsTritonChecksumTool` .

kpedro88:
@smuzaffar I agree that a unit test would be conceptually simpler. Originally I wanted to do it this way, but I didn't know how to make the unit test (which usually only runs when the CMSSW package itself is modified) run whenever someone makes a cms-data PR. But since we have to update cms-bot anyway, just incorporating a CMSSW unit test into the bot's PR tests (which is already done for a few other cases, I think) is probably simpler. I will work on rearranging things next week. That will also simplify this PR, since the checksum tool can be reabsorbed into the general config tool in CMSSW.

I will try to set up the Triton python client through our existing C++ external. I think there was some reason I hadn't done that originally, but hopefully it works. (pip installing in CMSSW via scram-venv worked fine for me when developing these scripts, but I suppose that's because I was using a CUDA-enabled CMSSW build.)

smuzaffar:
@kpedro88 , https://github.com/cms-sw/cmsdist/pull/8943 should build the triton-client python interface

kpedro88:
superseded by https://github.com/cms-sw/cmsdist/pull/8943, https://github.com/cms-sw/cmsdist/pull/8947, https://github.com/cms-sw/cmssw/pull/43696 in the new plan

PR #8939, opened 2024-01-15
PGO: Use env variables with fprofile-use instead of %q{VAR}

smuzaffar:


smuzaffar:
test parameters:
- full_cmssw = true

smuzaffar:
please test

PR #8943, opened 2024-01-16
Enable triton-client python interface

smuzaffar:
This enables the triton client python interface. Code like [a] runs in cmssw env with this change.

[a]
```
bash-4.4$ python3
Python 3.9.14 (main, Jun 21 2023, 12:13:52) 
[GCC 12.3.1 20230527] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from tritonclient import grpc
>>> from google.protobuf import text_format, message, descriptor
>>> model = grpc.model_config_pb2.ModelConfig()
>>> config='/cvmfs/cms.cern.ch/share/cms/data-HeterogeneousCore-SonicTriton/V00-02-00/HeterogeneousCore/SonicTriton/data/models/gat_test/config.pbtxt'
>>> with open(config,'r') as infile:
...   text_format.Parse(infile.read(), model)
... 
name: "gat_test"
platform: "pytorch_libtorch"
input {
  name: "x__0"
  data_type: TYPE_FP32
  dims: -1
  dims: 1433
}
input {
  name: "edgeindex__1"
  data_type: TYPE_INT64
  dims: 2
  dims: -1
}
output {
  name: "logits__0"
  data_type: TYPE_FP32
  dims: -1
  dims: 7
}

>>> 
```

### DIFF
--- a/triton-inference-client.spec
+++ b/triton-inference-client.spec
@@ -1,4 +1,5 @@
 ### RPM external triton-inference-client 2.25.0
+## INITENV +PATH PYTHON3PATH %{pkginstroot}/${PYTHON3_LIB_SITE_PACKAGES}
 ## INCLUDE cpp-standard
 %define branch r22.08
 %define github_user triton-inference-server
@@ -7,12 +8,14 @@
 
 Source0: git+https://github.com/%{github_user}/client.git?obj=%{branch}/%{client_tag}&export=%{n}-%{realversion}&output=/%{n}-%{realversion}.tgz
 Source1: git+https://github.com/%{github_user}/common.git?obj=%{branch}/%{common_tag}&export=common-%{realversion}&output=/common-%{realversion}.tgz
-BuildRequires: cmake git
-Requires: protobuf grpc cuda abseil-cpp re2
+BuildRequires: cmake git py3-wheel
+Requires: protobuf grpc cuda abseil-cpp re2 rapidjson
+Requires: py3-numpy py3-grpcio-tools py3-python-rapidjson
 
 %prep
 
 %setup -D -T -b 0 -n %{n}-%{realversion}
+sed -i -e 's|import os|import os,sys|' src/python/library/build_wheel.py
 %setup -D -T -b 1 -n common-%{realversion}
 
 %build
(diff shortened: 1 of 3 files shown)


smuzaffar:
please test

smuzaffar:

smuzaffar:
please test for el8_ppc64le_gcc12

triton-inference-client.spec
```diff
@@ -44,25 +47,46 @@ cmake ${PROJ_DIR} \
     -DCMAKE_CXX_STANDARD=%{cms_cxx_standard} \
     -DTRITON_ENABLE_CC_HTTP=OFF \
     -DTRITON_ENABLE_CC_GRPC=ON \
-    -DTRITON_ENABLE_PYTHON_HTTP=OFF \
```
kpedro88, line 47: If this part is only for building the C++ client (since there's a separate `buildpy` below), shouldn't these lines be kept here? (the defaults for all of these are OFF, right now, but I think it's better to be explicit, in case the default changes...)

triton-inference-client.spec
```diff
@@ -44,25 +47,46 @@ cmake ${PROJ_DIR} \
...
+
+make %{makeprocesses} VERBOSE=1
+
+rm -rf ../buildpy ; mkdir ../buildpy ; cd ../buildpy
+cmake ../%{n}-%{realversion}/src/python \
+    -DCMAKE_INSTALL_PREFIX="%{i}" \
+    -DCMAKE_INSTALL_LIBDIR=lib \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_CXX_STANDARD=%{cms_cxx_standard} \
+    -DTRITON_ENABLE_PYTHON_HTTP=OFF \
```
kpedro88, line 69: Following the above comment, I would include the corresponding lines here:
```
    -DTRITON_ENABLE_CC_HTTP=OFF \
    -DTRITON_ENABLE_CC_GRPC=OFF \
```

kpedro88:
thanks @smuzaffar ! Indeed, I think it was the lack of a rapidJSON external that had kept me from enabling the python client.

triton-inference-client.spec
smuzaffar: These `TRITON_ENABLE_PYTHON_*`  flags are not available for `triton/src/c++/CMakeLists.txt` ( which we use to build the c++ interface)  and `cmake` generate warning message that these are unused flags.

triton-inference-client.spec
smuzaffar: these flags are not available for `triton/src/python/CMakeLists.txt` which we use to build the python interface

triton-inference-client.spec
kpedro88: Ah, that is because they are in the top-level CMakeLists.txt, but we're using the project-specific ones for more control. Okay, never mind then.

smuzaffar:
please test

lets use the latest IB

smuzaffar:
@kpedro88 , should we try to update triton client ?
- triton client to https://github.com/triton-inference-server/client/tree/r23.03
- triton common https://github.com/triton-inference-server/common/tree/r22.12

kpedro88:
@smuzaffar we do not have any urgent need to update the triton client at the moment, so let's not complicate this PR.

smuzaffar:
+externals

PR #8958, opened 2024-01-23
Fix gsl-config so that by default it uses openblas lib

smuzaffar:
This fixes the `gsl-config` script to by default use openblas library (see https://github.com/cms-sw/cmssw/issues/42716 for details)


smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
+externals

PR #8960, opened 2024-01-23
cms-common: Fix for gsl-config

smuzaffar:
New cms-common fixes those gsl installations for which `gsl-config` contains reference to `gslcblas` library instead of `openblas`


smuzaffar:
please test

smuzaffar:
+externals

PR #8996, opened 2024-02-09
BuildRules: Fixes for multi-arch and PGO

smuzaffar:
Build rules are updated to
- PGO: Use relative source file path instead of full. This allows to generate and use PGO in different paths
- MULTIARCHS: fix generation of poisoned edmplugin cache for extra micro-arch. Thsi should fix the failing unit test in MUTLARCH IBs

### DIFF
--- a/scram-project-build.file
+++ b/scram-project-build.file
@@ -55,7 +55,7 @@ BuildRequires: dwz
 %endif
 
 %if "%{?configtag:set}" != "set"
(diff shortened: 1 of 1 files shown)


smuzaffar:
test parameters:
- full_cmssw = true

smuzaffar:
please test

smuzaffar:
+externals

PR #9005, opened 2024-02-12
Enable TF XLA, add tensorflow-xla-runtime.

riga:
Hi!

This PR enables TensorFlow XLA and adds a new tool `tensorflow-xla-runtime`, which provides the library needed by AOT compiled ML models.

Changes in particular:

- Enable XLA flag
- Add `tf2xla_supported_ops` to bazel targets, providing an executable that lists XLA and AOT compatible TF ops for the current env.
- Added `tensorflow-xla-runtime` spec, which is based on `tensorflow-source` which contains everything necessary to compile the runtime via cmake.
- Updates some python packages needed for some ML workflows.
  - `wrapt` is downgraded to from 1.15.0 to 1.14.1 since 1.15 introduces a bug in TF's SavedModel interface (issue linked next to `NO_AUTO_UPDATE` instruction)

The PR is required by https://github.com/cms-sw/cmssw/pull/43941.


cmssw-vectorization.file
```diff
@@ -1,2 +1,2 @@
-%define vectorized_packages zlib fastjet tensorflow-sources tensorflow OpenBLAS rivet gbl
+%define vectorized_packages zlib fastjet tensorflow-sources tensorflow tensorflow-xla-runtime OpenBLAS rivet gbl
```
smuzaffar: @riga, please also update https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_14_1_X/master/vectorization/cmsdist_packages.py#L14-L22 and include ` tensorflow-xla-runtime`  in the list

riga: Added :+1:

tensorflow-xla-runtime.spec
```diff
@@ -0,0 +1,32 @@
...
+
+# fix a single comparison between size_t and int
+sed -i -r 's/assert\(\(arg_size\(index\)/assert\(\(\(size_t\)arg_size\(index\)/' tensorflow/include/tensorflow/compiler/tf2xla/xla_compiled_cpu_function.h
+
+export CPATH="${CPATH}:${EIGEN_ROOT}/include/eigen3"
+export CPATH="${CPATH}:%{i}/tensorflow/include"
+export CPATH="${CPATH}:%{i}/tensorflow/include/third_party/eigen3"
+
+pushd tensorflow/xla_aot_runtime_src
+  cmake . -DCMAKE_CXX_FLAGS="-fPIC"
```
smuzaffar: In order to properly build for multi-vectorization please add `-msse3` here e.g. `-DCMAKE_CXX_FLAGS="-fPIC -msse3"`

smuzaffar:
please test

lets see the external build step

tensorflow-xla-runtime.spec
riga: Added :+1:

tensorflow-xla-runtime.spec
```diff
@@ -0,0 +1,32 @@
+### RPM external tensorflow-xla-runtime 2.12.0
```
smuzaffar, line 1: @riga, we need `tensorflow-xla-runtime` to be part of cmssw dependency chain, so please add it in https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_14_1_X/master/cmssw-tool-conf.spec

riga: Added :+1:

smuzaffar:
please test

pip/tensorflow.file
```diff
@@ -1,7 +1,7 @@
 Requires: py3-tensorboard py3-opt-einsum py3-tensorflow-estimator py3-wrapt py3-google-pasta py3-scipy
 Requires: tensorflow-sources
 %define PipPreBuildPy PIPFILE=${TENSORFLOW_SOURCES_ROOT}/tensorflow-%{realversion}-cp%{cms_python3_major_minor}-cp%{cms_python3_major_minor}-linux_%{_arch}.whl
-%define PipPostBuild rm -f %{i}/bin/tensorboard* ; ls %{i}/bin/* | xargs -i mv '{}' '{}3'
+%define PipPostBuild rm -f %{i}/bin/tensorboard* ; rm -r %{i}/lib/python%{cms_python3_major_minor_version}/site-packages/tensorflow/xla_aot_runtime_src ; ls %{i}/bin/* | xargs -i mv '{}' '{}3'
```
smuzaffar: @riga any specific reason to delete it? Looks like this is what you get out of tensorflow-sources wheel file. May be if we do not delete it here then you can just add dependency on `py3-tensorflow` and copy $PY3_TENSORFLOW_ROOT/$PYTHON3_LIB_SITE_PACKAGES/tensorflow/xla_aot_runtime_src to build `tensorflow-xla-runtime` ... right?

tensorflow-xla-runtime.spec
```diff
@@ -0,0 +1,32 @@
...
+
+# fix a single comparison between size_t and int
+sed -i -r 's/assert\(\(arg_size\(index\)/assert\(\(\(size_t\)arg_size\(index\)/' tensorflow/include/tensorflow/compiler/tf2xla/xla_compiled_cpu_function.h
+
+export CPATH="${CPATH}:${EIGEN_ROOT}/include/eigen3"
+export CPATH="${CPATH}:%{i}/tensorflow/include"
+export CPATH="${CPATH}:%{i}/tensorflow/include/third_party/eigen3"
+
+pushd tensorflow/xla_aot_runtime_src
+  cmake . -DCMAKE_CXX_FLAGS="-fPIC -msse3"
```
smuzaffar: I think you can add `-DBUILD_SHARED_LIBS=OFF` to disable shared library and directly build the archive lib. In this case you can avoid running `gcc -shared ....` command

tensorflow-xla-runtime.spec
```diff
@@ -0,0 +1,32 @@
...
+  make %{makeprocesses}
+  # this builds a shared library, but when used some symbols are missig (e.g `tsl::mutex::unlock()')
+  # so it does not seem intended to be used as a shared lib
+  gcc -shared -o libtf_xla_runtime.so -Wl,--whole-archive libtf_xla_runtime.a -Wl,--no-whole-archive
+popd
+
+%install
+
+mv tensorflow/include %{i}
```
smuzaffar: @riga, the contents of `tensorflow/include` are same as the one in `$PY3_TENSORFLOW_ROOT/$PYTHON3_LIB_SITE_PACKAGES/tensorflow/include` . I would suggest that instead of duplicating this why not we get the include from py3-tensorflow ( we can add an extra toolfile so that these headers are accessible in cmssw)

tensorflow-xla-runtime.spec
```diff
@@ -0,0 +1,32 @@
...
+
+mkdir -p %{i}/lib/archive
+mv tensorflow/xla_aot_runtime_src/libtf_xla_runtime.a %{i}/lib/archive/libtf_xla_runtime-static.a
+mv tensorflow/xla_aot_runtime_src/libtf_xla_runtime.so %{i}/lib/libtf_xla_runtime.so
```
smuzaffar: do we need the shared lib?

tensorflow-xla-runtime.spec
smuzaffar: also please add `-DCMAKE_CXX_STANDARD=%{cms_cxx_standard}`  just to make sure to use the correct c++std. Ypu need to include `cpp-standard.file` too see https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_14_1_X/master/tbb.spec#L2C12-L2C24

cmssw-vectorization.file
smuzaffar: ah sorry, I forgot to mention, please also add `scram-tools.file/tools/tensorflow-xla-runtime/vectorized.tmpl`  file same as https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_14_1_X/master/scram-tools.file/tools/tensorflow/vectorized.tmpl except that the tool `name` should be `tensorflow-xla-runtime_@TOOL_VECTORIZATION@`

pip/tensorflow.file
riga: Indeed. We changed this in f77258c.

tensorflow-xla-runtime.spec
riga: Both added in f77258c.

tensorflow-xla-runtime.spec
riga: In f77258c we changed the dependency of `tensorflow-xla-runtime` from `tensorflow-sources` to `py3-tensorflow` and added a separate tool file `py3-tensorflow/tensorflow-includes.xml` that we `use`d in the toolfile of the xla runtime.

tensorflow-xla-runtime.spec
riga: Not for now. We removed it in f77258c.

scram-tools.file/tools/tensorflow-xla-runtime/tensorflow-xla-runtime.xml
```diff
@@ -0,0 +1,10 @@
+<tool name="tensorflow-xla-runtime" version="@TOOL_VERSION@">
+  <client>
+    <environment name="TENSORFLOW_XLA_RUNTIME_BASE" default="@TOOL_ROOT@"/>
+    <environment name="LIBDIR" default="$TENSORFLOW_XLA_RUNTIME_BASE/lib/archive"/>
+    <environment name="INCLUDE" default="$TENSORFLOW_XLA_RUNTIME_BASE/include"/>
```
smuzaffar: @riga, I guess this is not needed any more ... right?

riga: Right. It's removed now.

smuzaffar:
test parameters:
- pull_request = cms-sw/cmssw#43941

smuzaffar:
please test

riga:
The reported header inconsistencies were fixed in https://github.com/cms-sw/cmssw/pull/43941.

smuzaffar:
please test

riga:
The broken unit tests should be fixed with [cmssw/9784e09](https://github.com/cms-sw/cmssw/pull/43941/commits/9784e0944c35660f2322c0af9ad7ddaa99a181d1).

smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
@riga, `tensorflow-xla-runtime` contains only an archive library. Enabling this tool for [multi-archs/vactorizations builds](https://github.com/cms-sw/cmsdist/pull/9005/files#diff-56c31d38da773f5fab39a62d89ad676228cf6d73cbee7a0d2265d052671ed984R1) is not going to work. CMSSW multi-arch env only works for shared library. So I would suggest to remove https://github.com/cms-sw/cmsdist/pull/9005/files#diff-56c31d38da773f5fab39a62d89ad676228cf6d73cbee7a0d2265d052671ed984 change.

makortel:
> `tensorflow-xla-runtime` contains only an archive library. Enabling this tool for [multi-archs/vactorizations builds](https://github.com/cms-sw/cmsdist/pull/9005/files#diff-56c31d38da773f5fab39a62d89ad676228cf6d73cbee7a0d2265d052671ed984R1) is not going to work. CMSSW multi-arch env only works for shared library.

Does this mean that in order to fully utilize the multi-arch build (i.e. wider vector instructions) for the AOT-compiled models we need some further development? (like making `tensorflow-xla-runtime` to produce a shared library, or extending CMSSW multiarch build rules to work with archive libraries, or something else)

smuzaffar:
yes @makortel

> extending CMSSW multiarch build rules to work with archive libraries

currently the multi-arch build works by linking the shared lib and then loading the correct micro-arch library at runtime. For archive libs  we need to change build rules to use the correct micro-arch library at build time. It is doable but need some development

> making tensorflow-xla-runtime to produce a shared library

I do not know if this will work for xla cmssw build. @riga , remind me why we have this archive libs instead of shared.

riga:
@smuzaffar The shared library had some symbols missing when linked to certain AOT models. However, we did not fully understand whether this is the intended behavior and the library is only meant to be used as a archive, or whether there is a way to build it correctly. This is unfortunately not documented.

If having only the static archive is a show stopper, we can look deeper in the *.so build.

> @riga, `tensorflow-xla-runtime` contains only an archive library. Enabling this tool for [multi-archs/vactorizations builds](https://github.com/cms-sw/cmsdist/pull/9005/files#diff-56c31d38da773f5fab39a62d89ad676228cf6d73cbee7a0d2265d052671ed984R1) is not going to work. CMSSW multi-arch env only works for shared library. So I would suggest to remove https://github.com/cms-sw/cmsdist/pull/9005/files#diff-56c31d38da773f5fab39a62d89ad676228cf6d73cbee7a0d2265d052671ed984 change.

Should this be reverted in any case?

makortel:
> If having only the static archive is a show stopper, we can look deeper in the *.so build.

My take is that it is not a showstopper, but something that probably should be worked on in the (near) future.

riga:
> My take is that it is not a showstopper, but something that probably should be worked on in the (near) future.

Ok. Since we plan a second PR with the production workflow for the compilation of models anyway, we would work on the shared library build then.

rappoccio:
+1 


Tests are already successful in the sibling PR in cmssw. 

https://github.com/cms-sw/cmssw/pull/43941

rappoccio:
merge

PR #9043, opened 2024-02-29
[DO NOT MERGE] Test making Eigen alignment setting consistent

makortel:
Following https://github.com/cms-sw/cmssw/issues/44188#issuecomment-1971628271 This PR attempts to use consistent alignment options for Eigen in the remaining externals depending on Eigen.

I skipped `professor2` because of not feeling proficient enough in RPM spec syntax, and it not being linked against in CMSSW.

I think we'd also want a better way to propagate these settings.

### DIFF
--- a/lwtnn.spec
+++ b/lwtnn.spec
@@ -1,4 +1,4 @@
-### RPM external lwtnn 2.13
+### RPM external lwtnn 2.14.1
 ## INCLUDE cpp-standard
 
 Source: https://github.com/lwtnn/lwtnn/archive/v%{realversion}.tar.gz
--- a/lwtnn.spec
+++ b/lwtnn.spec
@@ -17,7 +17,11 @@ cd ../build
 cmake ../%{n}-%{realversion} \
   -G Ninja \
   -DCMAKE_CXX_COMPILER="g++" \
-  -DCMAKE_CXX_FLAGS="-fPIC" \
+  %ifarch x86_64
+  -DCMAKE_CXX_FLAGS="-fPIC -DEIGEN_MAX_ALIGN_BYTES=64 -msse3" \
+  %else
+  -DCMAKE_CXX_FLAGS="-fPIC -DEIGEN_MAX_ALIGN_BYTES=64" \
+  %endif
   -DCMAKE_INSTALL_PREFIX:PATH="%{i}" \
   -DCMAKE_BUILD_TYPE=Release \
   -DBUILTIN_BOOST=OFF \
--- a/opencv.spec
+++ b/opencv.spec
@@ -23,6 +23,11 @@ cmake ../%{n}-%{realversion} \
     -DCMAKE_CXX_STANDARD=%{cms_cxx_standard} \
     -DCMAKE_INSTALL_LIBDIR=lib \
     -DWITH_EIGEN=ON \
+    %ifarch x86_64
+    -DCMAKE_CXX_FLAGS="-DEIGEN_MAX_ALIGN_BYTES=64 -msse3" \
+    %else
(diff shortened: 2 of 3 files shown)


makortel:
@cmsbuild, please test

lwtnn.spec
```diff
@@ -17,7 +17,11 @@ cd ../build
 cmake ../%{n}-%{realversion} \
   -G Ninja \
   -DCMAKE_CXX_COMPILER="g++" \
-  -DCMAKE_CXX_FLAGS="-fPIC" \
+  %ifarch x86_64
+  -DCMAKE_CXX_FLAGS="-fPIC -DEIGEN_MAX_ALIGN_BYTES=64 -msse3" \
+  %else
+  -DCMAKE_CXX_FLAGS="-fPIC -DEIGEN_MAX_ALIGN_BYTES=64" \
+  %endif
```
makortel, line 24: Somehow these settings do not end up being used in `lwtnn` compilation (see https://github.com/cms-sw/cmssw/issues/44188#issuecomment-1973588273, and I also confirmed the behavior locally).

@iarspider @aandvalenzuela Would you be able to take a look what needs to be done here in order to make the `-DEIGEN_MAX_ALIGN_BYTES=64` effective?

makortel: @iarspider @aandvalenzuela Would you be able to take a look?

iarspider:
I will try to figure it out.

lwtnn.spec
smuzaffar: @makortel , this is because lwtnn was overrding the CXX_FLAGS. They have fixed this in new versions https://github.com/lwtnn/lwtnn/commit/2647c25a433c706764fadb123a3a517e87bc6892

makortel: Thanks @smuzaffar! Let's see if the update to 2.14.1 works better.

makortel:
@cmsbuild, please test

makortel:
Based on the tests in https://github.com/cms-sw/cmssw/issues/44188#issuecomment-1984764982 and https://github.com/cms-sw/cmssw/issues/44188#issuecomment-1985807551 this PR seems to resolve https://github.com/cms-sw/cmssw/issues/44188. 

@smuzaffar @iarspider Would this PR be good-enough to be deployed as is? Or do we want these changes done in some better way? Some possible issues I could think of
- Build recipe of `professor2` should also be updated correspondingly
- Is this way proper also for the MULTIARCHS build?
  - If not, should that be addressed in this PR (or equivalent), or later?
- I'm not happy with the amount of copy-paste of this special setting

smuzaffar:
@makortel , I would like to redo this to avoid copy pasting the eigen flags and enabling these extra packages for multi-arch builds. I will do it on Monday

makortel:
Thanks @smuzaffar. I'll close this PR then (it did its purpose).

perrotta:
> @makortel , I would like to redo this to avoid copy pasting the eigen flags and enabling these extra packages for multi-arch builds. I will do it on Monday

Thank you @smuzaffar 
Please let me point out that such PR will be needed rather urgently by the TkAl team, and we'll request a new release as soon as that one (as well as [cms-sw/cmssw#44359](https://github.com/cms-sw/cmssw/pull/44359) and
[cms-sw/cmsdist#9057](https://github.com/cms-sw/cmsdist/pull/9057)) will get merged in 14_0_X

smuzaffar:
FYI @perrotta :  https://github.com/cms-sw/cmsdist/pull/9060

perrotta:
> FYI @perrotta : #9060

Thank you @smuzaffar !

smuzaffar:
@perrotta , consistent eigen flags PR for 14.0.X is https://github.com/cms-sw/cmsdist/pull/9062

PR #9066, opened 2024-03-12
Add tf_cuda_support tool if TF is build with cuda support

smuzaffar:
Following up https://github.com/cms-sw/cmssw/pull/44376 , this PR adds a new tool `tf_cuda_support` which will only be available in cmssw if TF is build with cuda/gpu support. One can use this tools to conditionally build/run cmssw tests which require TF GPU support.

### DIFF
--- a/tensorflow.spec
+++ b/tensorflow.spec
@@ -29,7 +29,7 @@ tar xfz ${%{tf_root}}/libtensorflow_cc.tar.gz -C %{i}
 mkdir -p %{i}/etc/scram.d
 cat << \EOF_TOOLFILE >%{i}/etc/scram.d/tf_cuda_support.xml
   <tool name="tf_cuda_support" version="1.0">
-  <tool>
+  </tool>
 EOF_TOOLFILE
 %endif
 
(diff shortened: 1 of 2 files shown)


smuzaffar:
please test

tensorflow.spec
```diff
@@ -25,6 +25,13 @@ esac
 %install
 
 tar xfz ${%{tf_root}}/libtensorflow_cc.tar.gz -C %{i}
+%if %{enable_gpu}
+mkdir -p %{i}/etc/scram.d
+cat << \EOF_TOOLFILE >%{i}/etc/scram.d/tf_cuda_support.xml
+  <tool name="tf_cuda_support" version="1.0">
+  <tool>
```
makortel: Should this be
```suggestion
  </tool>
```
?

smuzaffar:
please test 
thanks @makortel for extra eyes :-)

smuzaffar:
+externals

Note that as currently TF is build with GPU support so this new tool will not be available in cmssw

PR #9093, opened 2024-03-22
Add tools to build TF AOT models

riga:
This PR is a continuation of https://github.com/cms-sw/cmsdist/pull/9005 and adds all necessary tools to AOT compile TensorFlow models provided via external resources or data repositories.

- A new, small external package `cms_tfaot` is provided which forwards the compilation commands to the already existing cmsml package and dynamically builds tool files. It also has a `dev` mode which produces more verbose outputs with instructions for local compilations.
- The package also provides 2 test models which will be required by the unittests in PhysicsTools/TensorFlowAOT.
- The compilation instructions are handled by a `tfaot-compile.file` which only requires 2 variables. With that, each model should have its own spec file which can `##INCLUDE tfaot-compile.file`.
- The tensorflow-xla-runtime is now provided as a shared library. However, it seems like this is not necessarily the intention. When building the `*.so` normally, the plugin build process in CMSSW will complain about missing symbols from absl and tsl. For this reason we manually linked to 2 absl libraries and disabled complaints about unresolved symbols of the tsl library (shipped incomplete within the xla runtime source, i.e., some symbols are missing but also not required when built as a static lib) via `-Wl,--unresolved-symbols=ignore-in-shared-libs` in the tool file.

I enabled edit permissions, so feel free to move the `cms-tfaot` tool to cms-externals.

This PR is needed by https://github.com/cms-sw/cmssw/pull/44519.

@Bogdan-Wiederspan


py3-cms-tfaot.spec
```diff
@@ -0,0 +1,22 @@
+### RPM external py3-cms-tfaot 1.0.0
```
smuzaffar: @riga , for pip based packages it is batter to use the default way of building ( i.e. using https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_14_1_X/master/build-with-pip.file ). So please drop this spec and add `cmsdist/pip/cms-tfaot.file` with following contents
```
Requires: py3-cmsml py3-tensorflow
%define github_user riga
%define tag a2bbd06cbed0efa1fa191cc2f50a6b03067e59d2
%define branch master
%define source0 git+https://github.com/%{github_user}/cms-tfaot.git?obj=%{branch}/%{tag}&export=%{n}-%{realversion}&output=/%{n}-%{realversion}-%{tag}.tgz 
```

tensorflow-xla-runtime.spec
```diff
@@ -21,11 +22,17 @@ CXXFLAGS="-fPIC %{arch_build_flags} ${CMS_EIGEN_CXX_FLAGS}"
 %endif
 
 pushd tensorflow/xla_aot_runtime_src
+  # the CMakelists.txt file is not intended to produce a shared library so create a static one first
   cmake . -DCMAKE_CXX_FLAGS="${CXXFLAGS}" -DCMAKE_CXX_STANDARD=%{cms_cxx_standard} -DBUILD_SHARED_LIBS=OFF
   make %{makeprocesses}
+
+  # build the shared library, linking missing symbols from abseil
+  gcc -shared -o libtf_xla_runtime.so -Wl,--whole-archive libtf_xla_runtime.a -Wl,--no-whole-archive \
+    -L${ABSEIL_CPP_ROOT}/lib -labsl_strings -labsl_str_format_internal
```
smuzaffar: do you understand why there are missing symbols? May be it is not picking up our abseil-cpp properly?

smuzaffar: @riga, I changed the cmake file (CMakeLists.txt) to have
```
add_library(tf_xla_runtime SHARED
  $<TARGET_OBJECTS:tf_xla_runtime_objects>
)
``` 
and then looking at the missing `abseil` symbols, I get
```
                 U _ZN4absl12lts_2022062319str_format_internal10FormatPackB5cxx11ENS1_21UntypedFormatSpecImplENS0_4SpanIKNS1_13FormatArgImplEEE
                 U _ZN4absl12lts_2022062319str_format_internal13FormatArgImpl8DispatchIiEEbNS2_4DataENS1_24FormatConversionSpecImplEPv
                 U _ZN4absl12lts_2022062319str_format_internal13FormatArgImpl8DispatchISt17basic_string_viewIcSt11char_traitsIcEEEEbNS2_4DataENS1_24FormatConversionSpecImplEPv
                 U _ZN4absl12lts_202206239StrAppendEPNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEERKNS0_8AlphaNumE
```

all of these are defined in the `absl_strings` library.  We can patch [a] `xla_aot_runtime_src/CMakeLists.txt ` and then it should generate the shared library (properly linked to  absl::strings)

[a]
```
--- xla_aot_runtime_src/CMakeLists.txt    2024-03-24 08:28:34.000000000 +0100
+++ xla_aot_runtime_src/CMakeLists.txt      2024-03-25 11:17:58.108587945 +0100
@@ -14,6 +14,8 @@
   -Wno-sign-compare
 )
 
-add_library(tf_xla_runtime STATIC
+find_package(absl REQUIRED)
+add_library(tf_xla_runtime SHARED
   $<TARGET_OBJECTS:tf_xla_runtime_objects>
 )
+target_link_libraries(tf_xla_runtime absl::strings)
```

py3-cms-tfaot.spec
riga: 👍 Changed in 7838448.

tensorflow-xla-runtime.spec
riga: > do you understand why there are missing symbols? May be it is not picking up our abseil-cpp properly?

Only partially. Indeed it seems that it didn't find the abseil-cpp libraries which is now fixed. For the missing symbols of the tsl library I think the reason is that the `tensorflow_xla_runtime` sources (shipped with `py3-tensorflow`) only provide a subset of the implementations defined by headers.

tsl

These symbols aren't needed anyway - unlike the absl ones - hence the `-Wl,--unresolved-symbols=ignore-in-shared-libs` in the toolfile. (I guess one could also fetch and compile the missing sources, though at the risk of inflating the runtime with unnecessary symbols.)

<hr/>

I added a patch in 7838448 and removed the initial static build step.

scram-tools.file/tools/tensorflow-xla-runtime/tensorflow-xla-runtime.xml
```diff
@@ -1,9 +1,11 @@
 <tool name="tensorflow-xla-runtime" version="@TOOL_VERSION@">
   <client>
     <environment name="TENSORFLOW_XLA_RUNTIME_BASE" default="@TOOL_ROOT@"/>
-    <environment name="LIBDIR" default="$TENSORFLOW_XLA_RUNTIME_BASE/lib/archive"/>
+    <environment name="LIBDIR" default="$TENSORFLOW_XLA_RUNTIME_BASE/lib"/>
   </client>
-  <lib name="tf_xla_runtime-static"/>
+  <lib name="tf_xla_runtime"/>
+  <flags LDFLAGS="-Wl,--unresolved-symbols=ignore-in-shared-libs"/>
```
smuzaffar: @riga , we do not need these ld flags any more .. right?

smuzaffar: ah ok, I now have read your https://github.com/cms-sw/cmsdist/pull/9093#discussion_r1538034992 . I will check what else we can do to avoid the missing symbols. We should not add this flag here otherwise scram will not fail for missing symbols for every cmssw shared lib which depend on tensorflow-xml-runtime  (directly or in-directly)

riga: Yeah, this already felt like a dangerous thing to do. Do you think there is any other option besides a) fetching the missing source files to `tensorflow/tsl` before compiling, or b) installing the [full tsl](https://github.com/google/tsl) as a dependency (still seems a bit like work in progress)?

smuzaffar: @riga , all the missing TSL symbols are coming from https://github.com/tensorflow/tensorflow/blob/v2.12.1/tensorflow/compiler/xla/service/cpu/runtime_fork_join.cc e.g various macros calls `VLOG, CHECK_*` and `tsl::BlockingCounter`. If you think we can run without these missing symbols ( e.g. we never call these) so why not just then drop the `runtime_fork_join.cc ` from the compilation?

smuzaffar: you can also add `-Wl,-z,defs` to cmake CXXFLAGS to make sure that the shared library has no missing symbols. I mean at link time `make` will fail if there are any missing symbols.

smuzaffar:
test parameters:
- pull_request = cms-sw/cmssw#44519

smuzaffar:
please test

scram-tools.file/tools/tensorflow-xla-runtime/tensorflow-xla-runtime.xml
riga: > If you think we can run without these missing symbols ( e.g. we never call these) so why not just then drop the runtime_fork_join.cc  from the compilation?

Ok, so when we used the static lib before, the corresponding object files were likely dropped already, makes sense. And since we will potentially never use eigen thread pools for inference, we can skip it right away. Then I guess we are lucky that the tsl dependence only affects that part of the xla runtime source that we anyway don't need (well, at least for now).

We just verified that everything still works 👍

smuzaffar:
please test

tfaot-compile.file
```diff
@@ -0,0 +1,33 @@
...
+mv compiled_model/*.o %{i}/lib/
+
+mkdir -p %{i}/include/%{n}
+mv compiled_model/*.h %{i}/include/%{n}
+
+mkdir -p %{i}/etc/scram.d
+mv compiled_model/%{n}.xml %{i}/etc/scram.d/
+
+%post
+%{relocateConfig}etc/scram.d/%{n}.xml
```
smuzaffar, line 40: @riga, there is hard-coded build path in `tfaot-model-*/version/include/tfaot-model-test-multi/test_multi.h`. Though it is in comment section but it is better to run relocation on these generated header files. I would suggest to add the following line here to relocated all generated header files
```
%relocateConfigAll include/%{n} *.h
```



[a] 
- https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-e51cbd/38438/external_checks/relocate/tfaot-model-test-multi.txt
- https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-e51cbd/38438/external_checks/relocate/tfaot-model-test-simple.txt

e.g.
```
#ifndef TFAOT_MODEL_TEST_MULTI_H
#define TFAOT_MODEL_TEST_MULTI_H

/*
 * Auto-generated AOT wrapper for
 *   model path  : /data/cmsbld/jenkins/workspace/ib-run-pr-tests/testBuildDir/el8_amd64_gcc12/external/py3-cms-tfaot/1.0.0-b8bfd2f94522378f26f348963d4ff4ac/share/test_models/multi/saved_model
 *   prefix      : test_multi
 *   namespace   : tfaot_model
 *   class name  : test_multi
 *   batch sizes : 1, 2, 4
 */
```

riga: Added :+1:
Plus, the last commit includes some final changes to the aot python tools.
The accompanying cmssw PR is also updated with adjustments to the unit tests of the dev workflow which failed tonight.

smuzaffar:
please test

smuzaffar:

smuzaffar:
please test for el8_ppc64le_gcc12

smuzaffar:
please test for slc7_amd64_gcc12

smuzaffar:
please test for el8_ppc64le_gcc12

smuzaffar:

smuzaffar:
please test for el8_ppc64le_gcc12

riga:
The build arch is now forwarded to the aot compiler invocation.

https://github.com/riga/cmsdist/blob/0677c1b66b90326b47d191b54109f4c38d1af3c3/tfaot-compile.file#L10-L14

smuzaffar:

smuzaffar:
please test

tfaot-compile.file
```diff
@@ -17,7 +23,8 @@ cms_tfaot_compile \
   --aot-config "%{aot_config}" \
   --tool-name "%{n}" \
   --tool-base "%{i}" \
-  --output-directory compiled_model
+  --output-directory compiled_model \
+  --additional-flags="--target_triple %{build_arch}"
```
smuzaffar, line 27: Thanks @riga, looks like this worked ( atleast ppc64le build was successful)

smuzaffar:
+externals

Looks good

tfaot-compile.file
smuzaffar: @rida, I guess `cms_tfaot_compile` tool is used by developers to work with new models ... right? Do we want developers to explicitly pass `--additional-flags/--target_triple` or may be you can set some default values in side the script ?

riga: Exactly. When run with `--dev` the tool (almost) mimics what the `tfaot-compile.file` does. The default arch for the `--target_triple` is `x86_64` so I guess this should apply in most cases. However, in dev mode we could check which arch is currently used and issue a warning, together with instructions on how to change the triple.

riga: @smuzaffar I just added a warning in https://github.com/riga/cms-tfaot/commit/a03ccef81c610e051d3d7d172d675fe6c1ff17e2. Was your plan to move the project to cms-externals? If not, I would adjust the commit hash in *this* PR.

smuzaffar: > Was your plan to move the project to cms-externals? If not, I would adjust the commit hash in this PR.

right, we should move it to cms-externals. I have added you developers of `cms-externals`. Can you please transfer the repo there. If you do not want to transfer then I can fork it

riga: I moved it and updated the reference in the spec.

smuzaffar:
please test

lets run the final tests

smuzaffar:

smuzaffar:
+externals

smuzaffar:
please test

lets see if this can go in without cmssw PR

antoniovilela:
+1

PR #9097, opened 2024-03-25
[ROOT6] Updated root to tip of branch master

aandvalenzuela:
ROOT [version](https://github.com/root-project/root/blob/master/core/foundation/inc/ROOT/RVersion.hxx) is still `6.31.01`.


aandvalenzuela:
please test

PR #9109, opened 2024-04-02
[DQM] Bump DQMGUI tag + Apply patch for increasing index limits

nothingface0:
- Bumped DQMGUI version. The [new version](https://github.com/cms-DQM/dqmgui_prod/releases/tag/9.9.0) includes a monitoring script for the index size. 
  - To do so, I updated the git branch in the `spec` file from (`index128_cc7`) which is dedicated to the CC7 deployment. I'm not sure if it's needed. 
- Apply a patch (`dqmgui-index-size.patch`) to increase the limit of `PATHNAMES` (mostly for the `vocms0738` "offline" deployment) and `DATASETNAMES` (mostly for the `vocms0731` "dev" deployment). The patch was created by a simple `git diff` in the root directory of the `dqmgui_prod` source. I am not sure if it will just work, since the build procedure is not exactly clear to me. Let me know if I need to modify it.

This PR is also required for merging this PR: https://github.com/dmwm/deployment/pull/1310

CC: @syuvivida @tjavaid @antoniovagnerini

### DIFF
--- a/dqmgui.spec
+++ b/dqmgui.spec
@@ -1,4 +1,4 @@
-### RPM cms dqmgui 9.8.1
+### RPM cms dqmgui 9.9.0
 ## INITENV +PATH PATH %i/xbin
 ## INITENV +PATH %{dynamic_path_var} %i/xlib
 ## INITENV +PATH PYTHONPATH %i/${PYTHON_LIB_SITE_PACKAGES}
--- a/dqmgui.spec
+++ b/dqmgui.spec
@@ -17,6 +17,7 @@ Source0: git+https://github.com/cms-DQM/dqmgui_prod.git?obj=index128/%realversio
 Source2: svn://rotoglup-scratchpad.googlecode.com/svn/trunk/rtgu/image?module=image&revision=10&scheme=http&output=/rtgu.tar.gz
 Source3: http://opensource.adobe.com/wiki/download/attachments/3866769/numeric.tar.gz
 Patch0: dqmgui-rtgu
+Patch1: dqmgui-index-size
 
 Requires: python cherrypy py2-cheetah yui extjs gmake pcre boost root rootjs libpng libjpg classlib rotatelogs py2-pycurl py2-cjson libuuid d3 protobuf py2-argparse py2-pytest py2-nose jemalloc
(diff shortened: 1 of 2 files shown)


dqmgui.spec
```diff
@@ -8,7 +8,7 @@
 %define webdoc_files %{installroot}/%{pkgrel}/128/doc
 %define cvs cvs://:pserver:anonymous@cmscvs.cern.ch:2401/cvs_server/repositories/CMSSW?passwd=AA_:yZZ3e
 
-Source0: git+https://github.com/cms-DQM/dqmgui_prod.git?obj=index128/%realversion&export=Monitoring&output=/Monitoring.tar.gz
+Source0: git+https://github.com/cms-DQM/dqmgui_prod.git?obj=index128_cc7/%realversion&export=Monitoring&output=/Monitoring.tar.gz
```
dpapagiannis, line 11: I added this change since the tag we need (`9.9.0`) is on branch `index128_cc7` instead of `index128`, but I'm not exactly sure if this change is needed. Does this build procedure use the git tag directly, based on what's specified on the `spec` file's first line?

smuzaffar: @nothingface0 , yes if the source url has `obj=branch/tag`  then `cmsBuild` uses `tag` . In your case `Source0` has `obj=index128/%realversion` i.e. `%realversion` (specified in the first line) is treated as tag and `cmsBuild` will download sources using the tag. so no need to change the branch but is better to keep the correct branch in source url

smuzaffar:
@nothingface0 you need to ping @arooshap to start the tests

dqmgui.spec
dpapagiannis: Thank you @smuzaffar, so I leave it as `index128_cc7`, which is where the source code actually is, just in case.

arooshap:
please test

arooshap:
+1 to 2d0e9f1ea5c23bb06291590b731903a1898a0cfd on slc7_amd64_gcc630.<br>Build log can be found <a href="https://cern.ch/cms-dmwm-builds/9109/test_2d0e9f1-20240408-141003.log" target="_blank">here</a>.

PR #9143, opened 2024-04-17
Update ROCm to version 6.1.0

fwyzard:
Add amd-smi and ROCProfiler binaries and libraries.

### DIFF
--- a/rocm-rocrand.spec
+++ b/rocm-rocrand.spec
@@ -1,4 +1,4 @@
-### RPM external rocm-rocrand 5.6.1
+### RPM external rocm-rocrand 6.1.0
 ## NOCOMPILER
 
 %if "%{rhel}" == "7"
@@ -14,8 +14,8 @@
 # without any .0 in the directory name
 %define repoversion %(echo %{realversion} | sed -e's/\.0$//')
 
-Source0: https://%{repository}/%{repoversion}/main/rocrand-2.10.17.50601-93.el%{rhel}.%{_arch}.rpm
-Source1: https://%{repository}/%{repoversion}/main/rocrand-devel-2.10.17.50601-93.el%{rhel}.%{_arch}.rpm
+Source0: https://%{repository}/%{repoversion}/main/rocrand-3.0.1.60100-82.el%{rhel}.%{_arch}.rpm
+Source1: https://%{repository}/%{repoversion}/main/rocrand-devel-3.0.1.60100-82.el%{rhel}.%{_arch}.rpm
 Requires: rocm
 AutoReq: no
 
--- a/rocm.spec
+++ b/rocm.spec
@@ -1,4 +1,4 @@
-### RPM external rocm 5.6.1
+### RPM external rocm 6.1.0
 
 %if "%{rhel}" == "7"
 # allow rpm2cpio dependency on the bootstrap bundle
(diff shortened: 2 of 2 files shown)


fwyzard:
please test

fwyzard:
please test

fwyzard:
please test with cms-sw/cmssw#44777

fwyzard:
`runtestPhysicsToolsPatAlgos` has been failing for weeks.

@smuzaffar do you think it would be possible to teach the bot to ignore failures that are already present in the base release ?

smuzaffar:
> `runtestPhysicsToolsPatAlgos` has been failing for weeks.
> 
> @smuzaffar do you think it would be possible to teach the bot to ignore failures that are already present in the base release ?

we had discussed this in core sw meeting. One reason to not do this is that sometimes a PR is suppose to fix a failure but if it does not provide the right fix (and test still fails) then bot is going to ignore the error ( as we told it to ignore IB errors) and mark it as passed.

smuzaffar:

smuzaffar:
+externals

fwyzard:
> One reason to not do this is that sometimes a PR is suppose to fix a failure but if it does not provide the right fix (and test still fails) then bot is going to ignore the error ( as we told it to ignore IB errors) and mark it as passed.

What about ignoring the tests that are expected to fail (because they fail in the IB), but report if they actually pass ?
This way if a PR is supposed to fix a test, it should be easy to see that it does.

smuzaffar:
but if a PR is suppose to fix a test but it still fails during PR tests then bot will ignore the failure and mark PR test passed  ... right?

fwyzard:
> but if a PR is suppose to fix a test but it still fails during PR tests then bot will ignore the failure and mark PR test passed ... right?

In such case I would expect the person that submitted the PR specifically to fix a test, the relevant L2, or the release manager, to check if the bot reports the test being fixed.

smuzaffar:
> In such case I would expect the person that submitted the PR specifically to fix a test, the relevant L2, or the release manager, to check if the bot reports the test being fixed.

this is a big expectation :-) I will see if we can come up with reasonable solution

fwyzard:
@cms-sw/orp-l2 could you merge this and https://github.com/cms-sw/cmssw/pull/44777 for the next IB ?

rappoccio:
+1

fwyzard:
thanks

makortel:
Given https://github.com/cms-sw/cmssw/issues/44821 maybe we should revert this update before building 14_1_0_pre3? @cms-sw/orp-l2

fwyzard:
OK for me to revert.

The main reason for the update was to fix the linker warnings about executable stack, but we can live with them.

Does https://github.com/cms-sw/cmssw/issues/44821 mean that we have to start building HIP/ROCm from sources ?

makortel:
> Does [cms-sw/cmssw#44821](https://github.com/cms-sw/cmssw/issues/44821) mean that we have to start building HIP/ROCm from sources ?

That is certainly one way forward, but I'm not sure if we understand the full picture well enough to judge if that is the best (or only?) way forward. (e.g. I still wonder if there is something (non-)LTO-specific in this failure mode, or if the set of so-far-impacted IBs is just "luck")

smuzaffar:
@makortel , all the IBs with this failure are non-lto ( https://github.com/search?q=repo%3Acms-sw%2Fcms-bot%20BUILD_OPTS%3Dno-lto&type=code )

PR #9152, opened 2024-04-22
[ROOT632] Root update v6 32 00 patches

aandvalenzuela:
ROOT [version](https://github.com/root-project/root/blob/v6-32-00-patches/core/foundation/inc/ROOT/RVersion.hxx) is still `6.31.99`.


aandvalenzuela:
please test

PR #9157, opened 2024-04-24
deploy new site based os mismatch check hook

smuzaffar:
This deploys a new site based scram hook to check for scram/host OS mismatch. See https://github.com/cms-sw/cms-common/pull/11

PR #9154, opened 2024-04-24
for multi-arch set SCRAM_TARGET to auto MULTIARCH/SKYLAKE otherwise to default

smuzaffar:
This PR contains the following updates

- For special `MULTIARCH/SKYLAKE` IB/releases: set `SCRAM_TARGET` to `auto` so that scram can automatically select the best set of libraries at runtime. This is what current MULTIARCHS IBs do (so no change in env logic for MULTIARCH/SKYLAKE builds)
- For default IB with multi-arch support enabled (e.g. 14.1.0.pre3): set `SCRAM_TARGET` to `default` so that scram only set default (`sse3`) set libraries at runtime.
- SCRAM: Update document about SCRAM_TARGET selection and extra scram help on project specific build targets (https://github.com/cms-sw/SCRAM/commit/f06afeba5ec011dd62e723a6490513bfddec7012)
- Build rules: [scram/project/hook](https://github.com/cms-sw/cmssw-config/commit/ba4e63e664e1d074975839dc39694e3ca89f5e79) has been updated to only enable multi-arch builds in developers' area if release is built with `SCRAM_TARGET=auto`. With this change, scram will build multi-arch libs for `MULTIARCH/SKYLAKE` IB/releases but will only build default (`sse3`) set of libraries for other IBs/releases.

FYI @makortel , @fwyzard

PR #9158, opened 2024-04-24
[14.0.X] backport of multi-arch and scram os mismatch hook

smuzaffar:
backport of #9154

Also include changes from #9157 to deploy the new cms-common with os-mismatch hook. These changes do not change the way we build MULTIARCH or normal IBs. These changes will be needed when we enable multi-arch for default IBs(14.1.X) and 14.0.X


smuzaffar:
please test

PR #9190, opened 2024-05-14
Added mozearch clang-plugin to generate mozsearch indexing

smuzaffar:
Mozsearch is DXR replacement. This PR adds the mozsearch clang-plugin in cmssw distro which we can use later for generating the cmssw indexing

FYI @aandvalenzuela


smuzaffar:
please test

smuzaffar:
+externals

PR #9191, opened 2024-05-14
cmssw-osenv: make sure to use scram v2 in el6 env

smuzaffar:


smuzaffar:
please test workflows 554.0,25210.0,25213.18,1360.0

smuzaffar:
+externals

workflows 554.0,25210.0,25213.18,1360.0 now run properly. through we do get messages like 
```
ERROR: ld.so: object 'libPerfToolsAllocMonitorPreload.so' from LD_PRELOAD cannot be preloaded: ignored.
ERROR: ld.so: object 'libPerfToolsMaxMemoryPreload.so' from LD_PRELOAD cannot be preloaded: ignored.
```
as PR tests set `LD_PRELOAD=libPerfToolsAllocMonitorPreload.so:libPerfToolsMaxMemoryPreload.so` and inside el6 containers these libs are not loadable.

PR #9196, opened 2024-05-21
updated pip pkg: requests, Jinja2, tqdm, Werkzeug

smuzaffar:
As nonted by dependabot, this update fixes following isses
- Werkzeug debugger vulnerable to remote execution when interacting with attacker controlled domain
- Requests `Session` object does not verify requests after making first request with verify=False
- Jinja vulnerable to HTML attribute injection when passing user input as keys to xmlattr filter
- tqdm CLI arguments injection attack


smuzaffar:
please test

smuzaffar:
+externals

PR #9203, opened 2024-05-22
Python3: disable picking up packages from user site (~/.local/lib/python)

iarspider:
As suggested by @kpedro88 in [Core Software](https://mattermost.web.cern.ch/cms-o-and-c/pl/yrt5qp48kjgdtgmrdywuaaii6a) MM channel.

### DIFF
--- a/python3.spec
+++ b/python3.spec
@@ -47,6 +47,7 @@ make %{makeprocesses} install
 sed -i -e "s|^#!.*python%{pythonv} *$|#!/usr/bin/env python%{python_major}|" %{i}/bin/* %{i}/lib/python*/*.py
 sed -i -e 's|^#!/.*|#!/usr/bin/env python%{pythonv}m|' %{i}/lib/python*/config-*/python-config.py
 sed -i -e 's|^#! */usr/local/bin/python|#!/usr/bin/env python|' %{i}/lib/python*/cgi.py
+sed -i -e 's|^ENABLE_USER_SITE = .*|ENABLE_USER_SITE = False|g' %{i}/lib/python*/site.py
 
 # is executable, but does not start with she-bang so not valid
 # executable; this avoids problems with rpm 4.8+ find-requires


iarspider:
please test

python3.spec
```diff
@@ -47,6 +47,7 @@ make %{makeprocesses} install
 sed -i -e "s|^#!.*python%{pythonv} *$|#!/usr/bin/env python%{python_major}|" %{i}/bin/* %{i}/lib/python*/*.py
 sed -i -e 's|^#!/.*|#!/usr/bin/env python%{pythonv}m|' %{i}/lib/python*/config-*/python-config.py
 sed -i -e 's|^#! */usr/local/bin/python|#!/usr/bin/env python|' %{i}/lib/python*/cgi.py
+sed -i -e 's|^ENABLE_USER_SITE = .*|ENABLE_USER_SITE = False|g' %{i}/lib/python*/site.py
```
smuzaffar, line 50: Actually we do not want this. This disables `PYTHONUSERBASE` env which we use to install pip packages.
We can recommend users to set `PYTHONUSERBASE` to some non valid path. This avoid adding `~/.local` in the pythonpath. see code from `site.py`
```
# Copy of sysconfig._getuserbase()
def _getuserbase():
    env_base = os.environ.get("PYTHONUSERBASE", None)
    if env_base:
        return env_base

    def joinuser(*args):
        return os.path.expanduser(os.path.join(*args))

    if os.name == "nt":
        base = os.environ.get("APPDATA") or "~"
        return joinuser(base, "Python")

    if sys.platform == "darwin" and sys._framework:
        return joinuser("~", "Library", sys._framework,
                        "%d.%d" % sys.version_info[:2])

    return joinuser("~", ".local")

```

kpedro88: @smuzaffar does this imply that we should not recommend users to set `export PYTHONNOUSERSITE=True`?

What if we reverse the logic of this function? i.e., treat `PYTHONUSERBASE` as null by default and only consider it if the user deliberately set it to some value.

smuzaffar: humm, I did not know about `PYTHONNOUSERSITE`. So is `export PYTHONNOUSERSITE=True`  not enough to avoid adding `~/.local` to PYTHONPATH?

iarspider: I think yes: https://docs.python.org/3.3/library/site.html#site.ENABLE_USER_SITE

iarspider: https://docs.python.org/3.3/using/cmdline.html#envvar-PYTHONNOUSERSITE

smuzaffar: Instead of changing python behavior (which can break users code who were using user site customization or site-dir`), I think we should recommend users to set `PYTHONNOUSERSITE=True` to avoid local modules from `~/.lcoal`

- without PYTHONNOUSERSITE=True
```
bash-5.1$ python3 -c 'import site;print(site.getusersitepackages());import muzaffar;print(muzaffar.MYNAME)'
/afs/cern.ch/user/m/muzaffar/.local/lib/python3.9/site-packages
muzaffar
```
- with PYTHONNOUSERSITE=True
```
bash-5.1$ PYTHONNOUSERSITE=True python3 -c 'import site;print(site.getusersitepackages());import muzaffar;print(muzaffar.MYNAME)'
/afs/cern.ch/user/m/muzaffar/.local/lib/python3.9/site-packages
ModuleNotFoundError: No module named 'muzaffar'
```

iarspider:
please abort

iarspider:
please close

PR #9218, opened 2024-06-03
scram tool clean; support for building stack without an option tool

smuzaffar:
The PR proposes following two updates
- Allow to have rpm macros in scram toolfiles. This allows to properly generate the toolfiles using the package informaton atbuild time instead of setting env or doing sed/grep in `%post` section. This is done by adding all tools in a single file and then expanding that file with in the actual toolfile or tool-conf

- Support added to build the stack without optional tools e.g. on Risc-V we do not have rocm/cuda. So using new feature of [pkgtools](https://github.com/cms-sw/pkgtools/pull/209) , now we should be able to build the external software tool stack for Risc-V. 

- cpu_features update to version 0.9.0 which has risc-v support/fixes


smuzaffar:
test parameters:
- pull_request = cms-sw/pkgtools#209
- full_cmssw = true
- enable = gpu

smuzaffar:
please test

onnxruntime.spec
```diff
@@ -21,7 +19,7 @@ rm -rf ../build; mkdir ../build; cd ../build
 USE_CUDA=OFF
 %if "%{cmsos}" != "slc7_aarch64"
 if [ "%{cuda_gcc_support}" = "true" ] ; then
-USE_CUDA=ON
+USE_CUDA=%{!?without_cuda:ON}
```
iarspider: @smuzaffar Sholdn't `%{cuda_gcc_support}` evaluate to `false` if we don't have cuda?

smuzaffar: `cuda_gcc_support` is when we have cuda i.e. we can run but we do not have a working cuda/gcc combination

smuzaffar: and https://github.com/cms-sw/cmsdist/blob/6a36618cc31540695383e45442e6b65c4b3a3910/rpm-preamble.file#L259-L270 chage is going to force set `cuda_gcc_support` to `false` is sw stack is build without cuda

iarspider: Thanks for the explanation. Will this work in a hypotetical situation where we have CUDA support, but don't build with CUDA (i.e. won't a line `USE_CUDA=`, where nothing is assigned to `USE_CUDA`, break anything?)

smuzaffar: in the case it will work as `USE_CUDA=%{!?without_cuda:ON}` will never be executed when we build `without_cuda`

smuzaffar: this is true that in this case `USE_CUDA=%{!?without_cuda:ON}` change is not needed as for `without_cuda` this blockof code is not executed

smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
+externals

PR #9230, opened 2024-06-07
Force rebuild jemalloc to fix the aarch64 runtime issues

smuzaffar:
Last build of Jemalloc was on our HTCondor ARM node where the page size is 4K. This broke the RelVal/unittests on our vocms-arm nodes where page size if 64K. I have disable the use of HTCondor ARM nodes for IBs/Releases/PR tests. The force build of jemalloc will build it on our vocms-arm nodes

### DIFF
--- a/jemalloc-common.file
+++ b/jemalloc-common.file
@@ -39,6 +39,7 @@ make %{makeprocesses}
 %{?PostBuild:%PostBuild}
 
 %install
+
 %{?PreInstall:%PreInstall}
 make install
 %if "%{n}" != "jemalloc"


smuzaffar:

smuzaffar:
+extenals

ok, unit tests/addon test passed

fwyzard:
Can we set the page size explicitly at configuration time, to avoid this
problem in the future ?

.A

smuzaffar:
@fwyzard ,  @iarspider is testing that

PR #9231, opened 2024-06-08
Add CLUE external library.

rovere:
This addresses the issue [here](https://github.com/cms-sw/cmssw/issues/45039).

### DIFF
--- a/clue.spec
+++ b/clue.spec
@@ -1,7 +1,7 @@
 ### RPM external clue 1.0.0
 ## NOCOMPILER
 
-%define git_commit V_1_0_0
+%define git_commit V_%(echo %{realversion} | tr '.' '_')
 
 Source: https://gitlab.cern.ch/kalos/%{n}/-/archive/%{git_commit}/%{n}-%{git_commit}.tar.gz
 Requires: alpaka
--- a/clue.spec
+++ b/clue.spec
@@ -1,0 +1,17 @@
+### RPM external clue 1.0.0
+## NOCOMPILER
+
+%define git_commit V_1_0_0
+
+Source: https://gitlab.cern.ch/kalos/%{n}/-/archive/%{git_commit}/%{n}-%{git_commit}.tar.gz
+Requires: alpaka
+
+%prep
+%setup -n %{n}-%{git_commit}
+
+%build
+
+%install
+cp -ar clueLib/include %{i}/include
+
+%post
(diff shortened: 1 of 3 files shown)


fwyzard:
please test

smuzaffar:
@rovere , can you please rebase your changes on top of latest IB/CMSSW_14_1_X/master branch and add `clue` dependency in `cmssw-tools.spec`.

rovere:
@smuzaffar let me know if that's what you asked for.

cmssw-tools.spec
```diff
@@ -146,6 +146,7 @@ Requires: dasgoclient
 Requires: dablooms
 Requires: zlib
 Requires: rivet
+Requires: clue
```
fwyzard: I don't remember why `alpaka` is in the Linux-only section, but could you move `clue` to the same spot, right after `alpaka` itself ?

rovere: @fwyzard you are right, since clue depends on alpaka.
Did it, thanks.

clue.spec
```diff
@@ -0,0 +1,17 @@
+### RPM external clue 1.0.0
+## NOCOMPILER
+
+%define git_commit V_1_0_0
```
smuzaffar: @rovere , can you please also change this to
```
%define git_commit V_%(echo %{realversion} | tr '.' '_')
```

this way in future we only need to update the version in the first line :-)

smuzaffar:
> @smuzaffar let me know if that's what you asked for.

looks good

clue.spec
rovere: Done, thanks!

fwyzard: You could also change the release tag in GitHub to `1.0.0` :-)

fwyzard:
please test

srimanob:
Hi @rappoccio @antoniovilela 
Should this external PR also be merged? Thanks.

fwyzard:
Yes, this external is needed to go with https://github.com/cms-sw/cmssw/pull/45178 :

[`RecoLocalCalo/HGCalRecProducers/plugins/BuildFile.xml`, line 49](https://github.com/rovere/cmssw/blob/d996d3df95d583c9abbdf3d8d0f7b32b83bf18f6/RecoLocalCalo/HGCalRecProducers/plugins/BuildFile.xml#L49)
```xml
  <use name="clue"/>
```

smuzaffar:
+externals

@cms-sw/orp-l2 , I am merging this for 14.1.X as it is needed by  which is already merged

PR #9245, opened 2024-06-17
[ROOT6] ROOT Update Master

aandvalenzuela:
ROOT [version](https://github.com/root-project/root/blob/master/core/foundation/inc/ROOT/RVersion.hxx) is still `6.33.01`.


aandvalenzuela:
please test

PR #9253, opened 2024-06-19
Update professor2 to 2.4.2 and fixing Python3 env

DickyChant:
This is the draft PR to update `professor2` to the most recent version 2.4.2.

Some changes of the code are due to the change of the installation commands (`python setup.py` to `python -m pip install`), while the `python` in shebang has been updated to `python3`

@shimashimarin

### DIFF
--- a/pip/iminuit.file
+++ b/pip/iminuit.file
@@ -1,0 +1,1 @@
+Requires: py3-numpy py3-matplotlib py3-scipy py3-ipywidgets
--- a/professor2.spec
+++ b/professor2.spec
@@ -3,14 +3,14 @@
 ## INITENV +PATH PROF_VERSION %{realversion}
 ## INCLUDE cpp-standard
 
-Source: http://www.hepforge.org/archive/professor/professor-professor-%{realversion}.tar.gz
+Source: git+https://gitlab.com/hepcedar/professor.git?obj=main/professor-%{realversion}&export=professor-%{realversion}&output=/professor-%{realversion}.tgz
 Requires: py3-matplotlib root yoda eigen py3-iminuit
 BuildRequires: py3-cython py3-pip
 
 Patch0: professor2-ppc64-flag-change
 
 %prep
-%setup -n professor-professor-%{realversion}
+%setup -n professor-%{realversion}
 
 %ifarch ppc64le
 %patch0 -p1
--- a/professor2.spec
+++ b/professor2.spec
@@ -4,7 +4,7 @@
 ## INCLUDE cpp-standard
 
 Source: http://www.hepforge.org/archive/professor/professor-professor-%{realversion}.tar.gz
-Requires: py3-matplotlib root yoda eigen
+Requires: py3-matplotlib root yoda eigen py3-iminuit
 BuildRequires: py3-cython py3-pip
 
 Patch0: professor2-ppc64-flag-change
--- a/professor2.spec
+++ b/professor2.spec
@@ -1,6 +1,5 @@
 ### RPM external professor2 2.4.2
 ## INITENV +PATH PYTHON3PATH %i/lib/python%{cms_python3_major_minor_version}/site-packages
-## INITENV +PATH PROF_VERSION %{realversion}
 ## INCLUDE cpp-standard
 
 Source: git+https://gitlab.com/hepcedar/professor.git?obj=main/professor-%{realversion}&export=professor-%{realversion}&output=/professor-%{realversion}.tgz
--- a/professor2.spec
+++ b/professor2.spec
@@ -1,15 +1,16 @@
-### RPM external professor2 2.3.3
+### RPM external professor2 2.4.2
 ## INITENV +PATH PYTHON3PATH %i/lib/python%{cms_python3_major_minor_version}/site-packages
+## INITENV +PATH PROF_VERSION %{realversion}
 ## INCLUDE cpp-standard
 
-Source: http://www.hepforge.org/archive/professor/Professor-%{realversion}.tar.gz
+Source: http://www.hepforge.org/archive/professor/professor-professor-%{realversion}.tar.gz
 Requires: py3-matplotlib root yoda eigen
 BuildRequires: py3-cython py3-pip
 
 Patch0: professor2-ppc64-flag-change
 
 %prep
-%setup -n Professor-%{realversion}
+%setup -n professor-professor-%{realversion}
 
 %ifarch ppc64le
 %patch0 -p1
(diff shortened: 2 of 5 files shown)


iarspider:
please test

shimashimarin:
@DickyChant thanks for preparing this PR. From this cms talk thread [1], there is a denpenence on `iminuit`. 

Can you also add `iminuit` in the compilation?

Thank you,
Jie

[1] https://cms-talk.web.cern.ch/t/prof2-cmssw-13-x-x/40274/5

DickyChant:
> @DickyChant thanks for preparing this PR. From this cms talk thread [1], there is a denpenence on `iminuit`.
> 
> Can you also add `iminuit` in the compilation?
> 
> Thank you, Jie
> 
> [1] https://cms-talk.web.cern.ch/t/prof2-cmssw-13-x-x/40274/5

Nice catch...

I had the impression that `iminuit` can be installed simply by `pip`, let me see if I can have some quick workaround...

DickyChant:
Just did a test and the compilation looks fine, sorry for this additional layer of entropy...

iarspider:
please test

professor2.spec
```diff
@@ -1,15 +1,16 @@
-### RPM external professor2 2.3.3
+### RPM external professor2 2.4.2
 ## INITENV +PATH PYTHON3PATH %i/lib/python%{cms_python3_major_minor_version}/site-packages
+## INITENV +PATH PROF_VERSION %{realversion}
 ## INCLUDE cpp-standard
 
-Source: http://www.hepforge.org/archive/professor/Professor-%{realversion}.tar.gz
-Requires: py3-matplotlib root yoda eigen
+Source: http://www.hepforge.org/archive/professor/professor-professor-%{realversion}.tar.gz
```
mseidel42:
```suggestion
Source: git+https://gitlab.com/hepcedar/professor.git?obj=main/professor-%{realversion}&export=professor-%{realversion}&output=/professor-%{realversion}.tgz
```

Hi Sitian, I'd suggest to move to git, as the new tarball name looks more like an accident

professor2.spec
```diff
@@ -22,13 +23,22 @@ sed -i -e 's|-std=c[+][+]11|-std=c++%{cms_cxx_standard}|' pyext/setup.py
...
 
+%define build_flags CPPFLAGS=-I${EIGEN_ROOT}/include/eigen3 PYTHON=$(which python3) PROF_VERSION=%{realversion} PYTHONPATH=./${PYTHON3_LIB_SITE_PACKAGES}
 %build
-make %{build_flags}
+make PREFIX=%{i} %{build_flags}
 
 %install
 make install PREFIX=%{i} %{build_flags}
-mv %{i}/${PYTHON3_LIB_SITE_PACKAGES}/%{n}-%{realversion}*.egg %{i}/${PYTHON3_LIB_SITE_PACKAGES}/%{n}
+mv %{i}/${PYTHON3_LIB_SITE_PACKAGES}/%{n}-%{realversion}*.dist* %{i}/${PYTHON3_LIB_SITE_PACKAGES}/%{n}
```
mseidel42, line 40: Just to check with the usage of `%{n}`, is the python directory really `professor2-2.4.2` not `professor-2.4.2`?

As prof1 got deprecated we may consider renaming the whole package to just `professor` now? (question to cmsdist team)

professor2.spec
```diff
@@ -1,15 +1,16 @@
...
-Requires: py3-matplotlib root yoda eigen
+Source: http://www.hepforge.org/archive/professor/professor-professor-%{realversion}.tar.gz
+Requires: py3-matplotlib root yoda eigen py3-iminuit
 BuildRequires: py3-cython py3-pip
 
 Patch0: professor2-ppc64-flag-change
 
 %prep
-%setup -n Professor-%{realversion}
+%setup -n professor-professor-%{realversion}
```
mseidel42:
```suggestion
%setup -n professor-%{realversion}
```

professor2.spec
DickyChant: Yeah one of the confusing part, let me try

professor2.spec
DickyChant: Yes I did checked that (I got stuck by the egg file for a long while!)

DickyChant:
Addressed 👍

professor2.spec
DickyChant: In addition, the professor executables are explicitly importing `professor2` (e.g. https://gitlab.com/hepcedar/professor/-/blob/main/bin/prof2-I?ref_type=heads#L31), so we might want to keep the name without changing all those importing.

professor2.spec
```diff
@@ -1,15 +1,16 @@
-### RPM external professor2 2.3.3
+### RPM external professor2 2.4.2
 ## INITENV +PATH PYTHON3PATH %i/lib/python%{cms_python3_major_minor_version}/site-packages
+## INITENV +PATH PROF_VERSION %{realversion}
```
smuzaffar: @DickyChant , I think `PROF_VERSION` is not a `PATH` like variable. I suggest to change this to 
```
## INITENV SET PROF_VERSION %{realversion}
```

If this is needed in CMSSW env too then I would suggest to add the following in professor2.xml too
```
<runtime name="PROF_VERSION"  value="@PROF_VERSION@"/>
```
```

Comment (DickyChant) [reply professor2.spec]: You are indeed correct! I was using this to help somewhere for the Makefile to set the version. But I might actually be wrong about this since it might not be useful at all. I just forget to clean it up...

I will correct this with some test on whether we need it or not.

File: scram-tools.file/tools/professor2/professor2.xml [hunk 5]
```diff
@@ -8,4 +8,6 @@
 <use name="yoda"/>
 <use name="eigen"/>
 <runtime name="PATH" value="$PROFESSOR2_BASE/bin" type="path"/>
+<runtime name="PYTHONPATH" value="$PROFESSOR2_BASE/lib/python3.9/site-packages" type="path"/>
```
smuzaffar: We avoid having individual packages adding their paths to `PYTHONPATH`, so please remove this line. `professor2` is  already a dependency of https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_14_1_X/master/python_tools.spec so its `$PROFESSOR2_BASE/lib/python3.9/site-packages` will be available at runtime.

scram-tools.file/tools/professor2/professor2.xml
```diff
@@ -8,4 +8,6 @@
...
+<runtime name="LD_LIBRARY_PATH" value="$LIBDIR" type="path"/>
```
smuzaffar: this is also not needed. `$PROFESSOR2_BASE/lib` will be in LD_LIBRARY_PATH in cmssw env. So please remove this line

DickyChant: Thanks! I was confused indeed since I need to add those paths manually to get things work if I just build professor2... 

Was going to ask but you already pointed out, thanks!

smuzaffar:
please test

smuzaffar:

smuzaffar:
+externals

pip/iminuit.file
```diff
@@ -0,0 +1 @@
+Requires: py3-numpy py3-matplotlib py3-scipy py3-ipywidgets
```
smuzaffar, line 1: @DickyChant , do `iminuit` actually need `py3-scipy py3-ipywidgets` dependencies? I do not see any ref to these package in `imniut`

```
> pwd
.../py3-iminuit/1.2-9722ab16622cfb12a6caf4da6d1edccd
> grep -R -i -E 'scipy|ipywidgets'  .
./lib/python3.9/site-packages/iminuit/__init__.py:# http://docs.scipy.org/doc/numpy/reference/generated/numpy.testing.Tester.test.html
```

professor2.spec
```diff
@@ -22,13 +22,22 @@ sed -i -e 's|-std=c[+][+]11|-std=c++%{cms_cxx_standard}|' pyext/setup.py
...
+for i in `ls bin`
+do
+   echo "bin/${i}"
+   sed -i -e 's|/usr/bin/env python|/usr/bin/env python3|' "bin/${i}"
+done
 
+%define build_flags CPPFLAGS=-I${EIGEN_ROOT}/include/eigen3 PYTHON=$(which python3) PROF_VERSION=%{realversion} PYTHONPATH=./${PYTHON3_LIB_SITE_PACKAGES}
 %build
-make %{build_flags}
+make PREFIX=%{i} %{build_flags}
```
smuzaffar, line 36: @DickyChant , I do not remember why we are not building `professor2` with `-j N` option. Do you know if there is any issue building it in parallel?

DickyChant: Nice point... I think there are only few g++ commands to be executed, not sure if it is because someone found it not helping too much... I simply didn't change...

pip/iminuit.file
DickyChant: nice catch I didn't investigate clearly but just follow https://scikit-hep.org/iminuit/ and adds as much optional dependencies as possible since I thought in our real world case have these in well defined sequence could make it more robust...

professor2.spec
DickyChant: I just locally added an arbitrary "-j 5", by reading the log I think at most 3 jobs would be executed in parallel, which already agrees with what I found from the [Makefile](https://gitlab.com/hepcedar/professor/-/blob/main/Makefile?ref_type=heads) by a quick scan that not much targets being executed during `make`. Though I fully agree that we would benefit from building it in parallel indeed, at least a little bit.

PR #9255, opened 2024-06-20
Update DD4Hep to v01-29-00

### LINKED ISSUE #9254: Migrate DD4hep external library to v1-29
In the recent version of DD4hep an important fix is introduced: order of instantiation of geometry objects is fixed to be stable, no random change of this order anymore. This fix may fix issue cms-sw/cmssw#40205 and guarantees the same order of instantiation of sensitive detectors.

It is possible that this update from v01-27-2 to v1-29 will keep regression - subject of tests. So, I would not exclude of update to this version of DD4hep directly in the master. If there will be problems we may first switch to G4VECGEOM branch.

The page of dd4hep:  https://github.com/AIDASoft/DD4hep/blob/master/doc/ReleaseNotes.md

iarspider:
Resolves #9254

### DIFF
--- a/dd4hep.spec
+++ b/dd4hep.spec
@@ -1,9 +1,9 @@
-### RPM external dd4hep v01-27-02
+### RPM external dd4hep v01-29-00
 ## INCLUDE compilation_flags
 ## INCLUDE compilation_flags_lto
 ## INCLUDE cpp-standard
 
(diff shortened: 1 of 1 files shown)


iarspider:
please test

smuzaffar:
+externals

looks good

PR #9269, opened 2024-06-27
Make array-bounds a non-error for GCC13 as well

iarspider:
This is already a non-error in GCC12

### DIFF
--- a/scram-tools.file/tools/gcc/env.sh
+++ b/scram-tools.file/tools/gcc/env.sh
@@ -40,8 +40,8 @@ GCC_CXXFLAGS="$GCC_CXXFLAGS -fvisibility-inlines-hidden"
 GCC_CXXFLAGS="$GCC_CXXFLAGS -fno-math-errno --param vect-max-version-for-alias-checks=50"
 GCC_CXXFLAGS="$GCC_CXXFLAGS -Xassembler --compress-debug-sections"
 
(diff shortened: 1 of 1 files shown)


fwyzard:
@iarspider would you know how many false positive are we getting from this flag ?

smuzaffar:
May be it is related to this https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105523 ? Can we try latest GCC 13 ?

iarspider:
> @iarspider would you know how many false positive are we getting from this flag ?

At least five: 2 in FWCore, 3 in FlexiStorage.



> May be it is related to this https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105523 ? Can we try latest GCC 13 ?

The bug you've linked doesn't seem relevant (it's mostly about embedded/low-level development). But updating gcc may help anyway.

smuzaffar:
GCC 13.3 update did not improve any thing (https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-826f08/40116/build-logs/ ) . 

@fwyzard , instead of suppressing this error for full cmssw via this PR, any objections if we just suppress it for `RecoTracker/PixelSeeding`  package (from where these 3 FlexiStorage errors are triggered)?

fwyzard:
No objections, though it might be better to put the suppression directly in the FlexiStorage header.

PR #9275, opened 2024-06-28
Buildrules: Use sort instead of python3 to calculcate the dependency order

smuzaffar:
Using `sort` to calculcate the order is much faster then using `python3`.  Using `sort` to calculate the dependency order of all product takes around 40s which python takes over 4mins


smuzaffar:
test parameters:
- full_cmssw = true

smuzaffar:
please test

smuzaffar:
+externals

PR #9280, opened 2024-07-02
Add cms-cat.spec

clelange:
This is basically a copy of https://github.com/cms-sw/cmsdist/blob/HEAD/cmssw-osenv.spec, moving files into `cat` instead of `common` and cloning the contents of https://gitlab.cern.ch/cms-analysis/services/cms.cern.ch-cat (this is a public repository).

Let me know if I e.g. need to add this anywhere else (I saw that `cmssw-tools.spec` contains the following line related to `cmssw-osenv`:

```
## INSTALL_DEPENDENCIES cmsLHEtoEOSManager gcc-fixincludes cmssw-osenv cms-git-tools SCRAMV2
```

### DIFF
--- a/cms-cat.spec
+++ b/cms-cat.spec
@@ -24,6 +24,14 @@ mv * %{i}/cat
 cd ${RPM_INSTALL_PREFIX}/%{pkgrel}
 mkdir -p ${RPM_INSTALL_PREFIX}/cat ${RPM_INSTALL_PREFIX}/etc/%{pkgname}
 
+# Check if a newer revision is already installed
+if [ -f ${RPM_INSTALL_PREFIX}/etc/%{pkgname}/version ] ; then
+  oldrev=$(cat ${RPM_INSTALL_PREFIX}/etc/%{pkgname}/version )
+  if [ ${oldrev} -ge %{fakerevision} ] ; then
+    exit 0
+  fi
+fi
+
 rsync -a --delete ${RPM_INSTALL_PREFIX}/%{pkgrel}/cat/ ${RPM_INSTALL_PREFIX}/cat/
 
 echo %{fakerevision} > ${RPM_INSTALL_PREFIX}/etc/%{pkgname}/version
--- a/cms-cat.spec
+++ b/cms-cat.spec
@@ -1,0 +1,42 @@
+### RPM cms cms-cat 240702.0
+## NOCOMPILER
+## NO_VERSION_SUFFIX
+
+# ***Do not change minor number of the above version. ***
+
+%define commit 0e3b60982ec088096fa67c46db5e9de2290ea5fb
+%define branch master
+# We do not use a revision explicitly, because revisioned packages do not get
+# updated automatically when there are dependencies.
+%define fakerevision %(echo %realversion | cut -d. -f1)
+Source0: git://gitlab.cern.ch/cms-analysis/services/cms.cern.ch-cat.git?obj=%{branch}/%{commit}&export=cms-cat&output=/cms-cat-%{commit}.tgz
+
+%prep
+%setup -n %{n}
+
+%build
+
+%install
+mkdir -p %{i}/cat
+mv * %{i}/cat
+
+%post
+cd ${RPM_INSTALL_PREFIX}/%{pkgrel}
+mkdir -p ${RPM_INSTALL_PREFIX}/cat ${RPM_INSTALL_PREFIX}/etc/%{pkgname}
+
+#Check if a newer revision is already installed
+if [ -f ${RPM_INSTALL_PREFIX}/etc/%{pkgname}/version ] ; then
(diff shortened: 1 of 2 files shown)


cms-cat.spec
```diff
@@ -0,0 +1,42 @@
...
+#Check if a newer revision is already installed
+if [ -f ${RPM_INSTALL_PREFIX}/etc/%{pkgname}/version ] ; then
+  oldrev=$(cat ${RPM_INSTALL_PREFIX}/etc/%{pkgname}/version )
+  if [ ${oldrev} -ge %{fakerevision} ] ; then
+    exit 0
+  fi
+fi
+
+for file in $(find . -name '*' -type f -path '*/cat/*') ; do
+  cp -f ${file} ${RPM_INSTALL_PREFIX}/${file}
```
smuzaffar: @clelange , looks like there is a directory under `cat`, so this command fails with
```
RPM installation stderr cms-cat:
cp: cannot create regular file '/tmp/muzaffar/w/./cat/combine_alias/exec_script.sh': No such file or directory
cp: cannot create regular file '/tmp/muzaffar/w/./cat/combine_alias/alias_script.sh': No such file or directory
```

If only thing which should go under `${RPM_INSTALL_PREFIX}/cat` is the `cat` directory then why not use
```
rsync -a --delete ${RPM_INSTALL_PREFIX}/%{pkgrel}/cat/ ${RPM_INSTALL_PREFIX}/cat/
```
this will allow to cleanup any old files which you might remove in new verions

smuzaffar:
@clelange , yes please add `cms-cat` in `cmsw-tools.spec` ( as INSTALL_DEPENDENCIES package), this will allow to deploy it on cvmfs when we install cmssw release

cms-cat.spec
smuzaffar: note that we should only use `--delete` here if only `cms-cat` copies its contents in `${RPM_INSTALL_PREFIX}/cat` . If you plan to have multiple packages copying their contents in `${RPM_INSTALL_PREFIX}/cat` then we should not use `--delete` option here

clelange: OK, I see. I think we'll only deploy from a single repository for now, so I'll change to the suggest `rsync` command and will delete the rest.

What's the purpose of the second `cp` for-loop

```shell
for file in $(find . -name '*' -type l -path '*/cat/*') ; do
  cp -pRf ${file} ${RPM_INSTALL_PREFIX}/${file}
done
```

I guess this should also go? I'll remove it for now.

cms-cat.spec
```diff
@@ -24,19 +24,6 @@ mv * %{i}/cat
 cd ${RPM_INSTALL_PREFIX}/%{pkgrel}
 mkdir -p ${RPM_INSTALL_PREFIX}/cat ${RPM_INSTALL_PREFIX}/etc/%{pkgname}
 
-#Check if a newer revision is already installed
-if [ -f ${RPM_INSTALL_PREFIX}/etc/%{pkgname}/version ] ; then
-  oldrev=$(cat ${RPM_INSTALL_PREFIX}/etc/%{pkgname}/version )
-  if [ ${oldrev} -ge %{fakerevision} ] ; then
-    exit 0
-  fi
-fi
```
smuzaffar: @clelange , you need to keep this above block otherwise an old version of cms-cat (e.g. when it will be rebuild/installed for an old release cycle) will override the new version files

clelange: OK, re-added. Sorry, I don't know how to test this myself, so I'm relying on you here.

smuzaffar:
please test

smuzaffar:
+externals

Failure are not related to this change. `cms-cat` was properly built. Its installation for PR is available under
```
> find /cvmfs/cms-ci.cern.ch/week0/PR_a87b81c1/cat
/cvmfs/cms-ci.cern.ch/week0/PR_a87b81c1/cat
/cvmfs/cms-ci.cern.ch/week0/PR_a87b81c1/cat/combine_alias
/cvmfs/cms-ci.cern.ch/week0/PR_a87b81c1/cat/combine_alias/exec_script.sh
/cvmfs/cms-ci.cern.ch/week0/PR_a87b81c1/cat/combine_alias/alias_script.sh
/cvmfs/cms-ci.cern.ch/week0/PR_a87b81c1/cat/README.md
/cvmfs/cms-ci.cern.ch/week0/PR_a87b81c1/cat/combine_env.sh
```

version control file for cms-cat is also installed properly
```
> cat /cvmfs/cms-ci.cern.ch/week0/PR_a87b81c1/etc/cms-cat/version
240702
```

clelange:
@smuzaffar Thanks for your help and merging. When will this be available on `/cvmfs/cms.cern.ch/cat`?

smuzaffar:
@clelange , normally a new release build deploys such packages but for now I  have installed `cms-cat` and it should appear under cvmfs in 10-15 mins.

clelange:
Thank you very much!

PR #9290, opened 2024-07-09
Let icpx find the gcc installation

fwyzard:


fwyzard:
@smuzaffar this is how we make ROCm / `hipcc` find the GCC headers. The same expanded arguments work for `icpx`, but I'm not sure if `COMPILER_PATH` will be defined properly ?

fwyzard:
FYI @AuroraPerego

smuzaffar:
> @smuzaffar this is how we make ROCm / `hipcc` find the GCC headers. The same expanded arguments work for `icpx`, but I'm not sure if `COMPILER_PATH` will be defined properly ?

`COMPILER_PATH` is set as runtime env by gcc tools, so it should be defined and points to our installation of GCC root directory. ` --gcc-toolchain=$(COMPILER_PATH)` should work.

By the way, are we doing to add some cmssw code which we might not be able to compile outside CERN ( note that icx installed under /cvmfs/projects.cern.ch/ will only be accessible without CERN) ?

fwyzard:
Eventually, yes, maybe.

But we are also discussing a legal agreement with Intel in order to redistribute ICX/ICPX.

smuzaffar:
please test

smuzaffar:
we need a new IB with root 6.30.08. lets wait for tonight IB

smuzaffar:
please test

smuzaffar:
+externals

fwyzard:
@smuzaffar looking at the areas created by the cms bot, I don't see any changes to `icx-cxxcompiler.xml` 🤷🏻‍♂️ 

I'll check the IB tomorrow.

smuzaffar:
`icx-*compiler.xml`  files are not selected by default and they are available under `config/toolbox/$SCRAM_ARCH/tools/available`. As for PR tests this `available` directory is not used so that is why you do not see `icx-*.xml` in the PR create installation area. 

It will be available in tonight's IBs

fwyzard:
OK, thanks !

fwyzard:
Looks good in today's IB.

PR #9320, opened 2024-07-25
[riscv] changes needed for using system-gcc and disabling pkgs

smuzaffar:
- Allow to disable packages which do not build on riscv64 e.g
  - valgrind, tensorflow, dasgoclient,  openloops
- Make sure that scram tools use the correct gcc/g++/gfortran in system gcc is used


smuzaffar:
please test

smuzaffar:
+externals

PR #9323, opened 2024-07-26
[cepgen] Disable BoostWrapper addon

smuzaffar:
We were not building CepGen BoostWrapper as our containers (cmssw-elX) do not have boost installed on system (we always want to use boost from cms externals). But on systems with boost is installed (and where we can not use our containers e.g. riscv) cepgen picks up Boost from system and then packaging fail as it complains about wrong boost_system lib dependency. 

Somehow I was not able to convince cepgen build system to use our boost (may be we are missing something in our boost?) so for now we explicitly disable cepgen BoostWrapper

### DIFF
--- a/cepgen.spec
+++ b/cepgen.spec
@@ -3,10 +3,11 @@
 Source: https://github.com/cepgen/cepgen/archive/refs/tags/%{realversion}.tar.gz
 
 BuildRequires: cmake ninja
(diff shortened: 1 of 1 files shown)


smuzaffar:
FYI @forthommel

smuzaffar:
please test

smuzaffar:
+externals

PR #9326, opened 2024-07-26
dasgoclient: Updated to version v02.05.52

smuzaffar:
New dasgoclient is also build for riscv64


smuzaffar:
please test

iarspider:
This version needs Cython 3.x, but we can't update Cython - this will break numpy, and we can't update numpy because it will break Tensorflow.

iarspider:
https://github.com/scikit-learn/scikit-learn/blob/1.5.0/pyproject.toml#L97

smuzaffar:
+externals

PR #9327, opened 2024-07-27
[llvm] add zstd deps to make sure it is not picked up from system

smuzaffar:
We build our software stack in containers where zstd package is not installed but on shared system e.g. riscv64 nodes zstd might be available on system and llvm then picks it up and linked against it. This change will allow to build llvm agaisnt out zstd


smuzaffar:
please test

smuzaffar:
+externals

PR #9351, opened 2024-08-05
do not create symlinks if system gcc is used

smuzaffar:
Thsi should avoid creating symlinks for `/usr/lib/gcc/$(uname -m)-redhat-linux/<version>/plugin` if system gcc is used


smuzaffar:
please test

smuzaffar:
+externals

PR #9241, opened 2024-06-11
[TF] Update TF v2.16.1

smuzaffar:
Update tensorflow version to 2.16.1

PR #9388, opened 2024-09-03
[TF] Update TF v2.16.1 (without libfft)

iarspider:
Alternative version of #9241


iarspider:
@cmsbuild please test for CMSSW_14_2_TF_X/el8_aarch64_gcc12

iarspider:
@cmsbuild please test for CMSSW_14_2_TF_X/el8_amd64_gcc12

iarspider:
@cmsbuild please test for CMSSW_14_2_TF_X/el8_aarch64_gcc12

iarspider:
> `test TestGeneratorInterfacePythia8ConcurrentGeneratorFilter had ERRORS`

Generators' web service acting up, not caused by this PR.

iarspider:
@cmsbuild please test for CMSSW_14_2_TF_X/el8_aarch64_gcc12

iarspider:
@cmsbuild please test for CMSSW_14_2_TF_X/el8_amd64_gcc12

iarspider:
@cmsbuild please test for CMSSW_14_2_TF_X/el8_amd64_gcc12
No idea why it failed to build - works fine locally

iarspider:
@cmsbuild please test for CMSSW_14_2_TF_X/el8_amd64_gcc12

iarspider:
@cmsbuild please test for CMSSW_14_2_TF_X/el8_amd64_gcc12

iarspider:
@smuzaffar the backend is overwritten by file `~/.keras/keras.json`. It seems to be present on all cmsbuild machines, with timestamp around the date when the node was created. This can be overridden with environment variable at import time

smuzaffar:
> @smuzaffar the backend is overwritten by file `~/.keras/keras.json`. It seems to be present on all cmsbuild machines, with timestamp around the date when the node was created. This can be overridden with environment variable at import time

@iarspider , can you please check how was this created? I don't remember we deployed it via puppet

iarspider:
> > @smuzaffar the backend is overwritten by file `~/.keras/keras.json`. It seems to be present on all cmsbuild machines, with timestamp around the date when the node was created. This can be overridden with environment variable at import time
> 
> @iarspider , can you please check how was this created? I don't remember we deployed it via puppet

It is created here: https://github.com/keras-team/keras/blob/v3.5.0/keras/src/backend/config.py#L221-L241. `_KERAS_DIR` defaults to `~/.keras`, but can be overwritten by setting `$KERAS_HOME`.

iarspider:
@cmsbuild please test for CMSSW_14_2_TF_X/el8_amd64_gcc12

iarspider:
@cmsbuild please test for CMSSW_14_2_TF_X/el8_aarch64_gcc12

smuzaffar:
enable gpu

smuzaffar:
@cmsbuild please test for CMSSW_14_2_TF_X/el8_amd64_gcc12

iarspider:
@smuzaffar since the tests depend on Tensorflow (not Keras), the environment was not set. Should I add an explicit dependency on keras to [PhysicsTools/TensorFlow](https://github.com/cms-sw/cmssw/blob/master/PhysicsTools/TensorFlow/BuildFile.xml)? I don't think we can handle circular dependency `keras ↔ tensorflow`

smuzaffar:
@iarspider , can you try running unit tests locally after setting the KERAS_BACKEND=tensorflow env? 
Note that `## INITENV SET KERAS_BACKEND tensorflow` only add/set this env via `init.*sh` file. For scram one need to update the xml file to set it. So if setting `KERAS_BACKEND=tensorflow ` allows to fix the gou unit tests then please add `KERAS_BACKEND=tensorflow ` in to one of tf xml files

iarspider:
@cmsbuild please test for CMSSW_14_2_TF_X/el8_amd64_gcc12

scram-tools.file/tools/tensorflow/tensorflow.xml
```diff
@@ -3,6 +3,7 @@
     <environment name="TENSORFLOW_BASE" default="@TOOL_ROOT@"/>
     <environment name="LIBDIR" default="$TENSORFLOW_BASE/lib"/>
     <environment name="INCLUDE" default="$TENSORFLOW_BASE/include"/>
+    <environment name="KERAS_BACKEND" default="tensorflow"/>
```
smuzaffar: it should be `<runtime ..../>` type variable ( see ROOTSYS as an example)
did you run test locally to see if gpu unit tests passed after setting this?

iarspider: I tested it by setting the environment variable manually (not via toolfile).

iarspider:
please abort

iarspider:
@cmsbuild please test for CMSSW_14_2_TF_X/el8_amd64_gcc12

scram-tools.file/tools/tensorflow/tensorflow.xml
smuzaffar: for me all the unit tests still fails with error
```
##Failure Location unknown## : Error
Test name: testHelloWorldCUDA::test
uncaught exception of type std::exception (or derived).
- An exception of category 'UnavailableAccelerator' occurred while
   [0] Calling tensorflow::setBackend()
Exception Message:
Cuda backend requested, NVIDIA GPU visible to cmssw, but not visible to TensorF
low in the job
```

iarspider: Some tests that were failing previously worked after setting `KERAS_BACKEND`. Yes, I saw these failures as well - I thought I missed some setup step to make them work (in a container started with `--nv` flag)

iarspider: e.g. testTFConstSession was failing with `ValueError: Unable to import backend : theano`, but after setting KERAS_BACKEND it passed.

iarspider: Could the failure be due to 12.4 not being an officially tested CUDA version for TF 2.16.1 (and even 2.17) - [link](https://www.tensorflow.org/install/source#gpu) lists 12.3 as officially tested version?

iarspider: Running `python -c "import tensorflow as tf; print(tf.config.list_physical_devices('GPU'))"` prints this message:

```
successful NUMA node read from SysFS had negative value (-1), but there must be at least one NUMA node, so returning NUMA node zero. See more at https://github.com/torvalds/linux/blob/v6.0/Documentation/ABI/testing/sysfs-bus-pci#L344-L355
```

and returns an empty list `[]`. I googled this message, and there are basically three solutions:
1) Run TensorFlow using official docker image
2) Install TensorFlow using `conda` and prebuilt wheels
3) Force-connect NUMA node (as suggested in the document that TensorFlow prints out), namely run `sudo echo 0 | sudo tee -a /sys/bus/pci/devices/0000\:06\:10.0/numa_node` after each reboot. But that requires `sudo` rights (and, I would imagine, not in container, but on the host).

smuzaffar: are you sure you started `cmssw-el8` with `--nv` option? For me the following command
```
python3 -c "import tensorflow as tf; print(tf.config.list_physical_devices('GPU'))"
```
 runs fine (both for this PR and TF_X Ibs) and return
```
[PhysicalDevice(name='/physical_device:GPU:0', device_type='GPU')]
```

iarspider: Now it works for me as well, weird.

iarspider:
ChatGPT suggests using `Session::ListDevices` to check if GPU is available, instead of `mutable_device_count()`:

```cpp
#include "tensorflow/core/public/session.h"
#include "tensorflow/core/protobuf/config.pb.h"
#include "tensorflow/core/platform/env.h"
#include "tensorflow/core/common_runtime/device_factory.h"

#include <iostream>

int main() {
    // Initialize a session
    tensorflow::Session* session;
    tensorflow::SessionOptions options;
    
    // Try to create a new session
    tensorflow::Status status = tensorflow::NewSession(options, &session);
    if (!status.ok()) {
        std::cerr << "Error creating session: " << status.ToString() << std::endl;
        return -1;
    }

    // Retrieve the list of available devices
    std::vector<tensorflow::DeviceAttributes> devices;
    status = session->ListDevices(&devices);
    if (!status.ok()) {
        std::cerr << "Error listing devices: " << status.ToString() << std::endl;
        return -1;
    }

    // Check if any GPU devices are available
    bool gpu_available = false;
    for (const auto& device : devices) {
        std::cout << "Device name: " << device.name() << ", type: " << device.device_type() << std::endl;
        if (device.device_type() == "GPU") {
            gpu_available = true;
        }
    }

    if (gpu_available) {
        std::cout << "GPU is available and can be used." << std::endl;
    } else {
        std::cout << "No GPU devices are available." << std::endl;
    }

    // Clean up
    session->Close();
    delete session;

    return 0;
}
```

PR #9389, opened 2024-09-03
cmssw-osenv: added extra paths to mount; set LC_ALL=C

### LINKED ISSUE #17: Unset locales
I have the following locales inherited from my main machine, since SSH forwards them by default:
```
LANG=fr_BE.UTF-8
LC_ADDRESS=fr_BE.UTF-8
LC_IDENTIFICATION=fr_BE.UTF-8
LC_MEASUREMENT=fr_BE.UTF-8
LC_MONETARY=fr_BE.UTF-8
LC_NAME=fr_BE.UTF-8
LC_NUMERIC=fr_BE.UTF-8
LC_PAPER=fr_BE.UTF-8
LC_TELEPHONE=fr_BE.UTF-8
LC_TIME=fr_BE.UTF-8
```
I want to keep seeing messages in French if possible, so I do not want to unset them.

This is usually not an issue since every possible locale is enabled on a typical interactive node. However, the `cmssw-*` containers have only the following:
```
$ cmssw-cc7 -- locale -a
C
POSIX
en_US.utf8
```

This causes `perl` to complain quite loudly:
```
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
        LANGUAGE = (unset),
        LC_ALL = (unset),
        LC_PAPER = "fr_BE.UTF-8",
        LC_ADDRESS = "fr_BE.UTF-8",
        LC_MONETARY = "fr_BE.UTF-8",
        LC_NUMERIC = "fr_BE.UTF-8",
        LC_TELEPHONE = "fr_BE.UTF-8",
        LC_IDENTIFICATION = "fr_BE.UTF-8",
        LC_MEASUREMENT = "fr_BE.UTF-8",
        LC_TIME = "fr_BE.UTF-8",
        LC_NAME = "fr_BE.UTF-8",
        LANG = "fr_BE.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
```
This is shown in particular for every `scram` call.

Generating all locales like is done for interactive nodes would make the container very large for little gain. Instead, I think the script should reset the locale in the container to the default.

### LINKED ISSUE #18: Mount additional filesystems
`cmssw-env` hard-codes a list of filesystems to be mounted in the container:
https://github.com/cms-sw/cmssw-osenv/blob/35500d47783eb0aa2b9ed7bd4a71e4ee0fe90612/cmssw-env#L72

(The intent seems to be to mount them only if they exist, but they are always mounted.)

This list is complete at CERN, but other sites have more. For instance at DESY, one would want to test for:
```
/asap3
/beegfs
/gpfs
/nfs
/pnfs
```
These are all mass storage filesystems that the user may want to access. While users could modify `cmssw-env` for their site or set `MOUNT_POINTS`, this requires extra work. To my knowledge, this is also undocumented.

I propose to invert the logic: instead of whitelisting a set of known mount points, mount all folders in `/` that are not in the images:
```sh
# Mount all filesystems in / except for things already in the image.
blacklist="bin dev etc lib lib64 mnt opt proc root run sbin singularity sys usr var"
mounts=$(comm -23 <(ls /) <(echo $blacklist | tr " " "\n" | sort))  # https://stackoverflow.com/a/11964477
for dir in $mounts; do
    MOUNT_POINTS="${MOUNT_POINTS},${dir}"
done
```
This would be a more general solution than maintaining a hard-coded list of mount points and would work out of the box in many cases.

smuzaffar:
Fixes https://github.com/cms-sw/cmssw-osenv/issues/18 
Fixes https://github.com/cms-sw/cmssw-osenv/issues/17

### DIFF
--- a/cmssw-osenv.spec
+++ b/cmssw-osenv.spec
@@ -1,10 +1,10 @@
-### RPM cms cmssw-osenv 240816.0
+### RPM cms cmssw-osenv 240903.0
 ## NOCOMPILER
 ## NO_VERSION_SUFFIX
 
 # ***Do not change minor number of the above version. ***
 
-%define commit 35500d47783eb0aa2b9ed7bd4a71e4ee0fe90612
+%define commit 9e83f8bfe574077556190dadff1a9aeed2dfb00b
 %define branch master
 # We do not use a revision explicitly, because revisioned packages do not get
 # updated automatically when there are dependencies.


smuzaffar:
please test

smuzaffar:
+externals

testing locally the new `cmssw-el9` does not show any perl warnings. Also if any of `/asap3 /beegfs /gpfs /nfs /pnfs` exists then they are properly mounted

PR #9404, opened 2024-09-17
[TF][xz] version 5.6.2 and build with -fnofast-math

smuzaffar:
Used `-fno-fast-math` to build  `xz`  (see https://github.com/cms-sw/cmssw/issues/41773). 
Looks like new `xz` version `5.6.2` is relatively faster than `5.2.5`. ROOT bench mark test https://github.com/cms-sw/cmssw/issues/41773#issuecomment-2328555041 timings are below (avg of 5 runs)
- `xz 5.2.5 with -Ofast`:  8.251s
- `xz 5.6.2 with -Ofast -fno-fast-math`: 8.1s

Generated root file size is exactly same for both.

### DIFF
--- a/xz.spec
+++ b/xz.spec
@@ -1,10 +1,5 @@
-### RPM external xz 5.2.5
-
-%define tag 50f585dc3b7b9b94b6b7f7a4c29903602d1e2a2d
-%define branch cms/v%{realversion}
-%define github_user cms-externals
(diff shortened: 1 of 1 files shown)


makortel:
On a quick look of the release notes the 5.6.2 should also be free of the backdoor.

smuzaffar:
closing it in favor of #9410  ( which is for default branch )

PR #9409, opened 2024-09-17
Set default keras backend at runtime

iarspider:


tensorflow.spec
```diff
@@ -1,4 +1,5 @@
 ### RPM external tensorflow 2.16.1
+### INITENV SETV KERAS_BACKEND tensorflow
```
smuzaffar: @iarspider , previously this was add in the keras.file ( https://github.com/cms-sw/cmsdist/pull/9388/commits/1ba91fcf806f9fd4068e54ff36832bc0d0624482 ) . I do not remember which packages actually needed this ... do you? Was it one of `tfaot` package? Though it does not make any sense but I think it is safe to move this to `pip/keras.file`

iarspider: Sure.

iarspider:
please test

iarspider:
please test for CMSSW_14_2_TF_X/el8_amd64_gcc12

iarspider:
please test for CMSSW_14_2_TF_X/el8_aarch64_gcc12

smuzaffar:
looks good to go in TF IB

PR #9166, opened 2024-04-29
simplify micro arch selection; set default micro arch to x86-64-v2

smuzaffar:
This PR simplify the selection of default micro archtitecture which is now defined in `cmsdist/microarch_flags.file` file. Every package which needs to be enable micro-instraction should include it and pass `%{selected_microarch}` to the its build system.

This also updates the default micro arch to `x86-64-v2` ( previously it was `sse3`)

PR #9434, opened 2024-09-24
[TF] Do not use mcpu/mtune native if CMS arch_build_flags are not set

smuzaffar:
do not pass `--copt=-mcpu=native --copt=-mtune=native` to TF build if CMS `arch_build_flags` is not set. For `x86_64` though we set `--copt=-march=x86-64-v2` but as  `arch_build_flags` is not set for `x86_64` so we also end up passing  `--copt=-mcpu=native --copt=-mtune=native` . This can cause compiler to optimize tf build for the machine where it was build .

https://github.com/cms-sw/cmsdist/pull/9166/ PR has accidentally included `--copt=-mcpu=native --copt=-mtune=native` for `x86_64`. Before that `--copt=-mcpu=native --copt=-mtune=native` were added for non-x86_64 arch if  `arch_build_flags` was not set for those archs.


smuzaffar:
please test

smuzaffar:
+externals

PR #9451, opened 2024-10-03
[ROOT6] Updated root to tip of branch master

smuzaffar:


smuzaffar:
please test

smuzaffar:
@makortel @Dr15Jones , looks like `RNTupleCollectionWriter.hxx` was removed https://github.com/root-project/root/commit/0b939e261ab06cdea141ad30bef87b9c61454293 . We  use `RNTupleCollectionWriter` here https://github.com/search?q=repo%3Acms-sw%2Fcmssw+RNTupleCollectionWriter&type=code

Dr15Jones:
We don't use that for anything so it would be OK to break initially.

smuzaffar:
@jblomer  , can you please propose/help with changes in https://github.com/search?q=repo%3Acms-sw%2Fcmssw+RNTupleCollectionWriter&type=code ( to replace RNTupleCollectionWriter and MakeCollection ) ?

jblomer:
We unfortunately realized that the current collection writer and collection field breaks the RNTuple abstractions in ways that are hard to fix. So we decided to replace it altogether by "untyped records" and "untyped collections". The only use of the collection writer was the representation of TTree leaf count arrays in RNTuple. For the RNTupleImporter, we migrated to untyped records and collections (see: https://github.com/root-project/root/pull/16478).

For the NanoAOD outputter, the change will likewise be manageable but more than a two-liner. I'm happy to help but I think this should be discussed with @dan131riley.

Dr15Jones:
@smuzaffar given the RNTuple code for NanoAOD only exists to do tests, I suggest we just comment out

https://github.com/cms-sw/cmssw/blob/14a12c2bcc115ca89bccfeb2fc7aacde53111773/PhysicsTools/NanoAOD/plugins/BuildFile.xml#L40-L42

until we actually plan on using RNTuple for NanoAOD.

makortel:
I concur, for now it is more important to get https://github.com/root-project/root/pull/16595 than to keep the `NanoAODRNTupleOutputModule` functional (it needs major updates anyway that @dan131riley is working on).

smuzaffar:
please test with https://github.com/cms-sw/cmssw/pull/46268 for CMSSW_14_2_ROOT6_X

smuzaffar:
+externals
good to go in root6x ibs

PR #9469, opened 2024-10-17
add CMS_X_SANITIZER macros

smuzaffar:
These macros allows to check is code is compiled with `undefined` or `address` sanitizer


smuzaffar:
please test

PR #9484, opened 2024-10-28
[ROOT632] Updated root to tip of branch v6-32-00-patches

iarspider:
Technical change, should we skip @smuzaffar ? [changes](https://github.com/root-project/root/compare/9310258d4c...26f27fdf1b)


smuzaffar:
yes , we can skip this update

PR #9503, opened 2024-11-06
Update Intel tools to oneAPI 2025.0

fwyzard:

PR #9510, opened 2024-11-08
cleanup intel-vtune; remove duplicate toolfile

smuzaffar:
This PR contains the following changes (follow up on https://github.com/cms-sw/cmsdist/pull/9503#discussion_r1832188693)
- Remove duplicate `intel-vtune.xml` generation from the `intel-vtue.spec`
- Added `NO_VERSION_SUFFIX` to avoid adding any `-suffix` in version sothat xml file contains the realversion
- Added `NO_AUTO_DEPENDENCY` to avoid generation extra dep env file for `intel-vtune` stub package
- Set `INTEL_VTUNE_INSTALLDIR` as env pointing to actual pathof `intel-vtune` . This is then used by `intel-vtune.xml` 

FYI @fwyzard

### DIFF
--- a/scram-tools.file/tools/intel-vtune/intel-vtune.xml
+++ b/scram-tools.file/tools/intel-vtune/intel-vtune.xml
@@ -1,7 +1,7 @@
 <tool name="intel-vtune" version="@TOOL_VERSION@">
   <info url="https://www.intel.com/content/www/us/en/developer/tools/oneapi/vtune-profiler.html"/>
   <client>
-    <environment name="INTEL_VTUNE_BASE" default="@TOOL_ROOT@"/>
+    <environment name="INTEL_VTUNE_BASE" default="@INTEL_VTUNE_INSTALLDIR@"/>
     <environment name="BINDIR" default="$INTEL_VTUNE_BASE/bin64"/>
   </client>
   <runtime name="PATH" value="$INTEL_VTUNE_BASE/bin64" type="path"/>
--- a/intel-vtune.spec
+++ b/intel-vtune.spec
@@ -1,31 +1,14 @@
 ### RPM external intel-vtune 2025.0
 ## NOCOMPILER
+## NO_AUTO_DEPENDENCY
+## NO_VERSION_SUFFIX
 
 %define year %(echo %realversion | cut -d. -f1)
+
+## INITENV SET INTEL_VTUNE_INSTALLDIR /cvmfs/projects.cern.ch/intelsw/oneAPI/linux/x86_64/%{year}/vtune/%{realversion}
 
 Source: none
 
 %prep
-
 %build
-
 %install
-mkdir -p %i/etc/scram.d
-cat << \EOF_TOOLFILE >%i/etc/scram.d/intel-vtune.xml
-<tool name="intel-vtune" version="%{realversion}">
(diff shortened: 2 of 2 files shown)


smuzaffar:
please test

scram-tools.file/tools/intel-vtune/intel-vtune.xml
```diff
@@ -1,7 +1,7 @@
 <tool name="intel-vtune" version="@TOOL_VERSION@">
   <info url="https://www.intel.com/content/www/us/en/developer/tools/oneapi/vtune-profiler.html"/>
   <client>
-    <environment name="INTEL_VTUNE_BASE" default="@TOOL_ROOT@"/>
+    <environment name="INTEL_VTUNE_BASE" default="@INTEL_VTUNE_INSTALLDIR@"/>
```
fwyzard, line 4: is this needed because of `NO_AUTO_DEPENDENCY` ?

smuzaffar: No, not due to `NO_AUTO_DEPENDENCY`.  `TOOL_ROOT` points to the actual install path of a package in this case it will be `/our/cvmfs/install/dir/<arch>/externals/inten-vtune/<version>` but we want `INTEL_VTUNE_BASE` to point to hardcoded `/cvmfs/projects.cern.ch/intelsw/oneAPI/....` path. One way to achieve this is to override `TOOL_ROOT` in the `etc/prodile.d/init.sh` e.g. we could have done the following the intel-vtune.spec `%post`
```
echo "INTEL_VTUNE_ROOT=''/cvmfs/projects.cern.ch/intelsw/oneAPI/..." > $RPM_INSTALL_PREFIX/%{pkgrel}/etc/profile.d/init.sh
```

but as packaging system also use `<TOOL>_ROOT` so I would avoid pointing  `<TOOL>_ROOT` to something else. That is why spec file now saves `INTEL_VTUNE_INSTALLDIR /cvmfs/projects.cern.ch/intelsw/oneAPI/linux/x86_64/%{year}/vtune/%{realversion}` in the `init.sh` file which is then used by xml file

smuzaffar:
+externals

intel-vtune tool looks good now
```
> cat /cvmfs/cms-ci.cern.ch/week0/cms-sw/cmsdist/9510/42677/CMSSW_14_2_X_2024-11-07-1100/config/toolbox/el8_amd64_gcc12/tools/available/intel-vtune.xml 
<tool name="intel-vtune" version="2025.0">
  <info url="https://www.intel.com/content/www/us/en/developer/tools/oneapi/vtune-profiler.html"/>
  <client>
    <environment name="INTEL_VTUNE_BASE" default="/cvmfs/projects.cern.ch/intelsw/oneAPI/linux/x86_64/2025/vtune/2025.0"/>
    <environment name="BINDIR" default="$INTEL_VTUNE_BASE/bin64"/>
  </client>
  <runtime name="PATH" value="$INTEL_VTUNE_BASE/bin64" type="path"/>
  <runtime name="VTUNE_PROFILER_DIR" value="$INTEL_VTUNE_BASE"/>
  <runtime name="VTUNE_PROFILER_2025_DIR" value="$INTEL_VTUNE_BASE"/>
</tool>
```

smuzaffar:
merging it for 11h00 IB. Though relvals/addon tests are still running but as intel-vtune tool is not setup by default so this change is not going to effect relvals/addon

PR #9534, opened 2024-11-25
Added revision number for toolfiles

smuzaffar:
This PR adds tool file revision number.  This will allow to properly rebuild cmssw packages which depend on the tool. 
Revision number should be incremented for every change in toolfile itself.


smuzaffar:
please test

smuzaffar:
+externals

PR #9540, opened 2024-11-27
update cms-tfaot: added revision in toolfiles

smuzaffar:
Added missing revision in to cms-tfaot toolfiles


smuzaffar:
please test

PR #9563, opened 2024-12-10
Update NVIDIA gdrcopy to v2.4.3

fwyzard:
Bug fixes:
  - fix `NVIDIA_IS_OPENSOURCE` detection when compile with NVIDIA driver version 545 or newer;
  - fix compile error in `gdrdrv` when compiling on RHEL9.5.


fwyzard:
please test

PR #9577, opened 2024-12-13
Revert build rules: python import fix, prefer generated cfipython files

smuzaffar:
This reverts the change https://github.com/cms-sw/cmssw-config/commit/91e7bf620361fad2a8427d018e8839c50c183263


smuzaffar:
please test

smuzaffar:
+externals
this was already part of IB, so merging it and starting an IB . Once a full IB is available then I will build a patch IB to make sure that `TestIntegrationParameterSet` does not fail

PR #9605, opened 2025-01-09
Update valgrind to 3.24

slava77:
It looks like this recent version adds support (via LibVEX) to more AVX instructions. https://valgrind.org/docs/manual/dist.news.html

I couldn't run our reco production job with x86-64-v3 under valgrind v3.23 from the default cvmfs distribution, while it runs with a locally build 3.24


slava77:
@cmsbuild please test

smuzaffar:
+externals

PR #9610, opened 2025-01-13
[DEVEL]Build rules non default min microarch

smuzaffar:
This contains build rules improvement. By default, during the build phase, we use the default micro-arch edm tools to 
- register edm plugins
- generate edm plugin configuration
- Run edm class version checks

This break the build if the build node does not support default micro-arch e.g. building on node with x86-64-v2  only support for CMSSW releases where x86-64-v3 is default micro-arch.

Also we only build cuda/rocm code for default micro-arch and then link it for additional micro-archs. This works as long as default micro-arch is subset of all addtional micro-archs but fails otherwise.

New build rules has the following changes
- Use min supported micro-arch edm tools to register edm plugins, generate edn plugin configs and edm class version checks
- Build cuda/rocm (+ alpaka cuda/rocm backend) once using the min supported micro-arch flags 

Lets test the changes in DEVEL IBs first and then we can integrate it in main branch

### DIFF
--- a/scram-project-build.file
+++ b/scram-project-build.file
@@ -52,7 +52,7 @@ BuildRequires: dwz
 %endif
 
 %if "%{?configtag:set}" != "set"
-%define configtag       V09-07-00
+%define configtag       V09-07-01
 %endif
 
 %if "%{?buildarch:set}" != "set"
--- a/scram-project-build.file
+++ b/scram-project-build.file
@@ -52,7 +52,7 @@ BuildRequires: dwz
 %endif
 
 %if "%{?configtag:set}" != "set"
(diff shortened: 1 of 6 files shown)


smuzaffar:
test parameters:
- full_cmssw = true
- enable = multi-microarchs

smuzaffar:
please test for CMSSW_15_0_DEVEL_X/el8_amd64_gcc12

smuzaffar:
please test for CMSSW_15_0_DEVEL_X/el8_amd64_gcc12

smuzaffar:
lets merge it for devel IB. In case of massive failure I will revert it

PR #9611, opened 2025-01-13
refresh edm plugin cache for extra micro-archs too

smuzaffar:
Refresh edm plugins for extra micro-archs too


smuzaffar:
please test for CMSSW_15_0_DEVEL_X/el8_amd64_gcc12

PR #9617, opened 2025-01-16
Buildrule: fixes when extra micro-arch is subset of default microarch

smuzaffar:
This contains build rules improvement. By default, during the build phase, we use the default micro-arch edm tools to

    register edm plugins
    generate edm plugin configuration
    Run edm class version checks

This break the build if the build node does not support default micro-arch e.g. building on node with x86-64-v2 only support for CMSSW releases where x86-64-v3 is default micro-arch.

Also we only build cuda/rocm code for default micro-arch and then link it for additional micro-archs. This works as long as default micro-arch is subset of all addtional micro-archs but fails otherwise.

New build rules has the following changes

    Use min supported micro-arch edm tools to register edm plugins, generate edn plugin configs and edm class version checks
    Build cuda/rocm (+ alpaka cuda/rocm backend) once using the min supported micro-arch flags

These changes have been tested in DEVEL IBs

backport of #9610

### DIFF
--- a/scram-project-build.file
+++ b/scram-project-build.file
@@ -52,7 +52,7 @@ BuildRequires: dwz
 %endif
 
 %if "%{?configtag:set}" != "set"
-%define configtag       V09-06-11
+%define configtag       V09-07-01
 %endif
 
 %if "%{?buildarch:set}" != "set"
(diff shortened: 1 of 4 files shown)


smuzaffar:
please test

smuzaffar:
test parameters:
- addpkg = DataFormats, FWCore, Geometry

smuzaffar:
please test

smuzaffar:
+externals

PR #9667, opened 2025-02-02
Test the  g++ `-fimplicit-constexpr` option

fwyzard:
From the [GCC 12 manual](https://gcc.gnu.org/onlinedocs/gcc-12.4.0/gcc/C_002b_002b-Dialect-Options.html#index-fimplicit-constexpr):
> #### `-fimplicit-constexpr`
> Make inline functions implicitly constexpr, if they satisfy the requirements for a constexpr function. This option can be used in C++14 mode or later. This can result in initialization changing from dynamic to static and other optimizations.

@makortel [noted](https://github.com/cms-sw/cmssw/issues/47239#issuecomment-2629356875) that this flag makes the compilation very slow and memory consuming with GCC 12.x and 13.2, but is should be usable with GCC 13.3.


fwyzard:

smuzaffar:
test parameters:
- full_cmssw = true

smuzaffar:

lets rebuild full cmssw

fwyzard:
please test for el8_amd64_gcc13 with cms-sw/cmssw#47241

fwyzard:
please test with cms-sw/cmssw#47241 for el8_amd64_gcc13

PR #9679, opened 2025-02-09
Enable new CUDA visibility flags

fwyzard:
CUDA 12.8.0 introduce two flags that restrict the linkage and visibility of `__global__` functions (kernels) and global device variables.

Enabling them should make the CUDA device code and kernel launch interface more self-contained, potentially reducing conflicts across shared libraries.

The new behaviour will be the default in CUDA 13.0.

### DIFF
--- a/cuda-flags.file
+++ b/cuda-flags.file
@@ -26,6 +26,10 @@
 # allow __host__, __device__ attributes in lambda declaration
 %define nvcc_flags_lambda --extended-lambda
 
+# - in whole-program compilation mode (-rdc=false), force 'static' linkage for host side stub functions generated for __global__ function templates
+# - mark __global__ functions and function templates, and __constant__, __device__ and __managed__ variables and variable templates, with __attribute__((visibility("hidden")))
+%define nvcc_visibility_flags --static-global-template-stub=true --device-entity-has-hidden-visibility=true
+
 # build support for the various compute architectures
 %define nvcc_flags_cuda_archs %(echo $(for ARCH in %cuda_arch; do echo "-gencode arch=compute_$ARCH,code=[sm_$ARCH,compute_$ARCH]"; done)) -Wno-deprecated-gpu-targets
 
@@ -46,5 +50,5 @@
 
 # collect all CUDA flags
 %define nvcc_cuda_flags \
-  %{nvcc_flags_stdcxx} %{nvcc_flags_opt} %{nvcc_flags_debug} %{nvcc_flags_constexpr} %{nvcc_flags_lambda} \
+  %{nvcc_flags_stdcxx} %{nvcc_flags_opt} %{nvcc_flags_debug} %{nvcc_flags_constexpr} %{nvcc_flags_lambda} %{nvcc_visibility_flags} \
   %{nvcc_flags_cuda_archs} %{nvcc_flags_cuda_diag_suppress} %{nvcc_flags_cudafe_diag} %{nvcc_flags_gnu_version} %{nvcc_flags_cudart}


fwyzard:
enable gpu

fwyzard:
please test

fwyzard:
OK, onnxruntime does not like the new mode:

> ```
> /data/cmsbld/jenkins/workspace/ib-run-pr-tests/testBuildDir/BUILD/el8_amd64_gcc12/external/onnxruntime/1.20.1-e35ea1b21c6163717a6cadfc46ef1910/onnxruntime-1.20.1/onnxruntime/contrib_ops/cuda/bert/fastertransformer_decoder_attention/decoder_masked_multihead_attention_impl.h(24):
> error #20280-D:
> when "-static-global-template-stub=true" in whole program compilation mode ("-rdc=false"), a __global__ function template instantiation or specialization ("onnxruntime::contrib::cuda::masked_multihead_attention_kernel<float, (int)64, (int)4, (int)16, (int)64> ") must have a definition in the current translation unit.
> To resolve this issue, either use separate compilation mode ("-rdc=true"), or explicitly set "-static-global-template-stub=false" (but see nvcc documentation about downsides of turning it off)
> ```

fwyzard:
please test

onnxruntime.spec
```diff
@@ -36,7 +36,7 @@ cmake ../%{n}-%{realversion}/cmake -GNinja \
    -Donnxruntime_CUDNN_HOME="${CUDNN_ROOT}" \
    -Donnxruntime_NVCC_THREADS=1 \
    -DCMAKE_CUDA_ARCHITECTURES=$(echo %{cuda_arch} | tr ' ' ';' | sed 's|;;*|;|') \
-   -DCMAKE_CUDA_FLAGS="-DTHRUST_IGNORE_DEPRECATED_API -DCUB_IGNORE_DEPRECATED_API -Wno-deprecated-gpu-targets --static-global-template-stub=false -cudart shared" \
+   -DCMAKE_CUDA_FLAGS="-DTHRUST_IGNORE_DEPRECATED_API -DCUB_IGNORE_DEPRECATED_API %{nvcc_cuda_flags}" \
```
smuzaffar: @fwyzard , I would suggest to add 
```
CUDA_FLAGS=$(echo "-DTHRUST_IGNORE_DEPRECATED_API -DCUB_IGNORE_DEPRECATED_API %{nvcc_cuda_flags}" | tr '\n' ' ')
```
before the `cmake` command and then use `-DCMAKE_CUDA_FLAGS="${CUDA_FLAGS}"` here

fwyzard: thanks

fwyzard:
please test

onnxruntime.spec
```diff
@@ -21,6 +21,7 @@ rm -rf ../build; mkdir ../build; cd ../build
 USE_CUDA=OFF
 if [ "%{cuda_gcc_support}" = "true" ] ; then
   USE_CUDA=ON
+  CUDA_FLAGS=$(echo "-DTHRUST_IGNORE_DEPRECATED_API -DCUB_IGNORE_DEPRECATED_API %{nvcc_cuda_flags}" | tr '\n' ' ')
```
fwyzard: This is actually includes too many flags, some are already set by `CMAKE_CUDA_ARCHITECTURES=...` and `CMAKE_CUDA_RUNTIME_LIBRARY=Shared`.

fwyzard:
please test

smuzaffar:
please test

smuzaffar:
@fwyzard , should we consider this for integration in 16.0.X ?

fwyzard:
Sure, I don't think there should be any downsides.

Maybe let me try to run some GPU benchmarks first.

fwyzard:
please test

to rebuild the externals

fwyzard:
@smuzaffar I've compared the performance of the HLT with and without these changes, and the impact on the performance seems negligible:

#### `CMSSW_16_0_X_2025-10-26-0000`
  - 1099.2 ±   1.5 ev/s
  - 1099.9 ±   1.8 ev/s

#### `CMSSW_16_0_X_2025-10-26-0000` + #9679
  - 1103.5 ±   4.9 ev/s
  - 1100.5 ±   2.7 ev/s

fwyzard:
So it should be fine to merge this.

smuzaffar:
+externals

mandrenguyen:
+1

smuzaffar:
@fwyzard , looks like this broke the IBs . I will check why we did not get the build errors during PR tests. For now I propose to revert this change unless you have a quick fix

PR #9680, opened 2025-02-10
Disable arry-bound errors for GCC 14

smuzaffar:
Lets convert the array-bound errors to warnings for GCC 14.

I have opened https://gcc.gnu.org/bugzilla/show_bug.cgi?id=118817  issue , once that is fixed then we can revert this change


smuzaffar:
test parameters:
- full_cmssw = true

smuzaffar:

smuzaffar:
Lets gets this in fir gcc 14
 Only three packages with errors are remaining now

PR #9681, opened 2025-02-10
[CUDART_X] Make cuda-runtime provide libnvrtc-builtins

iarspider:
This fixes build failure of onnxruntime [link](https://cmssdt.cern.ch/jenkins/job/build-any-ib/200843/):

```
Checking local path dependency for rpm package external+onnxruntime+1.20.1-5b770e6b075607a7a6ec9767ece84472 just build.
RPM installation stderr onnxruntime:
error: Failed dependencies:
	libnvrtc-builtins.so.12.8()(64bit) is needed by external+onnxruntime+1.20.1-5b770e6b075607a7a6ec9767ece84472-1-1.x86_64

Failed to install RPM for onnxruntime
```

### DIFF
--- a/cuda-runtime.spec
+++ b/cuda-runtime.spec
@@ -6,6 +6,8 @@ Source99: install-cuda.sh
 ## INCLUDE cuda-version
 ### RPM external cuda-runtime %{cuda_version}
 
+Provides: libnvrtc-builtins.so.12.8()(64bit)
+
 %install
 #Copy runtime libs
 mkdir -p %i/lib64/stubs


iarspider:
please test for CMSSW_15_1_CUDART_X

iarspider:
please test

fwyzard:
I don't understand, this is not actually providing any additional library or symbols, just making the RPM package fake-provide something.

How does that help ?

smuzaffar:
@iarspider , cuda 12.8 added new `libnvrtc-builtins.so`  which is needed by our cuda-runtime distribution. I would suggest to add `nvrtc-builtins` in https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_15_1_X/cudart/cuda-runtime.spec#L1 

Note that https://docs.nvidia.com/cuda/eula/index.html allows to distribute `libnvrtc-builtins.so` so it should  be safe to include it in our rumtime cuda distribution

iarspider:
> @iarspider , cuda 12.8 added new `libnvrtc-builtins.so` which is needed by our cuda-runtime distribution. I would suggest to add `nvrtc-builtins` in https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_15_1_X/cudart/cuda-runtime.spec#L1
> 
> Note that https://docs.nvidia.com/cuda/eula/index.html allows to distribute `libnvrtc-builtins.so` so it should be safe to include it in our rumtime cuda distribution

Thanks!

PR #9702, opened 2025-02-20
[gcc14] Build meschach archive libs with -fPIC

smuzaffar:
This should fix the GCC14 build/link errors
```
ld.bfd: libmeschach.a(copy.o): relocation R_X86_64_32 against `.rodata' can not be used when making a shared object; recompile with -fPIC
ld.bfd: failed to set dynamic section sizes: bad value
```


smuzaffar:
please test

smuzaffar:
please test

PR #9705, opened 2025-02-20
Updated cryptography  to 44.0.1

smuzaffar:
Dropped slc7 support of old cryptography version
This should fix the `Vulnerable OpenSSL included in cryptography wheels` but as we build from sources, so our distribution should be fine


smuzaffar:
please test

PR #9732, opened 2025-03-06
set crab-pre to what is prod ATM

belforte:
in preparation for moving from IB to main CVMFS
FYI @sinonkt

### DIFF
--- a/crab-pre.spec
+++ b/crab-pre.spec
@@ -3,7 +3,7 @@
 #For any other change, increment version_suffix
 ##########################################
 %define version_suffix 00
-%define crabclient_version v3.241218
+%define crabclient_version v3.241125
 ### RPM cms crab-pre %{crabclient_version}.%{version_suffix}
 %define crabserver_version v3.241219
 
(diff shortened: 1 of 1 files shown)


smuzaffar:
please test

belforte:
@smuzaffar I am sorry if this is due to my trying basically roll back `crab-pre` to an earlier tag. I can surely change `00` suffix to `01` but am not sure that I understand correctly. Is that what I should do ?

smuzaffar:
@belforte , going back to an old version is not going to help. The way crab wrapper scripts work is that they search for greatest version  and use it. so crab-pre is not going to pick up v3.241125 as newer tag e.g v3.241218 is already installed in ib area. 

But if you want this for production cvmfs area which still has v3.241030.00 as latest crab-pre then yes just update the suffix

belforte:
thanks @smuzaffar . Yes, that's what I want, so I have updated the prefix. The problem here is that it has been too long since CRAB Client was updated in production and I got confused. Sorry for extra work for you.

smuzaffar:
please test

smuzaffar:
+externals

PR #9761, opened 2025-03-27
[Do not merge][15_1_X] Enable ROOT debug

makortel:
For debugging in https://github.com/cms-sw/cmssw/issues/47697#issuecomment-2758909834

Cherry-picked from https://github.com/cms-sw/cmsdist/pull/9733

PR #9765, opened 2025-03-28
Generate cuda.pcm if CUDA is available

fwyzard:
If CUDA is available, let ROOT generate the `cuda.pcm` file after the build, and include it in the distribution.
This fixes the current problem where a missing dictionary for a type that includes a CUDA header leads to a fatal exception.


fwyzard:
please test

fwyzard:
please test with cms-sw/cmssw#47731

fwyzard:
type bugfix

mandrenguyen:
+1

mandrenguyen:
merge

makortel:
Looks like this PR could have caused problems https://github.com/cms-sw/cmssw/issues/47740

makortel:
Just to note another hurdle https://github.com/cms-sw/cmsdist/pull/9761#issuecomment-2787371103 (that I personally didn't quite understand yet)

PR #9766, opened 2025-03-29
Generate cuda.pcm if CUDA is available [15.0.x]

fwyzard:
If CUDA is available, let ROOT generate the `cuda.pcm` file after the build, and include it in the distribution.
This fixes the current problem where a missing dictionary for a type that includes a CUDA header leads to a fatal exception.


fwyzard:
please test

fwyzard:
backport #9765

fwyzard:
type bugfix

makortel:
hold

Because of https://github.com/cms-sw/cmssw/issues/47740

makortel:
@cmsbuild, please test workflow 180.1,181.1,182.1

makortel:
@cmsbuild, please test

According to https://github.com/cms-sw/cmssw/issues/47740#issuecomment-2768546002 these could work now

PR #9772, opened 2025-04-05
Rework ROCm installation and update to version 6.3.4

fwyzard:


fwyzard:
enable gpu

fwyzard:
please test

rocm-rocrand.spec
```diff
@@ -13,17 +13,30 @@
...
+
+# generate the Source statements for the list of packages
+%(curl -s %{baseurl} |
+  sed -n -e's#<a href="\([^"]*.rpm\)".*#\1#p' |
+  grep -v -e '-asan' |
+  grep -v -e '-debug' |
+  grep -v -e '-rpath' |
+  grep -v -F '%{realversion} -' |
+  grep "$(for P in %{packages}; do echo -n ^$P-%{rpm_version}.rpm'\|'; done; echo 'do_not_match')" |
+  awk '{ printf "Source%d: %s/%s\n", NR-1, "'${baseurl}'", $0; }')
```
smuzaffar: @fwyzard , you need to use `%{expand: }` to expand the output of this code e.g. something like
```
%{expand:%(
  curl -s %{baseurl} |
    sed -n -e's#<a href="\([^"]*.rpm\)".*#\1#p' |
    grep -v -e '-asan' |
    grep -v -e '-debug' |
    grep -v -e '-rpath' |
    grep -v -F '%{realversion} -' |
    grep "$(for P in %{packages}; do echo -n ^$P-%{rpm_version}.rpm'\|'; done; echo 'do_not_match')" |
    awk '{ printf "Source%d: %s/%s\n", NR-1, "'${baseurl}'", $0; }'
)}
```

might work. I would not suggest to generate the list of sources on fly, this means every time this package is rebuilt then we might end up with 
- different set of sources ( which will trigger a rebuilt)
- change in order to sources will also trigger a rebuild.

One other way to simplify this is to create a new package e.g. rocm-souerces.spec which runs the `curl` and dump the list of sources in a file then `rocm` package can depend of it and in its setup section it can loop of all the sources from `$ROCM_SOURCES_ROOT/somefile.txt` to install/unpack the rpms

fwyzard: Mhm, OK, first let me try if `%expand` works.
The list of packages should not change over time - at least, the repositories seem to be fixed after being released, so far.
I've added a `sort` to make sure there are no changes in the order of the sources.

fwyzard:
please test

smuzaffar:
looks like `$(for ...)` is not working properly . Following changes should allso you to properly declare the sources
```
--- a/rocm.spec
+++ b/rocm.spec
@@ -5,9 +5,9 @@
 # allow rpm2cpio dependency on the bootstrap bundle
 %undefine drop_bootstrap_lib
 %define drop_bootstrap_lib true
-%define repository repo.radeon.com/rocm/yum
+%define repository https://repo.radeon.com/rocm/yum
 %else
-%define repository repo.radeon.com/rocm/rhel%{rhel}
+%define repository https://repo.radeon.com/rocm/rhel%{rhel}
 %endif

 # ROCm branch, e.g. 5.6.x, 6.2.x
@@ -52,15 +52,16 @@
   roctracer-devel

 # generate the Source statements for the list of packages
+%define pack_regexp %(for P in  %{packages}; do echo -n ^$P-%{rpm_version}.rpm'\|'; done; echo 'do_not_match')
 %{expand:%(curl -s %{baseurl} |
   sed -n -e's#<a href="\([^"]*.rpm\)".*#\1#p' |
   grep -v -e '-asan' |
   grep -v -e '-debug' |
   grep -v -e '-rpath' |
   grep -v -F '%{realversion} -' |
-  grep "$(for P in %{packages}; do echo -n ^$P-%{rpm_version}.rpm'\|'; done; echo 'do_not_match')" |
+  grep -E "%{pack_regexp}" |
   sort |
-  awk '{ printf "Source%d: %s/%s\n", NR-1, "'${baseurl}'", $0; }')}
+  awk '{ printf "Source%%d: %%s/%%s\n", NR-1, "'%{baseurl}'", $0; }')}
```

fwyzard:
ROCm 6.4.0 has been recently announced, and includes a local installer similar to that used by CUDA, which hopefully simplifies the installation.

Closing the PR for now, will open a new one for 6.4.0 next week.

PR #9792, opened 2025-04-16
build with frame pointer

### LINKED ISSUE #1358: Test FP_X IB for profiling and debugging
Testing shows that there is a small variation in throughput.
```
With FP	        Without FP
0.028078	0.027357
0.027773	0.027983
0.027863	0.027773
0.028051	0.027590
0.027882	0.027213
0.028045	0.027224
0.028011	0.027305
0.027984	0.027319
0.028063	0.027345
	
0.027972	0.027457   Avg
0.000107	0.000268   StdDev
```

AuroraPerego:
This is to request a build with frame pointers enabled so that we can profile CMSSW with [Adaptyst](https://github.com/Adaptyst/Adaptyst).

FYI @maksgraczyk

### DIFF
--- a/scram-tools.file/tools/gcc/gcc-cxxcompiler.xml
+++ b/scram-tools.file/tools/gcc/gcc-cxxcompiler.xml
@@ -9,7 +9,7 @@
     </client>
     <flags CPPDEFINES="GNU_GCC _GNU_SOURCE @GCC_CPPDEFINES@"/>
     <flags CXXSHAREDOBJECTFLAGS="-fPIC @GCC_CXXSHAREDOBJECTFLAGS@"/>
-    <flags CXXFLAGS="-O3 -pthread -pipe -Werror=main -Werror=pointer-arith"/>
+    <flags CXXFLAGS="-O3 -g -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -pthread -pipe -Werror=main -Werror=pointer-arith"/>
     <flags CXXFLAGS="-Werror=overlength-strings -Wno-vla @GCC_CXXFLAGS@"/>
     <flags CXXFLAGS="-felide-constructors -fmessage-length=0"/>
     <flags CXXFLAGS="-Wall -Wno-non-template-friend -Wno-long-long -Wreturn-type"/>


scram-tools.file/tools/gcc/gcc-cxxcompiler.xml
```diff
@@ -9,7 +9,7 @@
     </client>
     <flags CPPDEFINES="GNU_GCC _GNU_SOURCE @GCC_CPPDEFINES@"/>
     <flags CXXSHAREDOBJECTFLAGS="-fPIC @GCC_CXXSHAREDOBJECTFLAGS@"/>
-    <flags CXXFLAGS="-O3 -pthread -pipe -Werror=main -Werror=pointer-arith"/>
+    <flags CXXFLAGS="-O3 -g -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -pthread -pipe -Werror=main -Werror=pointer-arith"/>
```
smuzaffar, line 12: @AuroraPerego , we are planning to build a dedicated IB with frame pointers. I will setup an IB for this. 
By the way, why enable `-g` ?

maksgraczyk:
Let me chime in as the Adaptyst lead developer: for getting the most complete stack traces during profiling, it's not just CMSSW itself that needs to be compiled with frame pointers, but also as many of its dependencies as possible such as ROOT and TBB.

Thank you @AuroraPerego for the PR by the way!

scram-tools.file/tools/gcc/gcc-cxxcompiler.xml
maksgraczyk: @smuzaffar, debug symbols are needed for getting function names and source code lines and paths (if available) for every stack trace obtained during profiling with Adaptyst.

smuzaffar: `-g` can not go in to production builds . In past it also broke other optimizations

smuzaffar:
byt he way, I have opened https://github.com/cms-sw/cmsdist/pull/9793 to enable frame pointers for special `CMSSW_15_1_FP_X` builds

scram-tools.file/tools/gcc/gcc-cxxcompiler.xml
rovere: Ciao @smuzaffar I vaguely recall we used to have a separate build with all debugging symbols?

smuzaffar: yes, we still have e.g. `CMSSW_15_1_DBG_X` IBs

rovere: can we add frame pointers there as well?

smuzaffar: sure we can, let s first wait how enabling frame-pointer behaves in `FP_X` IBs

makortel:
The idea of `FP_X` IB was to test the performance impact of frame pointers before enabling them in production (or being "forced" to use them with RHEL 10).

gartung:
Use of Adaptsys requires a kernel with two parameters configured
https://adaptyst.web.cern.ch/install/#requirements
as well as setting the command itself to have access to some kernel parameters through group permissions as mentioned here
https://adaptyst.web.cern.ch/docs/adaptyst/running-adaptyst/#profiling
for running as a non-root user.

makortel:
About 

> debug symbols are needed for getting function names and source code lines and paths (if available) for every stack trace obtained during profiling with Adaptyst.

are debug symbols needed already for function names?

gartung:
> are debug symbols needed already for function names?

Not that I recall. Debug symbols are needed for lines in the body of the functions.

maksgraczyk:
> > are debug symbols needed already for function names?
> 
> Not that I recall. Debug symbols are needed for lines in the body of the functions.

You do need debug symbols if you want to have all function names in your stack trace.

makortel:
> > > are debug symbols needed already for function names?
> > 
> > 
> > Not that I recall. Debug symbols are needed for lines in the body of the functions.
> 
> You do need debug symbols if you want to have all function names in your stack trace.

Why? Other tools like VTune, IgProf, or gdb don't need debug symbols for function names. Or is the point "**all** function names", i.e. names for some functions could be missing without debug symbols?

maksgraczyk:
Yes, I mean "**all** function names", with an emphasis on "all". I know that some function names can be resolved without debug symbols, but I have personally seen cases in gdb where stack traces have question marks instead of proper names unless the corresponding debug symbols are available (in non-CMSSW software).

dan131riley:
> Why? Other tools like VTune, IgProf, or gdb don't need debug symbols for function names. Or is the point "**all** function names", i.e. names for some functions could be missing without debug symbols?

Without debug symbols, the debugger/profiler is relying on the symbols used to resolve run-time linkage dependencies.  It follows that symbols that don't have external visibility (e.g., file local statics) won't necessarily have symbol resolution.  Inline functions also won't unless the optimization level is reduced.

Also note that "-g" level debugging symbols don't have a significant on program performance, just disk space usage, but compilation with "-g" and high optimization levels may be confusing to debug as the compiler can rearrange the control flow.  Lowering the optimization level makes debugging/profiling easier to understand, but will impact performance (by quite a lot).

makortel:
> Yes, I mean "**all** function names", with an emphasis on "all". I know that some function names can be resolved without debug symbols, but I have personally seen cases in gdb where stack traces have question marks instead of proper names unless the corresponding debug symbols are available (in non-CMSSW software).

Thanks. Can Adaptsys resolve the function names that are "easily available" (as in gdb) without debug symbols?

maksgraczyk:
> Thanks. Can Adaptsys resolve the function names that are "easily available" (as in gdb) without debug symbols?

Adaptyst should be able to resolve such names. If you're hesitant about including "-g", I can first try the tool with CMSSW (+ dependencies) just with frame pointers and go from there.

maksgraczyk:
> Use of Adaptsys requires a kernel with two parameters configured
https://adaptyst.web.cern.ch/install/#requirements

One small note here @gartung: it's true that Adaptyst requires a Linux 5.8+ kernel built with ```CONFIG_DEBUG_INFO_BTF=y``` and ```CONFIG_FTRACE_SYSCALLS=y```, but chances are that your setup already comes with these options as the linked instructions indicate (you need to check if ```/sys/kernel/btf``` and ```/sys/kernel/tracing/events/syscalls``` are available, taking into account that ```/sys/kernel/tracing``` may have to be mounted first).

For example, my AlmaLinux 9 node set up with an OpenStack image provided by CERN already satisfies the aforementioned kernel requirements without any kernel recompilation.

smuzaffar:
`-g` is not only a lot disk ( in past we were even not able to build RPM as the size of cmssw RPM was too big) but also 2 to 3 time slower build . So current 6 hours release build could take 10-15 hours to build.

We also noitced that when root is build in debug mode then relvals take much more time ( but may be cmake is enabling `-O0 -g` for debug builds).

Anyway, I would not suggest to add `-g` for production build, we can have a separate build with frame-pointers and -g

smuzaffar:
byt he way, FP_X IB is available now ( without -g though).

gartung:
Are all externals built with frame-pointer or just CMSSW libraries?

gartung:
> > Use of Adaptsys requires a kernel with two parameters configured
> > https://adaptyst.web.cern.ch/install/#requirements
> 
> One small note here @gartung: it's true that Adaptyst requires a Linux 5.8+ kernel built with `CONFIG_DEBUG_INFO_BTF=y` and `CONFIG_FTRACE_SYSCALLS=y`, but chances are that your setup already comes with these options as the linked instructions indicate (you need to check if `/sys/kernel/btf` and `/sys/kernel/tracing/events/syscalls` are available, taking into account that `/sys/kernel/tracing` may have to be mounted first).
> 
> For example, my AlmaLinux 9 node set up with an OpenStack image provided by CERN already satisfies the aforementioned kernel requirements without any kernel recompilation.

I am more concerned about the permissions needed running as non-root user. We can probably grant those permissions with the correct VM configuration.

gartung:
Testing shows that there is a small variation in throughput.
```
With FP	        Without FP
0.028078	0.027357
0.027773	0.027983
0.027863	0.027773
0.028051	0.027590
0.027882	0.027213
0.028045	0.027224
0.028011	0.027305
0.027984	0.027319
0.028063	0.027345
	
0.027972	0.027457   Avg
0.000107	0.000268   StdDev
```

makortel:
Thanks Patrick. With a naive ratio uncertainty propagation those numbers would indicate the frame pointer build resulting `1 +- 1 %` higher throughput than the no-frame-pointer case (which is the present default), i.e. I'd say within the variations the performance in the two cases is the same.

Then it's a good question if this CMSSW-only test would be sufficient, or we'd want to test building the entire software stack with frame pointers.

gartung:
I suppose the next step is to build Adaptyst as a cmsdist package. When I built it I used my AL9 desktop and installed rpms as needed.

smuzaffar:
closing this PR, we now have special CMSSW_15_1_FP_X IBs with frame-pointers enabled

PR #9793, opened 2025-04-16
added frame_pointer build option

smuzaffar:
Changes needed to build SCRAM based projects with frame-pointer. By default frame pointers are omitted. For special IBs ( e.g. `FP_X` we can build with `--build-option frame_pomiter`


smuzaffar:
please test

by default CMSSW should not be build with frame-pointer

smuzaffar:
please test

smuzaffar:
+externals

by default frame-pointer flags are not added. these will be added when cmssw is build with `--build-option frame_pointer`

PR #9836, opened 2025-05-08
[GCC14] build gcc with frame-pointers; sync changes from el10/gcc14

smuzaffar:
Backport chaqnges from el10 branch
- Improved gcc spec ( splitted in to gcc-prerequisites)
- Bukld GCC with `--enable-frame-pointer` which by default enables frame-pointers for all packages built with this GCC on X86_64.  Note that this does not enable frame-pointers for aarch64
- updated cmake to version 3.31. which has a fix for newer curl
- sync curl changes from el10 branch to reduce the differences between el10 and g14 branch

### DIFF
--- a/cmake.spec
+++ b/cmake.spec
@@ -1,4 +1,4 @@
-### RPM external cmake 3.31.5
+### RPM external cmake 3.31.7
 %define downloaddir %(echo %realversion | cut -d. -f1,2)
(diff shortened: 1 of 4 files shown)


smuzaffar:
test parameters:
- full_cmssw = true

smuzaffar:

smuzaffar:
+externals

cmssw build and tests passed. comparison will take some time as it is comparing default (gcc12) vs gcc14

PR #9818, opened 2025-04-29
Update RDMA Core, OFI, UCX, and Open MPI

fwyzard:
#### ROCm
  * add ROCr headers (used by OFI), remove debuginfo files.

#### RDMA Core
  * update RDMA Core Userspace Libraries to version 57.0
  * make libibverbs plugins relocatable

#### OFI
  * add Libfabric OpenFabrics version 2.1.0

#### UCX
  * update UCX to version 1.18.1

#### Open MPI
  * update Open MPI to version 4.1.8

PR #9843, opened 2025-05-09
ROCm: various updates

fwyzard:
Include additional libraries and tools in the ROCm package, needed to build UCX and MPI with ROCm support.

Enable unified memory on Instinct MI100, MI210/250, and MI300 GPUs:
  - compile the device code without an explicit `xnack` setting, which supports running with `xnack` enabled or disabled;
  - attempt to enable `xnack` support setting `HSA_XNACK=1`.


fwyzard:
enable gpu

fwyzard:
please test

fwyzard:

fwyzard:
The ROCm failures are [a known issue](https://github.com/cms-sw/cmssw/issues/46624).

fwyzard:
Rebased after merging #9818.

fwyzard:
please test

fwyzard:
hold

We may need to change the `xnack` settings.

fwyzard:
please test

fwyzard:
please test

fwyzard:
please test

fwyzard:
please test

fwyzard:
unhold

fwyzard:
please test

fwyzard:
please test

fwyzard:
please test

fwyzard:

fwyzard:
@smuzaffar could you check if the various changes to the tools are correct ?

scram-tools.file/tools/rocm/env.sh
```diff
@@ -1,2 +1,3 @@
 #!/bin/bash -e
 export COMPILER_HOST=$(gcc -dumpmachine)
+export HSA_XNACK=1
```
smuzaffar: @fwyzard , looks like this is not used any more. Env set here is used by toolfile generation script which replaced `@ENV_VAR@`  in the toolfile to `$ENV_VAR`.

smuzaffar:
@fwyzard , changes look good but I do not understand how `HSA_XNACK` env is being used. 

Also rocm distribution size is now `24GB` instead of `2G` of previous  `rocm + rocm-rocrand` distribution

[a]
````
> du -sh rocm/6.3.2-f1d309965d0216445a39746d4167bb44
24G     rocm/6.3.2-f1d309965d0216445a39746d4167bb44

> du -sh rocm/6.3.2-94b981ba216f4b76c08c130cf3731d10 rocm-rocrand/6.2.4-fec0133700f9e26e6827e25662f9399d
1.8G    rocm/6.3.2-94b981ba216f4b76c08c130cf3731d10
179M    rocm-rocrand/6.2.4-fec0133700f9e26e6827e25662f9399d
```

scram-tools.file/tools/rocm/env.sh
fwyzard: Ah... how do I make `cmsenv` export `HSA_XNACK=1` ?

fwyzard:
> Also rocm distribution size is now `24GB` instead of `2G` of previous `rocm + rocm-rocrand` distribution

Yes, that's the downside of bundling the extra libraries :-(
I think I've added the bare minimum that is required by PyTorch, based on its CMake checks.

How much of an issue is the size ?

scram-tools.file/tools/rocm/env.sh
smuzaffar: for cmsenv, you just need to add `<runtime name="HSA_XNACK" value="1"/>` in the `rocm.xml` toolfile. If you want it to be configurable then use `<runtime name="HSA_XNACK" value="@HSA_XNACK@"/>` in the toolfile and add some  login the `scram-tools.file/tools/rocm/env.sh` to set `HSA_XNACK` env ( e.g. by checking the `$TOOL_ROOT/lib` directory for hsa libs? )

fwyzard: OK, thanks.
No, it doesn't need to be configurable.

fwyzard:
please test

makortel:
> How much of an issue is the size ?

IIRC 24 GB increase would about double the disk size of the CMSSW distribution (of one release). Naively anyone creating a container with CMSSW installed would be impacted (well, anyone installing CMSSW "locally", but I suppose the use cases for containers would be the ones suffering most). Maybe (theoretically, at least) those use cases could cut the ROCm out (if they are not trying to use AMD GPUs)?

If this is how the future will look like, perhaps we should think about options to avoid bundling everything together.

fwyzard:
OK, for the time being I've only added here the minimal set of libraries for UCX and MPI.
I'll push the larger set of libraries used by PyTorch to a separate PR, and we can discuss if we want them and how to handle the large size.

fwyzard:
please test

fwyzard:

fwyzard:
@smuzaffar could you merge these changes ?

They are needed to rebase the MPI PR, and to make progress debugging the issues on LUMI.

Thanks !

smuzaffar:
+externals

Thanks @fwyzard  for the cleanup up. `rocm` now just has `800MB` more ( 2.8GB as compare to 2GB previous).

PR #9846, opened 2025-05-12
Set CLING_PREBUILT_MODULE_PATH so that rootcling can find cuda.pcm from full cmssw

smuzaffar:
This add CLING_PREBUILT_MODULE_PATH to cmssw toolfile so that `cuda.pcm` can be found from full cmssw lib's directory for a patch release. This should hopefully fix the failing unit test in ROOT6 IBs ( https://cmssdt.cern.ch/SDT/cgi-bin/logreader/el8_aarch64_gcc12/CMSSW_15_1_ROOT6_X_2025-05-09-2300/unitTestLogs/HeterogeneousCore/CUDATest#/ )


smuzaffar:
+externals

PR tests do not use this file as this is only used for patch releases. I am merging it and will build ROOT6 IBs at 11h00 and then a patch release later to make sure every thing works

PR #9863, opened 2025-05-14
ONNX Runtime: depend explicitly on Eigen

fwyzard:
Make the dependency on Eigen explicit, and let CMake find the CMS version.

PR #9864, opened 2025-05-14
ONNX Runtime: depend explicitly on Eigen [15.0.x]

fwyzard:
Make the dependency on Eigen explicit, and let CMake find the CMS version.

### DIFF
--- a/onnxruntime.spec
+++ b/onnxruntime.spec
@@ -8,7 +8,7 @@ Source: git+https://github.com/%{github_user}/%{n}.git?obj=%{branch}/%{tag}&expo
 Patch0: onnxruntime-gcc13
 
 BuildRequires: cmake ninja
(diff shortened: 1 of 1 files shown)


fwyzard:
backport #9863

fwyzard:
please test

fwyzard:
type bugfix

smuzaffar:
+externals

smuzaffar:
@cms-sw/orp-l2 , this is a backport for 15.0.X. @fwyzard can tell how important/urgent it is for 15.0.X

fwyzard:
It's not urgent, but it seems to be necessary to avoid breaking ONNX with other updates.

We don't need a release specifically for this, but please include it at the next opportunity.

mandrenguyen:
+1

PR #9872, opened 2025-05-16
alpaka: use the HEAD of the develop branch

fwyzard:
This PR is to be used to regularly test that the HEAD of the alpaka `develop` branch is compatible with CMSSW.

### DIFF
--- a/alpaka.spec
+++ b/alpaka.spec
@@ -1,9 +1,11 @@
-### RPM external alpaka 2.0.0
+%define git_remote https://github.com/alpaka-group/alpaka
+%define git_branch develop
+%define git_commit %(git ls-remote %{git_remote}.git refs/heads/%{git_branch} | cut -c-40)
+%define git_short  %(echo %{git_commit} | cut -c-11)
+### RPM external alpaka %{git_branch}-%{git_short}
 ## NOCOMPILER
 
-%define git_commit %{realversion}
-
-Source: https://github.com/cms-externals/%{n}/archive/%{git_commit}.tar.gz
+Source: %{git_remote}/archive/%{git_commit}.tar.gz
 
 %prep
 %setup -n %{n}-%{git_commit}


fwyzard:
please test

fwyzard:
please test

fwyzard:
@smuzaffar @iarspider if these tests work, do you think we could run something like this more regularly ?
Something like, once a day, and only if any changes were pushed to alpaka ?

fwyzard:
please test

smuzaffar:
> @smuzaffar @iarspider if these tests work, do you think we could run something like this more regularly ? Something like, once a day, and only if any changes were pushed to alpaka ?

@fwyzard , this should work and should use the latest commit of alpaka's selected branch. For now we can run manually once a day but we will soon hit the commit status limit.

fwyzard:
If it's manual, I can trigger it when something gets merged in alpaka, and rebase it once in a while to update the commit.

fwyzard:
More in general I was wondering if we could set up some automatic testing of the dependencies that we have, like also ROOT, to spot potential problems earlier.

smuzaffar:
for root we have dedicated IBs and we manually update root commits once a week. so ROOT automatic testing it done via daily IBs. We can setup new ALPAKA IBs where alpaka get latest commit automatically and we build such IB few times a week

fwyzard:
> for root we have dedicated IBs and we manually update root commits once a week

Eh, so not very much automatic :)

fwyzard:
Mhm, I would not add another set of IBs to test alpaka, doing PR-like tests should be enough for now.

I'm mostly interested in making sure that alpaka does not break the syntax we use.

smuzaffar:
correct, we do want to run few IBs for every root update to catch any threading issue

fwyzard:
Note: the latest tests correspond to https://github.com/alpaka-group/alpaka/commit/e906b1c0c449ebab02b45db6475246c972251aad .

fwyzard:
please test

fwyzard:
please test

fwyzard:
please test

fwyzard:
please test with cms-sw/cmssw#49042 for el9_amd64_gcc13

fwyzard:
please test

fwyzard:
please test

fwyzard:
enable gpu

fwyzard:
please test

PR #9883, opened 2025-05-21
Disabled alpaka-cuda backend if nvcc is not accessible

smuzaffar:


smuzaffar:
test parameters:
- addpkg = HeterogeneousCore

smuzaffar:
please test

fwyzard:
Do we have something similar for `alpaka-rocm` when `hipcc` is not available, like on ARM ?

smuzaffar:
build rules already disable alpaka-rocm if rocm tool is not setup ( e.g. on aarch64). If we decided to go with rocm build/development tools and rocm-runtime then I can also update build rules to check for build/development tool and disable rocm backend if missing but for now just check for rocm tool is enough to enable/disable rocm backend

smuzaffar:
+externals

PR #9889, opened 2025-05-26
Updated root to tip of branch v6-32-00-patches

smuzaffar:
This include the backport https://github.com/root-project/root/pull/18830 ( see https://github.com/cms-sw/cmssw/issues/46341 for details ). Once properly tested in 15.1.X IBs, we would like to backport this to 15.0.X too

### DIFF
--- a/root.spec
+++ b/root.spec
@@ -3,8 +3,8 @@
 ## INITENV SET ROOTSYS %{i}
 ## INCLUDE compilation_flags
 ## INCLUDE cpp-standard
-%define tag 101b78bfce1a48ef030d70816bb459e71973732d
-%define branch cms/v6-32-00-patches/9880139790
+%define tag 63c27767f1be350ef3d684242da34f0d396cf97d
+%define branch cms/v6-32-00-patches/69384a6a78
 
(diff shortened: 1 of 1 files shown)


smuzaffar:
please test

smuzaffar:
@makortel , this updates ROOT 6.32 for 15.1.X default IBs. Hopefully this should fix the issue seen in https://github.com/cms-sw/cmssw/issues/46341 . I do not know if there is any easy way to test the fix ... do you have any idea/test which I can run to make sure that the issue is fixed?

smuzaffar:
+externals

lets get this in 15.1.X. We need to backport these fixes in to 15.0.X too

makortel:
> I do not know if there is any easy way to test the fix ... do you have any idea/test which I can run to make sure that the issue is fixed?

Not really. Since the data race is/was rare, it seems the best we can do is to wait ~2 months and see if we get more such crashes. Written that, I'd consider https://github.com/cms-sw/cmssw/issues/46341 done and close the issue (although maybe wait after the fix gets backported to 15_0_X?), and open a new one if we see the problem again.

PR #9903, opened 2025-06-02
GCC14: Update to version 14.3.1

smuzaffar:


smuzaffar:
test parameters:
- full_cmssw = true

smuzaffar:

smuzaffar:

smuzaffar:

smuzaffar:
@makortel , as mentioned in https://github.com/root-project/root/issues/18939 , ROOT 6.32 does not build with newer GCC versions ( GCC 13.4, 14.3, 15.1) due to llvm not properly supporting c++20.

This PR builds ROOT 6.32 by forcing root/llvm to use c++20 (which root already does for root 6.36 and above: see https://github.com/root-project/root/pull/17735). For moving forward, I have suggested in Core SW meeting that may be for GCC14 we can already move to ROOT 6.36. 

For GCC 13 builds, as currently we are using GCC 13.3 so ROOT 6.32 will work but in couple of days GCC 13.4 will come out in few days and I would like to move to GCC 13.4 (which already contains patches we had applied in our build) when ROOT 6.32 gets the  backport of https://github.com/root-project/root/pull/17735

makortel:
> For moving forward, I have suggested in Core SW meeting that may be for GCC14 we can already move to ROOT 6.36.

I'm ok with that. Would it be implemented such that on GCC 14 we only build ROOT636 IB, or that the default IB for GCC14 would use ROOT 6.36?

smuzaffar:
> > For moving forward, I have suggested in Core SW meeting that may be for GCC14 we can already move to ROOT 6.36.
> 
> I'm ok with that. Would it be implemented such that on GCC 14 we only build ROOT636 IB, or that the default IB for GCC14 would use ROOT 6.36?

ah this is a good question. I was thinking to just build default IB of gcc14 (for all OS el8 - el10) with ROOT 6.36 (we do not build releases for GCC 14 yet).

makortel:
> > > For moving forward, I have suggested in Core SW meeting that may be for GCC14 we can already move to ROOT 6.36.
> > 
> > 
> > I'm ok with that. Would it be implemented such that on GCC 14 we only build ROOT636 IB, or that the default IB for GCC14 would use ROOT 6.36?
> 
> ah this is a good question. I was thinking to just build default IB of gcc14 (for all OS el8 - el10) with ROOT 6.36 (we do not build releases for GCC 14 yet).

If practical from the IB management point of view, I'd find it more clear to use the ROOT636 IB for gcc14.

smuzaffar:
> If practical from the IB management point of view, I'd find it more clear to use the ROOT636 IB for gcc14.

sounds reasonable, so we will not have any default GCC 14 IBs but we will have ROOT636 GCC 14 IBs for amd64/aarch64 and el8/el9/el10

smuzaffar:
please test

smuzaffar:
please test for CMSSW_15_1_ROOT636_X/el8_amd64_gcc14

PR #9907, opened 2025-06-02
[13.1.X] Update correctionlib to 2.7

nsmith-:
Backporting to older CMSSW versions before we auto-update packages regularly.
If this works well and is correct I'll propagate it to other versions.

### DIFF
--- a/pip/requirements.txt
+++ b/pip/requirements.txt
@@ -15,6 +15,7 @@ absl-py==1.4.0
 aiohttp==3.8.4
 aiosqlite==0.18.0
 aiosignal==1.3.1
+annotated-types==0.7.0
 anyio==3.6.2
 appdirs==1.4.4
 argon2-cffi==21.3.0
@@ -58,7 +59,7 @@ click==8.1.3
 clikit==0.6.2
 cmsml==0.1.2
 contourpy==1.0.7
-correctionlib==2.2.2
+correctionlib==2.7.0
 crashtest==0.4.1
 certifi==2022.12.7
 cffi==1.15.1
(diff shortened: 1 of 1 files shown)


smuzaffar:
@nsmith- , this should first go in to development IB, i.e. CMSSW_15_1_X, before backporting it to production releases like CMSWW_13.1.X. Can you please open a PR for IB/cMSSW_15_1_X/master branch?

nsmith-:
@smuzaffar the version of correctionlib is already at a much more recent version in the the development IB:
https://github.com/cms-sw/cmsdist/blob/c4e8bacca1ab2b32fe91b6eee4d9e0197ed4dd01/pip/requirements.txt#L66
This is a backport. If the issue is with 2.6.4 vs. 2.7 I can also bump it to 2.7 there as well

smuzaffar:
>  If the issue is with 2.6.4 vs. 2.7 I can also bump it to 2.7 there as well

@nsmith- , yes please first bump it in 15.1.X.

also I see you have added few more pip packages . Make sure to get those updates for 15.1.X too and note that newly added packages are not build automatically. You have to add them as dependency of package which actually need/use these

nsmith-:
It looks like another auto-update to 15_1_X has happened in the meantime and now correctionlib is v2.7 there:
https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_15_1_X/master/pip/requirements.txt#L70
So are we ok to backport this now?

nsmith-:
> also I see you have added few more pip packages . Make sure to get those updates for 15.1.X too and note that newly added packages are not build automatically. You have to add them as dependency of package which actually need/use these

All those are just dependencies declared by the package or its dependencies. Again, in 15_1_X, they are all already there. I don't really understand what you are saying here. Why can't we just let pip resolve the environment from this requirements file?

smuzaffar:
please test

Do we still need this in 13.1.X ?

smuzaffar:
please test

PR #9926, opened 2025-06-12
[GCC13] convert array-bounds error to warnings for gcc 13.4

smuzaffar:


smuzaffar:
test parameters:
- addpkg = CondCore/CondDB, CondFormats/RunInfo

smuzaffar:

smuzaffar:
+externals

changes synced from gcc14 branch. This change is needed to avoid the false-positive arrary-bound errors. Previously we were converting GCC 13.3 array-bound errors to warning but recently we have update tp gcc 13.4 so that is why we need to get this change in gcc 13 branch too

PR #9934, opened 2025-06-18
make sure to download unique wheel for each arch

smuzaffar:


smuzaffar:
please test with cms-sw/pkgtools#222

PR #9941, opened 2025-06-23
changed dependency of hdf5 and sherpa to more general mpi.xml

Annnnya:
In the process of making it possible to use any mpi library, we also should try changing dependencies for other tools which currently depend on openmpi


fwyzard:
please test

smuzaffar:
+externals

fwyzard:
type ngt

PR #9997, opened 2025-08-07
pass the G4HepEm_EARLY_TRACKING_EXIT flag

SeverinDiederichs:
G4HepEm is build with the flag `-DG4HepEm_EARLY_TRACKING_EXIT=ON` but it seems that it must be also passed via the scram-tools such that the flag is available in the AdePT header files that we require for the integration into cmssw.

Please check and confirm @civanch @smuzaffar 

Note that we do have a local workaround (by passing it via the `BuildFile.xml`), so we do not need a new IB build *yet* (we will need it after a few more fixes are pushed to AdePT and AdePT is tagged again).

### DIFF
--- a/scram-tools.file/tools/g4hepem/g4hepem_interface.xml
+++ b/scram-tools.file/tools/g4hepem/g4hepem_interface.xml
@@ -6,4 +6,5 @@
     <environment name="INCLUDE" default="$G4HEPEM_INTERFACE_BASE/include/G4HepEm"/>
   </client>
   <use name="geant4core"/>
+  <flags CXXFLAGS="-DG4HepEm_EARLY_TRACKING_EXIT=ON"/>
 </tool>
--- a/scram-tools.file/tools/g4hepem/g4hepem_interface.xml
+++ b/scram-tools.file/tools/g4hepem/g4hepem_interface.xml
@@ -1,4 +1,4 @@
-<tool name="g4hepem_interface" version="@TOOL_VERSION@" revision="3">
+<tool name="g4hepem_interface" version="@TOOL_VERSION@" revision="4">
   <info url="https://github.com/mnovak42/g4hepem"/>
   <client>
     <environment name="G4HEPEM_INTERFACE_BASE" default="@TOOL_ROOT@"/>


smuzaffar:
thanks @SeverinDiederichs for the change. Can you please increment the `revision` at https://github.com/cms-sw/cmsdist/pull/9997/files#diff-a52f9f8ece5b4c4becb791607710695ce5d987f0d5df942dfae82e93900f7e36R1 (just using `revision="4"` should be enough)

smuzaffar:
@cmsbuild, please test

smuzaffar:
please test using addpkg SimG4Core,Alignment/LaserAlignmentSimulation

civanch:
@SeverinDiederichs , would it be possible make this compilation flag the default for G4HepEm? We expect in long term G4HepEM work simultaneously in CPU or CPU+GPU.

civanch:
@smuzaffar , I maybe wrong but in this branch Geant4 with all dependences is built in static mode,  so all components should be recompiled and included into this static build.

smuzaffar:
@civanch , this PR only changes scram tool file so that change is not going to trigger a rebuild of geant4 or any other external. This will only rebuild cmssw packages which use g4hepem.

failure here are due to the fact the this branch uses geant4 ref6 and https://github.com/cms-sw/cmssw/pull/48665 changes work do not work in ref7. We can either move to ref7 for g4adept branch of we need to add some protection in cmssw code so that https://github.com/cms-sw/cmssw/pull/48665 can work for any geant4 version

smuzaffar:
> @SeverinDiederichs , would it be possible make this compilation flag the default for G4HepEm? We expect in long term G4HepEM work simultaneously in CPU or CPU+GPU.

@civanch , should we update our g4hepem spec to enable this flag for all IBs?

SeverinDiederichs:
> > @SeverinDiederichs , would it be possible make this compilation flag the default for G4HepEm? We expect in long term G4HepEM work simultaneously in CPU or CPU+GPU.
> 
> @civanch , should we update our g4hepem spec to enable this flag for all IBs?

Enabling this flag will add a virtual function call for each step (which is used to get tracks from the G4HepEmTrackingManager on CPU to AdePT on GPU). The reason for the compile time flag is to avoid the overhead of a virtual function call in a CPU-only simulation. 

We have not measured the overhead, so I cannot tell if this will actually impact the performance

civanch:
@smuzaffar , for me not clear what happens, I inspect Monopole classes and it seems there is no G4StepperType.hh header directly or indirectly. Do you think that compilation will work if we will use in G4ADEPT G4 version 11.3ref07 ?

smuzaffar:
@civanch , the build error is [a] , this is because we now have https://github.com/cms-sw/cmssw/pull/48665 change. This works for old geant4 and geant4 11.3.2 ref7 only. For ref6 , `G4MagIntegratorStepper` class do not have `G4StepperType StepperType() const = 0;` that is why build fails for ref6.

As I wrote , we either need to move to ref7 for G4ADEPT IB, or have a separate cmssw branch for G4ADEPT without https://github.com/cms-sw/cmssw/pull/48665 change. I would prefer to move to g4 ref7 for adept IBs

[a]
```
[src/SimG4Core/MagneticField/interface/FieldStepper.h:24](https://github.com/cms-sw/cmssw/blob/CMSSW_15_1_G4ADEPT_X_2025-08-07-1100/SimG4Core/MagneticField/interface/FieldStepper.h#L24):17: error: 'G4StepperType FieldStepper::StepperType() const' marked 'override', but does not override
    24 |   G4StepperType StepperType() const override { return kTDormandPrince45; };
```

smuzaffar:
I am merging it as build errors are not due to this change. We need to move to geant4 ref7 for g4adept branch too 
(see https://github.com/cms-sw/cmsdist/pull/9997#issuecomment-3167156999 )

civanch:
@smuzaffar , yes, I agree.

PR #10000, opened 2025-08-08
Do not set PYTHONHOME via gdb; set python ignore-env for gdb

smuzaffar:
Instead of SCRAM unsetting PYTHONHOME, lets update GDB to not set PYTHONHOME


smuzaffar:
please test

PR #10004, opened 2025-08-09
GDB: build using relocatable python libdir

smuzaffar:
This should avoid setting PYTHONHOME in gdb. Using python path relative to gdb install should allow gdb to automatically relocate python at runtime


smuzaffar:
please test

PR #10008, opened 2025-08-11
[G4AdePT] add AdePT host library

SeverinDiederichs:
The AdePT host library also needs to be exposed as otherwise it wasn't linking as symbols from the host library were missing:

```
undefined reference to `AdePTTrackingManager::AdePTTrackingManager(AdePTConfiguration*, int)'
collect2: error: ld returned 1 exit status
```

This is fixed by exposing both   `<lib name="AdePT_G4_integration"/>` for the host symbols and `<lib name="AdePT_G4_integration_final"/>` for the device symbols.

No new IB needs to be created, as I can fix this locally for now, until a new IB is needed.


smuzaffar:
please test

smuzaffar:
+externals

good to go in G4ADEPT IBs

PR #10018, opened 2025-08-15
auto generate/repalce tool name and base variables

smuzaffar:
This PR simplifies the creation of toolfiles for new externals. For new externals, one can just copy the one of the existing toolfile and just replace the lib name or paths. `get_tools` will automatically replace `@TOOL@` (tool name e.g. abseil-cpp) and `@UC_TOOL@` (upper case toolname e.g. ABSEIL_C) with correct tool name

### DIFF
--- a/scram-tools.file/tools/abseil-cpp/abseil-cpp.xml
+++ b/scram-tools.file/tools/abseil-cpp/abseil-cpp.xml
@@ -1,4 +1,4 @@
-<tool name="abseil-cpp" version="@TOOL_VERSION@" revision="1">
+<tool name="@TOOL@" version="@TOOL_VERSION@" revision="2">
   <info url="https://github.com/abseil/abseil-cpp"/>
   <lib name="absl_bad_any_cast_impl"/>
   <lib name="absl_bad_optional_access"/>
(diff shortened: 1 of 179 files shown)


smuzaffar:
please test

fwyzard:
Since there are _a lot_ of uses of `@UC_TOOL@_BASE`, maybe you could introduce also `@TOOL_BASE@` for it ?

smuzaffar:
> Since there are _a lot_ of uses of `@UC_TOOL@_BASE`, maybe you could introduce also `@TOOL_BASE@` for it ?

sounds reasonable

smuzaffar:
please test

scram-tools.file/tools/abseil-cpp/abseil-cpp.xml
```diff
@@ -60,8 +60,8 @@
   <lib name="absl_time"/>
   <lib name="absl_time_zone"/>
   <client>
-    <environment name="ABSEIL_CPP_BASE" default="@TOOL_ROOT@"/>
-    <environment name="LIBDIR" default="$ABSEIL_CPP_BASE/lib"/>
-    <environment name="INCLUDE" default="$ABSEIL_CPP_BASE/include"/>
+    <environment name="@TOOL_BASE@" default="@TOOL_ROOT@"/>
```
fwyzard, line 63: I guess now every tool file will have this:
```xml
    <environment name="@TOOL_BASE@" default="@TOOL_ROOT@"/>
```

Maybe there is a way to make it automatic ?

smuzaffar: yes, can be done , may be for next round :-)

smuzaffar:
+externals

cmssw build and unit tests ran fine.

PR #10040, opened 2025-08-26
Upgrade OpenMPI version from 4.1.9 to 5.0.8

ghyls:
OpenMPI 5.0.8 is the current stable release and provides features would like to take advantage of (e.g. User-Level Fault Mitigation), and that are not available in OpenMPI 4.x

Note that OpenMPI 5.0.8's spec file requires the changes proposed in this pull request to build:
https://github.com/cms-sw/pkgtools/pull/224

### DIFF
--- a/openmpi.spec
+++ b/openmpi.spec
@@ -1,8 +1,8 @@
-### RPM external openmpi 4.1.9a1-20250505
+### RPM external openmpi 5.0.8
 ## INITENV SET OPAL_PREFIX %{i}
-%define branch v4.1.x
-%define tag e6d2cb856f3fc649aa01bd5b688a003b3b33db7d
-Source: git+https://github.com/open-mpi/ompi.git?obj=%{branch}/%{tag}&export=%{n}-%{realversion}&output=/%{n}-%{realversion}.tgz
+%define branch v5.0.x
+%define tag v%{realversion}
+Source: git+https://github.com/open-mpi/ompi.git?obj=%{branch}/%{tag}&export=%{n}-%{realversion}&submodules=1&output=/%{n}-%{realversion}.tgz
 BuildRequires: autotools flex
 %{!?without_cuda:Requires: cuda}
 Requires: libfabric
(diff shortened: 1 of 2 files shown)


iarspider:
please test with https://github.com/cms-sw/pkgtools/pull/224

iarspider:
@fwyzard I guess the failure is not related to pkgtools PR?

fwyzard:
Indeed, looks like the pkgtools changes worked as expected.
We need some updates within CMSSW to support OpenMPI 5.x.

smuzaffar:
please test

openmpi.spec
```diff
@@ -28,12 +28,9 @@ AUTOMAKE_JOBS=%{compiling_processes} ./autogen.pl
   --prefix=%i \
   --disable-dependency-tracking \
   --enable-ipv6 \
-  --enable-mpi-cxx \
```
smuzaffar, line 31: @ghyls , as `libmpi_cxx.so` is not built any more so I guess we have to remove https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_15_1_X/master/scram-tools.file/tools/openmpi/openmpi.xml#L3  from the tool definition so that scram do not link this library .... right?

fwyzard: Right.

scram-tools.file/tools/openmpi/openmpi.xml
```diff
@@ -1,6 +1,5 @@
 <tool name="@TOOL@" version="@TOOL_VERSION@" revision="2">
```
smuzaffar: @ghyls , can you please update/increment the `revision` of toolfile

ghyls: Thanks for noticing, yes.

smuzaffar:
please test

mandrenguyen:
+1

smuzaffar:
+externals

PR #10044, opened 2025-08-27
[RFC] Build pytorch with ROCm + python interface 

iarspider:
This is a proof of concept PR, showing how to build pytorch with ROCm. 

@smuzaffar let me know if I should just run `pip install <wheel file>` as install step, instead of splitting a package into two.

One thing to note is that `aotriton` (or, to be exact, it's dependency `triton`) downloads it's dependencies - a custom LLVM and some CUDA tools (nvcc, for example). This can be disabled, but then we will need to provide these ourselves - which means keeping track of LLVM updates (triton needs a very specific commit of LLVM).


iarspider:
We still can't build CUDA and ROCm together (or merge them later):

```
$ ldd .../libtorch.so
        libtorch_hip.so => /build/razumov/pytorch-rocm/BUILD/el8_amd64_gcc12/external/py3-torch/2.8.0-70389357bdba905eef4e2a952c7a2db8/lib/python3.9/site-packages/torch/lib/libtorch_hip.so (0x00007f3a26421000)
```

iarspider:
enable gpu

iarspider:
please test

iarspider:
please test

iarspider:
please test

iarspider:
please test

fwyzard:
> We still can't build CUDA and ROCm together (or merge them later):
> 
> ```
> $ ldd .../libtorch.so
>         libtorch_hip.so => /build/razumov/pytorch-rocm/BUILD/el8_amd64_gcc12/external/py3-torch/2.8.0-70389357bdba905eef4e2a952c7a2db8/lib/python3.9/site-packages/torch/lib/libtorch_hip.so (0x00007f3a26421000)
> ```

Is it possible to build `libtorch.so` with a different soname, like for example `libworch_hip_wrapper.so` ?
So it would look like
```
$ ldd .../libworch_hip_wrapper.so
        libtorch_hip.so => /build/razumov/pytorch-rocm/BUILD/el8_amd64_gcc12/external/py3-torch/2.8.0-70389357bdba905eef4e2a952c7a2db8/lib/python3.9/site-packages/torch/lib/libtorch_hip.so (0x00007f3a26421000)
```
?

iarspider:
@fwyzard sorry, didn't properly explain what the problem is: `libtorch.so` requires either `libtorch_cuda.so` (if pytorch is built with cuda) or `libtorch_hip.so` (if built with rocm). And you can only build with one - either with CUDA, or with ROCm. So I'm not sure if we can easily merge two builds of pytorch (as @smuzaffar suggested - to avoid having two packages `pytorch-cuda` and `pytorch-rocm`).

iarspider:
please test

fwyzard:
Indeed... what I meant is that we could at least have 2 (or 3) packages:
  - `pytorch-cuda`
  - `pytorch-hip` (or more accurately `pytorch-rocm` ?)
  - and maybe, `pytorch-cpu` built without any GPU support.
 
.A

iarspider:
> Indeed... what I meant is that we could at least have 2 (or 3) packages:
> 
>     * `pytorch-cuda`
> 
>     * `pytorch-hip` (or more accurately `pytorch-rocm` ?)
> 
>     * and maybe, `pytorch-cpu` built without any GPU support.
> 
> 
> .A

I think one doesn't need a separate `pytorch-cpu`, can use either `-cuda` or `-rocm`.

fwyzard:
I guess you're right.

One reason that having `pytorch-cpu` might be useful is to keep the dependencies clean, in case the CUDA-enabled version runs into issues when running on a machine without CUDA or with some weird drivers.

Another reason is if it would make it easier to implement some way to *not* use GPUs in PyTorch-based modules in CMSSW.

iarspider:
Somehow I've missed that we need Python 3.10+ for this update. I will retarget this PR to PY312 branch.

PR #10058, opened 2025-09-04
openmpi: Set needed env to avoid openmpi calling setenv at runtime

smuzaffar:
This change set the following env variables
 - `HFI_NO_BACKTRACE=1`:  Default is`1` if not set (https://github.com/open-mpi/ompi/blob/main/opal/runtime/opal_init.c#L116-L118)
 - `IPATH_NO_BACKTRACE=1`: Default is also `1` if not set (https://github.com/open-mpi/ompi/blob/main/opal/runtime/opal_init.c#L113-L115)
 - `PMIX_PREFIX=install_path` 

This should be enough for openmpi to not call `setenv` at runtime

FYI @makortel , hopefully this should fix the failiing DEVEL IBs relvals

### DIFF
--- a/openmpi-opt.file
+++ b/openmpi-opt.file
@@ -1,0 +1,2 @@
+%define HFI_NO_BACKTRACE 1
+%define IPATH_NO_BACKTRACE 1
--- a/openmpi-setenv-fix.patch
+++ b/openmpi-setenv-fix.patch
@@ -1,0 +1,17 @@
+diff --git a/ompi/runtime/ompi_mpi_init.c b/ompi/runtime/ompi_mpi_init.c
+index 19c0999..66e6553 100644
+--- a/ompi/runtime/ompi_mpi_init.c
++++ b/ompi/runtime/ompi_mpi_init.c
+@@ -349,8 +349,10 @@ int ompi_mpi_init(int argc, char **argv, int requested, int *provided,
+     /* deal with OPAL_PREFIX to ensure that an internal PMIx installation
+      * is also relocated if necessary */
+ #if OPAL_USING_INTERNAL_PMIX
+-    if (NULL != (evar = getenv("OPAL_PREFIX"))) {
+-        opal_setenv("PMIX_PREFIX", evar, true, &environ);
++    if (NULL == getenv("PMIX_PREFIX")) {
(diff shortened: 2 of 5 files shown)


smuzaffar:
please test

smuzaffar:
please test workflow 534.0,536.0 for CMSSW_15_1_DEVEL_X

testing for devel IBs but as PR tests do not use `cmsTraceFunction ` so it is not going to check the if this change actually fix the relvals but I will re-run locally once the PR area is available on cvmfs

smuzaffar:
please test workflow 534.0,536.0 for CMSSW_15_1_DEVEL_X

makortel:
> This should be enough for openmpi to not call `setenv` at runtime

Nice to see the tracer is catching something :) This also reminded me of the issue https://github.com/cms-sw/cmssw/issues/45165 where removing OpenMPI from Sherpa was brought up.

fwyzard:
We may need to set more variables, I stumbled upon this online:
image

fwyzard:
Although that was for OpenMPI 4.x, something may have changed in OpenMPI 5.x.

smuzaffar:
@fwyzard `PRTE_PREFIX` env is set by the mpi executable and it is set by using the `OPAL_PREFIX` (which we already set). We mostly care for the env which is set by libs (loaded by cmsRun as runtime).

smuzaffar:
latest change fixes setenv calls from openmpi but I have found https://gitlab.com/sherpa-team/sherpa/-/blob/v2.2.15/ATOOLS/Org/Run_Parameter.C?ref_type=tags#L234-237 where sherpa calls setenv for setting LD_LIBRARY_PATH.  I see this code was removed in commit https://gitlab.com/sherpa-team/sherpa/-/commit/6ead62d7a2758612f8965fb5b61df8c012cf9cae . I can try to apply https://gitlab.com/sherpa-team/sherpa/-/commit/6ead62d7a2758612f8965fb5b61df8c012cf9cae change for cms or better if we can move sherpa 3.0 (which contains this fix). FYI @cms-sw/generators-l2

smuzaffar:
please test

smuzaffar:
+externals

lets gets this in and for https://github.com/cms-sw/cmsdist/pull/10058#issuecomment-3255956279 we can open a new PR

PR #10065, opened 2025-09-11
Update tag for L1Trigger-Phase2L1ParticleFlow to V00-09-00

mmusich:
forward port of https://github.com/cms-sw/cmsdist/pull/10064 to `IB/CMSSW_16_0_X/master`.
See https://github.com/cms-sw/cmssw/issues/48897.

PR #10066, opened 2025-09-11
Update tag for L1Trigger-Phase2L1ParticleFlow to V00-09-00

mandrenguyen:
Missing tag update from master release, see https://github.com/cms-sw/cmssw/issues/48897#issuecomment-3279221671

Thanks for spotting it, @mmusich


mmusich:
ah, I already opened https://github.com/cms-sw/cmsdist/pull/10065

mandrenguyen:
@mmusich is faster on the draw then I am

PR #10080, opened 2025-09-15
[RFC] Add tool to automatically update most of ROCm subpackages' versions

iarspider:
FYI @fwyzard


fwyzard:
OK for me...

You could also look into the ROCm offline installer: https://rocm.docs.amd.com/projects/install-on-linux/en/latest/install/rocm-offline-installer.html .

We are also thinking whether to try and build ROCm from sources, but that's going to take some time.

fwyzard:
Good timing, ROCm 7.0 just came out 😁

PR #10116, opened 2025-10-07
[PY312] Dropped flatbuffers patches

smuzaffar:
We do not need flatbuffer patches for python 3.12 where we already have a newer version of flatbuffers 24.3.25. 
- flatbuffers-7422.patch was not applied previously for CMSSW_15_1_X/py312 so it should be safe to drop it for CMSSW_16_0_X/py312
- https://patch-diff.githubusercontent.com/raw/google/flatbuffers/pull/7227.diff : was applied only for gcc13 for older version of flatbuffers but as we are using a newer flatbuffers so we do need itany more


smuzaffar:
please test for CMSSW_16_0_PY312_X

lets see if bot is able to test it (I am afraid it might not as there is no PY312 IB)

smuzaffar:
lets merge it and build an IB for PY312

PR #10056, opened 2025-09-03
[PY312] Build PyTorch with rocm, with cuda or without accelerators

iarspider:
TODO:
- [x] Toolfiles
- [x] pytorch-x packages
- [ ] multiarch

PR #10123, opened 2025-10-14
[PY312] Build PyTorch with rocm, with cuda or without accelerators

iarspider:
Continuation of https://github.com/cms-sw/cmsdist/pull/10056


iarspider:
please test

iarspider:
please test

iarspider:
please test

iarspider:
please test

iarspider:
please test

iarspider:
please test

iarspider:
please test

valsdav:
Hi @iarspider @smuzaffar! Just for my understanding,  what is the current technical issue with the simultaneous ROCm and CUDA support for PyTorch compilation? Is there something we can do to help on this issue?  

Thanks a lot and sorry for asking again

fwyzard:
My understanding is that PyTorch does not support building for both CUDA and ROCm at the same time.

The idea would be to have 2 or 3 independent builds from the same source tree (1 for CUDA, 1 for ROCm, maybe 1 for CPU-only), and somehow pick the correct one based on what CMSSW wants to use (CUDA, ROCm, or CPU-only).

I'm not sure if there was any progress on this front, thought. I think if you or the ML group wants to help, that would be very welcome :-)

iarspider:
@valsdav one can't build ROCm and CUDA versions from a single source tree because for ROCm the build process modifies the source code heavily (renames classes, methods, files) before actual compilation happens. We never tested if we can manually merge ROCm and CUDA installations after building, but I would expect a lot of headache and weird side effects to happen if someone tries. So, as @fwyzard said, the idea is to have 2 or 3 builds (CPU, CUDA, ROCm) and somehow switch them at runtime.

smuzaffar:
> Hi @iarspider @smuzaffar! Just for my understanding, what is the current technical issue with the simultaneous ROCm and CUDA support for PyTorch compilation? Is there something we can do to help on this issue?

@valsdav , with scram dynamic runtime env we should be able to get torch for cpu/cuda/rocm in cmssw. I have a rough idea how to do it, let me first get toruch 2.10.0 in PY312 IBs and then I will see what we can do to get toruch build or rocm/cuda

makortel:
> with scram dynamic runtime env we should be able to get torch for cpu/cuda/rocm in cmssw.

I _think_ the PyTorch backend selection logic should honor CMSSW's `process.options.accelerators` configuration parameter (https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideEDMParametersForModules#The_options_Parameter_Set).

smuzaffar:
this has been replaced by #10390

PR #8901, opened 2023-12-20
Add L1METML extra

jmduarte:
This PR adds one new package needed for a forthcoming CMSSW PR for an ML-based Level-1 trigger MET regression algorithm: [L1METML](https://github.com/cms-hls4ml/L1METML).

This follows similar PRs for AXOL1TL https://github.com/cms-sw/cmsdist/pull/8379 and CICADA https://github.com/cms-sw/cmsdist/pull/8307

@quinnanm @vloncar @thesps

PR #10129, opened 2025-10-16
EMTF NN v1

pallabidas:
EMTF NN model v1.0.0

model release: https://github.com/cms-hls4ml/EMTF_NN/releases/tag/v1.0.0

### DIFF
--- a/scram-tools.file/tools/EMTF_NN/emtf_nn.xml
+++ b/scram-tools.file/tools/EMTF_NN/emtf_nn.xml
@@ -1,6 +1,6 @@
-<tool name="emtf_nn" version="@TOOL_VERSION@" revision="1">
+<tool name="@TOOL@" version="@TOOL_VERSION@" revision="1">
   <client>
-    <environment name="EMTF_NN_BASE" default="@TOOL_ROOT@"/>
-    <environment name="LIBDIR"       default="$EMTF_NN_BASE/lib64"/>
+    <environment name="@TOOL_BASE@" default="@TOOL_ROOT@"/>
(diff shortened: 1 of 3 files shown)


smuzaffar:
@pallabidas , as this is new external, so you need to provide its scram's xml definition and also need to add it as dependency on cmssw. see https://github.com/cms-sw/cmsdist/pull/8901/files as an example , note that now cmssw deps should go in to https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_16_0_X/master/cmssw-tools.spec file.

scram-tools.file/tools/EMTF_NN/emtf_nn.xml
```diff
@@ -0,0 +1,6 @@
+<tool name="emtf_nn" version="@TOOL_VERSION@" revision="1">
+  <client>
+    <environment name="EMTF_NN_BASE" default="@TOOL_ROOT@"/>
+    <environment name="LIBDIR"       default="$EMTF_NN_BASE/lib64"/>
```
smuzaffar:
```suggestion
<tool name="@TOOL@" version="@TOOL_VERSION@" revision="1">
  <client>
    <environment name="@TOOL_BASE@" default="@TOOL_ROOT@"/>
    <environment name="LIBDIR"       default="@TOOL_BASE@/lib64"/>
```

@pallabidas , can you also please apply this change? build/packaging system will replace `@VAR@` with the correct values.

smuzaffar:
please test

smuzaffar:
+externals

PR #10147, opened 2025-10-24
toolfile fix: make sure there is new line at the end of toolfile

smuzaffar:
The scram toolfile does not have new line at the end then generating toolfile or tool-conf fails. This change makes sure that we always add new line if it is missing


smuzaffar:
please test

smuzaffar:
+externals

PR #10160, opened 2025-10-31
Update AdePT to v0.1.8

SeverinDiederichs:
This PR updates AdePT to v0.1.8, which includes an important fix for the B field and the possibility to add experiment-specific SteppingActions. Thus, a new flag must be passed to use the CMS GPU SteppingAction.

This will require a new IB, as I cannot use a local AdePT build within CMS that has interface changes.


smuzaffar:
Please test

civanch:
@smuzaffar , it seems, that the fix is working, is it possible to merge today, so it will be included into the next ADEPT IB?

smuzaffar:
+externals

good to go in special G4ADEPT IBs

PR #10161, opened 2025-11-01
Build ROOT with the external version of Nlohmann Json

fwyzard:
ROOT comes with Nlohmann Json v3.10, while other CMS externals use v3.11.
This may cause conflicts when building an external or CMS package that use both ROOT and Json, because the version being used depend on the order of the `-I` flags passed to the compiler.

### DIFF
--- a/cepgen.spec
+++ b/cepgen.spec
@@ -4,6 +4,8 @@ Source: https://github.com/cepgen/cepgen/archive/refs/tags/%{realversion}.tar.gz
 
 BuildRequires: cmake ninja
 Requires: gsl OpenBLAS hepmc hepmc3 lhapdf pythia6 root bz2lib zlib xz python3
+# ROOT uses the json package, and seems to require that it be availble also when other packages use ROOT
+Requires: json
 
 %prep
 %setup -n %{n}-%{realversion}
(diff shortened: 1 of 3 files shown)


fwyzard:
type bugfix

fwyzard:
please test

fwyzard:
please test

fwyzard:
please test

smuzaffar:
@fwyzard , as `%{cmake_prefix_path}` only uses the direct dependencies of a package, so looks like you need to add `json` dependency in `dd4hep and cepgen`

fwyzard:
please test

fwyzard:
this seems ... random

fwyzard:
please test

fwyzard:
> @fwyzard , as `%{cmake_prefix_path}` only uses the direct dependencies of a package, so looks like you need to add `json` dependency in `dd4hep and cepgen`

The problem I've been running into is that on my local machine everything works (possibly because the same version of json is installed as a system package).

If it's only these two packages I guess adding the extra dependency by hand is not too much of a problem.

If there are more packages, I think this should somehow be fixed in the package for ROOT itself.

fwyzard:
please test with cms-sw/cmssw#49278

fwyzard:
please test with cms-sw/cmssw#49278

fwyzard:
please test with cms-sw/cmssw#49278

fwyzard:
please test with cms-sw/cmssw#49278 for el9_amd64_gcc13

smuzaffar:
looks good but this should only go in after https://github.com/cms-sw/cmssw/pull/49278 has been integrated.

fwyzard:
@smuzaffar now that cms-sw/cmssw#49278 is merged, could we go ahead with this too ?

smuzaffar:
please test

lets build without https://github.com/cms-sw/cmssw/pull/49278 to sure every thing works

smuzaffar:
+externals

PR #10167, opened 2025-11-03
sync vecgeom/geant4 changes from g4 branch

smuzaffar:


smuzaffar:
+externals

this should fix the build/link errors for G4ADEPT Its

PR #10178, opened 2025-11-06
Update clang version for ROCm 7

fwyzard:


fwyzard:
please test for CMSSW_16_0_ROOT636_X/el8_amd64_gcc14

smuzaffar:
@fwyzard , we need to update openmpi or at least get https://github.com/open-mpi/ompi/commit/aa5577441ff1ab7f97f8b63e442b37457c7bd997 patch and apply it to existing version

fwyzard:
We are already at 5.0.8 which is the latest release... there is a release candidate for 5.0.9, but I don't know if that's OK to use.

fwyzard:
https://github.com/cms-sw/cmsdist/pull/10180 has the minimal patch

fwyzard:
Please test for CMSSW_16_0_ROOT636_X/el8_amd64_gcc14 with #10180

fwyzard:
Please test with #10180

smuzaffar:
> We are already at 5.0.8 which is the latest release... there is a release candidate for 5.0.9, but I don't know if that's OK to use.

Looks like  5.0.9 is available (https://github.com/open-mpi/ompi/tree/v5.0.9 )

smuzaffar:
Please test for CMSSW_16_0_ROOT636_X/el8_amd64_gcc14 with https://github.com/cms-sw/cmsdist/pull/10180

smuzaffar:
@fwyzard , while we are here, can you please also drop the condition `%if 0%{?rhel} < 10` (https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_16_0_X/g14/rocm.spec#L55) ? The fix should be need for el10 too. We did not see the error in el10 IBs because el10 software stack was build without ROCm. Now as version 7.0.2 is available for el10, I would like to enable it for el10

fwyzard:
Sure.

fwyzard:
> Looks like 5.0.9 is available (https://github.com/open-mpi/ompi/tree/v5.0.9 )

Ah, thanks for looking into it - let's see if https://github.com/cms-sw/cmsdist/pull/10182 works.

smuzaffar:
lets get this gcc14 IBs

PR #10205, opened 2025-11-17
OpenLDAP: Added missing dependency on uuid

smuzaffar:
Testing full rebuild of CMS software stack showed that `openldap` was missing dependency on `libuuid` which was picked up from system.


smuzaffar:
please test

smuzaffar:
+externals

external build without dependency issue

PR #10215, opened 2025-11-20
[dqmgui] Changes needed for root638 and std=c++20

smuzaffar:
Changes needed to build dqmgui using root638 and std=c++20


smuzaffar:
merging it as there is no ROOT638_X IBs , so can not run tests via PR tests

akritkbehera:
Compilation were failing under C++20 because some occurrences of isnan and fabs. Locally I replaced them with std::isnan and std::fabs to fix that.

makortel:
I just want to note that AFAIK the upcoming ROOT 6.38 is a short term support release (i.e. all support stops when 6.40 will be released in May 2026), so I'd not move CMSSW to it. I have nothing against 6.38-specific IB though if that brings value over the ROOT master IBs.

PR #10221, opened 2025-11-24
Update MPICH to version v4.3.2

fwyzard:


fwyzard:
please test

fwyzard:
@smuzaffar @iarspider something seems wrong with the EL10 images ?

fwyzard:
type bugfix

MPICH v4.3.1 may fail to run if built with GPU support (like we do) and used on machines with GPUs disabled (which we may want to do).

smuzaffar:
please test

@fwyzard , issue is https://cern.service-now.com/service-portal?id=outage&n=OTG0158718 which is causing `cmssw-elX` to properly use the unpacked images. I have updated bot to keep going with testing even if there are cms-env check fails. This will allow to run all the tests but will mark the checking of cms-env as failed. For now we can ignore such failures

fwyzard:

smuzaffar:
+externals

ftenchini:
+1

PR #10223, opened 2025-11-26
Add switch to build_options to use C++23

iarspider:


smuzaffar:
+externals

this is nt used by any thing  , so should be good to go in

PR #10244, opened 2025-12-08
Celeritas: 0.6.3: Use actual github release artifact

smuzaffar:


smuzaffar:
please test

smuzaffar:
please test for CMSSW_16_0_G4VECGEOM_X

celeritas.spec
```diff
@@ -17,7 +15,7 @@ mkdir ../build
 cd ../build
 
 cmake ../%{n}-%{realversion} \
-  -DCeleritas_GIT_DESCRIBE="%{celeritas_gitversion};;" \
+  -DCeleritas_GIT_DESCRIBE="%{realversion};;" \
```
sethrj: This is now unnecessary since you're using the artifact rather than a shallow clone.### 
```suggestion
```

sethrj (approved):
Thanks for the very quick response @smuzaffar !

celeritas.spec
smuzaffar: looks like artifact does not work as it extract every thing under top level directory. Normally packaging system assumes that unpacking a tar will create a single directory with the contents inside. I am now trying the github tar resources https://github.com/celeritas-project/celeritas/archive/refs/tags/v%{realversion}.tar.gz which properly creates celeritas-0.6.3 directory

celeritas.spec
```diff
@@ -1,7 +1,5 @@
-### RPM external celeritas v0.6.0
-%define celeritas_gitversion %(echo %{realversion} | sed -e 's|^v||;s|-.*||')
-%define tag dfa4cde7d7d65bf656b17a24c59fcc030aa6b0d9 
-Source: git+https://github.com/celeritas-project/celeritas?obj=develop/%{tag}&export=%{n}-%{realversion}&output=/%{n}-%{realversion}.tgz
+### RPM external celeritas 0.6.3
+Source: https://github.com/celeritas-project/celeritas/archive/refs/tags/v%{realversion}.tar.gz
```
sethrj: Can you please use

which is hashed, released as an immutable artifact, and is guaranteed to have the correct versioning information embedded (not relying on the `GIT_DESCRIBE` hack below)?
```suggestion
Source: https://github.com/celeritas-project/celeritas/releases/download/v%{realversion}/celeritas-%{realversion}.tar.gz
```

smuzaffar:
please test

smuzaffar:
please test for CMSSW_16_0_G4VECGEOM_X

celeritas.spec
sethrj: That's strange, it looks like both versions behave the same (and both successfully install with spack)... I'll see if I can figure out what makes it different for the future.

smuzaffar:
please test for CMSSW_16_0_G4ADEPT_X

celeritas.spec
smuzaffar: I have updated it to use the github release artifact (I hust had to use the correct `%setup` options to unpack it.).
By the way , I see cmake messages like
```
-- Check if compiler accepts -pthread
-- Check if compiler accepts -pthread - yes
-- Found Geant4: /build/muz/x/w/el8_amd64_gcc13/external/geant4/11.2.2-507472096f383abf946fff1eb5c9dba6/lib64/cmake/Geant4/Geant4Config.cmake (found version "11.2.2")
-- Found nlohmann_json: /build/muz/x/w/el8_amd64_gcc13/external/json/3.11.3-3bf93f232c63d55abe3d04e52aa0a168/share/cmake/nlohmann_json/nlohmann_jsonConfig.cmake (found suitable version "3.11.3", minimum required is "3.7.0")
-- Found Python: /build/muz/x/w/el8_amd64_gcc13/external/python3/3.9.14-e16d2924e9eb9db8fddd14e187cf7209/bin/python3.9 (found suitable version "3.9.14", minimum required is "3.6") found components: Interpreter
-- Could NOT find G4VG: found neither G4VGConfig.cmake nor g4vg-config.cmake (Required is at least version "1.0.3")
-- Executing download step for cli11
-- Downloading...
   dst='/build/muz/x/w/BUILD/el8_amd64_gcc13/external/celeritas/0.6.3-c2a460fcce8f29afb6e805e22622a4d2/build/_deps/cli11-tmp/v2.5.0.tar.gz'
   timeout='none'
   inactivity timeout='none'
-- Using src='https://github.com/CLIUtils/CLI11/archive/refs/tags/v2.5.0.tar.gz'
-- verifying file...
       file='/build/muz/x/w/BUILD/el8_amd64_gcc13/external/celeritas/0.6.3-c2a460fcce8f29afb6e805e22622a4d2/build/_deps/cli11-tmp/v2.5.0.tar.gz'
-- Downloading... done
-- extracting...
     src='/build/muz/x/w/BUILD/el8_amd64_gcc13/external/celeritas/0.6.3-c2a460fcce8f29afb6e805e22622a4d2/build/_deps/cli11-tmp/v2.5.0.tar.gz'
     dst='/build/muz/x/w/BUILD/el8_amd64_gcc13/external/celeritas/0.6.3-c2a460fcce8f29afb6e805e22622a4d2/build/_deps/cli11-src'
-- extracting... [tar xfz]
-- extracting... [analysis]
-- extracting... [rename]
-- extracting... [clean up]
-- extracting... done
-- Executing update step for cli11
-- Executing patch step for cli11
-- Executing download step for g4vg
-- Downloading...
   dst='/build/muz/x/w/BUILD/el8_amd64_gcc13/external/celeritas/0.6.3-c2a460fcce8f29afb6e805e22622a4d2/build/_deps/g4vg-tmp/g4vg-1.0.5.tar.gz'
   timeout='none'
   inactivity timeout='none'
-- Using src='https://github.com/celeritas-project/g4vg/releases/download/v1.0.5/g4vg-1.0.5.tar.gz'
-- verifying file...
       file='/build/muz/x/w/BUILD/el8_amd64_gcc13/external/celeritas/0.6.3-c2a460fcce8f29afb6e805e22622a4d2/build/_deps/g4vg-tmp/g4vg-1.0.5.tar.gz'
-- Downloading... done
-- extracting...
     src='/build/muz/x/w/BUILD/el8_amd64_gcc13/external/celeritas/0.6.3-c2a460fcce8f29afb6e805e22622a4d2/build/_deps/g4vg-tmp/g4vg-1.0.5.tar.gz'
     dst='/build/muz/x/w/BUILD/el8_amd64_gcc13/external/celeritas/0.6.3-c2a460fcce8f29afb6e805e22622a4d2/build/_deps/g4vg-src'

```

I guess g4vg is for Vecgeom, why it downloads VG ? we it should use one from our externals. We do set CMAKE_PREFIX_PATH to include path to our VG

smuzaffar:
please test

smuzaffar:
please test for CMSSW_16_0_G4ADEPT_X

smuzaffar:
please test for CMSSW_16_0_GEANT4_X

smuzaffar:
+externals

G4VECGEOM and G4ADEPT externals build now

celeritas.spec
kpedro88: @sethrj can you comment on this?

sethrj: Thanks for the ping, I didn't see this. G4VG is a small library with only geant4 and VecGeom as dependencies that constructs the VG geometry. It's relatively new but I'm surprised we didn't add an external already; it predates 0.6. We should add the external as it's also needed for adept.

sethrj: @whokion could you add a G4VG external?

whokion: @sethrj I am a bit confused about why we need to add g4vg as an external package and dependency, given that both Celeritas and AdePT already use g4vg internally. Are we intending to introduce an explicit external g4vg dependency for Celeritas and AdePT? If so, we also need to update celeritas.xml to remove the g4vg lib from the celeritas build.

kpedro88: Right now, each external (Celeritas and AdePT) download and build g4vg separately. This is not the recommended practice. Any shared external dependency should be declared as its own package so that it is only downloaded and built once, and then the dependent externals link against it.

whokion: I can add g4vg as an external dependency, but then we would need to disable the internally provided g4vg (which is currently the default). We would also need to add the g4vg dependency in celeritas.spec and make the necessary updates in celeritas.xml to follow the recommended practice. AdePT would likely need to apply similar updates as well. As the first step, I may add g4vg as an external for now.

kpedro88: Yes, these steps would all be necessary.

sethrj: Sounds great! @whokion when you do this, could you please add these cmake arguments:
```sh
-DCELERITAS_BUILTIN_CLI11:BOOL=OFF \
-DCELERITAS_BUILTIN_G4VG:BOOL=OFF \
-DCELERITAS_BUILTIN_GTest:BOOL=OFF \
-DCELERITAS_BUILTIN_covfie:BOOL=OFF \
-DCELERITAS_BUILTIN_nlohmann_json:BOOL=OFF \
-DCELERITAS_BUILD_APPS:BOOL=OFF \
```
to ensure we're not accidentally building or "vendoring" any of these dependencies? This also prevents the unnecessary building of the front-end apps.

sethrj: And to be clear @whokion : the built-in G4VG isn't a default; it's automatically selected based on the available packages. The `...BUILTIN...` variables can be used to forcibly enable/disable those.

PR #10255, opened 2025-12-15
Geant4: Fix typo in member variable assignment

smuzaffar:
Integrated https://github.com/cms-externals/geant4/pull/102

There was a bug/typo in geant4 code which GCC 14 and earlier did not catch but with GCC15 we get build errors. As CMS does not use this part of Geant4 code so it should be safe to get this in default 16.0.X Its


smuzaffar:
please test

smuzaffar:
Please test

smuzaffar:
+externals

PR #10256, opened 2025-12-17
Add g4vg external to support celeritas and adept

whokion:
Currently, g4vg is installed internally as part of the celeritas and adept builds. Making g4vg available as a standalone external package will allow both celeritas and adept to use a common external library. Once this external is successfully added, celeritas will be updated to use the g4vg external accordingly. This change was requested by @kpedro88.

### DIFF
--- a/g4vg.spec
+++ b/g4vg.spec
@@ -1,0 +1,40 @@
+### RPM external g4vg v1.0.5
+%define g4vg_gitversion %(echo %{realversion} | sed -e 's|^v||;s|-.*||')
+%define tag d377c26ea725d35d222f8b8dbd1094d0bb9a7a76
+Source: git+https://github.com/celeritas-project/g4vg?obj=main/%{tag}&export=%{n}-%{realversion}&output=/%{n}-%{realversion}.tgz
+
+%define package_build_flags -Wall -Wextra -pedantic
+## INCLUDE geant4-deps
+Requires: geant4 vecgeom
+
+%prep
+%setup -n %{n}-%{realversion}
+
+%build
+
+rm -rf ../build
+mkdir ../build
+cd ../build
+
+cmake ../%{n}-%{realversion} \
+  -DCMAKE_INSTALL_PREFIX=%{i} \
+  -DCMAKE_CXX_STANDARD:STRING="%{cms_cxx_standard}" \
+  -DCMAKE_AR=$(which gcc-ar) \
+  -DCMAKE_RANLIB=$(which gcc-ranlib) \
+  -DCMAKE_BUILD_TYPE=%{cmake_build_type} \
+  -DCMAKE_CXX_FLAGS="%{build_flags}" \
+  -DCMAKE_C_FLAGS="%{build_flags}" \
+  -DCMAKE_STATIC_LIBRARY_CXX_FLAGS="%{build_flags}" \
+  -DCMAKE_STATIC_LIBRARY_C_FLAGS="%{build_flags}" \
+  -DCMAKE_PREFIX_PATH="%{cmake_prefix_path}" \
+  -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
+  -DBUILD_SHARED_LIBS=ON
+
+make %{makeprocesses} VERBOSE=1
+
+%install
+cd ../build
+make %{makeprocesses} install VERBOSE=1
+
+%post
+%{relocateConfig}lib64/cmake/G4VG/G4VGConfig.cmake
--- a/scram-tools.file/tools/g4vg/g4vg.xml
+++ b/scram-tools.file/tools/g4vg/g4vg.xml
@@ -1,0 +1,12 @@
+<tool name="@TOOL@" version="@TOOL_VERSION@" revision="1">
+  <info url="https://github.com/celeritas-project/g4vg"/>
+  <lib name="g4vg"/>
+  <client>
+    <environment name="@TOOL_BASE@" default="@TOOL_ROOT@"/>
+    <environment name="INCLUDE" default="$@TOOL_BASE@/include"/>
+    <environment name="LIBDIR" default="$@TOOL_BASE@/lib64"/>
+  </client>
+  <use name="geant4core"/>
+  <use name="vecgeom_interface"/>
+  <use name="vecgeom"/>
+</tool>


kpedro88:
@whokion ideally the update to celeritas.spec would also be made in this PR

whokion:
@kpedro88 I would like to proceed in two steps, as the first step should not affect anything, and make sure it build successfully.

whokion:
@smuzaffar Thanks for changing the base branch to IB/CMSSW_16_1_X/master. Please let me know if there is anything else I should follow up on.

g4vg.spec
```diff
@@ -0,0 +1,40 @@
+### RPM external g4vg v1.0.5
+%define g4vg_gitversion %(echo %{realversion} | sed -e 's|^v||;s|-.*||')
```
smuzaffar, line 2: @whokion , looks like this is not needed ( at least I do not see it's use in the spec file).

g4vg.spec
```diff
@@ -0,0 +1,40 @@
...
+%define tag d377c26ea725d35d222f8b8dbd1094d0bb9a7a76
+Source: git+https://github.com/celeritas-project/g4vg?obj=main/%{tag}&export=%{n}-%{realversion}&output=/%{n}-%{realversion}.tgz
+
+%define package_build_flags -Wall -Wextra -pedantic
+## INCLUDE geant4-deps
+Requires: geant4 vecgeom
```
smuzaffar, line 8: @whokion , please removed `vecgeom` dependency from here. It should be conditionally added via [geant4-deps](https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_16_1_X/master/geant4-deps.file) if geant4 was built with vecgeom.

scram-tools.file/tools/g4vg/g4vg.xml
```diff
@@ -0,0 +1,12 @@
...
+  <info url="https://github.com/celeritas-project/g4vg"/>
+  <lib name="g4vg"/>
+  <client>
+    <environment name="@TOOL_BASE@" default="@TOOL_ROOT@"/>
+    <environment name="INCLUDE" default="$@TOOL_BASE@/include"/>
+    <environment name="LIBDIR" default="$@TOOL_BASE@/lib64"/>
+  </client>
+  <use name="geant4core"/>
+  <use name="vecgeom_interface"/>
+  <use name="vecgeom"/>
```
smuzaffar, line 11: @whokion , please remove these two dependencies. They will be automaticlaly added via `geant4core` is geant4 was built with vecgeom

kpedro88:
@smuzaffar in the case when Geant4 is not built with VecGeom, is VecGeom still built? The point of this external is specifically to "build a VecGeom geometry representation from a Geant4 user application using the in-memory Geant4 volume representation", so I'm not sure if it even makes sense to provide it in the non-VecGeom Geant4 case...

smuzaffar:
> @smuzaffar in the case when Geant4 is not built with VecGeom, is VecGeom still built?

No, as vecgeom dependency is only added via geant4 so in that case we do not build vecgeom.

This external is used by `celeritas` ... right? I would suggest that we update `celeritas.spec` to have something like
```
%{?enable_vecgeom:Requires g4vg}
```

and then in the `cmake` configure command enable it 
```
%if %{enable_vecgeom}
  -DCELERITAS_USE_VecGeom=ON
  #flags to enable G4VG
%else
  -DCELERITAS_USE_VecGeom=OFF
  #flags to disable G4VG
%endif
```

smuzaffar:
note that having a spec file in cmsdist does not mean that we will built it. We only build externals which are directly of indirectly dependencies of cmssw-tool-conf. So having something like `%{?enable_vecgeom:Requires g4vg}` in celeritas.spec  will built it if vecgeom is enabled.

whokion:
@smuzaffar, @kpedro88 Thank you very much for your suggestion and the related comments. I will close this MR for now and reopen a new one with the suggested changes in Celeritas later after further testing.

sethrj:
I wasn't aware of this PR (sorry!)

Some notes to reiterate with an ongoing conversation with @kpedro88 :
- Celeritas and ADePT require G4VG
- G4VG requires VecGeom and Geant4
- G4VG is *agnostic* to configuration settings and versions for either VG or G4 (no issues with/without CUDA, or VG 1/2)

kpedro88:
@smuzaffar the upshot is that Celeritas itself (as well as g4vg) should only be built if VecGeom is available. How should we express this condition in the spec files?

g4vg.spec
```diff
@@ -0,0 +1,40 @@
+### RPM external g4vg v1.0.5
+%define g4vg_gitversion %(echo %{realversion} | sed -e 's|^v||;s|-.*||')
+%define tag d377c26ea725d35d222f8b8dbd1094d0bb9a7a76
+Source: git+https://github.com/celeritas-project/g4vg?obj=main/%{tag}&export=%{n}-%{realversion}&output=/%{n}-%{realversion}.tgz
```
sethrj, line 4: Can you just use the release download from https://github.com/celeritas-project/g4vg/releases/tag/v1.0.5:
```
https://github.com/celeritas-project/g4vg/releases/download/v1.0.5/g4vg-1.0.5.tar.gz
```

smuzaffar:
@kpedro88 , https://github.com/cms-sw/cmsdist/pull/10277 should not build/include `Celeritas` if `Vecgeom` is not enabled ( i.e. `geant4` is build without `vecgeom`). Once that change worked then you can just update `Celeritas` spec with assumption that it will only be build (in cmssw software stack) if geant4 is built with vecgeom

kpedro88:
thanks @smuzaffar ! so we can copy this approach for g4vg also?

smuzaffar:
if `g4vg` is a separate package (i.e. has its own `g4vg.spec`) and is only included in `celeritas.spec` then no need to do any thing special for it. It will be only built if  `celeritas` is built

smuzaffar:
I guess we can reopen this PR to get g4vg build. for now I would suggest to add `Requires: g4vg` in celetitas.spec so that it gets build. Later you can change `celeritas` to actually make use of it

whokion:
@sethrj Could you take over the MR to add g4vp (with a fresh branch of your fork from cms-sw:IB/CMSSW_16_1_X)?  - somehow, I am not able to add the remote cmsdist master to my local repo.  And you are the owner of g4vg.  Thanks.

sethrj:
I'll see if I can, sure...

PR #10257, opened 2025-12-18
 For gcc15 added cuda 13.1.0 as this release introduces support for G…

raoatifshad:
Added CUDA 13.1.0 support since CUDA 13.x introduces compatibility
with GCC 15 and Clang 20, which is required for newer toolchains.

Additionally, address a known CUDA installer issue related to the
default log file path (/tmp/cuda-installer.log). When the file is
created by another user, subsequent installations may fail due to
permission errors.

To avoid this, explicitly create the required directory before the
installation and clean it up afterward. This ensures the installer
can write logs safely in multi-user build environments.

References:
- https://developer.nvidia.com/blog/whats-new-and-important-in-cuda-toolkit-13-0/#:~:text=The%20CUDA%20Toolkit%2013.0%20release%20introduces%20support,removes%20support%20for%20ICC%20and%20MSVC%202017
- Known installer issue with cuda-installer.log
  https://forums.developer.nvidia.com/t/change-the-path-of-install-log-when-installing-cuda-toolkit/183179

### DIFF
--- a/cuda.spec
+++ b/cuda.spec
@@ -1,8 +1,8 @@
-### RPM external cuda 12.9.1
+### RPM external cuda 13.1.0
 ## INITENV +PATH LD_LIBRARY_PATH %i/lib64
 
 %define runpath_opts -m compute-sanitizer -m drivers -m nvvm
-%define driversversion 575.57.08
+%define driversversion 590.44.01
 
 %ifarch x86_64
 Source0: https://developer.download.nvidia.com/compute/cuda/%{realversion}/local_installers/%{n}_%{realversion}_%{driversversion}_linux.run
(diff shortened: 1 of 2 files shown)


fwyzard:
Why do we use GCC 15 in the `g14` branch ?

smuzaffar:
@fwyzard , title os PR is misleading, we still use GCC 14 in this branch.
As we do not have GCC 15  IBs, so we can not test it. This PR I would like to test with c++23 IBs ( which are using this branch). If you want, we can try testing it for default gcc13 IBs too

fwyzard:
OK, but then I'm confused - CUDA 13.x does not support C++23 ...

fwyzard:
As for testing in the main branch, I have no objections to _testing_ it, but we cannot move to CUDA 13.x until we agree to drop support for Pascal and Volta GPUs :-(

smuzaffar:
> OK, but then I'm confused - CUDA 13.x does not support C++23 ...

that is what I wanted to test, there is good chance that it will fail :-)

PR #10263, opened 2026-01-05
[ROOT636] Updated root to tip of branch v6-36-00-patches

akritkbehera:


akritkbehera:
Please test

smuzaffar:
@makortel , this change includes https://github.com/root-project/root/compare/3a3891e8b30512fca238abf66b74ce4459f3b37c...v6-36-00-patches root changes on top of what already have in 16.1.X builds. It contains a fix https://github.com/root-project/root/commit/e5ee22f71b2e5db5817ae7e7f33a7193270c6501 for https://github.com/root-project/root/issues/20706 . Should we consider this change for 16.1.X default builds ( and may be a backport to 16.0.X ?)

makortel:
> Should we consider this change for 16.1.X default builds ( and may be a backport to 16.0.X ?)

I agree, also for the backport. While the framework doesn't use `TTree::Merge()` (what I recall), users might, and the changes seems simple enough to be comfortable with a backport.

smuzaffar:
+externals

good to go in 16.1.X IBs

PR #8280, opened 2023-02-01
Add  -fno-reciprocal-math -mrecip=none to Ofast to solve the AMD/Intel differences

silviodonato:
About one year ago we discovered differences of few percents between AMD and Intel in few HLT paths when rerunning the HLT [1]. The situation got worse when we added the parking di-electron paths with dZ path, which gave a difference of ~8% [2].

After several failed attempts, @gparida managed to solve the differences following @VinInn suggestion [3] to use `-Ofast  -fno-reciprocal-math -mno-recip`.

@gparida replaced `<use name="ofast-flag"/>` with `<flags   CXXFLAGS="-Ofast  -fno-reciprocal-math -mrecip=none"/>` in all `BuildFiles.xml` (except `test` folders to save compilation time)
```
"RecoEgamma/EgammaElectronAlgos/BuildFile.xml"
"RecoEgamma/EgammaPhotonAlgos/BuildFile.xml"
"RecoPixelVertexing/PixelTriplets/BuildFile.xml"
"RecoPixelVertexing/PixelTriplets/plugins/BuildFile.xml"
"RecoTracker/FinalTrackSelectors/BuildFile.xml"
"RecoTracker/FinalTrackSelectors/plugins/BuildFile.xml"
"RecoTracker/TkDetLayers/BuildFile.xml"
"RecoTracker/TkHitPairs/BuildFile.xml"
"RecoVertex/PrimaryVertexProducer/BuildFile.xml"
"TrackingTools/GsfTools/BuildFile.xml"
"TrackingTools/GsfTools/plugins/BuildFile.xml"
"TrackingTools/GsfTracking/BuildFile.xml"
"TrackingTools/KalmanUpdators/BuildFile.xml"
"TrackingTools/MaterialEffects/BuildFile.xml"
"TrackingTools/MaterialEffects/plugins/BuildFile.xml"
"TrackingTools/TrackFitters/BuildFile.xml"
"TrackingTools/TrajectoryState/BuildFile.xml"
```
and found no AMD/Intel differences @gparida please share your results here (which release did you use?).

So I made this PR to implement this change directly at `cmsdist` level 
(but I didn't test it compiling the externals).

I don't know exactly the impact on timing, but I assume it will be much smaller than 1%, that was the impact of the full removal of Ofast [4]

[1] https://its.cern.ch/jira/browse/CMSHLT-2257
[...]

PR #10267, opened 2026-01-11
Do not use `-Ofast`

fwyzard:
`-Ofast` implies `-funsafe-math-optimizations`, which affects the global FP state for all programs: when used at link time, it may include libraries or startup files that change the default FPU control word or other similar optimisations.

`-Ofast` enables, directly or indirectly:
  - ❌ `-fallow-store-data-races`
  - ✔ `-fno-semantic-interposition`
  - ✔ `-fno-math-errno`
  - ❌ `-funsafe-math-optimizations`
  - ❌ `-ffinite-math-only`
  - ✔ `-fno-rounding-math` (_default_)
  - ✔ `-fno-signaling-nans` (_default_)
  - ❔ `-fcx-limited-range`
  - ✔ `-fexcess-precision=fast`
  - ✔ `-fno-signed-zeros`
  - ✔ `-fno-trapping-math`
  - ✔ `-fassociative-math`
  - ❌ `-freciprocal-math` (https://github.com/cms-sw/cmsdist/pull/8280)

We should not use `-fallow-store-data-races`, `-ffinite-math-only`.

We should not use `-funsafe-math-optimizations` at link time, as it affects the global FP state. It's probably easier to replace it with the individual options `-fno-signed-zeros`, `-fno-trapping-math` and `-fassociative-math`.

We may revisit `-fcx-limited-range` if we do use Complex arithmetics.

`-freciprocal-math` was already disabled explicitly, see cms-sw/cmsdist#8280.

### DIFF
--- a/scram-tools.file/tools/gcc/ofast-flag.xml
+++ b/scram-tools.file/tools/gcc/ofast-flag.xml
@@ -1,12 +1,24 @@
-  <tool name="ofast-flag" version="1.0" revision="1">
-    <flags CXXFLAGS="-Ofast -fno-reciprocal-math"/>
-    <ifarch match="ppc64le|amd64|x86_64">
-      <flags CXXFLAGS="-mrecip=none"/>
-    </ifarch>
-    <ifarchitecture name="slc6_">
-      <ifcompiler name="llvm">
-        <flags CXXFLAGS="-fno-builtin"/>
-      </ifcompiler>
-    </ifarchitecture>
-    <flags NO_RECURSIVE_EXPORT="1"/>
-  </tool>
+<tool name="ofast-flag" version="2.0" revision="1">
+  <!--
+  As of GCC version 13.3.0, `-Ofast` enables (directly or indirectly):
+    - `-fallow-store-data-races`
+    - `-fno-semantic-interposition`
+    - `-fno-math-errno`
+    - `-funsafe-math-optimizations`
+    - `-ffinite-math-only`
+    - `-fno-rounding-math` (enabled by default)
+    - `-fno-signaling-nans` (enabled by default)
+    - `-fcx-limited-range`
+    - `-fexcess-precision=fast`
+    - `-fno-signed-zeros`
+    - `-fno-trapping-math`
+    - `-fassociative-math`
+    - `-freciprocal-math` (disabled in cms-sw/cmsdist#8280)
+  We disable `-fallow-store-data-races`, `-funsafe-math-optimizations`, `-ffinite-math-only` and `-freciprocal-math` (see cms-sw/cmsdist#8280).
+  -->
+  <flags CXXFLAGS="-fno-semantic-interposition -fno-math-errno -fno-rounding-math -fno-signaling-nans -fexcess-precision=fast -fno-signed-zeros -fno-trapping-math -fassociative-math -fno-reciprocal-math"/>
+  <ifarch match="ppc64le|amd64|x86_64">
+    <flags CXXFLAGS="-mrecip=none"/>
+  </ifarch>
+  <flags NO_RECURSIVE_EXPORT="1"/>
+</tool>
--- a/xz-bootstrap.spec
+++ b/xz-bootstrap.spec
@@ -5,7 +5,7 @@ Source0: http://tukaani.org/xz/xz-%{realversion}.tar.gz
 %setup -n xz-%{realversion}
 
 %build
-./configure CFLAGS='-fPIC -D_FILE_OFFSET_BITS=64 -Ofast' --prefix=%{i} --disable-static
+./configure CFLAGS='-fPIC -D_FILE_OFFSET_BITS=64 -O3' --prefix=%{i} --disable-static
 make %{makeprocesses}
 
 %install
--- a/xz.spec
+++ b/xz.spec
@@ -7,7 +7,7 @@ BuildRequires: autotools
 %setup -n %{n}-%{realversion}
 
 %build
-./configure CFLAGS='-fPIC -Ofast' --prefix=%{i} --disable-static --disable-nls --disable-rpath --disable-dependency-tracking --disable-doc
+./configure CFLAGS='-fPIC -O3' --prefix=%{i} --disable-static --disable-nls --disable-rpath --disable-dependency-tracking --disable-doc
 make %{makeprocesses}
 
 %install


fwyzard:
- `-fallow-store-data-races`
    > Allow the compiler to perform optimizations that may introduce new data races on stores, without proving that the variable cannot be concurrently accessed by other threads.  Does not affect optimization of local data.  It is safe to use this option if it is known that global data will not be accessed by multiple threads.
    > 
    > Examples of optimizations enabled by `-fallow-store-data-races` include hoisting or if-conversions that may cause a value that was already in memory to be re-written with that same value.  Such re-writing is safe in a single threaded context but may be unsafe in a multi-threaded context.
    > Note that on some processors, if-conversions may be required in order to enable vectorization.

  - `-fno-semantic-interposition`
    > Some object formats, like ELF, allow interposing of symbols by the dynamic linker.  This means that for symbols exported from the DSO, the compiler cannot perform interprocedural propagation, inlining and other optimizations in anticipation that the function or variable in question may change.  While this feature is useful, for example, to rewrite memory allocation functions by a debugging implementation, it is expensive in the terms of code quality.
    > With `-fno-semantic-interposition` the compiler assumes that if interposition happens for functions the overwriting function will have precisely the same semantics (and side effects).  Similarly if interposition happens for variables, the constructor of the variable will be the same.
    > The flag has no effect for functions explicitly declared inline (where it is never allowed for interposition to change semantics) and for symbols explicitly declared weak.

  - `-ffast-math`
    > Sets the options `-fno-math-errno`, `-funsafe-math-optimizations`, `-ffinite-math-only`, `-fno-rounding-math` (_default_), `-fno-signaling-nans` (_default_), `-fcx-limited-range` and `-fexcess-precision=fast`.
    >
    > This option causes the preprocessor macro `__FAST_MATH__` to be defined.

  - `-fno-math-errno`
    > Do not set `errno` after calling math functions that are executed with a single instruction, e.g., `sqrt`.  A program that relies on IEEE exceptions for math error handling may want to use this flag for speed while maintaining IEEE arithmetic compatibility.

  - `-funsafe-math-optimizations`
    > Allow optimizations for floating-point arithmetic that (a) assume that arguments and results are valid and (b) may violate IEEE or ANSI standards.
    > When used at link time, it may include libraries or startup files that change the default FPU control word or other similar optimizations.
    >
    > Enables `-fno-signed-zeros`, `-fno-trapping-math`, `-fassociative-math` and
     `-freciprocal-math`.

  - `-ffinite-math-only`
    > Allow optimizations for floating-point arithmetic that assume that arguments and results are not NaNs or +-Infs.

  - `-frounding-math` (_disabled by default_)
    > Disable transformations and optimizations that assume default floating-point rounding behavior.  This is round-to-zero for all floating point to integer conversions, and round-to-nearest for all other arithmetic truncations.  This option should be specified for programs that change the FP rounding mode dynamically, or that may be executed with a non-default rounding mode.  This option disables constant folding of floating-point expressions at compile time (which may be affected by rounding mode) and arithmetic transformations that are unsafe in the presence of sign-dependent rounding modes.

  - `-fsignaling-nans` (_disabled by default_)
    > Compile code assuming that IEEE signaling NaNs may generate user-visible traps during floating-point operations.  Setting this option disables optimizations that may change the number of exceptions visible with signaling NaNs.  This option implies `-ftrapping-math`.
    >
    > This option causes the preprocessor macro `__SUPPORT_SNAN__` to be defined.

  - `-fcx-limited-range` (complex numbers)
    > When enabled, this option states that a range reduction step is not needed when performing complex division.  Also, there is no checking whether the result of a complex multiplication or division is `NaN + I*NaN`, with an attempt to rescue the situation in that case.

  - `-fexcess-precision=fast`
    > This option allows further control over excess precision on machines where floating-point operations occur in a format with more precision or range than the IEEE standard and interchange floating-point types.  By default, `-fexcess-precision=fast` is ineffect; this means that operations may be carried out in a widerprecision than the types specified in the source if that wouldresult in faster code, and it is unpredictable when rounding to the types specified in the source code takes place.
    > When compiling C or C++, if `-fexcess-precision=standard` is specified then excess precision follows the rules specified in ISO C99 or C++; in particular, both casts and assignments cause values to be rounded to their semantic types (whereas `-ffloat-store` only affects assignments).
    > This option is enabled by default for C or C++ if a strict conformance option such as `-std=c99` or `-std=c++17` isused.

  - `-fno-signed-zeros`
    > Allow optimizations for floating-point arithmetic that ignore the signedness of zero.  IEEE arithmetic specifies the behavior of distinct +0.0 and −0.0 values, which then prohibits simplification of expressions such as x+0.0 or 0.0*x (even with `-ffinite-math-only`).  This option implies that the sign of a zero result isn't significant.

  - `-fno-trapping-math`
    > Compile code assuming that floating-point operations cannot generate user-visible traps.  These traps include division by zero, overflow, underflow, inexact result and invalid operation.  This option requires that `-fno-signaling-nans` be in effect.  Setting this option may allow faster code if one relies on "non-stop" IEEE arithmetic, for example.

  - `-fassociative-math`
    > Allow re-association of operands in series of floating-point operations.  This violates the ISO C and C++ language standard by possibly changing computation result.  NOTE: re-ordering may change the sign of zero as well as ignore NaNs and inhibit or create underflow or overflow (and thus cannot be used on code that relies on rounding behavior like `(x + 2**52) - 2**52`.  May also reorder floating-point comparisons and thus may not be used when ordered comparisons are required.  This option requires that both `-fno-signed-zeros` and `-fno-trapping-math` be in effect. Moreover, it doesn't make much sense with `-frounding-math`.

  - `-freciprocal-math`
    > Allow the reciprocal of a value to be used instead of dividing by the value if this enables optimizations.  For example `x / y` can be replaced with `x * (1/y)`, which is useful if `(1/y)` is subject to common subexpression elimination.  Note that this loses precision and increases the number of flops operating on the value.

fwyzard:
enable gpu

fwyzard:
please test

makortel:
While I agree with the premise of the PR description, I wonder if these options

> We should not use `-fallow-store-data-races`, `-ffinite-math-only`.
> 
> We should not use `-funsafe-math-optimizations` at link time, as it affects the global FP state. It's probably easier to replace it with the individual options `-fno-signed-zeros`, `-fno-trapping-math` and `-fassociative-math`.

impact the GCC auto-vectorization? (that, IIRC, was one major motivation for using `-ffast-math` in the first place)

fwyzard:
IMHO `-funsafe-math-optimizations` should not be used, and should be replaced by (a subset of) the options it enables.

We _could_ look into enabling it at compile time and _not_ at link time, but I'm a bit wary of this mixed approach :-(

fwyzard:
`-fallow-store-data-races` seems dangerous from its description, but indeed I don't have any concrete evidence against it.

fwyzard:
`-ffinite-math-only` is pretty annoying, because it disables also explicit checks like `isnan()`.
While we do have a workaround in CMSSW, it is also used in externals like _e.g._ Eigen, and we cannot easily patch that.

dan131riley:
> We _could_ look into enabling it at compile time and _not_ at link time, but I'm a bit wary of this mixed approach :-(

Does that even work with LTO?  (OTOH, I thought mucking with the global FP state in the shared library init was removed in gcc13.)

fwyzard:
> thought mucking with the global FP state in the shared library init was removed in gcc13

that would be good :-)

It's still mentioned in [the 13.4 documentation](https://gcc.gnu.org/onlinedocs/gcc-13.4.0/gcc/Optimize-Options.html#index-funsafe-math-optimizations), though 🤔

fwyzard:
Let's see what is the impact of removing `-Ofast` only from `libxz` in https://github.com/cms-sw/cmsdist/pull/10273 ...

dan131riley:
> > thought mucking with the global FP state in the shared library init was removed in gcc13
> 
> that would be good :-)
> 
> It's still mentioned in [the 13.4 documentation](https://gcc.gnu.org/onlinedocs/gcc-13.4.0/gcc/Optimize-Options.html#index-funsafe-math-optimizations), though 🤔

'-fast' at link time of an executable (e.g., cmsRun) will include initialization code that changes the FP state, but from the [gcc13 release notes](https://gcc.gnu.org/gcc-13/changes.html#:~:text=%2DOfast%2C%20%2Dffast%2Dmath%20and%20%2Dfunsafe%2Dmath%2Doptimizations%20will%20no%20longer%20add%20startup%20code%20to%20alter%20the%20floating%2Dpoint%20environment%20when%20producing%20a%20shared%20object%20with%20%2Dshared.)

> -Ofast, -ffast-math and -funsafe-math-optimizations will no longer add startup code to alter the floating-point environment when producing a shared object with -shared.

dan131riley:
> `-fallow-store-data-races` seems dangerous from its description, but indeed I don't have any concrete evidence against it.

I'm not absolutely sure, but from my reading I believe what the description means is that it can introduce data races for programs that assume a strongly-ordered memory model.  It should be ok for programs that conform to the C++ memory model with appropriate use of atomics, and it may allow some vectorization optimizations that would otherwise be forbidden.

srlantz:
FWIW, vectorization of trig functions has been important to mkFit timing performance. The most recent versions of mkFit accomplish this vectorization through VDT, which is a header-only library. The latest version of VDT (0.4.6, April 2025) changed the preferred optimization flags from -O3 -ffast-math to just -O3. Therefore I wonder if @dpiparo has any insights on this.

fwyzard:
> Therefore I wonder if @dpiparo has any insights on this.

Or @VinInn .

makortel:
> The latest version of VDT (0.4.6, April 2025) changed the preferred optimization flags from -O3 -ffast-math to just -O3.

I guess we should update VDT anyhow
https://github.com/cms-sw/cmsdist/blob/95e4f6acff72c50aa6acf55c21c234daf260b1a6/vdt.spec#L1

srlantz:
Single-threaded performance tests with mkFit standalone (VDT included) confirms that deleting `-Ofast` and instead substituting `-O3` and the specific options selected by this PR results in a 40% slowdown on the step that is most sensitive to vectorization: initialStep or iteration 0, where all Si strip hits are available for track finding. (Test input: 100 ttbar events at 14TeV, pileup 50, phase 1 detector.)

If the `-Ofast` suboptions that are omitted by this PR are re-added to the list, i.e., `-fallow-store-data-races -funsafe-math-optimizations -ffinite-math-only`, then the prior timing performance is recovered. Of these, `-funsafe-math-optimizations` is observed to restore more than half of the missing performance, while `-fallow-store-data-races` gives an improvement of only 1-2%.

Note that the suboptions of `-Ofast` that are selected for this PR seem somewhat inconsistent:
- The option `-fno-signed-zeros` appears twice in the list.
- The option `-fassociative-math` requires both `-fno-signed-zeros` and `-fno-trapping-math`, but only the first of these appears. Therefore, `-fassociative-math` is ignored, and warnings about it are issued.

Finally, according to [gcc-13 documentation](https://gcc.gnu.org/onlinedocs/gcc-13.4.0/gcc/Optimize-Options.html#index-funsafe-math-optimizations), the option `-funsafe-math-optimizations` enables `-fno-signed-zeros`, `-fno-trapping-math`, `-fassociative-math`, and `-freciprocal-math`. However, substituting`-fno-signed-zeros -fno-trapping-math -fassociative-math` for `-funsafe-math-optimizations -fno-reciprocal-math` (as required by #8280 ) does not restore full performance. So these sets of options are not equivalent. It seems that the combination `-funsafe-math-optimizations -fno-reciprocal-math` does something more, and whatever it does seems to be important to mkFit's timing performance.

fwyzard:
Ah, thanks, I wrongly put `-fno-signed-zeros -fno-signed-zeros` instead of `-fno-signed-zeros -fno-trapping-math`.

fwyzard:
please test

fwyzard:
OK, yes, browsing through the GCC code, `-funsafe-math-optimizations` sets its own internal flag, that enables additional transformations (_e.g._ `a × (b + c)` <-> `a × b + a × c`, _etc_).

fwyzard:
@srlantz how does `-O3 -ffast-math -no-finite-math-only -fno-reciprocal-math` look like ?

srlantz:
> how does `-O3 -ffast-math -no-finite-math-only -fno-reciprocal-math` look like ?

About 20% slower, similar to omitting `-funsafe-math-optimizations`

srlantz:
@fwyzard note also that we generally include `-fno-math-errno` in all standalone compilations as I believe this is common to CMSSW builds

fwyzard:
> @fwyzard note also that we generally include `-fno-math-errno` in all standalone compilations as I believe this is common to CMSSW builds

Sure, but that's already implied by `-ffast-math`.

fwyzard:
> > how does `-O3 -ffast-math -fno-finite-math-only -fno-reciprocal-math` look like ?
> 
> About 20% slower, similar to omitting `-funsafe-math-optimizations`

Pity, because `-ffinite-math-only` is _really_ annoying.

srlantz:
> > > how does `-O3 -ffast-math -fno-finite-math-only -fno-reciprocal-math` look like ?
> > 
> > 
> > About 20% slower, similar to omitting `-funsafe-math-optimizations`
> 
> Pity, because `-ffinite-math-only` is _really_ annoying.

Yeah, `isnan()` is a hard habit to break, one naturally inserts it and expects it to work instead of silently doing nothing 😝

PR #10273, opened 2026-01-12
Do not use `-Ofast` in libxz

fwyzard:


fwyzard:
please test

fwyzard:
please test

what ?

fwyzard:
please test

fwyzard:
please test

mandrenguyen:
+1 
As long as it passes tests, I trust that you have a good reason for making this change.   For reference would you please explain in the PR description why it's preferable to run without this flag for this particular library?  I leave it for @cms-sw/externals-l2 to merge whenever they are happy.

fwyzard:
hold

Sorry, I forgot about it, but I didn't mean to merge it just yet.
The idea was to use the build from the bot to test the impact, but then I got sidetracked.

fwyzard:
It's only specific to this library because this is the only library for which we are currently using `-Ofast`.

mandrenguyen:
Thanks I seem to recall some discussion about this flag in the context of reproducibility, perhaps in the context of GPU workflows.

PR #10275, opened 2026-01-15
Update the xtd library

fwyzard:
Now `xtd` should include all functions used in GPU software in CMSSW, as well as most C99/C++11 math functions.


fwyzard:
enable gpu

fwyzard:
please test

fwyzard:

fwyzard:

The failures are due to cms-sw/cmssw#49795, and the 16.0.x backport (#10276) does not show any problems.

smuzaffar:
+externals

NVIDIA H100 relvals error are due to https://github.com/cms-sw/cmssw/issues/49795 and the fact that h100 relvals were run on HTCondor node with cuda version 13.0 which somehow does not go well with cuda 12.9 from our software stack.

PR #10323, opened 2026-02-05
Drop meschach from cms software stack

smuzaffar:
As [Alignment/Cocoa packages are now removed](https://github.com/cms-sw/cmssw/pull/50016) from CMSSW 16.1.X, so lets also drop `meschach ` external too


smuzaffar:
please test using full cmssw

PR #10335, opened 2026-02-10
Set env variable LIT_OPTS in llvm.spec

akritkbehera:
Add LIT_OPTS environment variable for make processes to not consume all cores on a machine


akritkbehera:
Please test

smuzaffar:
+externals

PR #10337, opened 2026-02-11
Add L1TSC82ProngJetModel

henripetrow:
This PR adds one new package for an upcoming CMSSW PR for the Phase 2 Level 1 Trigger 2-Prong Jet Tagging Model.


scram-tools.file/tools/L1TSC82ProngJetModel/l1tsc82prongjetmodel.xml
```diff
@@ -0,0 +1,6 @@
+<tool name="l1tsc82prongjetmodel" version="@TOOL_VERSION@" revision="1">
```
smuzaffar: @henripetrow , please add `.xml` extension to this file i.e. it should be `l1tsc82prongjetmodel.xml`

smuzaffar:
assign l1

@cms-sw/l1-l2 , if no objections on this addition then please start the tests here

L1TSC82ProngJetModel/spec
```diff
@@ -0,0 +1,13 @@
+### RPM external L1TSC82ProngJetModel 0.0.0
+Source: https://github.com/cms-hls4ml/%{n}/archive/refs/tags/v%{realversion}.tar.gz
```
makortel, line 2: On a quick look the repository seems to contain binary files (`.so, `.o`), I'd suggest to clean them up.

Also many files seem to contain code that is not in a unique namespace increasing the risk for ODR violations. For example
* https://github.com/cms-hls4ml/L1TSC82ProngJetModel/blob/v0.0.0/L1TSC82ProngJetModel_v0/NN/parameters.h
* https://github.com/cms-hls4ml/L1TSC82ProngJetModel/blob/v0.0.0/L1TSC82ProngJetModel_v0/NN/weights/b14.h (and other files in `weights` directory)

scram-tools.file/tools/L1TSC82ProngJetModel/l1tsc82prongjetmodel.xml
henripetrow: Added the .xml extension.

L1TSC82ProngJetModel/spec
henripetrow: Removed binary files. Looking on changing the namespace.

smuzaffar:
please test

@henripetrow , I have updated the tag to `0.0.1` where you have cleaned up the shared objects and added the namespace

smuzaffar:
please test

smuzaffar:

smuzaffar:
+externals

PR #10345, opened 2026-02-14
[gcc-tool] Change GCC version command to dump full version

smuzaffar:


smuzaffar:
please test

smuzaffar:
+externals

this change is only for using system gcc. cmssw build worked, so lets get this in and rebuild riscv64 IB

PR #10346, opened 2026-02-14
use dumpfullversion so that for system gcc we also get Major.Minor.Patch

smuzaffar:
System `gcc -dumpversion` returns only major version number 
```
> which gcc
/usr/bin/gcc
> gcc -dumpversion
11
> gcc -dumpfullversion
11.5.0
```

while GCC build by our software stack uses sources directly from github instead of tar file so our GCC returns full gcc version
```
> which gcc
/cvmfs/cms.cern.ch/el8_amd64_gcc13/external/gcc/13.4.0-6908cfdf803923e783448096ca4f0923/bin/gcc
> gcc -dumpversion
13.4.0
```

This change proposes to use `gcc -dumpfullversion` so that we properly get the full gcc version.  This should properly set `TBB_USE_GLIBCXX_VERSION` when system gcc is uses e.g. for RISCV64 build


smuzaffar:
please test

PR #10356, opened 2026-02-18
stdcxx-fs: do not explicitly link against stdc++fs for system compiler

smuzaffar:


smuzaffar:
please test

smuzaffar:
not needed, we should drop the `stdcxx-fs` tool as `stdc++fs` is now part of `stdc++` since c++17.

PR #10359, opened 2026-02-19
Include Google Benchmark and Test in the CMSSW distribution

fwyzard:
These are used to build the ACTS and Traccc tests.

### DIFF
--- a/cmssw-tools.spec
+++ b/cmssw-tools.spec
@@ -15,7 +15,6 @@ Requires: celeritas
 %endif
 Requires: crab
 Requires: cmssw-wm-tools
-Requires: google-benchmark
 Requires: catch2
 Requires: starlight
 Requires: alpgen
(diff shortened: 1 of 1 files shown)


fwyzard:
please test

cmssw-tools.spec
```diff
@@ -46,6 +46,8 @@ Requires: geant4data
 Requires: g4hepem
 Requires: glimpse
 Requires: gmake
+Requires: google-benchmark
```
smuzaffar, line 48: this is already included (may be you can cleanup the previous entry at line #18)

fwyzard:
please test

cmssw-tools.spec
fwyzard: done, thanks for the suggestion

smuzaffar:
+externals

PR #10376, opened 2026-02-25
Update pacparser to version 1.5.0

smuzaffar:
Update pacpasrser to version 1.5.0
- GCC 14 patch is not needed any more as pacparser now does not depend on spidermonkey
- Patch added for pymod as on riscv64, the `--root=/` was causing pacparser wronly install under python3 path


smuzaffar:

smuzaffar:
+externals

PR #10379, opened 2026-02-26
[prmon]Build and deploy prmon along with cmssw-tools

smuzaffar:
Submission Infrastructure (SI) requested to include [prmon](https://github.com/HSF/prmon) in our env. This is used for monitoring a running process.  It is build as a static executable to we can build it once for x86_64, aarch64 and riscv64 and then we can use it for any <os>_<arch>_<compiler>. Though it is deployed under normal `<CMS>/$SCRAM_ARCH/cms/prmon/<version>` but then at post install time it gets copied under `<CMS>/share/$(uname -m)/cms/prmon/<version>` so that it can be used by any $SCRAM_ARCH.

### DIFF
--- a/cmssw-tools.spec
+++ b/cmssw-tools.spec
@@ -1,7 +1,7 @@
-### RPM cms cmssw-tools 5.0
+### RPM cms cmssw-tools 6.0
 # With cmsBuild, change the above version only when a new tool is added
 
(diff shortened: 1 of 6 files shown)


smuzaffar:
please build

smuzaffar:
please build for el8_aarch64_gcc13

smuzaffar:
+externals
deployment look good, lets get this in

PR #10384, opened 2026-02-26
fix prmon-env.csh to avoid nasted back-ticks

smuzaffar:


smuzaffar:
please test

smuzaffar:
+externals

this deploys a fixed version of prmon-env script for csh

PR #10386, opened 2026-03-01
Enble HIP/ROCm build for Traccc in Acts

fwyzard:
Includes in the CMS branch:
  - https://github.com/acts-project/acts/pull/5178
  - https://github.com/acts-project/traccc/pull/1269

### DIFF
--- a/acts.spec
+++ b/acts.spec
@@ -4,7 +4,7 @@
 ## INCLUDE rocm-flags
 ## INCLUDE geant4-deps
 
-%define tag         cc436985e
+%define tag         1af234d8a
 %define branch      cms/%{realversion}
 %define github_user cms-externals
 Source: git+https://github.com/%{github_user}/%{n}.git?obj=%{branch}/%{tag}&export=%{n}-%{realversion}&output=/%{n}-%{realversion}-%{tag}.tgz
@@ -12,9 +12,6 @@ Source99: scram-tools.file/tools/eigen/env
 
 # Do not build the Acts and Traccc tests
 %define build_test 0
-
-# ROCm support is not yet building correctly
-%define without_rocm 1
 
 BuildRequires: cmake
 Requires: boost


fwyzard:
enable gpu

fwyzard:
please test

fwyzard:
hold

Traccc developers are suggesting a different fix

fwyzard:
unhold

Now it uses a backport of the fix proposed by Traccc developers.

fwyzard:
please test

mandrenguyen:
+1

mandrenguyen:
@cms-sw/externals-l2 feel free to merge

smuzaffar:
@fwyzard , so this increases the `acts` distribution size from 330MB to 5.4GB https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-da2840/51710/external-tools.html

Is this expected?

fwyzard:
it's not expected

fwyzard:
I'll check what else gets included.

fwyzard:
Ah... it seems to be including the _data_ for the Acts and Traccc tests.
I'll update the spec file to exclude them from the package.

After seeing that I'm wondering, do you think we could push them to some dedicated data package?
We don't use them for anything in CMSSW, but they could be useful for some standalone tests.

Do we have the option of building a data package, deploy it to CVMFS, but don't include it in CMSSW ?

fwyzard:
I've removed building the tests and downloading the data.

fwyzard:
please test

smuzaffar:
+externals

looks good.

smuzaffar:
> Do we have the option of building a data package, deploy it to CVMFS, but don't include it in CMSSW ?

@fwyzard , you can just add a new spec (e.g. acts-data.spec) which can download and deploy the needed data. In the `%post` section you can sync the contents in to shared/externals/acts-data/<realversion>` so that we have one copy of this data for archs.

fwyzard:
> you can just add a new spec (e.g. acts-data.spec) which can download and deploy the needed data. In the `%post` section you can sync the contents in to `shared/externals/acts-data/` so that we have one copy of this data for archs.

OK, I'm not sure I understand the details, but I'll give it a try.

smuzaffar:
@fwyzard , normally when we build a package then its contents are installed under `<path>/<architecture>/external/<package-name>/<realversion>-<hash>` . If we do not do any thing special for data packages than data files have multiple copies ( one for each architecture). This is fine for CVMFS but to reduce the disk space usage, specially when deployed on online HLT form, we create a shared copy of data files so multiple architectures only have one set of data files deployed on disk. This is achieved by copying the data contents of `<path>/<architecture>/external/<package-name>/<realversion>-<hash>` in to `<path>/shared/external/<package-name>/<realversion>` e.g. something like this
```
mkdir -p <path>/shared/external/<package-name>/<realversion>
for dir in $(ls <path>/shared/external/<package-name>/<realversion> | grep -v etc) ; do
  rsync -a <path>/<architecture>/external/<package-name>/<realversion>-<hash>/${dir}/  <path>/shared/external/<package-name>/<realversion>/${dir}/
  rm -rf <path>/<architecture>/external/<package-name>/<realversion>-<hash>/${dir}
  ln -s ../../../../shared/external/<package-name>/<realversion>/${dir} <path>/<architecture>/external/<package-name>/<realversion>-<hash>/${dir}
done
```

fwyzard:
Ah, now I understand, thanks.

With this approach, are the content of `.../shared/external/<package-name>/<realversion>` ever deleted ?

smuzaffar:
> With this approach, are the content of .../shared/external/<package-name>/<realversion> ever deleted ?

No, currenlty these are not deleted but we can update cmspkg to remove these if there are no ref/symlink from any of the architecture install directories to share install directory

PR #10390, opened 2026-03-02
Split torch tool for cpu, cuda and rocm

smuzaffar:
This PR proposes to build Torch separately for `cpu, cuda and rocm`. For now torch-rocm is not enabled as it requires full distribution of rocm (nearly 11GB)[a]. We should be able to build most/all of [a] from sources using https://github.com/ROCm/rocm-libraries . 

For now this PR proposes to split torch in cpu and cuda part as a proof of concept that we can dynamically select env depending on available resources. Currently this PR proposes the following
- `py3-torch`: Only cpu c++ API and python interface
- `py3-torch-cuda`: cuda and cpu  c++ API and python interface
- `py3-torch-<extensions>` (e.g. `cluster`, `scatter` and `sparse`): cpu only python interface
- `py3-torch-<extensions>-cuda` (e.g. `cluster`, `scatter` and `sparse`): cuda and cpu python interface

As we do not have `py3-torch-rocm` yet ( which needs full 11GB of rocm distribution), on amdgpu hosts we fall back to cpu only interface. SCRAM will dynamically set the env e.g. on
- cpu only host:
  - LD_LIBRARY_PATH: `<path>/lib`
  - PYTHON3PATH:`<path>/lib/python3.12/site-packages`
- host with cuda:
  - LD_LIBRARY_PATH: `<path>/lib/scram_cuda:<path>/lib`(where <path>/lib/scram_cuda contains libs for cuda and cpu torch)
  - PYTHON3PATH: `<path>/lib/python3.12/site-packages/scram_cuda:<path>/lib/python3.12/site-packages`

for rocm, for now we fall back to cpu only libs/python modules.

I also have added triton and aotriton spec so that if we have to build the torch-rocm fully from sources then we can use aotirton. For now torch-rocm downloads prebuild aotriton bundle at build time

This need https://github.com/cms-sw/cmssw-config/pull/117 too

[a]
```
 hipblas-common-devel
  miopen-hip miopen-hip-devel
  hipfft hipfft-devel
  hipsparse hipsparse-devel
  rocprim-devel hipcub-devel
  rocthrust-devel
  hipsolver hipsolver-devel
  rocsolver  rocsolver-devel
  hipblaslt hipblaslt-devel
  hipsparselt hipsparselt-devel
  roctrace roctrac-devel
  rocblas rocblas-devel
```


smuzaffar:
enable gpu

smuzaffar:
please test with cms-sw/cmssw-config#117 using full cmssw

smuzaffar:
please test

smuzaffar:
please test with https://github.com/cms-sw/cmssw-config/pull/117 using full cmssw

smuzaffar:
please test with https://github.com/cms-sw/cmssw-config/pull/117 using full cmssw

smuzaffar:
@valsdav @makortel , this PR splits the torch tools in separate packages
py3-torch: which has cpu implementation only
py3-torch-cuda: which has cuda and cpu implementation only

Once we have torch-rocm working then we can add `py3-torch-rocm` too.  All unit tests (full cmssw) for NVIDA, ROCM and CPU worked fine. Note that this is for PYTHON 3.12 IBs and GCC15 IBs.

Let me know if you have any concern about this approach . I followed MULTI arch/target way of setting env dynamically so we do not need a big change in SCRAM or build rules.

fwyzard:
With this approach, is there a way to request explicitly one backend, _e.g._ the CPU one, in parallel to the default one, that could be CPU, CUDA or ROCm ?

This feature has been very useful for the GPU-vs-CPU validation of alpaka-based modules.

smuzaffar:
I do not know how touch works internally. This approach is same as our existing torch distribution ( which has cuda and cpu libs). Now with this change we can have one of the following env
- cpu only: on hosts with neither nvidia nor amd gpus
- cuda + cpu: on hosts with nvidia cpus. On host with both nvidia and amd gous, cuda is preferred
- rocm + cpu: on host with only amd gpus

https://github.com/cms-sw/cmssw/blob/master/PhysicsTools/PyTorch/test/torch-cpu-cuda.py unit test uses both cpu and cuda backends and It worked fine on host with nvidia [a]. So I think one can explicitly request the backend. This unit test works using torch python, I am sure there must be something for C++ API too. 


[a] https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-b5dec0/51775/nvidia_h100UnitTests/src/PhysicsTools/PyTorch/test/test-torch-cpu-cuda/testing.log
```
===== Test "test-torch-cpu-cuda" ====
Testing on cpu
Time: 9.32814645767212 seconds
Testing on cuda
Time: 0.17551016807556152 seconds

---> test test-torch-cpu-cuda succeeded
TestTime:20
^^^^ End Test test-torch-cpu-cuda ^^^^
```

fwyzard:
Great, thanks.

Then I think the external is fine, and now we have to figure out the glue in CMSSW.

smuzaffar:

smuzaffar:
please test with https://github.com/cms-sw/cmssw-config/pull/117 using full cmssw for el9_amd64_gcc15

smuzaffar:
please test with https://github.com/cms-sw/cmssw-config/pull/117 using full cmssw

smuzaffar:
please test with cms-sw/cms-bot#2702 using full cmssw

smuzaffar:
+externals 

good to go in Python 3.12 based IBs ( i.e. PY312 and GCC15 IBs)

PR #10408, opened 2026-03-10
Added support for gpu backend specific packages

smuzaffar:


smuzaffar:
please test with https://github.com/cms-sw/cms-bot/pull/2702

smuzaffar:
+externals

PR externals were build without errors. The change here is protected by `gpu_backend_specific_packages` so it should not break externals which do not define `gpu_backend_specific_packages`

PR #10409, opened 2026-03-11
set env for various build system so that each package respect the build jobs count

smuzaffar:
many pypi packages run `nproc` jobs to build. This is fine on small machines with 8, 16 core but on machine with hundreds of core this break build as we start to hit various limitation e.g. number of files open, number of parallel process a user can run. This PR proposes to set various environment various which build system like cmake, meson , ninja can use. In most cases e do pass `-j N` to build system but many pypi packages have different way of controlling the parallel jobs. This change will atleast control those pypi packages which use cmake or ninja to build


smuzaffar:
please test

smuzaffar:
+externals

PR #10416, opened 2026-03-14
Add missing dependency on compression libraries

fwyzard:
The missing dependency can be noted when building a package that depends on hepmc3 on a system with an older version of zstd.

### DIFF
--- a/scram-tools.file/tools/hepmc3/hepmc3.xml
+++ b/scram-tools.file/tools/hepmc3/hepmc3.xml
@@ -1,6 +1,11 @@
 <tool name="@TOOL_FILENAME@" version="@TOOL_VERSION@" revision="2">
   <lib name="HepMC3"/>
   <lib name="HepMC3search"/>
+  <!-- header-only dependencies through bxzstr -->
+  <use name="bz2lib"/>
+  <use name="xz"/>
+  <use name="zlib"/>
+  <use name="zstd"/>
   <client>
     <environment name="@TOOL_BASE@" default="@TOOL_ROOT@"/>
     <environment name="LIBDIR" default="$@TOOL_BASE@/lib64"/>


fwyzard:
please test

fwyzard:
urgent

@smuzaffar @mandrenguyen @ftenchini can we get this in pre3 ?

fwyzard:
type bugfix

ftenchini:
+1

ftenchini:
Sorry, I managed to notice it at the last moment but I aborted the build in time.

smuzaffar:
@fwyzard , I do not think this is going to fix the issue. Looks like hepmc3 only include the header files of compresion tools e.g. https://gitlab.cern.ch/hepmc/HepMC3/-/blob/master/include/HepMC3/bxzstr/compression_types.hpp?ref_type=heads#L14-17
```
#include "bz_stream_wrapper.hpp"
#include "lzma_stream_wrapper.hpp"
#include "z_stream_wrapper.hpp"
#include "zstd_stream_wrapper.hpp"
```

I would sugest to just update the scram-tools.file/tools/hepmc3/hepmc3.xml file and add 
```
<use name="zstd"/>
<use name="bz2lib"/>
<use name="zlib"/>
<use name="xz"/>
```

fwyzard:
without adding the dependency to the rpm ?

smuzaffar:
yes, for now there is no need to add the dependency in the spec itself. dependency in spec is only needed if we enable the tests but for now configuration of test is broken. 

For a quick fix I suggest to only fix the tool file and after pre3 we can patch hepmc3 to fix the tests configuration for comparesion tools

fwyzard:
ok

fwyzard:
please test

fwyzard:
@smuzaffar actually this does not seem to help, at least for externals that use hepmc3, its dependency on the compression libraries, and need to discover those via CMake.

fwyzard:
(but neither did my previous approach)

fwyzard:
For the moment the workaround I used is to add those dependency explicitly, even if they come through hepmc3.

smuzaffar:
> @smuzaffar actually this does not seem to help, at least for externals that use hepmc3, its dependency on the compression libraries, and need to discover those via CMake.

correct, the toolfile fix only fixes the cmssw build where packages which use hepm3 will add path of our zlib, zstd, lzma etc.

For externals which use hepmc3, they still need to properly depend on these compresion tools and use the correct cmake parameters e.g. `CMAKE_PREFIX_PATH` to find and use these externals

smuzaffar:
+externals

lets get this in, we still need to fix the externals (hepmc3 and its dependent packages) to use our compression tools properly

PR #10431, opened 2026-03-19
improved data package install/uninstall and size check macros

smuzaffar:
This PR updates the rpm macros for managing shared data package. It also contains a helper spec file template which one can use to build data packages which are then installed in share/package path e.g.  

- Following data package unpacks its contents during the rpm build phase and then at the rpm install time its contents are move to share/package directory and symlinks are created in the actual arch/package directory 
```
### RPM external testdata v1.0
Source: https://foo.bar/foo-%{realversion}.tar.gz
## INCLUDE data/shared-data-package
```

- Following data package unpacks its contents during rpm install time in to share/package path and then create symlinks for arch/package
```
### RPM external testdata-tar v1.0
Source: https://foo.bar/foo-%{realversion}.tar.gz
%define unpack_at_install 1
## INCLUDE data/shared-data-package
```

a ref for each installation is stored in share/package/.links path. At n-install time this ref is deleted and once there is no more ref left then actual share/package is deleted too.

For packages which unpacks its content during the RPM install time, we also calculate the required disk size for installation (by unpacking the tar file and calculating its size in MB) and feed this information to `rpmbuild`  so that package's pre-install can make sure that there is enough disk available before starting the installation.

### DIFF
--- a/data/shared-data-package.file
+++ b/data/shared-data-package.file
@@ -1,0 +1,43 @@
+## NOCOMPILER
+AutoReqProv: no
+
+%pre
+%{?package_pre_pre}
+%if 0%{?unpack_at_install}
+%check_data_package_size
+%endif
+%{?package_post_pre}
+
+%prep
+%{?package_pre_prep}
+%{?package_post_prep}
+
+%build
+%{?package_pre_build}
+%{?package_post_build}
+
+%install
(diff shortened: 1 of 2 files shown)


smuzaffar:
please test

smuzaffar:
+externals

PR #10432, opened 2026-03-19
Added cms-ninja which runs ninja under make jobnserver to avoid recursive ninja running NxN jobs

smuzaffar:
During our build monitoring we noticed that few packages were consuming (on 16 core node) 1600% of CPU even when they are build with `-j 8`. It turned out that some of the packages e.g. `llvm`, `root` were calling `ninja`internally for its sub-projects. As `ninja` does not have its on job-server, so during the build we were runing `N+N` jobs.

Also some of the python packages e.g. `py3-contourpy`, `py3-iminuit`, `py3-numpy`, `py3-pandas`, `py3-scipy` were calling `ninja` without `-j N` this causes `ninja` to use all available cpus. On a host with a lot of memory and 8 to 16 core this was not an issue but building CMSSW software stack on host with more than 760 cpus causes the host to go in to bad state.

This PR proposes to add new `cms-ninja` wrapper script which runs the actual ninja under gmake jobserver. This allows recursive ninja to still use `N` cores. After this change building full software stach with `-j 8` none of the externals exceeded `800%` cpu. As `ninja` only support `fifo` bbase jobserver which is only available in gamke 4.4 so that is why this PR also adds gamke 4.4 (just to be used by `cms-ninja` only)

Only other external which still uses more cpu than asked for is Tensorflow. Tensorflow which uses bazel as build system still for a short period of time (30 seconds out of 6000sec build time) uses 900%-1600% cpu.


smuzaffar:
please test

smuzaffar:
+externals

PR #10457, opened 2026-04-01
[ROOT639] Disable stripping of ROOT's `libCling.so`

hahnjo:
https://github.com/root-project/root/commit/4731fd56e2f682b318f3b620d876a066f1e1ec12 automatically strips `libCling.so` in Release builds. To turn it off, define `CMAKE_STRIP=/usr/bin/true`, which is no-op.

### DIFF
--- a/root.spec
+++ b/root.spec
@@ -56,6 +56,7 @@ cmake ../%{n}-%{realversion} \
   -DCMAKE_CXX_COMPILER=g++ \
   -DCMAKE_Fortran_COMPILER=gfortran \
   -DCMAKE_LINKER=ld \
+  -DCMAKE_STRIP=/usr/bin/true \
   -DCMAKE_VERBOSE_MAKEFILE=TRUE \
   -Droot7=ON \
   -Dfail-on-missing=ON \


hahnjo:
@makortel it turns out there is an "easy" way to turn off stripping of `libCling.so` about which we discussed recently: defining `CMAKE_STRIP=/usr/bin/true` makes the step a no-op.

@smuzaffar if that sounds reasonable, it would be good to confirm in the PR build that `libCling.so` is indeed not stripped and retains the symbol tables.

fwyzard:
I would make _a lot_ more sense to make the "strip libCling.so" optional with its own cmake flag.

hahnjo:
> I would make _a lot_ more sense to make the "strip libCling.so" optional with its own cmake flag.

Why? There already is a *standard* CMake flag which allows to turn off all stripping.

smuzaffar:
please test

fwyzard:
> There already is a _standard_ CMake flag which allows to turn off all stripping.

As far as I can find, CMake does not have any flag to say "strip the symbols" or "don't strip the symbols", that's something the application build rules decide to do.

It does expose an internal variable to let you customise what command to run to strip symbols.

That's definitely not the same as having a "standard CMake flag" to _disable_ stripping the symbols.

Since it's an internal flag that may be used across different sub-projects, what is the impact of setting it to "do nothing" ?

smuzaffar:
@hahnjo , I agree with @fwyzard. Just overriding `CMAKE_STRIP` can have side effects e.g I see that it is used in various places in root https://github.com/search?q=repo%3Aroot-project%2Froot%20CMAKE_STRIP&type=code  (passed to llvm-project cmake configuration)

May be having `-DDISABLE_LIBCLING_STRIP=ON` or something like this is better.

By the way, I confirm that without stripping (i.e. this change)  `libCling.so` size is `120M` whereas with striping the size was `103M`

PR #10474, opened 2026-04-07
[15.1.X] Update L1TSC4NGJetModel.spec to v1.0.1

Chriisbrown:
This PR updates the Phase-2 L1 trigger Jet Tagger model to v1.01 to use fully emulated inputs

Release: https://github.com/cms-hls4ml/L1TSC4NGJetModel/tree/main/L1TSC4NGJetModel_v1_0_1

Model tested in emulation and FW

Corresponding CMSSW PR: https://github.com/cms-sw/cmssw/pull/50174

### DIFF
--- a/L1TSC4NGJetModel.spec
+++ b/L1TSC4NGJetModel.spec
@@ -1,4 +1,4 @@
-### RPM external L1TSC4NGJetModel 0.0.0
+### RPM external L1TSC4NGJetModel 1.0.1
 Source: https://github.com/cms-hls4ml/%{n}/archive/refs/tags/v%{realversion}.tar.gz
 Requires: hls4mlEmulatorExtras hls
 BuildRequires: gmake


quinnanm:
please test

quinnanm:
hi @Chriisbrown looks like you need to make the v1.0.1 release here: https://github.com/cms-hls4ml/L1TSC4NGJetModel/releases

Chriisbrown:
@quinnanm thanks for the spot, v1.0.1 release now available: https://github.com/cms-hls4ml/L1TSC4NGJetModel/releases

quinnanm:
please test

smuzaffar:
+externals

smuzaffar:
oops, I did not realize that this PR was for 15.1.X

@Chriisbrown , I accidently have merged this. This change should have gone to 17.0.X (development IB) first. I have opened https://github.com/cms-sw/cmsdist/pull/10485   for 17.0.X . I suggest that please also open backport prs for 16.1.X and 16.0.X

Chriisbrown:
Unsure if we want a backport into 16.X as this model is only needed with changes coming into CMSSW in this PR: https://github.com/cms-sw/cmssw/pull/50174 which is targeting CMSSW_17_0_X

smuzaffar:
So, should I revert this change for 15.1.X then ?

Chriisbrown:
The current [15_1_X ](https://github.com/cms-sw/cmssw/blob/1a653f48c678a2d6cbfc75a4df2532211ddf0349/L1Trigger/Phase2L1ParticleFlow/plugins/L1TSC4NGJetProducer.cc#L117)CMSSW points by default to the v0.0.0 tag of this model, so if the v1.0.1 tag is also available but unused it doesn't make a difference I think. So long as 15_1_X points to v0.0.0 by default and 17_0_X points to v1.0.1 by default the user will get the correct model. However, if we want to avoid any confusion where a user could inadvertently point to the v1.0.1 model in 15_1_X we should revert the change.

smuzaffar:
For now, I have reverted it for 15.1.X  . We can open a backport PR if it is needed in old release cycles

PR #10132, opened 2025-10-17
[G4Adept] Only link vecgeom_final library

smuzaffar:

PR #10476, opened 2026-04-07
[G4ADEPT] update AdePT specs to v0.3.3

SeverinDiederichs:
This PR updates AdePT to v0.3.0 including the following changes:

- the interfacing AdePT libraries were renamed to `adept_g4integration` and `adept_g4integration_final`
- `covfie` and `g4vg` are declared as a dependency, to avoid an additional build of them within AdePT (for `g4vg` this requires an additional CMake flag, for `covfie` it does not, as AdePT only builds it if it was not provided).
- removing of the deprecated cmake flag `ADEPT_ASYNC_MODE`, which is now the only available transport backend in AdePT
- removing of the CMake flag for `-DBUILD_SHARED_LIBS=OFF`, as it has no effect. AdePT is build as a static library if Geant4 is provided as a static library and as a shared library, if Geant4 is provided as a shared library. In case both a static and a shared Geant4 is provided (as it is the case in CMS), AdePT is built as a static library.


adept.spec
```diff
@@ -7,7 +7,7 @@ Source99: scram-tools.file/tools/adept/env
 
 ## INCLUDE geant4-deps
 ## INCLUDE cuda-flags
-Requires: geant4 g4hepem
+Requires: geant4 g4hepem g4vg covfie
```
fwyzard: I think this won't work - we are building covfie as part of acts, not as a standalone package.

In order to share it with adept I guess we need to make covfie a separate package, and make both acts and adept depend on it ?

fwyzard:
please test

adept.spec
SeverinDiederichs: Ah, I was not aware of that. I agree, in that case a separate package would be needed - otherwise we can also continue to use the fetched covfie within AdePT, whatever you prefer

fwyzard: How urgently would you want this update ?

makortel:
> * In case both a static and a shared Geant4 is provided (as it is the case in CMS), AdePT is built as a static library.

We ending up using a static library sounds potentially problematic (for longer-term future).

adept.spec
SeverinDiederichs: I think it would be good if it gets into the next IB, so I can check a couple of things with AdePT within CMSSW. 
I can also just delete the requirement of covfie and we fix that properly once covfie is available as a separate package.

SeverinDiederichs:
> > * In case both a static and a shared Geant4 is provided (as it is the case in CMS), AdePT is built as a static library.
> 
> We ending up using a static library sounds potentially problematic (for longer-term future).

I was wondering about this:

Currently CMS builds Geant4 both as a static and as a shared library but for examples forces G4HepEm to be build only as a static library. From the removed flag seemed to me that it would also enforce to build AdePT as a static library.

So what is the longer term strategy?

SeverinDiederichs:
I updated the PR to remove the explicit covfie usage, updated to the latest version of AdePT, and changed the target to `CMSSW_17_0_X/g4adept`.

It would be fantastic if this PR could make it into the upcoming IB

smuzaffar:
please test

SeverinDiederichs:
Apparently the G4VG version within cmssw is not compatible with the VecGeom version used within AdePT, so I turned back to using the built-in G4VG

smuzaffar:
> > > * In case both a static and a shared Geant4 is provided (as it is the case in CMS), AdePT is built as a static library.
> > 
> > 
> > We ending up using a static library sounds potentially problematic (for longer-term future).
> 
> I was wondering about this:
> 
> Currently CMS builds Geant4 both as a static and as a shared library but for examples forces G4HepEm to be build only as a static library. From the removed flag seemed to me that it would also enforce to build AdePT as a static library.
> 
> So what is the longer term strategy?

@SeverinDiederichs , in cmssw, along with normal small shared libs and plugins,  we also have a simulation plugin which links against static geant4 ( and static build of any package which depend on geant4 e.g. g4hepem, dd4hep, celeritas, g4vg, adept). The single big simiulation plugin gave us 3-5% improved throughput and linking static geant4 gave additional 3-5% gain. To keep this kind of build working we either build geant4 dependent packages in either in static only mode or build both static and shared version ( e.g. dd4hep) and link static version in big simulation library and link shared version in normal small cmssw edm plugs and shared libs.

smuzaffar:
> Apparently the G4VG version within cmssw is not compatible with the VecGeom version used within AdePT, so I turned back to using the built-in G4VG

We have g4vg version 1.0.6 (built staticlly) in cmssw externals while Adept is using version 1.0.3. Is it just the g4vg version or the missing shared libs of g4vg?

SeverinDiederichs:
> > Apparently the G4VG version within cmssw is not compatible with the VecGeom version used within AdePT, so I turned back to using the built-in G4VG
> 
> We have g4vg version 1.0.6 (built staticlly) in cmssw externals while Adept is using version 1.0.3. Is it just the g4vg version or the missing shared libs of g4vg?

You are right, I checked this again and the G4VG version itself is not the issue.

The undefined symbols are VecGeom BooleanVolume symbols. They are present in the regular `libvecgeom.a`, but I could not find them in `libvecgeom_final.a`. From the SCRAM toolfile, the current `vecgeom` tool seems to export only `vecgeom_final`, so when AdePT links against the external static `libg4vg.a`, the final plugin link misses those host-side VecGeom symbols.

This may also explain why Celeritas can still be fine: its external build goes through CMake, where the `G4VG::g4vg -> VecGeom::vecgeom` dependency is preserved. I am not completely sure whether the same final SCRAM link path is exercised there.

For now I switched AdePT back to the bundled G4VG as this worked before.

If you prefer, I can instead use the external G4VG setup work in this PR by exposing `libvecgeom.a` also where `libg4vg.a` is linked.

smuzaffar:
@SeverinDiederichs , in https://github.com/cms-sw/cmsdist/pull/10132 I had to drop the `libvecgeom` from the scram tool file as there were some dupilcate symbols. Feel free to add it back to vecgeom xml if it is needed

SeverinDiederichs:
> @SeverinDiederichs , in #10132 I had to drop the `libvecgeom` from the scram tool file as there were some dupilcate symbols. Feel free to add it back to vecgeom xml if it is needed

Thanks for the explanation! I think the regular `libvecgeom` is needed here for the host-side symbols used by the external static G4VG.

I'll try adding it back to the `vecgeom` SCRAM tool. Where did the duplication appear?

SeverinDiederichs:
I updated the AdePT version to the latest. It would be great if this could be tested again @smuzaffar

akritkbehera:
please test

smuzaffar:
please test

SeverinDiederichs:
@smuzaffar I added the libvecgeom again, I hope that this finally solves the issue, otherwise I am a bit surprised..

smuzaffar:
please test

smuzaffar:
please test using full cmssw

smuzaffar:
@SeverinDiederichs , there are link errors like
```
AdePTTrackingManager.cc:(.text+0x1aa3): undefined reference to `adept::transport::detail::GetGPUStepBatchFromBuffer(unsigned int, unsigned int, adept::transport::GPUstate&, bool&)'
```
this symbol is provided by `adept_transport` library. So I think you need to add `adept_transport` to the adept.xml tool file. I tried building it locally by adding `adept_transport` to the adept tool but then link failed with message
```
/cvmfs/cms-ib.cern.ch/sw/x86_64/nweek-02941/el8_amd64_gcc13/external/gcc/13.4.0-6908cfdf803923e783448096ca4f0923/bin/../lib/gcc/x86_64-redhat-linux-gnu/13.4.0/../../../../x86_64-redhat-linux-gnu/bin/ld.bfd: /build/muz/x/CMSSW_17_0_G4ADEPT_X_2026-05-10-2300/external/el8_amd64_gcc13/lib/libadept_transport.a(AdePTTransport.cu.o): relocation R_X86_64_32 against `.rodata.str1.8' can not be used when making a shared object; recompile with -fPIC
```

I guess you also need to add `-fPIC` in 
```
  -DCMAKE_STATIC_LIBRARY_CXX_FLAGS="%{build_flags}" \
  -DCMAKE_STATIC_LIBRARY_C_FLAGS="%{build_flags}" \
```
 I would also suggest that try to use `-DADEPT_USE_BUILTIN_G4VG=FALSE` i.e. use the g4vg from external

SeverinDiederichs:
Ok, @smuzaffar I followed your suggestions, I hope this fixes the issue. Thank you for your help!

smuzaffar:
please test

smuzaffar:
it still fails with link error below. I will look in to it once the build area is available on cvmfs
```
AdePTGeometryBridge.cpp:(.text+0x4f28): undefined reference to `vgdml::Parser::Load(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, double, bool)'
/data/cmsbld/jenkins/workspace/ib-run-pr-tests/testBuildDir/el8_amd64_gcc13/external/gcc/13.4.0-6908cfdf803923e783448096ca4f0923/bin/../lib/gcc/x86_64-redhat-linux-gnu/13.4.0/../../../../x86_64-redhat-linux-gnu/bin/ld.bfd: /data/cmsbld/jenkins/workspace/ib-run-pr-tests/CMSSW_17_0_G4ADEPT_X_2026-05-10-2300/external/el8_amd64_gcc13/lib/libadept_g4integration_final.a(cmake_device_link.o): in function `__cudaRegisterLinkedBinary_e3457b9c_25_G4IntegrationRdcFacade_cu_fb072f85_2982988':
link.stub:(.text+0x42): undefined reference to `__fatbinwrap_e3457b9c_25_G4IntegrationRdcFacade_cu_fb072f85_2982988'
```

SeverinDiederichs:
> it still fails with link error below. I will look in to it once the build area is available on cvmfs
> 
> ```
> AdePTGeometryBridge.cpp:(.text+0x4f28): undefined reference to `vgdml::Parser::Load(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, double, bool)'
> /data/cmsbld/jenkins/workspace/ib-run-pr-tests/testBuildDir/el8_amd64_gcc13/external/gcc/13.4.0-6908cfdf803923e783448096ca4f0923/bin/../lib/gcc/x86_64-redhat-linux-gnu/13.4.0/../../../../x86_64-redhat-linux-gnu/bin/ld.bfd: /data/cmsbld/jenkins/workspace/ib-run-pr-tests/CMSSW_17_0_G4ADEPT_X_2026-05-10-2300/external/el8_amd64_gcc13/lib/libadept_g4integration_final.a(cmake_device_link.o): in function `__cudaRegisterLinkedBinary_e3457b9c_25_G4IntegrationRdcFacade_cu_fb072f85_2982988':
> link.stub:(.text+0x42): undefined reference to `__fatbinwrap_e3457b9c_25_G4IntegrationRdcFacade_cu_fb072f85_2982988'
> ```

Thanks. I think for vgdml I understand it:

The `vgdml::Parser::Load(...)` symbol should come from VecGeom's GDML library, `libvgdml.a`. Since VecGeom is built with GDML enabled, we probably need to export `vgdml` from the VecGeom SCRAM tool as well.

For the cuda linking issue, maybe the order is the problem?

The current setup put `adept_transport` between `adept_g4integration` and `adept_g4integration_final`, maybe this breaks the cuda rdc? So maybe it should be:

```
  adept_g4integration
  adept_g4integration_final
  adept_transport
  adept_transport_final
```

If you confirm, I would push that

smuzaffar:
@SeverinDiederichs , you are right.
- adding `vgdml` in vecgeom.xml fixed the first link error 
- For cuda symbol changing the order of adept libs fixed the linker.  I tested 
```
  adept_g4integration_final 
  adept_g4integration
  adept_transport 
  adept_transport_final
```
and it fixed the link error.

SeverinDiederichs:
Great, thank you for the tests @smuzaffar, I updated the PR now, so it should finally work!

smuzaffar:
please test using full cmssw

smuzaffar:
+externals

PR #10479, opened 2026-04-09
Adding Pytorch Geometric to CMSSW

avk-ds:
Pytorch geometric needs now an additional library to work  pyglib. This PR includes this additional library in cmsdist to allow the evaluation of pytorch geometric model in the PyTorch inference in CMSSW.

### DIFF
--- a/pyg-lib.spec
+++ b/pyg-lib.spec
@@ -41,5 +41,6 @@ make %{makeprocesses} VERBOSE=1
 cd ../build
 mkdir -p %{i}/lib
 cp libpyg.so %{i}/lib/
-mkdir -p %{i}/include/pyg_lib
-cp -r ../%{n}-%{realversion}/pyg_lib/csrc %{i}/include/pyg_lib/
+cp libmetis/libmetis.so %{i}/lib/
+mkdir -p %{i}/include
+cp -r ../%{n}-%{realversion}/pyg_lib/csrc %{i}/include/pyg_lib
--- a/pyg-lib.spec
+++ b/pyg-lib.spec
@@ -39,4 +39,7 @@ cmake ../%{n}-%{realversion} \
 make %{makeprocesses} VERBOSE=1
 %install
 cd ../build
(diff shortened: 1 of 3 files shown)


avk-ds:
Hi 
@akritkbehera, @cmsbuild, @iarspider, @raoatifshad, @smuzaffar @ftenchini, @mandrenguyen, @sextonkennedy I don't think a notification was sent after the latest changes. Please have a look.

valsdav:
please test

valsdav:
Hi @smuzaffar, I see that the structure of the packages to build pytorch-cluster, sparse etc has changed. 
We need now to include this additional package to make pytorch geometric work. I think we are missing some files here but I'm not sure where we need to include this.  may we ask your help please?

smuzaffar:
@valsdav @avk-ds , I opened https://github.com/cms-sw/cmsdist/pull/10498 which now properly follows how we build torch extension in cmssw. basically we now have separate packages for torch-cpu, torch-cuda and torch-rocm (not yet in cmssw) so all the torch extension packages should also have separate package for cpu, cuda and rocm.

By the way, the torch extensions `pytorch-cluster, pytorch-scatter, pytorch-sparse` in cmssw only have python interface. If their c++ interface is needed (e.g. header files and library) then please let me know.

This PR can be closed in favor of #10498

valsdav:
Thanks a lot @smuzaffar !

I see the new structure makes sense. 

For the inference in CMSSW we will need to load these operators as a C++ library, so we will need the C++ part as well.   As fas as I understand at the moment only the pyglib was missing and the other libraries were found by Avik while loading models in CMSSW. But he needed to manually setup the LD_LIBRARY_PATH to find the lib.   Is this related to the fact that only the pytorch interface is now included?   Sorry for my confusion here

smuzaffar:
closing this as this external has been integrated in cmssw 17.0.X via https://github.com/cms-sw/cmsdist/pull/10498

PR #10493, opened 2026-04-18
Update GPU architectures

fwyzard:
#### NVIDIA CUDA
  - drop support for Pascal (sm 6.0);
  - add support for Blackwell (sm 10.0 and 12.0).

#### AMD ROCm
  - drop support for Instinct MI100 (gfx908) and Radeon Pro W6800 (gfx1030).

### DIFF
--- a/cuda-flags.file
+++ b/cuda-flags.file
@@ -2,8 +2,8 @@
 ## INCLUDE cpp-standard
 # define the CUDA compilation flags in a way that can be shared by SCRAM-based and regular tools
 
-# build support for Pascal (6.x), Volta (7.0), Turing (7.5), Ampere (8.x), Lovelace (8.9) and Hopper (9.0)
-%define cuda_arch 60 70 75 80 89 90
+# build support for Volta (7.0), Turing (7.5), Ampere (8.x), Lovelace (8.9), Hopper (9.0) and Blackwell (10.0, 12.0)
+%define cuda_arch 70 75 80 89 90 100 120
 
 # LIBOMPTARGET_NVPTX_COMPUTE_CAPABILITIES style for listing the supported CUDA compute architectures
 %define omptarget_cuda_archs %(echo $(for ARCH in %cuda_arch; do echo "$ARCH"; done) | sed -e"s/ /,/g")
--- a/rocm-flags.file
+++ b/rocm-flags.file
@@ -1,8 +1,8 @@
 ### FILE rocm-flags
 # define the ROCm compilation flags in a way that can be shared by SCRAM-based and regular tools
 
-# build support for Instinct MI100 (gfx908), Instinct MI210/MI250X (gfx90a), Instinct MI300X (gfx942), Radeon Pro W6800 (gfx1030), Radeon Pro W7800/W7900 (gfx1100), and Radeon Pro W7600 (gfx1102)
-%define rocm_archs gfx908:sramecc+ gfx90a:sramecc+ gfx942:sramecc+ gfx1030 gfx1100 gfx1102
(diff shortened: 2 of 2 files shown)


fwyzard:
enable gpu

fwyzard:
please test

fwyzard:
assign heterogeneous

fwyzard:
The error
```
WARNING: Download from https://github.com/bazelbuild/rules_java/releases/download/5.3.5/rules_java-5.3.5.tar.gz failed: class java.io.IOException GET returned 504 Gateway Time-out
```
looks like a transient issue, and pretty much unrelated to these changes.

fwyzard:
please test

fwyzard:
The [`RunInfo` failures](https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-bd38d3/52738/unitTests/failed.html) should be unrelated.

fwyzard:
please test

makortel:
I know we need to do this eventually, but at this moment is there something specific that motivates dropping Volta? I see some number of 7.0 GPUs in the global pool at the moment (number-wise actually less than 6.X). I suppose we should run this through O&C management (I can do that on Wednesday). Given the small number of resources compared to >= 7.5 I would not expect objections.

fwyzard:
Nothing urgent: we said we would do it after the end of Run 3 data taking, and the last data taking release should be 16.1.x.

The NVIDIA part could be further postponed.

makortel:
No clear objections were raised in the O&C meeting, but a question on "why V100 needs to be dropped now" was raised. Apparently there are V100 resources that "would be nice" to be kept supported as long as feasible.

One option would be to drop Pascal now, and Volta later when that is really necessary.

fwyzard:
I don't mind keeping Volta.

I'm confused because a couple of months ago I was told that it was OK for O&C to drop Pascal and Volta. I suggested keeping them until the end of Run 3, and here we are 🤷🏻‍♂️ 

If we keep Volta, I would keep also Pascal for the time being. We will need to drop both relatively soon anyway, once we move to CUDA 13.x.

fwyzard:
please test

makortel:
The last time I asked about this question within O&C was in November 26, but that was about Pascal only (whose dropping was ok).

fwyzard:
Remove again Pascal, see also #10516 .

fwyzard:
please test

cuda-flags.file
```diff
@@ -2,8 +2,8 @@
 ## INCLUDE cpp-standard
 # define the CUDA compilation flags in a way that can be shared by SCRAM-based and regular tools
 
-# build support for Pascal (6.x), Volta (7.0), Turing (7.5), Ampere (8.x), Lovelace (8.9) and Hopper (9.0)
-%define cuda_arch 60 70 75 80 89 90
+# build support for Volta (7.0), Turing (7.5), Ampere (8.x), Lovelace (8.9), Hopper (9.0) and Blackwell (10.0, 12.0)
+%define cuda_arch 70 75 80 89 90 100 120
```
smuzaffar, line 6: @fwyzard , can we get this change in a separate PR? The rocm 7.2.2 seems to break a lot of GPU relvals so I am not sure if `rocm 7.2.2` can goin soon. Having a separate PR for cuda change will help integrating onnxruntime

fwyzard: Sure, I'll make a separate PR.

fwyzard:
enable gpu

fwyzard:
please test

fwyzard:
+heterogeneous

smuzaffar:
+externals

PR #10504, opened 2026-04-24
Update tag for L1TriggerScouting-OnlineProcessing to V00-01-00 [`16_1_X`]

missirol:
backport of #10503

This integrates the update in https://github.com/cms-data/L1TriggerScouting-OnlineProcessing/pull/1 into 16_1_X.

It goes along with the update in https://github.com/cms-sw/cmssw/pull/50791 (though the latter does not strictly depend on this PR, because the cms-data file in question is not tested centrally at the moment).


mandrenguyen:
merge

PR #10555, opened 2026-05-13
Remove obsolte python directory.

smuzaffar:
`Dependabot` now also reads `setup.py` files and generate security alerts . We have this very old (over 9 year) `python` directory for build pip packages but we do not use it any more. Due to this `Dependabot` has  generated over 500 security alerts. This PR proposes to drop this unused directory


smuzaffar:
please test

smuzaffar:
+externals

externals build without any errors so this `python` directory is not used in any way. Merging this shoudl resolve the over 500 `Dependabot` reported issues

PR #10562, opened 2026-05-18
[DEVEL] Update xrootd version to latest release 6.0.1

akritkbehera:


akritkbehera:
please test for CMSSW_17_1_DEVEL_X

akritkbehera:
please test for CMSSW_17_0_DEVEL_X

akritkbehera:
please test for CMSSW_17_0_DEVEL_X

smuzaffar:
+externals

Lets test yhis in Devel ibs

nsmith-:
How far back in cmssw versions do we plan to backport xrootd updates? I'm particularly keen on xrootd 6.0 because it enables a much better http access interface, that should be usable transparently by downstream code (ROOT, Pool source)

smuzaffar:
@nsmith- , currently it is only in 20.1.X (development release cycle). If needed this can be backported to 20.0.X and 17.0.X release cycles.

As there are no cmssw changes to go with it , so unless xrootd 6.x does not support old compilers/OS, we can also back port it to CMSSW_10_6_X and above where we already have XROOTD 5.7.X versions. I would suggest that please open an issue for https://github.com/cms-sw/cmssw/issues for such backport

PR #10601, opened 2026-06-01
[Do not merge]Change config file download URLs in rivet.spec

akritkbehera:
Updated URLs for config.guess and config.sub to point to cmsrep.


smuzaffar:
Hold

No need to merge this now. We can consider this when we have other externals changes for 10.6.X ib.

PR #10631, opened 2026-06-13
Update SCRAMV1 and V2 to support unsetting env e.g GIT_SSL_CAINFO

smuzaffar:
`git` breaks in old releas cycles as `GIT_SSL_CAINFO` env set for releases in `slc5/slc6/slc7` points to an expired ca-bundle. SCRAM should now allow to unset an environment via scram site hooks e.g. we can unset `GIT_SSL_CAINFO` for existing releses and let git pick it up system


smuzaffar:
please test

PR #10658, opened 2026-06-17
Add optimization flags to `ucx.spec` while keeping ucx portable

ghyls:


ucx.spec
```diff
@@ -25,15 +30,8 @@ Requires: xpmem
   --disable-doxygen-html \
   --enable-compiler-opt \
   --enable-cma \
-  --enable-mt \
   --with-pic \
   --with-gnu-ld \
-%ifarch x86_64
-  --with-march=x86-64-v2 \
```
smuzaffar, line 32: @ghyls , by default now we build for `x86-64-v3`. Should we enable multi-vectorization of ucx too i.e. just like few other externals (https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_20_1_X/master/cmssw-vectorization.file)  we build `ucx` also for `x86-64-v3` and `x86-64-v2` (where by default `x86-64-v3` libs are used)

fwyzard: Naively I think we should build one version for `x86-64-v2` (ssse4.2) and one for `x86-64-v3` (avx/avx2/fma).

However @ghyls maybe want to try both to check if there is an actual difference in performance.

smuzaffar: if we enable vectorization for `ucx` then both `x86-64-v2` and `x86-64-v3` (default) will be available in CMSSW env  and then at rumtime scram will pick up the best microarch lib to use of one can explicitly use non-default v2 if needed

smuzaffar:
@ghyls , @fwyzard I have opened https://github.com/cms-sw/cmsdist/pull/10663 with vectorization support for ucx

smuzaffar:
please test

lets run tests (just in case @ghyls want to test it locally)

smuzaffar:
by the way, `ucx` by default build with `-g` ( https://github.com/openucx/ucx/blob/master/config/m4/compiler.m4#L13 ) . Do we want to keep it for production?

fwyzard:
@ghyls was suggesting to have two builds (now with vectorisation, three): one with debug flags and one without debug flags and full optimisations.

I think we could keep `-g` only in the debug build.

smuzaffar:
closing it in favor of #10663

PR #10661, opened 2026-06-18
use cryptography==46.0.7 for el8

smuzaffar:
As `cryptography>=47.0.0` requires OpenSSL 3 i.e. el9 and above so for `el8` we use old version of cryptography.


smuzaffar:
please build

smuzaffar:
please build for el8_amd64_gcc13

smuzaffar:
+externals

externals were properly built for el9 and el8 (using the correct version of cryptography). Merging it as this should fix the el8 IBs

PR #10680, opened 2026-06-24
setup cms env for scram-based project during build

smuzaffar:
Setup cms env during build so that scram hooks can find scram and cmsos


smuzaffar:
please test

smuzaffar:
+externals

PR #10695, opened 2026-07-01
Bump crabclient to latest

belforte:
after build in crab-dev was good, push these minor changes to prod

### DIFF
--- a/crab-prod.spec
+++ b/crab-prod.spec
@@ -3,8 +3,8 @@
 #For any other change, increment version_suffix
 ##########################################
 %define version_suffix 00
-%define crabclient_version v3.251020
+%define crabclient_version v3.260630
 ### RPM cms crab-prod %{crabclient_version}.%{version_suffix}
-%define crabserver_version v3.251015
+%define crabserver_version v3.260622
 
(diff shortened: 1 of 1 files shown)


smuzaffar:
please test

smuzaffar:
@belforte , just like crab clients `pre, dev and prod` , are there `pre, dev, prod` crab server instances? If yes then any combination of crab client and server should work? I can easily update our integration build jobs to tests all these combinations. Currently we are only tested crab clients `pre, dev and prod` against `prod` crab server

belforte:
THanks but that's not needed. Note that we use `crab-pre` to indicate previous client in case we need to fall back. Even if it never happened ! So `pre` for client has  a different meaning from `pre` for server (where it means `preprod` instance).  So we risk more confusion than gain. For client we have IB as "preprod".

We strive, and usually succeed, to deploy a backward/forward compatible client before changing server, but this time I forgot about updating the "previous client". In case of major screw-up we can always roll back server to be compatible with older client, that only takes a few minutes.

smuzaffar:
+externals

PR #10730, opened 2026-07-15
configtag V09-09-07: Modify Tool_DependencyCudaDLINK for conditional linking

### LINKED ISSUE #10548: CUDA dlink in cmssw generates duplicate registration entries
observed with CMSSW_16_1_0_pre4:
if I add a simple kernel in RecoTracker/LST/plugins/alpaka/LSTPixelHitsKernels.dev.cc and a module calling it RecoTracker/LST/plugins/alpaka/LSTPixelHitsFromSoAProducer.cc to the existing packages
we end up with duplicate registration symbols `__cudaRegisterLinkedBinary_*LSTEvent_dev_cc*`
one in pluginRecoTrackerLSTPluginsPortableCudaAsync.so and one in libRecoTrackerLSTCoreCudaAsync.so.
The former also has the new kernel registration in `__cudaRegisterLinkedBinary_*_LSTPixelHitsKernels_dev_cc_*`.
Due to the `LSTEvent_dev_cc` symbol duplication the kernel function registration is eventually missed for `_LSTPixelHitsKernels_dev_cc_`
This eventually leads to a runtime error `CUDA_ERROR_INVALID_HANDLE`

To reproduce, this branch can be used:
HSF-India-Pune-2026-LST:CMSSW_16_1_0_pre4_LST_pixelDirect-task5-pixsoa2soa
https://github.com/HSF-India-Pune-2026-LST/cmssw/pull/17/changes

Runtime is not necessarily required; after compilation the following indicates a problem
```
nm -C -l tmp/el9_amd64_gcc13/src/RecoTracker/LSTCore/src/alpaka/RecoTrackerLSTCoreCudaAsync/RecoTrackerLSTCoreCudaAsync_cudadlink.o | grep _LSTEvent_dev_cc_
0000000000000010 T __cudaRegisterLinkedBinary_fe01fecd_15_LSTEvent_dev_cc_75f6ad43_4100264      RecoTrackerLSTCoreCudaAsync_cudadlink.reg.c:2

nm -C -l tmp/el9_amd64_gcc13/src/RecoTracker/LST/plugins/RecoTrackerLSTPluginsPortableCudaAsync/RecoTrackerLSTPluginsPortableCudaAsync_cudadlink.o | grep _LSTEvent_dev_cc_
00000000000000e0 T __cudaRegisterLinkedBinary_fe01fecd_15_LSTEvent_dev_cc_75f6ad43_4100264      RecoTrackerLSTPluginsPortableCudaAsync_cudadlink.reg.c:3
```

One solution that I could come up with is to drop the present dependency on `LSTEvent_dev_cc` in the `RecoTracker/LST/plugins/alpaka` dlink step. Here I manually edited the command from `nvcc -dlink ... -lRecoTrackerLSTCoreCudaAsync_nv ...LSTPixelHitsKernels.dev.cc.o -o ...RecoTrackerLSTPluginsPortableCudaAsync_cudadlink.o` to drop the  `-lRecoTrackerLSTCoreCudaAsync_nv`.
The first direct evidence that this worked is the `__cudaRegisterLinkedBinary_*LSTEvent_dev_cc*` does not appear anymore in `RecoTrackerLSTPluginsPortableCudaAsync_cudadlink.o`
The runtime is also OK.

The question is: do we really need the dependent `cudadlink.o` to know about dev.cc/fatbins in another library? If not, these extra `-lRecoTrackerLSTCoreCudaAsync_nv` dependencies during dlink could be safely removed.

more notes in the [mattermost](https://mattermost.web.cern.ch/cms-o-and-c/pl/qtmrdwkyi3fmmfmrxxqxnghorr)

smuzaffar:
fixes #10548

### DIFF
--- a/scram-project-build.file
+++ b/scram-project-build.file
@@ -56,7 +56,7 @@ BuildRequires: dwz
 %endif
 
 %if "%{?configtag:set}" != "set"
-%define configtag       V09-09-06
+%define configtag       V09-09-07
 %endif
 
 %if "%{?buildarch:set}" != "set"


smuzaffar:
enable gpu

smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
please test using full cmssw

smuzaffar:
+externals

fwyzard:
Please revert until a proper integration test has been implemented.

fwyzard:
I take the opportunity to point out that this kind of [changes](https://github.com/cms-sw/cmsdist/pull/10730/changes) is impossible to review.

smuzaffar:
@fwyzard , full cmssw CUDA GPU unit tests and PR Relvals passed. The change only effects the alpaka cuda backend (the raw cuda build still behaves as previously) so do you still think it is not the correct fix to avoid ? 

Sorry my understanding of GPUs re limited and of course I can revert it but what else we expect from PR tests?

fwyzard:
Shazad, thanks for your work on this issue.

The problem is that my in understanding of how linking GPU software actually works is also limited.

That's why we should have a simple integration test that reproduces the problem, before merging a potential fix.

By the way, I don't think we should have different rules for nativa CUDA and for the alpaka backend - unless the integration test shows that the problem happens only with alpaka.

PR #10735, opened 2026-07-20
Implement CUDA 13 flags to parallelise device-code compilation

Parsifal-2045:
This PR implements a few new `nvcc` flags to speed-up device code compilation. In particular:
- [--split-compile number](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#split-compile-number-split-compile) to try and automatically divide heavy TUs into smaller chunks
- [--threads](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#threads-number-t) to allow parallel compilation of device-code TUs
- [--jobserver](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#jobserver-jobserver) to make it so both the split and the threads rely on the same jobs spawned by Make

I have performed some local measurements on a node equipped with 2 AMD EPYC 9534 using [#51085](https://github.com/cms-sw/cmssw/pull/51085) to pull about 25% of CMSSW and especially the `RecoTracker/PixelSeeding` package as one of the slowest to compile for device. I have also disabled rocm compilation to isolate the effects on the CUDA side. I attach here the results from time for each compilation step (starting from `scram b clean` before every pass):
```
Default compilation time:
real    13m16.469s
user    569m49.105s
sys     254m48.762s
Default RecoTracker/PixelSeeding compilation time:
real    9m39.174s
user    33m22.067s
sys     7m16.893s
New flags compilation time:
real    8m25.389s
user    510m24.473s
sys     107m21.013s
New flags RecoTracker/PixelSeeding compilation time:
real    2m32.060s
user    34m28.999s
sys     7m36.785s
```

Note: **Requires https://github.com/cms-sw/cmssw-config/pull/120 to be able to correctly parse the `--jobserver` flag**

### DIFF
--- a/cuda-flags.file
+++ b/cuda-flags.file
@@ -30,6 +30,12 @@
 # - mark __global__ functions and function templates, and __constant__, __device__ and __managed__ variables and variable templates, with __attribute__((visibility("hidden")))
 %define nvcc_visibility_flags --static-global-template-stub=true --device-entity-has-hidden-visibility=true
 
+# parallel device code compilation, integrated with make
+# split-compile=0 tries to automatically divide heavy TUs into multiple parallel compilation units
+# threads=0 uses all available CPU cores for parallel compilation
+# jobserver allows parallel compilation to be integrated with make's jobserver, so that the number of parallel jobs is limited to the number of jobs make is allowed to run in parallel
+%define nvcc_flags_parallel_compile --split-compile 0 --threads 0 --jobserver
+
 # build support for the various compute architectures
 %define nvcc_flags_cuda_archs %(echo $(for ARCH in %cuda_arch; do echo "-gencode arch=compute_$ARCH,code=[sm_$ARCH,compute_$ARCH]"; done)) -Wno-deprecated-gpu-targets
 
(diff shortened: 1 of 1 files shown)


akritkbehera:
please test with https://github.com/cms-sw/cmssw-config/pull/120 using full cmssw

Parsifal-2045:
type ngt

akritkbehera:
please test with https://github.com/cms-sw/cmssw-config/pull/120 using full cmssw

akritkbehera:
Let's rebuild and see

Parsifal-2045:
I am a bit puzzled by the failure, it looks like nothing here should be compiled using `nvcc` (or `CUDA_FLAGS`)

akritkbehera:
Yes it's not. It's puzzling I am taking a look!

smuzaffar:
@akritkbehera , the probklem is the way we build externals locally. For externals which are already installed on cvmfs we create symlinks in the build area and then torch fails when it tries to copy the aotriton libs (https://github.com/pytorch/pytorch/blob/aa034647b537aee3afaec28fd3054c566762d940/cmake/External/aotriton.cmake#L279-L284) . I would suggest to patch torch cmake file and avoid the install step e.g.
```
 if(DEFINED ENV{AOTRITON_INSTALLED_PREFIX})
-  install(DIRECTORY
-          $ENV{AOTRITON_INSTALLED_PREFIX}/lib
-          $ENV{AOTRITON_INSTALLED_PREFIX}/include
-          DESTINATION ${__AOTRITON_INSTALL_DIR})
   set(__AOTRITON_INSTALL_DIR "$ENV{AOTRITON_INSTALLED_PREFIX}")
   message(STATUS "Using Preinstalled AOTriton at ${__AOTRITON_INSTALL_DIR}")
```

the `install` command here copies the symlink `$ENV{AOTRITON_INSTALLED_PREFIX}/lib` instead of contents of `$ENV{AOTRITON_INSTALLED_PREFIX}/lib` . See if removing the `install` command works . If torch really need aotriton libs/include in the local `${__AOTRITON_INSTALL_DIR}` then may be just first try to read the symlink and then use the expanded paths in the `install` command

akritkbehera:
I have a PR open to fix py-torch-rocm.

akritkbehera:
please test with https://github.com/cms-sw/cmssw-config/pull/120

akritkbehera:
yay it's building CMSSW!

akritkbehera:
+externals

makortel:
Looks like this PR broke the build, see https://github.com/cms-sw/cmssw/issues/51570

PR #10737, opened 2026-07-22
Address the issues that are causing the RISCV IBs to fail.

akritkbehera:
With these changes I was able to build RISCV IBs using QEMU emulation layer. Since the RISCV node is down for now. Let's test these changes in AMD64 before we merge.


akritkbehera:

akritkbehera:
The tests seems to pass. Let's get this merged and see if we can build a RISC IBs

akritkbehera:
+externals

PR #10740, opened 2026-07-24
Update protobuf and abseil-cpp

akritkbehera:


akritkbehera:
please test with https://github.com/cms-sw/cmsdist/pull/10740

akritkbehera:
So it's not possible without updating Tensorflow! Tried it locally grpc needs the update and updating grpc breaks tensorflow 2.17.0

PR #10742, opened 2026-07-24
Update adept and g4vg due to update in VecGeom.

akritkbehera:


akritkbehera:
please test for CMSSW_20_1_G4ADEPT_X

akritkbehera:
I think these are failures related to version update. If there is failure in IBs we can debug it there.

akritkbehera:
+externals

PR #10764, opened 2026-08-06
Build LLVM only for the current CPU architecture

fwyzard:


fwyzard:
please test

akritkbehera:
Maybe also we can update llvm to the latest 22.1.8 while we are at it?

fwyzard:
OK for me, but I won't be able to do that.

Maybe we leave this PR as-is, to test only the architecture change, and you can open another one on top to update to the new version ?

fwyzard:
please test for el9_aarch64_gcc13

fwyzard:
please test for fc42_riscv64_gcc15

llvm.spec
```diff
@@ -56,7 +77,7 @@ cmake %{_builddir}/llvm-%{realversion}-%{llvmCommit}/llvm \
   -DCOMPILER_RT_INCLUDE_TESTS=OFF \
   -DLLVM_INCLUDE_TESTS=OFF \
   -DLLVM_HOST_TRIPLE=${host_triple} \
-  -DLLVM_TARGETS_TO_BUILD:STRING="X86;PowerPC;AArch64;RISCV;NVPTX" \
+  -DLLVM_TARGETS_TO_BUILD:STRING="%{llvmArch}%{llvmNvGpuArch}" \
```
hahnjo: You can also simply use `host` or `Native`: https://github.com/llvm/llvm-project/blob/50ff635538a52197ce708ce65c3fb1752be0f20f/llvm/cmake/config-ix.cmake#L593

mandrenguyen:
+1

fwyzard:
please test

smuzaffar:
please test

smuzaffar:
+externals

PR #10799, opened 2026-08-20
Build `libamdhip64.so` with debug symbols

fwyzard:
The size increase is order of 30MB, if there is no impact on the runtime performance I would actually suggest to keep this.


fwyzard:
test parameters:
  - enable = gpu
  - gpu = amd

fwyzard:
please test

fwyzard:
Included in #10800.

PR #10800, opened 2026-08-21
Prevent a deadlock in ROCm

fwyzard:
Prevent a deadlock between `hipEventSynchronize` and `hipMallocAsync`.

`hipMallocAsync` may try to acquire the device lock while it holds the memory pool lock, if the pool needs to allocate more memory.

`hipEventSynchronize` will try to acquire the memory pool lock while holding the device lock, to free the unused memory in the pool.

This may cause a deadlock if the two operations happen on different threads at the same time.

This change makes `MemoryPool::ReleaseFreedMemory` (called by `hipEventSynchronize`) return immediately if it cannot acquire the pool lock.

We also build `libamdhip64.so` with debug symbols.

### DIFF
--- a/patches/rocm-systems-issue10529.patch
+++ b/patches/rocm-systems-issue10529.patch
@@ -1,0 +1,36 @@
+commit 7e4352963e725c5ec3ef2bc541dcb3255dc2ecd3 (HEAD -> therock-7.14-issue10529, fwyzard/therock-7.14-issue10529)
+Author: Andrea Bocci <andrea.bocci@cern.ch>
+Date:   Fri Aug 21 12:25:16 2026 +0200
+
+    Prevent a deadlock between hipEventSynchronize and hipMallocAsync
+    
+    hipMallocAsync may try to acquire the device lock while it holds the memory
+    pool lock, if the pool needs to allocate more memory.
+    
+    hipEventSynchronize will try to acquire the memory pool lock while holding the
+    device lock, to free the unused memory in the pool.
+    
+    This may cause a deadlock if the two operations happen on different threads at
+    the same time.
+    
+    This change makes MemoryPool::ReleaseFreedMemory (called by hipEventSynchronize)
+    return immediately if it cannot acquire the pool lock.
+
+diff --git a/projects/clr/hipamd/src/hip_mempool_impl.cpp b/projects/clr/hipamd/src/hip_mempool_impl.cpp
+index 0983241188..0dc99953ef 100644
+--- a/projects/clr/hipamd/src/hip_mempool_impl.cpp
++++ b/projects/clr/hipamd/src/hip_mempool_impl.cpp
+@@ -339,7 +339,12 @@ void MemoryPool::ReleaseAllMemory() {
+ 
+ // ================================================================================================
+ void MemoryPool::ReleaseFreedMemory() {
+-  std::scoped_lock lock(lock_pool_ops_);
++  // Do not block if the lock cannot be acquired, to avoid a deadlock if a different thread holds
++  // the pool lock and tries to acquire the device lock, e.g. hipMallocAsync requesting new memory.
++  std::unique_lock lock(lock_pool_ops_, std::try_to_lock);
++  if (! lock.owns_lock()) {
++    return;
++  }
+ 
+   free_heap_.ReleaseAllMemory();
+ }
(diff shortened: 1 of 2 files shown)


fwyzard:
enable gpu

fwyzard:
please test

smuzaffar:
@fwyzard , please apply the patch for rocm-hip directly (instead of applying it to rocm-sources which trigger re-build of all rocm packages). Just clean up the rocm-sources changes and change rocm-hip to 
```
diff --git a/rocm/rocm-hip.spec b/rocm/rocm-hip.spec
index 4ec29cb9e..5b5404a2d 100644
--- a/rocm/rocm-hip.spec
+++ b/rocm/rocm-hip.spec
@@ -4,9 +4,11 @@
 ## INITENV SET HIP_PATH %{i}
 ## INITENV SET HIP_CLANG_PATH ${ROCM_LLVM_ROOT}/lib/llvm/bin
 ## INITENV HIP_PLATFORM amd
+Patch0: patches/rocm-systems-issue10529
 BuildRequires: py3-CppHeaderParser
 Requires: rocm-llvm rocm-core rocr-runtime rocprofiler-register numactl python3 rocm-comgr
 Provides: perl(URI::Escape)
+%define ROCMPostPrep patch -p3 -i %{PATCH0}
 %define rocm_project clr
```

fwyzard:
I assume the failures related to ticl dictionaries in the Alignment package are unrelated.

fwyzard:
> @fwyzard , please apply the patch for rocm-hip directly (instead of applying it to rocm-sources which trigger re-build of all rocm packages). Just clean up the rocm-sources changes and change rocm-hip to
> 
> ```
> diff --git a/rocm/rocm-hip.spec b/rocm/rocm-hip.spec
> index 4ec29cb9e..5b5404a2d 100644
> --- a/rocm/rocm-hip.spec
> +++ b/rocm/rocm-hip.spec
> @@ -4,9 +4,11 @@
>  ## INITENV SET HIP_PATH %{i}
>  ## INITENV SET HIP_CLANG_PATH ${ROCM_LLVM_ROOT}/lib/llvm/bin
>  ## INITENV HIP_PLATFORM amd
> +Patch0: patches/rocm-systems-issue10529
>  BuildRequires: py3-CppHeaderParser
>  Requires: rocm-llvm rocm-core rocr-runtime rocprofiler-register numactl python3 rocm-comgr
>  Provides: perl(URI::Escape)
> +%define ROCMPostPrep patch -p3 -i %{PATCH0}
>  %define rocm_project clr
> ```

@smuzaffar doesn't this make building one package change the content of another (the sources) ?
Is that allowed ?
What if the sources are on a readonly place like CVMFS ?
what if rocm-hip is built twice from the same sources ?

fwyzard:
Ah, I see that the sources are copied every time when building:
```bash
rsync -a ${ROCM_SOURCES_ROOT}/rocm-systems/projects/%{rocm_project}/ %{_builddir}/rocm-systems/projects/%{rocm_project}
```

fwyzard:
test parameters:
  - enable = gpu
  - gpu = amd

fwyzard:
please test

fwyzard:

fwyzard:
@smuzaffar I would go ahead and merge this, as it should fix the random hangs and timeouts with AMD GPUs.

I will try to check the impact on performance next week.

smuzaffar:
+externals

unit test failures are not related to this change ( tests are already failing in Ibs)

mmusich:
@fwyzard @smuzaffar 
IIUC, https://github.com/cms-sw/cmssw/issues/51677 will apply to all active release cycles.
Does it make sense to backport this PR to at least 20_0_X and 17_0_X where we might still have active HLT development in order to have smooth tests ?

fwyzard:
I am in favour, however those cycles are still using ROCm 7.1.
Do we need some more validation first?

mmusich:
> I am in favour, however those cycles are still using ROCm 7.1.

thanks. Does it mean we would need to update the ROCm version in the previous cycles for this to work or that we should first validate this change on top of the current native version?

fwyzard:
> Does it mean we would need to update the ROCm version in the previous cycles for this to work or that we should first validate this change on top of the current native version?

In 17.0.x and 20.0.x we don't build ROCm from sources, so we cannot apply a minimal fix.

fwyzard:
@akritkbehera how much work is it to backport the work to use ROCm 7.14 from sources (including this fix and your latest update for rocgdb) to CMSSW 17.0.x and 20.0.x ?

PR #10804, opened 2026-08-27
[rocm-llvm] make sure to deploy clang.cfg and host-triple.cfg

smuzaffar:
FYI @akritkbehera , rocm-comgr failed to build on cmsdev machines (for `el8_amd64_*`) as `clang` from rocm-llvm was picking up gcc from system and complain about missing `-lgcc -lgcc_s` . Things worked in our container (el8) as we have extra gcc packages installed there and `clang` was able to find `gcc and gcc_s` from system. The proper fix is that we should have `clang.cfg` deployed with the correct gcc-toolchain. This change fixes this issue


smuzaffar:
please test

fwyzard:

smuzaffar:
+externals

PR #10812, opened 2026-09-01
ROCM 10.0

smuzaffar:


smuzaffar:
enable gpu

smuzaffar:
please test using full cmssw

smuzaffar:
please test using full cmssw

smuzaffar:
please test with cms-sw/cmssw#51801 using full cmssw

smuzaffar:
please test with https://github.com/cms-sw/cmssw/pull/51801 using full cmssw

smuzaffar:
@fwyzard , this Pull requests proposes to update ROCM to their latest version `10.0`. Do you thnk we are ready to move to this version in 20.1.X or are there any known blockers?

fwyzard:
I would say that we should run some benchmarks to check for regressions, but I'm not sure if and when I would have the time to do it.

PR #10813, opened 2026-09-01
Avoid patching by openmpi since that causes failure in ASAN IBs

akritkbehera:


akritkbehera:
please test for CMSSW_20_1_ASAN_X

akritkbehera:
Similar issue as already reported here https://github.com/open-mpi/ompi/issues/13069 if needed we can conditionally set this only for ASAN_X IBs

OpenMPI overwrites a libasan entry point assuming it has enough space in memory. That holds for GCC13, but in GCC14 the libasan entry points are tightly packed 5-byte stubs, so the overwrite runs past them and this fails.

`export OMPI_MCA_memory="^patcher"` disables the patcher entirely, so the entry points are never overwritten.

fwyzard:
According to [these comments](https://github.com/open-mpi/ompi/issues/13069#issuecomment-2640203950) it's ok to disable the patcher for the sanitizer builds, but it should not be disabled for the production environment.

Can the environment variable be set only for ASAN/UBSAN ?

fwyzard:
@ghyls FYI, this is what is causing the failures you were looking into this morning.

fwyzard:
hold

Can the environment variable be set only for ASAN/UBSAN ?

akritkbehera:
> hold
> 
> Can the environment variable be set only for ASAN/UBSAN ?

Yes we can do that.

smuzaffar:
please test

smuzaffar:
please test for CMSSW_20_1_ASAN_X

fwyzard:
unhold

Do we need it also for UBSAN ?

smuzaffar:
> unhold
> 
> Do we need it also for UBSAN ?

We do not see these openmpi failures for UBSAN ( https://cmssdt.cern.ch/SDT/cgi-bin/showBuildLogs.py/el9_amd64_gcc14/www/wed/20.1.UBSAN-wed-23/CMSSW_20_1_UBSAN_X_2026-08-26-2300?utests ) , so for now, no we do not need this for UBSAN

smuzaffar:
new env variable is properly set only for ASAN env
- In non-ASAN  IB
```
Singularity> cmsenv
Singularity>  echo $CMSSW_VERSION
CMSSW_20_1_X_2026-09-01-2300
Singularity> env | grep OMPI_MCA_
OMPI_MCA_accelerator=null
```
- IN ASAN_X IB
```
Singularity> cmsenv
Singularity> echo $CMSSW_VERSION
CMSSW_20_1_ASAN_X_2026-08-31-2300
Singularity> env | grep OMPI_MCA_
OMPI_MCA_memory=^patcher
OMPI_MCA_accelerator=null
```

akritkbehera:
please test for CMSSW_20_1_ASAN_X

smuzaffar:
+externals

good to go in. This fixes the ASAN/openmpi related unit tests

PR #10814, opened 2026-09-01
FWLite: Added HeterogeneousCore/AlpakaCore dependency

smuzaffar:
https://github.com/cms-sw/cmssw/pull/50503 Added HeterogeneousCore/AlpakaCore dependency  which causes fwlite build to fail with error
```
src/HeterogeneousCore/TrivialSerialisation/plugins/alpaka/GenericClonerPortable.cc:41:10: fatal error: HeterogeneousCore/AlpakaCore/interface/alpaka/EDMetadataSentry.h: No such file or directory
   41 | #include "HeterogeneousCore/AlpakaCore/interface/alpaka/EDMetadataSentry.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

### DIFF
--- a/fwlite_build_set.file
+++ b/fwlite_build_set.file
@@ -129,6 +129,7 @@ Geometry/CaloGeometry
 Geometry/CommonDetUnit
 Geometry/CommonTopologies
 Geometry/GEMGeometry
+HeterogeneousCore/AlpakaCore
 HeterogeneousCore/AlpakaInterface
 HeterogeneousCore/TrivialSerialisation
 PhysicsTools/FWLite


makortel:
No, we should not add dependence to `HeterogeneousCore/AlpakaCore` to FWLite (because it would pull in the full framework).

makortel:
It seems like already before https://github.com/cms-sw/cmssw/pull/50503 the `HeterogeneousCore/TrivialSerialization/plugins` contained code that depends on the full framework. Were we building it in FWLite?

smuzaffar:
@makortel , looks like fwlite build set already has the FW code https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_20_1_X/master/fwlite_build_set.file#L100-L114 , so there is no new FW dependency

makortel:
> @makortel , looks like fwlite build set already has the FW code https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_20_1_X/master/fwlite_build_set.file#L100-L114 , so there is no new FW dependency

Ah, right. I wonder why many of those were included.

An alternative would be to not build the `HeterogeneousCore/TrivialSerialization/plugins` in FWLite.

smuzaffar:
In one of Core SW meetinsg, we discussed if we should build plugins or not. I do not remember what was the decision . @makortel do you recall any thing

fwyzard:
Do we build any other `alpaka` directory for FWLite ?

smuzaffar:
yes by default we build every fwlite package mentioned iin https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_20_1_X/master/fwlite_build_set.file ( except the test sub-directory of these packages). Some of these do contains alpaka sub-directories e.g
```
DataFormats/AlpakaCommon/interface/alpaka  DataFormats/EcalDigi/src/alpaka          DataFormats/HcalDigi/interface/alpaka          DataFormats/ParticleFlowReco/plugins/alpaka              HeterogeneousCore/TrivialSerialisation/plugins/alpaka
DataFormats/AlpakaCommon/src/alpaka        DataFormats/EcalRecHit/interface/alpaka  DataFormats/HcalDigi/plugins/alpaka            DataFormats/ParticleFlowReco/src/alpaka                  HeterogeneousCore/TrivialSerialisation/src/alpaka
DataFormats/BeamSpot/interface/alpaka      DataFormats/EcalRecHit/plugins/alpaka    DataFormats/HcalDigi/src/alpaka                DataFormats/Portable/interface/alpaka                    HeterogeneousCore/TrivialSerialisation/test/alpaka
DataFormats/BeamSpot/plugins/alpaka        DataFormats/EcalRecHit/src/alpaka        DataFormats/HcalRecHit/interface/alpaka        DataFormats/Portable/test/alpaka
DataFormats/BeamSpot/src/alpaka            DataFormats/HGCalReco/interface/alpaka   DataFormats/HcalRecHit/plugins/alpaka          DataFormats/TrivialSerialisation/test/alpaka
DataFormats/EcalDigi/interface/alpaka      DataFormats/HGCalReco/plugins/alpaka     DataFormats/HcalRecHit/src/alpaka              HeterogeneousCore/AlpakaInterface/test/alpaka
DataFormats/EcalDigi/plugins/alpaka        DataFormats/HGCalReco/src/alpaka         DataFormats/ParticleFlowReco/interface/alpaka  HeterogeneousCore/TrivialSerialisation/interface/alpaka

```

fwyzard:
Ah... but do we do also the device compilations with `nvcc` and `hipcc` ?
Or only the host case ?

makortel:
> In one of Core SW meetinsg, we discussed if we should build plugins or not. I do not remember what was the decision . @makortel do you recall any thing

I don't remember. In the leading order I think the plugins should not be built.

smuzaffar:
> Ah... but do we do also the device compilations with `nvcc` and `hipcc` ? Or only the host case ?

yes we build device code too. Device code is automatically enabled if cuda/rocm externals are available and I see that in fwlite tools we have dependency on both rocm and cuda. We can cleanup https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_20_1_X/master/fwlite-tools.spec to remove rocm/cuda

smuzaffar:
We will update bot to for now drop the HeterogeneousCore/*/plugins for FWLite. We will also try to clean up fwlite to not depend on Framework and drop the rocm and cuda dependencies

PR #10817, opened 2026-09-01
Build Sherpa for different microarchs

fwyzard:
Patch Sherpa to avoid c++20 warnings about asymmetric comparison operator.


fwyzard:
please test

fwyzard:
Not clear if this is actually needed or useful.

PR #10819, opened 2026-09-02
Update Boost version to 1.92.0 in spec file

smuzaffar:
Update Boost patch


smuzaffar:
please test using full cmssw

smuzaffar:
+externals

we will merge it after 20.1.0.pre3 is released

PR #10825, opened 2026-09-03
md5: To avoid collision: rename md5.h to edm_md5

### LINKED ISSUE #2: Rename the header to avoid collisions

smuzaffar:
Resolves https://github.com/cms-externals/md5/issues/2

md5.h is rename to edm_md5 and all functions are moved under edm_md5 namespace

### DIFF
--- a/md5.spec
+++ b/md5.spec
@@ -1,22 +1,22 @@
-### RPM external md5 1.0.0
-%define tag d97a571864a119cd5408d2670d095b4410e926cc
-%define branch cms/1.0.0
+### RPM external md5 2.0.0
+%define tag 1ed14d187d793216fb8345363f590bf3effd95e2
+%define branch cms/%{realversion}
 %define github_user cms-externals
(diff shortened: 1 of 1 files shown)


smuzaffar:
please build

smuzaffar:
please test with https://github.com/cms-sw/cmssw/pull/51818

smuzaffar:
ignore tests-rejected with manual-override

smuzaffar:
+externals


this should go in with https://github.com/cms-sw/cmssw/pull/51818

mandrenguyen:
+1

PR #10826, opened 2026-09-03
Fix PyTorch ROCm library linking

EmanueleCoradin:
#### PR description

This PR adds two additional linker settings:

- `FORCE_LINK` ensures that the ROCm backend library is kept in the final dependency chain even when no symbol from it is referenced directly.
- `-L$TOOL_BASE/lib` gives priority to the libraries shipped with the ROCm PyTorch package when linking a ROCm target. This ensures that `libtorch_hip`, `libtorch_cpu`, `libc10`, etc. all come from the same ROCm PyTorch build, instead of accidentally mixing ROCm-specific libraries with the generic PyTorch libraries.

### DIFF
--- a/scram-tools.file/tools/py3-torch-rocm/pytorch-rocm.xml
+++ b/scram-tools.file/tools/py3-torch-rocm/pytorch-rocm.xml
@@ -1,10 +1,12 @@
-<tool revision="3">
+<tool revision="4">
   <client>
     <environment name="INCLUDE" default="$TOOL_BASE/include"/>
     <environment name="INCLUDE" default="$TOOL_BASE/include/torch/csrc/api/include"/>
     <environment name="LIBDIR" default="$TOOL_BASE/lib"/>
   </client>
   <flags SKIP_TOOL_SYMLINKS="1"/>
+  <flags FORCE_LINK="1"/>
+  <flags LDFLAGS="-L$TOOL_BASE/lib"/>
   <use name="pytorch"/>
   <use name="rocm"/>
   <lib name="torch_hip"/>


EmanueleCoradin:
type ngt

fwyzard:
@EmanueleCoradin can you drop `[NGT]` from the PR title ?

scram-tools.file/tools/py3-torch-rocm/pytorch-rocm.xml
```diff
@@ -1,10 +1,12 @@
-<tool revision="3">
+<tool revision="4">
   <client>
     <environment name="INCLUDE" default="$TOOL_BASE/include"/>
     <environment name="INCLUDE" default="$TOOL_BASE/include/torch/csrc/api/include"/>
     <environment name="LIBDIR" default="$TOOL_BASE/lib"/>
   </client>
   <flags SKIP_TOOL_SYMLINKS="1"/>
+  <flags FORCE_LINK="1"/>
+  <flags LDFLAGS="-L$TOOL_BASE/lib"/>
```
fwyzard, line 9: I don't know what it the right approach, but I don't think it is this one.

EmanueleCoradin: Hello! I worked on PR #51819 and discovered some errors with the current status of cmsdist and cmssw-config.
After a lot of trials, I came up with a configuration that is working.
I decided to open the draft PRs to collect feedback from you and the other core experts, as I am not knowledgeable enough to come up with different solutions. I am preparing more detailed explanations, still in progress

smuzaffar: I have opened https://github.com/cms-sw/cmssw-config/pull/121 to fix part of the issue and https://github.com/cms-sw/cmssw/issues/51822 to see what to do with `edmPluginRefresh` whcih loads all plugins in single process

EmanueleCoradin: I closed the PR as you are taking care of the problems. Thanks!

PR #10839, opened 2026-09-09
[DEVEL] Fix dependency for extra micro-arch biglib

smuzaffar:


smuzaffar:
enable gpu,multi_microarchs

smuzaffar:
please test using full cmssw

PR #10844, opened 2026-09-10
Update CUDA to version 13.4.1

fwyzard:
Update CUDA to version 13.4.1 (general availability).

Major changes since CUDA 13.3:
  - add support for GCC 16 and clang 22 as host compilers;
  - the NVIDIA drivers are now packaged separately.

Patch the `cuda-gdb` shell script rather than adding a second wrapper.

### DIFF
--- a/cuda.spec
+++ b/cuda.spec
@@ -1,14 +1,16 @@
-### RPM external cuda 13.3.1
+### RPM external cuda 13.4.1
 ## INITENV +PATH LD_LIBRARY_PATH %i/lib64
 
 %define runpath_opts -m compute-sanitizer -m drivers -m nvvm
-%define driversversion 610.43.02
+%define driversversion 615.71.09
 
 %ifarch x86_64
-Source0: https://developer.download.nvidia.com/compute/cuda/%{realversion}/local_installers/%{n}_%{realversion}_%{driversversion}_linux.run
+Source0: https://developer.download.nvidia.com/compute/cuda/%{realversion}/local_installers/%{n}_%{realversion}_linux.run
+Source1: https://developer.download.nvidia.com/compute/nvidia-driver/redist/nvidia_driver/linux-x86_64/nvidia_driver-linux-x86_64-%{driversversion}-archive.tar.xz
 %endif
 %ifarch aarch64
-Source0: https://developer.download.nvidia.com/compute/cuda/%{realversion}/local_installers/%{n}_%{realversion}_%{driversversion}_linux_sbsa.run
+Source0: https://developer.download.nvidia.com/compute/cuda/%{realversion}/local_installers/%{n}_%{realversion}_linux_sbsa.run
+Source1: https://developer.download.nvidia.com/compute/nvidia-driver/redist/nvidia_driver/linux-sbsa/nvidia_driver-linux-sbsa-%{driversversion}-archive.tar.xz
 %endif
 Requires: python3
 AutoReq: no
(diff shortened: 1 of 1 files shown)


fwyzard:
@smuzaffar @akritkbehera, if this works for GCC 14 in C++20 mode, how do I test it with GCC 15 and GCC 16 in C++23 mode ?

fwyzard:
enable gpu

fwyzard:
please test

fwyzard:
please test for el9_amd64_gcc16 with cms-sw/cms-bot#2838

fwyzard:
please test for el9_amd64_gcc15 with cms-sw/cms-bot#2838

fwyzard:
The failure with GCC 16 is a general CUDA issue, already reported here: https://forums.developer.nvidia.com/t/nvcc-13-4-rejects-conforming-libstdc-string-with-g-16/382852 .

smuzaffar:
torch still fails with `Error: Internal Compiler Error (codegen): "unsupported float variant!"` for gcc15 with c++23

fwyzard:
> torch still fails with `Error: Internal Compiler Error (codegen): "unsupported float variant!"` for gcc15 with c++23

Is that because of gcc15 or because of c++23 ?

smuzaffar:
I think this is due to c++23, last time we tested cuda update with CPP23/GCC14 and we have seen the same ICE for that.

fwyzard:
Would it be useful to have CUDA with gcc15 or gcc16 in c++20 mode ?

fwyzard:
What I mean is that, given that CUDA still does not really support c++23, we could at least update the compiler to gcc 15 or gcc 16 and remain in c++20 mode 🤷🏻‍♂️

smuzaffar:
I have no objections moving default gcc15/gcc16 to c++20. May be we can add extra CPP23 IBs for gcc15 and gcc16 just to make sure we build the sw stack with c++23 and newer compilers

fwyzard:
> I have no objections moving default gcc15/gcc16 to c++20.

OK, let's bring this up at the next Core Software meeting.

> May be we can add extra CPP23 IBs for gcc15 and gcc16 just to make sure we build the sw stack with c++23 and newer compilers.

Yes, absolutely.

makortel:
> > I have no objections moving default gcc15/gcc16 to c++20.
> 
> OK, let's bring this up at the next Core Software meeting.

FWIW I'm fine with the suggestion (in case you don't want to wait for the next core meeting that might take a while).

> > May be we can add extra CPP23 IBs for gcc15 and gcc16 just to make sure we build the sw stack with c++23 and newer compilers.
> 
> Yes, absolutely.

This sounds good to me as well.

smuzaffar:
ok, I will change GCC15/16 OBs to use default c++20 and also open CPP23 IBs for GCC15/16

fwyzard:
please test

smuzaffar:

smuzaffar:
@fwyzard , this looks good to go it. Do you want to run any local tests first otherwise once GCC15 tests here and GCC16 tests via https://github.com/cms-sw/cms-bot/pull/2843#issuecomment-5757833274 are done then I can merge it?

fwyzard:
I'd like to run some tests first, hopefully done in a couple of days.

fwyzard:
Rebased.

fwyzard:
please test

